PR #20, opened 2023-02-28
Fix SuperLUSolver memory bug

hughcars:
SuperLUSolver has a memory bug.

The parent class requires the usage of a pointer to the memory declared in the child class. Changing to a composition model ensures that the `mfem::SuperLUSolver` is deleted first, before the `Aint` variable has become invalid. This requires some changes to mfem branch too

*Description of changes:*

Change from inheritance to composition. The class stores a `solver` variable and dispatches to it.

*Issue #, if available:*
none.

### DIFF
--- a/palace/deps/patch/mfem/patch_direct.diff
+++ b/palace/deps/patch/mfem/patch_direct.diff
@@ -1,5 +1,5 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index 3ff5b9462..e7e1e8212 100644
+index ef3ba4f6f..8fe537b57 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
 @@ -78,7 +78,7 @@ if (MFEM_USE_CONDUIT OR
@@ -11,7 +11,7 @@ index 3ff5b9462..e7e1e8212 100644
    # Just needed to find the MPI_Fortran libraries to link with
    set(XSDK_ENABLE_Fortran ON)
  endif()
-@@ -313,6 +313,7 @@ endif()
+@@ -312,6 +312,7 @@ endif()
  if (MFEM_USE_MUMPS)
    if (MFEM_USE_MPI)
      find_package(MUMPS REQUIRED mumps_common pord)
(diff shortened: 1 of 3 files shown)


palace/linalg/superlu.cpp
```diff
@@ -36,40 +36,40 @@ int GetNpDep(int np, bool use_3d)
...
 
 SuperLUSolver::SuperLUSolver(MPI_Comm comm, int sym_fact_type, bool use_3d, int print_lvl)
-  : mfem::SuperLUSolver(comm, GetNpDep(Mpi::Size(comm), use_3d))
+  : solver(comm, GetNpDep(Mpi::Size(comm), use_3d))
 {
   // Configure the solver.
   if (print_lvl > 1)
   {
-    if (npdep_ > 1)
+    if (GetNpDep(Mpi::Size(comm), use_3d) > 1)
```
sebastiangrimberg, line 44: Why not `solver.npdep_`?

palace/deps/patch/mfem/patch_direct.diff
```diff
@@ -183,37 +183,6 @@ index f8af98a72..0d4503723 100644
  #cmakedefine MFEM_USE_BENCHMARK
  
  // Enable Enzyme for AD
-diff --git a/config/cmake/modules/FindButterflyPACK.cmake b/config/cmake/modules/FindButterflyPACK.cmake
```
sebastiangrimberg, line 186: You probably did not mean to remove this file from the patch along with `FindZFP.cmake`.

palace/linalg/superlu.cpp
hughcars: I switched to the above originally when npdep was a protected variable, just hadn't updated it.

sebastiangrimberg (approved):

palace/linalg/superlu.hpp
```diff
@@ -41,7 +41,22 @@ class SuperLUSolver : public mfem::Solver
...
+  }
+  void MultTranspose(const mfem::Vector &x, mfem::Vector &y) const override
+  {
+    solver.MultTranspose(x, y);
+  }
+  void ArrayMultTranspose(const mfem::Array<const mfem::Vector *> &X,
+                          mfem::Array<mfem::Vector *> &Y) const override
+  {
+    solver.ArrayMultTranspose(X, Y);
+  }
```
hughcars, line 59: If we're not using the methods in any of the compilation, suggests we don't even really need them, or at least that mfem apparently doesn't deem completing them to be necessary for implementing an operator.

sebastiangrimberg: The methods have (unoptimized) implementations in the base `mfem::Operator` class.

PR #24, opened 2023-03-07
Fix bug for second-order absorbing boundary condition

sebastiangrimberg:
For boundary element integrals, the dimension of the curl is 1D in the reference space and thus the coefficient needs to be scalar for compatibility. For matrix-valued material properties, we need to use the normal-projected coefficient direction instead of the full matrix which is `space_dim x space_dim`.

Also updates `mfemcoefficients.hpp` classes to store small `mfem::Vector` and `mfem::Matrix` objects as member variables.

### DIFF
--- a/palace/utils/mfemcoefficients.hpp
+++ b/palace/utils/mfemcoefficients.hpp
@@ -58,12 +58,15 @@ class BdrCurrentVectorCoefficient : public mfem::VectorCoefficient,
 private:
   const mfem::ParGridFunction &B;
   const MaterialOperator &mat_op;
+  mfem::Vector C1, W, VU, VL, nor;
 
 public:
   BdrCurrentVectorCoefficient(const mfem::ParGridFunction &gf, const MaterialOperator &op,
                               const std::map<int, int> &l2s)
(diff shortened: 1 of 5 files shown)


palace/utils/mfemcoefficients.hpp
```diff
@@ -117,37 +116,35 @@ class BdrChargeCoefficient : public mfem::Coefficient, public BdrGridFunctionCoe
 private:
   const mfem::ParGridFunction &E;
   const MaterialOperator &mat_op;
+  mfem::Vector C1, W, VU, VL, nor;
```
hughcars, line 119: Maybe mark as `mutable` given these are internal working variables, their `const` status needn't block making the class `const`.

hughcars (approved):

palace/utils/mfemcoefficients.hpp
sebastiangrimberg: Done in [b26aead](https://github.com/awslabs/palace/pull/24/commits/b26aead1d3858a4bf54900a877b3b273a2228a14).

PR #44, opened 2023-05-26
Build PETSc without HDF5 for Spack CI test

sebastiangrimberg:
Example CI workflow failure: https://github.com/awslabs/palace/actions/runs/4998666186/jobs/9139924714.

Also updates {fmt} for superbuild to v10.0.0 to match Spack and fixes a corresponding compilation error.


palace/utils/communication.hpp
```diff
@@ -17,46 +17,115 @@ namespace mpi
 {
 
 template <typename T>
-struct CommTrace
```
hughcars, line 20: I think there's a way to reduce the size of this file a good bit, and to place all the switches in one location, a quick example of the technique:
```
template <typename T>
struct MpiTypeOf {    
    static constexpr int get() {
        if (std::is_same<T, double>::value) return 0;
        else if (std::is_same<T, int>::value) return 1;
        else if (std::is_same<T, long long>::value) return 2;
        else return -1;
    }
    static constexpr int value { get() };
};

template <typename T>
constexpr int MpiType = MpiTypeOf<T>::value;
```
demo: https://godbolt.org/z/qrc8r7j4j

Adding another type is one line then, and all the supported types are line adjacent to each other. The call site then also doesn't have the `()` too.

sebastiangrimberg: Indeed I had tried this route first, but with `MPI_Datatype` it's unfortunately not so simple as they are not defined as `constexpr`: https://github.com/open-mpi/ompi/issues/10017. Thus specialization it is.

hughcars: The constexpr-ness isn't necessary for that trick, as the compile time information that is being operated on, is the base types, not the MPI type, it actually gets shorter as there's no need for the indirection.
```
template <typename T>
struct MpiTypeOf {    
    static int get() {
        if (std::is_same<T, double>::value) return 0;
        else if (std::is_same<T, int>::value) return 1;
        else if (std::is_same<T, long long>::value) return 2;
        else return -1;
    }
};

template <typename T>
static const int MpiType = MpiTypeOf<T>::get();
```
demo: https://godbolt.org/z/TcodWx5e6

sebastiangrimberg: Losing `constexpr`, we lose the compile time error for an unsupported datatype. I prefer the function template specialization.

hughcars (approved):

PR #53, opened 2023-06-14
Github Actions CI tests with ARPACK in addition to SLEPc

sebastiangrimberg:
Adds ARPACK (from [ARPACK-NG](https://github.com/opencollab/arpack-ng)) to CI tests. Also a few other included fixes for CI:

- Fixes a bug which just showed up for CI builds using AOCL
- Better naming of CI job variants (replace `true`/`false` with actual keywords)
- Automatically add runtime library path in `palace` wrapper for shared library builds


.github/workflows/build-and-test-linux-aarch64.yml
```diff
@@ -86,6 +86,10 @@ jobs:
               NUM_PROC_BUILD=$NUM_PROC_BUILD_MAX
             fi
 
+            [[ "${{ matrix.build-shared }}" == 'shared' ]] && BUILD_SHARED='ON' || BUILD_SHARED='OFF'
```
hughcars, line 89: With these short circuits, I'm reading this as `([[ "${{ matrix.build-shared }}" == 'shared' ]] && BUILD_SHARED='ON') || BUILD_SHARED='OFF'`, as otherwise there's no need to include `BUILD_SHARED`. Might consider adding braces (if that's an option).

.github/workflows/build-and-test-macos.yml
```diff
@@ -16,31 +16,31 @@ jobs:
...
-            with-solver: strumpack
-            with-eigensolver: slepc
+          # - compiler: clang
+          #   mpi: openmpi
+          #   math-libs: openblas
+          #   build-shared: shared
+          #   with-64bit-int: int32
+          #   with-openmp: serial
+          #   with-solver: superlu
+          #   with-eigensolver: slepc
```
hughcars, line 26: If these are being commented out, might as well delete the lines

hughcars (approved):

.github/workflows/build-and-test-linux-aarch64.yml
sebastiangrimberg: Being bash, there's plenty of ways to achieve the same effect: https://stackoverflow.com/questions/3953645/ternary-operator-in-bash. I'll leave this as is for now because it seems the no-braces is widely used.

.github/workflows/build-and-test-macos.yml
sebastiangrimberg: This particular CI job should not be failing but for some unknown reason was. My plan is to re-enable in downstream PR to fix this issue.

PR #56, opened 2023-06-27
Update build system to prepare for libCEED

### LINKED ISSUE #61: SuperLU_DIST requires newer CMake version
*Description: A clear and concise description of what the bug is.*
SuperLU_DIST (optional dependency) currently requires a newer CMake version (3.18.1 or higher) than that of Palace (3.13).

Would be nice if this were added to the documentation (or just the minimum version raised)

*To reproduce: Please provide minimal example that reproduces the error. For existing
examples, please provide a link.*
On a system with with older CMake
```
    # Compile Palace
    git clone https://github.com/awslabs/palace.git --recurse-submodules
    cd palace || exit
    mkdir build
    cd build || exit
    cmake ../ -DCMAKE_INSTALL_PREFIX=/opt/palace \
            -DPALACE_WITH_OPENMP:BOOL=TRUE \
            -DCMAKE_CXX_FLAGS="-O3 -fopenmp -funroll-loops -march=$MARCH" \
            -DCMAKE_C_FLAGS="-O3 -fopenmp -funroll-loops -march=$MARCH -D_POSIX_C_SOURCE=199309L" \
            -DCMAKE_Fortran_FLAGS="-O3 -fPIC -funroll-loops -march=$MARCH" \
            -DPALACE_WITH_MUMPS:BOOL=TRUE
    make -j "$(nproc)"
```

*Error message: Paste the complete error message or log, if applicable.*
```
[ 31%] Creating directories for 'superlu_dist'
[ 32%] Performing download step (git clone) for 'superlu_dist'
Cloning into 'superlu_dist'...
done.
HEAD is now at 09cb1430 Rename constant name EMPTY to SLU_EMPTY in util_dist.h, to avoid name clash with system defs.
[ 33%] No patch step for 'superlu_dist'
[ 34%] No update step for 'superlu_dist'
[ 35%] Performing configure step for 'superlu_dist'
CMake Error at CMakeLists.txt:8 (cmake_minimum_required):
  CMake 3.18.1 or higher is required.  You are running version 3.17.5


-- Configuring incomplete, errors occurred!
make[2]: *** [extern/CMakeFiles/superlu_dist.dir/build.make:130: extern/superlu_dist-cmake/src/superlu_dist-stamp/superlu_dist-configure] Error 1
make[1]: *** [CMakeFiles/Makefile2:268: extern/CMakeFiles/superlu_dist.dir/all] Error 2
make: *** [Makefile:104: all] Error 2
make: *** No rule to make target 'install'.  Stop.
```

*Environment: Any environment details, such as operating system, compiler, etc.*
OS: centos:7.9.2009
CMake 3.17.5
MPICH 4.0.2

sebastiangrimberg:
Refactor CMake superbuild to move MFEM back to an `ExternalProject` and always build some set of dependencies as part of a Palace build (in preparation for libCEED, which cannot be build using `FetchContent`).

Also removes Git submodules, pulling source directly from the remote repo via CMake.

Experimented with using Scotch and PT-Scotch as replacements for METIS/ParMETIS in the superbuild, which was fun, but ended up reverting as there wasn't any noticeable difference and the latter are more broadly used still.

TODO
- [x] Update relevant parts of the documentation and user guide

This is the second part of #49 and #58.

Commit 09965f5 resolves #61.


sebastiangrimberg:
Ping @hughcars for review.

CMakeLists.txt
```diff
@@ -52,8 +55,14 @@ if(NOT DEFINED BUILD_SHARED_LIBS)
   set(BUILD_SHARED_LIBS OFF CACHE BOOL "Global flag to cause add_library() to create shared libraries if ON")
 endif()
 
+# For now, SLEPc is always required
```
hughcars, line 58: Do you have a time in mind when this would be made optional again? If not, you could just drop the comment

cmake/ExternalGitTags.cmake
```diff
@@ -0,0 +1,239 @@
...
+set(EXTERN_PARMETIS_GIT_BRANCH
+  "main" CACHE STRING
+  "Git branch for external ParMETIS build"
+)
+set(EXTERN_PARMETIS_GIT_TAG
+  "8ee6a372ca703836f593e3c450ca903f04be14df" CACHE STRING  # 03/26/2023
+  "Git tag for external ParMETIS build"
+)
+
+# PETSc (for SLEPc)
```
hughcars, line 139: So the future here is that we won't be using the PETSc linear algebra aside from SLEPc? So no more `PetscParVector`?

cmake/ExternalMFEM.cmake
```diff
@@ -0,0 +1,252 @@
...
+  endif()
+  if(PALACE_WITH_SUPERLU)
+    list(APPEND MFEM_DEPENDENCIES superlu_dist)
+  endif()
+else()
+  set(MFEM_DEPENDENCIES)
+endif()
+if(PALACE_WITH_GSLIB)
+  list(APPEND MFEM_DEPENDENCIES gslib)
+endif()
```
hughcars, line 25: Why isn't this inside of the above if/else? Is GSLIB not an external dependency?

cmake/ExternalSLEPc.cmake
```diff
@@ -70,30 +72,49 @@ if(NOT "${BLAS_LAPACK_LIBRARIES}" STREQUAL "")
...
   TEST_BEFORE_INSTALL TRUE
 )
+
+
+# Configure SLEPc eigenvalue solver (most options come from PETSc)
+set(SLEPC_OPTIONS
+  "--prefix=${CMAKE_INSTALL_PREFIX}"
+  "--with-feast=0"
+  "--with-arpack=0"
+  # "--with-slepc4py=1
```
hughcars, line 99: debris?

palace/CMakeLists.txt
```diff
@@ -100,28 +58,58 @@ if(PALACE_WITH_OPENMP)
...
-  message(FATAL_ERROR "PETSc could not be found, be sure to set PETSC_DIR")
+find_package(MFEM REQUIRED CONFIG)
+message(STATUS "Found MFEM: ${MFEM_VERSION} in ${MFEM_DIR}")
+if(NOT MFEM_USE_MPI)
+  message(FATAL_ERROR "Build requires MFEM with MPI support")
 endif()
+# if(MFEM_CXX_FLAGS)
+#   # Pull compiler flags from MFEM for OpenMP and optimizations
+#   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${MFEM_CXX_FLAGS}")
+# endif()
```
hughcars, line 85: debris?

palace/linalg/petsc.cpp
```diff
@@ -5,7 +5,7 @@
 
 #include <petsc.h>
 #include <petscblaslapack.h>
-#include <general/forall.hpp>
+#include <mfem/general/forall.hpp>
 // #include "linalg/hypre.hpp"
```
hughcars, line 9: delete?

palace/linalg/strumpack.cpp
```diff
@@ -8,25 +8,64 @@
...
+    case config::LinearSolverData::CompressionType::ZFP:
+      return strumpack::CompressionType::LOSSY;
+    case config::LinearSolverData::CompressionType::BLR_HODLR:
+      return strumpack::CompressionType::BLR_HODLR;
+      break;
+    case config::LinearSolverData::CompressionType::ZFP_BLR_HODLR:
+      return strumpack::CompressionType::ZFP_BLR_HODLR;
+      break;
+    case config::LinearSolverData::CompressionType::NONE:
+    default:
```
hughcars, line 34: Generally for switching over enum `default:` should be avoided, as without it the compiler will warn in future if any unhandled enum cases are introduced.

hughcars (commented):
I'm not sure this is related to this PR, rather than just my computer, but building now takes a minimum of 3 attempts (Calling `make -j6`, when I have 8 processors) in order to complete the build. Triggers a lot of resource exhausted errors such as: 
```
[ 41%] Building C object CMakeFiles/HYPRE.dir/parcsr_ls/amg_hybrid.c.o
[ 42%] Building C object CMakeFiles/HYPRE.dir/parcsr_ls/aux_interp.c.o
[ 11%] Building C object libmetis/CMakeFiles/metis.dir/contig.c.o
[ 12%] Building C object libmetis/CMakeFiles/metis.dir/debug.c.o
[ 58%] Performing build step for 'slepc'
[ 14%] Building C object libmetis/CMakeFiles/metis.dir/fm.c.o
make[5]: *** read jobs pipe: Resource temporarily unavailable.  Stop.
make[5]: *** Waiting for unfinished jobs....
make[5]: *** read jobs pipe: Resource temporarily unavailable.  Stop.
make[5]: *** Waiting for unfinished jobs....
[ 42%] Building C object CMakeFiles/HYPRE.dir/parcsr_ls/F90_hypre_laplace.c.o
make[4]: *** [libmetis/CMakeFiles/metis.dir/all] Error 2
make[3]: *** [all] Error 2
make[2]: *** [extern/METIS-cmake/src/metis-stamp/metis-build] Error 2
make[1]: *** [extern/CMakeFiles/metis.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[4]: *** [CMakeFiles/HYPRE.dir/all] Error 2
make[3]: *** [all] Error 2
make[2]: *** [extern/hypre-cmake/src/hypre-stamp/hypre-build] Error 2
make[1]: *** [extern/CMakeFiles/hypre.dir/all] Error 2
gmake[5]: Entering directory '/Users/hughcars/AWS/palace/review_build/extern/slepc'
gmake[6]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
==========================================
Starting make run on c889f3baddb0.ant.amazon.com at Thu, 13 Jul 2023 09:39:07 -0400
Machine characteristics: Darwin c889f3baddb0.ant.amazon.com 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:52:24 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6000 arm64
-----------------------------------------
Using SLEPc directory: /Users/hughcars/AWS/palace/review_build/extern/slepc
Using PETSc directory: /Users/hughcars/AWS/palace/review_build
Using PETSc arch: installed-arch-darwin-c-debug-complex
```

I suspect this is because everything is building more in parallel than before, as even with that, the whole process appears to be much faster. So, don't think it's worth doing anything about, just a data point to observe.

CMakeLists.txt
sebastiangrimberg: In a following PR (very soon), the complex-valued linear algebra functionality is refactored to not use PETSc and thus SLEPc will become completely optional again and this will be removed. For now, we have refactored the build system to only install PETSc for SLEPc but this isn't reflected in the Palace source yet.

cmake/ExternalGitTags.cmake
sebastiangrimberg: That's right! See above.

cmake/ExternalMFEM.cmake
sebastiangrimberg: We always build our own GSLIB, similar to NekRS and other codes. This is the same way we handle MFEM, for example. Namely there's a patch applied to GSLIB to allow for dynamic linkage and better header installation organization.

cmake/ExternalSLEPc.cmake
sebastiangrimberg: I'll leave this in case it becomes useful at some point.

palace/CMakeLists.txt
sebastiangrimberg: Likewise, leaving this in case it becomes useful (this functionality is not new here).

palace/linalg/petsc.cpp
sebastiangrimberg: This file is going to get removed very shortly as the PETSc wrapper classes get removed so I'll leave it for now.

palace/linalg/strumpack.cpp
sebastiangrimberg: Good call, I'll address this.

sebastiangrimberg:
> Sorry for the tardiness. I'm not sure this is related to this PR, rather than just my computer, but building now takes a minimum of 3 attempts (Calling `make -j6`, when I have 8 processors) in order to complete the build.

This is interesting, I have seen these errors before (only on macOS, never Linux) but actually they seem to have gone away following the changes in this PR. One reason could be if I updated CMake which might be worth trying if that's where the issue is coming from. The other would be to try starting from scratch with a smaller `-j` value and see if it goes away?

palace/linalg/strumpack.cpp
sebastiangrimberg: Resolved throughout in [d8bc2f1](https://github.com/awslabs/palace/pull/56/commits/d8bc2f1d1d5ed87110a5cb64f723ace3e3519388).

sebastiangrimberg:
Intel MPI was updated from 2021.9 to 2021.10 in the last 12 hours and broke something in the PETSc configuration when 64-bit integers are used. See https://gitlab.com/petsc/petsc/-/merge_requests/6705.

Should be fixed in [df3788f](https://github.com/awslabs/palace/pull/56/commits/df3788f21ea7846700f4bf70071b4dd8dfd7bfe9).

PR #58, opened 2023-06-27
Draft: Partial-assembly development branch, part 1c

sebastiangrimberg:


cmake/ExternalLibCEED.cmake
```diff
@@ -0,0 +1,104 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
hughcars, line 1: `LibCEED` rather than `LIBCEED`? Guessing this inconsistency with the other `LIB` is CEED's doing.

cmake/ExternalMFEM.cmake
```diff
@@ -231,8 +249,9 @@ message(STATUS "MFEM_OPTIONS: ${MFEM_OPTIONS_PRINT}")
 set(MFEM_PATCH_FILES
   "${CMAKE_SOURCE_DIR}/extern/patch/mfem/patch_mesh_part.diff"
   "${CMAKE_SOURCE_DIR}/extern/patch/mfem/patch_mesh_vis.diff"
-  "${CMAKE_SOURCE_DIR}/extern/patch/mfem/patch_submesh.diff"
   "${CMAKE_SOURCE_DIR}/extern/patch/mfem/patch_direct_solvers.diff"
+  "${CMAKE_SOURCE_DIR}/extern/patch/mfem/patch_pa_prereq.diff"
+  "${CMAKE_SOURCE_DIR}/extern/patch/mfem/patch_pa_libceed.diff"
```
hughcars, line 254: Just for the record, which branches in mfem are these patches based on? Also, the submesh patch is gone because it's now on main?

palace/CMakeLists.txt
```diff
@@ -132,7 +132,7 @@ add_subdirectory(models)
 add_subdirectory(utils)
 
 # Add binary extension for build architecture
-if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64|arm")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|arm")
```
hughcars, line 135: Seeing a few of these ${} removals around the cmake files, looks like it's related to cmake's string/list expansion rules, out of interest why the change?

hughcars (approved):

cmake/ExternalMFEM.cmake
sebastiangrimberg: Yep ND and RT `SubMesh` support is in MFEM `master`.

Here the `pa_prereq` patch is based on `sjg/assembly-cleanup-prereqs` (diff from `master`) and `pa_libceed` is from there to `sjg/libceed-openmp-dev`.

palace/CMakeLists.txt
sebastiangrimberg: Just trying to be consistent across all the code. CMake `if` supports the first argument being a string or a variable so I've just expanded out all the strings for `STREQUAL` versus using the variable for `MATCHES`. In hindsight I think just always using the variable for `if` is better so I'll go make that change.

sebastiangrimberg: Fixed more completely in [aa70d1e](https://github.com/awslabs/palace/pull/58/commits/aa70d1e7c681917ab5f2e5858aa035a36dd3fb70).

PR #65, opened 2023-07-18
Add singularity container definition file

### LINKED ISSUE #60: Offer Singularity / Apptainer container
I suggest supporting Singularity (or nowadays also Apptainer) containers in addition to the Spack manager currently supported. At least I find Singularity more widely supported than Spack.

It seems [easy to containerise Spack projects](https://spack.readthedocs.io/en/latest/containers.html) such that a user can easily do it themselves, but would be nice a purpose-built definition file would be provided or built in CI.

nikosavola:
*Description of changes:*

Added definition file for compiling from source with all dependencies in an HPC-supported container based on CentOS 7.

I put the definition file in the root folder, can be moved if somewhere else makes more sense.

Closes #60


sebastiangrimberg:
This is very cool @nikosavola, and nice to see how simple it was to put together. A few comments:

- Is it simpler to just install Spack inside the Singularity build and use this to install Palace? This way you wouldn't need to, for example, build MPICH yourself.
- It would be nice to have an associated CI workflow which installs Singularity, builds the container, and runs the tests using the container build. Is this doable?
- Regarding directory structure, it might be nice to create a top-level `singularity` directory which contains `singularity.def` and perhaps a short README referring to the relevant sections of the documentation, for example. MFEM seems to set a good example for this for the case of Docker: https://github.com/CEED/MFEM/tree/master/config/docker.

nikosavola:
> This is very cool @nikosavola, and nice to see how simple it was to put together. A few comments:
> 
> * Is it simpler to just install Spack inside the Singularity build and use this to install Palace? This way you wouldn't need to, for example, build MPICH yourself.
> * It would be nice to have an associated CI workflow which installs Singularity, builds the container, and runs the tests using the container build. Is this doable?
> * Regarding directory structure, it might be nice to create a top-level `singularity` directory which contains `singularity.def` and perhaps a short README referring to the relevant sections of the documentation, for example. MFEM seems to set a good example for this for the case of Docker: https://github.com/CEED/MFEM/tree/master/config/docker.

Installing with spack would result in less lines, but I think it's worthwhile to have a documented install without a package manager. I don't have much experience with spack but this should be more flexible and when it's documented manually, it should be easy for a user to change the compiler or math library for example.

The CI workflow makes sense and should be relatively easy, maybe as another issue?

I'll move the definition file to the new folder.

sebastiangrimberg:
Makes sense regarding Spack, I'm fine with that choice.

For the CI, I would just add the new workflow in this PR. This way we can test to make sure the container is being built correctly. If you want to get started, you can base it off of the `spack.yml` workflow and I can help if you get stuck. It seems like https://github.com/singularityhub/github-ci might be a good reference. Thanks a lot for looking at this!

nikosavola:
I now added that the Singularity container is built and tested in CI. Additionally,
* an older version of MPICH is used by default for less compiling (especially in the CI). The newer version can still be compiled by providing an env to `singularity build`.
* Julia is installed in the container (mainly for testing)
* The built Singularity image is pushed to GitHub registry when pushing to main or creating a tag. It would be visible [here](https://github.com/orgs/awslabs/packages?repo_name=palace). The image is still built and tested in PRs but not pushed.

I think these should work now but requires some reviewing to do.

sebastiangrimberg:
Thanks for this @nikosavola, great to see you got the CI test working as well. I will review this this weekend/next week!

sebastiangrimberg:
Hi @nikosavola, I've pushed a branch `sjg/singularity-container-updates` which contains a few proposed changes on top of this PR. Most if not all are totally minor and just for consistency with other scripts in the repo as well as clarifying some documentation. If you approve of these, feel free to fast-foward merge them onto your PR here. Otherwise, happy to discuss if you see anything that you don't agree with.

The only other comment I have is regarding the Docker image you use. As you probably know, CentOS and the Docker CentOS image are deprecated and likewise CentOS 7 is quite old (released almost 10 years ago). It might be nice to switch to a different image, might I suggest Amazon Linux 2023: https://hub.docker.com/_/amazonlinux. This will also enable you to avoid installing newer versions of packages from source, like CMake, for example.

nikosavola:
Hi @sebastiangrimberg, I now merged your suggestions, they make sense.

CentOS7 is indeed old, I should check out Rocky Linux at some point if it is a good replacement for HPC-applications and has a decent sized community.

nikosavola:
Looks like it is failing, I can take a look

sebastiangrimberg:
Thanks for the catch in [5e2adce](https://github.com/awslabs/palace/pull/65/commits/5e2adce70b251f2e0a7c8316ba06a9d56c5b2328), sorry for messing that up yesterday. The pipeline should be fixed now after fixing an annoying Intel bug in [9242ab0](https://github.com/awslabs/palace/pull/65/commits/9242ab022437560cfd2fb336d0df2e09f85e668c).

In my opinion I think it is worth switching from CentOS 7 to another more recent OS image before merging this PR. It should not be too bad and should just involve updating to the appropriate package manager command and package names. I can even take a stab at this if you'd like, just let me know.

nikosavola:
Hi, you can look at switching the OS if you're interested, works for me. I see that Amazon Linux makes sense for an Amazon project but Rocky Linux looks cool as well.

sebastiangrimberg:
OK! Hi @nikosavola, I've updated the build to use Rocky Linux 9 and made some other improvements to the workflow. Namely:

- The container now builds with the *local* Palace source (from the PR) rather than cloning and building off of `main`. If the user does not mount the directory containing the local source code, then the definition file will clone from `main` like before.
- The test workflow is handled outside of the container completely which simplifies things (no need for the container to know anything about Julia, just needs to provide the `palace` executable).

Let me know how these changes look to you, and perhaps test things locally if you can. Once I get your approval I think this is ready to merge.

nikosavola:
I tested that it can be built locally and runs. Looks good to me, can be merged!

PR #68, opened 2023-07-18
Update mesh files for cavity example

sebastiangrimberg:
- Add tetrahedral, prism, and hexahedral meshes for the cavity resonator problem.
- Update convergence study results for new cavity example meshes.
- Add metal thickness meshes for CPW example, and update results for new meshes.
- Updates and fixes for all example Julia code, including improvements for mesh generation code, and fixes for errors when trying to run plot generation code.

This depends on https://github.com/awslabs/palace/pull/197, which handles the boundary conditions correctly for wave ports which are directly neighboring each other in the mesh.


sebastiangrimberg:
@hughcars, this is ready for review now (finally). I've updated the cavity convergence data for the new meshes as we talked about. I have also expanded the scope to include updates to the CPW example (generate meshes with metal thickness), as well as some broader fixes encountered when trying to use the Julia wrappers to regenerate meshes/plots.

test/examples/runtests.jl
```diff
@@ -49,7 +49,7 @@ if "spheres" in cases
         np=numprocs,
         rtol=reltol,
         atol=abstol,
-        excluded_columns=["Maximum", "Minimum"]
```
hughcars, line 52: The number of elements ought to be large enough that the mean should be stable if the norm is stable, so this is a bit surprising. Roughly how big were the misses that meant `Mean` needed to be added?

docs/src/examples/cpw.md
```diff
@@ -101,11 +105,10 @@ sweeps.
 
 Second, there is a discrepancy between the results using lumped ports and those with wave
 ports, namely the lumped port excitation exhibits much higher reflection that that for wave
```
hughcars, line 107: that that

examples/cavity/cavity.jl
```diff
@@ -44,57 +44,59 @@ function solve_cavity_resonator(
...
 )
     @assert refinement >= 0
     @assert order > 0
     @assert geo_order > 0
     @assert mesh_type ∈ [0, 1, 2]
 
     cavity_dir = @__DIR__
-    fileroot = string("cavity_p", order, "_h", refinement)
+    file_root = string("cavity_p", order, "_h", refinement)
+    cd(cavity_dir)
```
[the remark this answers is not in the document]
sebastiangrimberg: Agreed. Let me see what I can do to improve this.

test/examples/runtests.jl
sebastiangrimberg: Here's an example from the `cavity_impedance.json` case:

```
Regression test error at row 1, column Mean: 0.0001699570272 ≉ 0.0001700702508
```

Very small errors, but there are only 12 elements in the mesh (96 after refinement) and the eigenvalue solve is particularly sensitive to very small differences between repeated runs. I will just exclude `"Mean"` from the cavity examples instead of adding it to the exclusion list for all examples. Does that sound OK?

sebastiangrimberg:
The new CPW example meshes revealed another issue with wave ports, addressed in [8ea655f](https://github.com/awslabs/palace/pull/68/commits/8ea655f77a524f0c46f58da3cc863bbf8e220192). Basically the computed eigenvectors from ARPACK and SLEPc differed by a phase `e^{i\phi}`, while still satisfying the power normalization constraint. This commit correctly implements a phase constraint, such that for the given direction vector `t`, the normalized eigenvector satisfies `t^T x > 0` with `t^T x` real-valued (by scaling `x` by `t^T x / |t^T x|`).

examples/cavity/cavity.jl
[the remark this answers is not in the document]
sebastiangrimberg: Fixed in [3b1375f](https://github.com/awslabs/palace/pull/68/commits/3b1375fa08d4223cce9c6ca723e42a2c00cb898a)

test/examples/runtests.jl
sebastiangrimberg: Resolved in [8c7b4b6](https://github.com/awslabs/palace/pull/68/commits/8c7b4b61b721c84d30904f93945c2a2d3ef33a6d) by just tightening the linear solver tolerance.

sebastiangrimberg: Closing the loop here:

This has nothing to do with linear solver tolerance/mesh resolution. The cylindrical cavity resonator modes have a rotational invariance; different eigenmode solvers may converge to a mode which has a rotation about the cylinder center axis by some nonzero angle. In this case, the global norm of the error esimate will still be the same, but the elementwise contributions will not agree. Thus, it's correct to ignore the `"Mean"` field (along with `"Min"` and `"Max"`), and we were probably just getting lucky earlier with the tests passing.

An alternative approach would be to simulate only half the cylinder, with a symmetry plane (PEC or PMC depending on the modes sought). This removes the rotational symmetry and is an option tested in [cbcb4dd](https://github.com/awslabs/palace/pull/68/commits/cbcb4ddb5d0040858bfe989bd54caa1c593258f0).

PR #76, opened 2023-07-25
Fix configuration file parsing for obscure `std::regex` bug

sebastiangrimberg:
For configuration files with large lists of boundary or domain attributes (for example, some 20k entries in one case), the range expansion step of configuration file parsing encounters a segmentation fault. This only shows up with GCC (not Clang) and appears to be a C++ STL bug. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66456 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61582 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86164.

This PR rewrites this step of the parser to avoid use of `std::regex` instead.

### DIFF
--- a/palace/utils/iodata.cpp
+++ b/palace/utils/iodata.cpp
@@ -71,7 +71,10 @@ std::stringstream PreprocessFile(const char *filename)
     // Handle the given string which is only numeric with possible hyphens.
     int num;
     auto [ptr, ec] = std::from_chars(str.data(), str.data() + str.length(), num);
-    MFEM_VERIFY(ptr != str.data(), "Invalid integer conversion in range expansion!");
+    MFEM_VERIFY(
+        ec == std::errc(),
+        "Invalid integer conversion in range expansion"
+            << (ec == std::errc::result_out_of_range ? " (integer out of range)!" : "!"));
     if (ptr == str.data() + str.length())
     {
       return std::string(str);
@@ -79,7 +82,10 @@ std::stringstream PreprocessFile(const char *filename)
     // Range specified, expand the bounds.
     int num2;
     auto [ptr2, ec2] = std::from_chars(ptr + 1, str.data() + str.length(), num2);
-    MFEM_VERIFY(ptr2 != ptr + 1, "Invalid integer conversion in range expansion!");
+    MFEM_VERIFY(
+        ec2 == std::errc(),
+        "Invalid integer conversion in range expansion"
+            << (ec2 == std::errc::result_out_of_range ? " (integer out of range)!" : "!"));
     std::string rng;
     while (num < num2)
     {
--- a/palace/utils/iodata.cpp
+++ b/palace/utils/iodata.cpp
@@ -3,11 +3,13 @@
 
 #include "iodata.hpp"
 
+#include <charconv>
 #include <fstream>
 #include <functional>
 #include <iostream>
 #include <regex>
 #include <stack>
+#include <string_view>
 #include <mfem.hpp>
 #include <nlohmann/json.hpp>
 #include "utils/communication.hpp"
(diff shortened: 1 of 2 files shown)


palace/utils/iodata.cpp
```diff
@@ -16,14 +20,131 @@
...
+
+  // Perform integer range expansion for arrays ([a - b, c] = [a-b,c] =
+  // [a,a+1,...,b-1,b,c]). The whole file is now one line and arrays have no spaces after
+  // whitespace stripping.
+  std::stringstream output;
+  auto RangeExpand = [](std::string_view str) -> std::string
+  {
+    // Handle the given string which is only numeric with possible hyphens.
+    int num;
+    auto [ptr, ec] = std::from_chars(str.data(), str.data() + str.length(), num);
```
hughcars, line 73: Looking at https://en.cppreference.com/w/cpp/utility/from_chars I think the `MFEM_VERIFY` could make use of the `ec` instead
```
MFEM_VERIFY(ec == std::errc(), "Invalid integer conversion in range expansion!");
```
this would also check for overflowing integer values, (which hopefully is a theoretical issue, but who knows).

palace/utils/iodata.cpp
```diff
@@ -16,14 +20,131 @@
...
+    MFEM_VERIFY(ptr != str.data(), "Invalid integer conversion in range expansion!");
+    if (ptr == str.data() + str.length())
+    {
+      return std::string(str);
+    }
+    // Range specified, expand the bounds.
+    int num2;
+    auto [ptr2, ec2] = std::from_chars(ptr + 1, str.data() + str.length(), num2);
+    MFEM_VERIFY(ptr2 != ptr + 1, "Invalid integer conversion in range expansion!");
```
hughcars, line 82: Add similar checks on `ec2` from above here, e.g. if this were `1-5000000000` I think this would bug in a weird way.

palace/utils/iodata.cpp
```diff
@@ -16,14 +20,131 @@
...
+      else if (*it == '[')
+      {
+        output << std::string(start, it);
+        start = it;
+        inside = true;
+      }
+    }
+    output << std::string(start, file.end());
+  }
+  return output;
```
hughcars, line 136: Beyond the scope of this PR but storing large strings of comma separated integers, only to parse them later on into vectors of `int` is a bit roundabout. I think for numbers less than or equal to 6 digits this comes out ahead (digits + space + ,) compared to the 8 bytes of an integer. I think this is still the best approach though, as otherwise this would get way more complicated, so any perf hit here is moot really.

hughcars (commented):
Mostly trivial/comments, the only thing of interest would be using the error codes from `std::from_chars`.

palace/utils/iodata.cpp
sebastiangrimberg: Fixed in [1325511](https://github.com/awslabs/palace/pull/76/commits/13255116fa34de49bb39efbb4cd74c35c0c7a95f).

palace/utils/iodata.cpp
```diff
@@ -16,14 +20,131 @@
...
+    std::regex rgx(R"((([\"'])(?:(?=(\\?))\3.)*?\2))"
+                   R"(|,+(?=\s*?[\}\]]))");
+    file = std::regex_replace(file, rgx, "$1");
+  }
+
+  // Perform integer range expansion for arrays ([a - b, c] = [a-b,c] =
+  // [a,a+1,...,b-1,b,c]). The whole file is now one line and arrays have no spaces after
+  // whitespace stripping.
+  std::stringstream output;
+  auto RangeExpand = [](std::string_view str) -> std::string
```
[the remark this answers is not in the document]
sebastiangrimberg: Will leave it in the name of improved readability.

palace/utils/iodata.cpp
sebastiangrimberg: The JSON spec requires the arrays to be comma separated lists of integers, so there isn't really a way around this before passing to the JSON parser.

PR #77, opened 2023-07-25
Fixing a range of compiler warnings for shadow/unused variables

hughcars:
*Description of changes:*
- Turned on some extra warnings (Wshadow) whilst hunting a bug on another branch, there were a few extra errors here that produced some noise. This PR resolves them, I don't believe there were any correctness issues, due to type resolution picking the correct variable in many cases. 
- The change to SurfacePostOperator is the most significant. The substructs were in error marking the GetCoefficient arguments as shadowing `mat_op` and `local_to_shared`. Given these subtypes were already being leaked through the accessor methods for their containers, I just moved the types in to the palace namespace. 
- In a few other spots, the ambiguity was resolved by switching to no argument or static methods, and in others by changing a local variable name.


hughcars:
Closed as the code base has moved so much underneath that all these have been addressed.

Issue #2, opened 2023-01-23
Adaptive mesh refinement


sebastiangrimberg:
Implement `solve-estimate-mark-refine` adaptive mesh refinement (AMR) process for all simulation types, as well as coarsening and load-balancing steps.

Use MFEM's functionality to support both conforming refinement for simplex meshes and nonconfirming refinement for meshes with hexahedra/prisms/pyramids. Requires AMR support for high-order H1, H(curl), and H(div) finite element spaces.

hughcars:
Various ongoing work within MFEM to make this happen:
* https://github.com/mfem/mfem/pull/3611
* https://github.com/mfem/mfem/pull/3587
* https://github.com/mfem/mfem/pull/3541

PR #84, opened 2023-08-21
[Draft] Modify mesh IO to support nonconformal mesh initialization

hughcars:
- The partitioning is done using the MeshPartitioner where possible, but for nonconformal meshes will now use the standard ParMesh constructor.
- Change to use Mesh& where a parameter out value is not allowed to be invalidated
- Move #if 0 to behind constexpr false to ensure can continue to compile
- Introduces a draft amr config file interface, behind a feature barrier, as this is used for determining whether `ParMesh` or `MeshPartitioner` is used.

*Issue #, if available:*
https://github.com/awslabs/palace/issues/2

### DIFF
--- a/palace/utils/configfile.cpp
+++ b/palace/utils/configfile.cpp
@@ -397,10 +397,61 @@ void RefinementData::SetUp(json &model)
     }
   }
 
+  auto adapt = refinement->find("Adaptation");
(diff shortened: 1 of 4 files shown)


hughcars:
Being rolled into other PR.

PR #85, opened 2023-08-21
Introduce Flux Error Estimation

hughcars:
Introduce flux error estimation based on a global ZZ solve.

- Introduces the FluxProjector class, which is used for computing a "coarse flux" in
a vector L2 space, then performing a global mass matrix inversion to compute a
"smooth" flux that better represents the underlying physical quantity. This is achieved
through a multigrid hierarchy.
- Introduces the CurlFluxErrorEstimator and GradFluxErrorEstimator classes, which utilize the
FluxProjector class to compute elemental error estimates comparing the coarse numerical flux
from the solution against a smoothed flux from a global mass matrix inversion.
- Introduces the CurlFluxErrorEstimator class which utilizes a flux projector to compute a
smooth flux in an ND space. Provides methods for real valued vectors, and complex vectors.
The flux is assumed to take the form μ⁻¹∇ × V, where V is either A or E depending on the solver mode.
- Introduces the GradFluxErrorEstimator class which utilizes a flux projector to compute a
smooth flux in an H1^d space. The flux is assumed to be of the form ϵ ∇ ϕ
- Introduce the CurlFluxCoefficient and GradFluxCoefficient classes which compute the numerical Curl
and Grad fluxes given a solution.

*Issue #, if available:*
https://github.com/awslabs/palace/issues/2


palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+  }
+  return {estimates, normalization};
+}
+
+GradFluxErrorEstimator::GradFluxErrorEstimator(
+    const IoData &iodata, const MaterialOperator &mat_op,
+    const std::vector<std::unique_ptr<mfem::ParMesh>> &mesh,
+    mfem::ParFiniteElementSpace &fes)
+  : mat_op(mat_op), fes(fes),
+    smooth_flux_fecs(ConstructFECollections<mfem::H1_FECollection>(
```
sebastiangrimberg, line 250: As above.

palace/fem/multigrid.hpp
```diff
@@ -131,37 +131,27 @@ std::vector<std::unique_ptr<FECollection>> inline ConstructFECollections(
...
-// marked.
+// finite element collections. Uses geometric multigrid and p multigrid.
 template <typename FECollection>
 inline mfem::ParFiniteElementSpaceHierarchy ConstructFiniteElementSpaceHierarchy(
     int mg_max_levels, bool mg_legacy_transfer, int pa_order_threshold,
     const std::vector<std::unique_ptr<mfem::ParMesh>> &mesh,
-    const std::vector<std::unique_ptr<FECollection>> &fecs,
-    const mfem::Array<int> *dbc_marker = nullptr,
-    std::vector<mfem::Array<int>> *dbc_tdof_lists = nullptr)
+    const std::vector<std::unique_ptr<FECollection>> &fecs, int dim = 1)
```
sebastiangrimberg, line 139: Since we never use this to construct vector FE spaces, the `dim` argument can be removed. Further I'm against all of these changes as I've stated previously, since the previous code was fine. Optional arguments for the essential dofs are perfectly acceptable and I don't see how these changes improve things. All the changes to this file can be reverted for this PR.

palace/linalg/fluxprojector.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+  // Intermediate storage vector used in Mult.
+  mutable Vector tmp;
+
+public:
+  FluxProjector(mfem::ParFiniteElementSpaceHierarchy &smooth_flux_fes, double tol = 1e-12,
+                int max_it = 200, int print_level = 1, int pa_order_threshold = 1);
+
+  // Given a vector of dof defining the flux, compute the smooth flux IN PLACE.
+  void Mult(Vector &x) const
```
sebastiangrimberg, line 40: Seems like the in-place projection isn't ever used, so I'd remove it entirely.

palace/linalg/errorestimator.hpp
```diff
@@ -0,0 +1,91 @@
...
+  std::vector<mfem::DenseMatrix> smooth_to_coarse_embed;
+
+public:
+  // Constructor for using geometric and p multigrid.
+  CurlFluxErrorEstimator(const IoData &iodata, const MaterialOperator &mat_op,
+                         const std::vector<std::unique_ptr<mfem::ParMesh>> &mesh,
+                         mfem::ParFiniteElementSpace &fes);
+
+  // Compute elemental error indicators given a complex vector of true DOF.
+  IndicatorsAndNormalization operator()(const ComplexVector &v,
```
sebastiangrimberg, line 49: I don't see the downside of giving these functions names so when you call them it is more clear what's going on.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+    scalar_mass_matrices(fes.GetNE()), smooth_to_coarse_embed(fes.GetNE())
+{
+  mfem::MassIntegrator mass_integrator;
+  for (int e = 0; e < fes.GetNE(); e++)
+  {
+    // Loop over each element, and save an elemental mass matrix.
+    // Will exploit the fact that vector L2 mass matrix components are independent.
+    const auto &coarse_fe = *coarse_flux_fes.GetFE(e);
+    auto &T = *coarse_flux_fes.GetElementTransformation(e);
+    mass_integrator.AssembleElementMatrix(coarse_fe, T, scalar_mass_matrices[e]);
```
sebastiangrimberg, line 43: This operation is expensive and is incompatible with partial assembly. There must be a way to avoid this? I can think about it some more as well. Representing this integration and the below projection as partially assembled operators should be possible.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+    const auto &smooth_fe = *smooth_flux_fes.GetFinestFESpace().GetFE(e);
+    coarse_fe.Project(smooth_fe, T, smooth_to_coarse_embed[e]);
+  }
+}
+
+IndicatorsAndNormalization CurlFluxErrorEstimator::operator()(const ComplexVector &v,
```
sebastiangrimberg, line 50: Seems like this is a generalization of the real-valued case. Can this be templated and the functions combined? This should be possible, see for example most of the solvers in `linalg/`.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+  const auto smooth_flux_rhs =
+      ComplexVector(rhs_from_coef(smooth_flux_fes.GetFinestFESpace(), real_coef),
+                    rhs_from_coef(smooth_flux_fes.GetFinestFESpace(), imag_coef));
+
+  // Given the RHS vector of non-smooth flux, construct a flux projector and perform mass
+  // matrix inversion in the appropriate space, giving f = M⁻¹ f̂.
+  auto build_flux = [](const FluxProjector &proj, const ComplexVector &flux_coef)
+  {
+    // Use a copy construction to match appropriate size.
+    ComplexVector flux(flux_coef);
```
sebastiangrimberg, line 82: Ideally this vector can be allocated once for all as a class member.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+    proj.Mult(flux_coef, flux);
+    return flux;
+  };
+  auto smooth_flux = build_flux(smooth_projector, smooth_flux_rhs);
+
+  // Given a complex solution represented with a PetscParVector, build a ComplexGridFunction
```
sebastiangrimberg, line 88: Maybe worth a `ctrl-f` for all `PetscParVector` to catch anything you've missed.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+  // for evaluation.
+  auto build_func = [](const ComplexVector &f, mfem::ParFiniteElementSpace &fes)
+  {
+    mfem::ParComplexGridFunction flux(&fes);
```
sebastiangrimberg, line 92: Again, this is another vector allocation that can be allocated once for all solves.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+  Vector smooth_vec, coarse_vec, sub_vec, estimates(nelem);
+  estimates = 0.0;
+  double normalization = 0.0;
+  for (int e = 0; e < fes.GetNE(); e++)
+  {
+    // real
+    smooth_flux_func.real().GetElementDofValues(e, smooth_vec);
+    coarse_flux_func.real().GetElementDofValues(e, coarse_vec);
+
+    const int ndof = coarse_vec.Size() / 3;
```
sebastiangrimberg, line 115: In the interest of eventual generality, all of these `3` magic numbers should be `mesh.SpaceDimension()`.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+  mfem::ParComplexGridFunction coarse_flux_func(&coarse_flux_fes);
+  coarse_flux_func.real().ProjectCoefficient(real_coef);
+  coarse_flux_func.imag().ProjectCoefficient(imag_coef);
+
+  // Loop over elements, embed the smooth flux into the coarse flux space, then compute
+  // squared integral using a component-wise mass matrix.
+  Vector smooth_vec, coarse_vec, sub_vec, estimates(nelem);
+  estimates = 0.0;
+  double normalization = 0.0;
+  for (int e = 0; e < fes.GetNE(); e++)
```
sebastiangrimberg, line 109: In thinking about the cost of allocating `scalar_mass_matrices` and `smooth_to_coarse_embed` above, I'm confused about this loop. Why is this not done matrix free by just looping over elements, looping over integration points on each element, and using `GridFunction::GetValue` or `GetVectorValue` to reconstruct the two quantities? This is essentially what is already happening when you call `coarse_flux_func.real().ProjectCoefficient(real_coef);` above, so you are doing double work here. See `GridFunction::ComputeElementLpErrors`.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,395 @@
...
+    Vector flux(rhs.Size());
+    flux = 0.0;
+
+    // Apply the flux projector component wise.
+    const int ndof = flux.Size();
+    const int stride = ndof / 3;
+    MFEM_ASSERT(ndof % 3 == 0, "!");
+
+    Vector flux_comp, rhs_comp;
+    for (int i = 0; i < 3; i++)
```
sebastiangrimberg, line 317: As you are already only solving the H1^d equations block-wise, there's no need to actually evaluate the H1^d finite element space and you should be able to reuse the existing scalar space. When constructing the RHS vector, you can make the `GradFluxCoefficient` inherit from `mfem::Coefficient` instead of `VectorCoefficient`, and take as argument to the constructor the component you want to compute, for example, x/y/z, and generate three RHS vectors on the scalar space.

sebastiangrimberg (changes requested):
Looking good @hughcars, I went ahead and made an initial pass as the tests are passing now.

Several style and organizational comments, but one large issue which came up towards the end in the `linalg/errorestimator.cpp` classes. I see you are allocating a `FiniteElementSpace` (L2) for the discontinuous flux, and constructing these `scalar_mass_matrices` and `smooth_to_coarse_embed` objects to perform the estimation. For reasons noted above, this is something to avoid and I think the loops over elements once the smooth flux `GridFunction` is computed needs to be reworked. This can be done fully matrix-free, as `mfem::GridFunction::ComputeElementLpErrors`. There may be good reason you have decided to go your current route vs. the other implementation, but maybe this was done prematurely. I would imagine it looking something like:

```
... compute mfem::ParGridFunction smooth_flux, and using CurlFluxCoefficient flux_coef ...

for e in 0, ... , NE-1:
    T = mesh.GetElementTransformation(e);
    ir = mfem::IntRules.Get(T.GetGeometryType(), q_order);
    for ip in ir.GetNPoints():
        T.SetIntPoint(ip);
        U = smooth_flux.GetVectorValue(T, ip);
        V = flux_coef.Eval(T, ip);

        // Integrate over the element: \int_e (U-V)^T (U-V) dV = \sum_i w_i ||U_i - V_i||_2^2
        // (w_i = ip.weight * T->Weight(), see GridFunction::ComputeLpError)

        // For the normalization, use \int_e ||U||_2^2 dV = sum_i w_i ||U_i||^2

    end
end
```

The loop over integration points can use dense matrix arithmetic using the overrides to `GetVectorValue` and `Eval` which return a `DenseMatrix` where each column is the vector at an integration point. I am fairly certain this approach is an improvement in the long run over what you have there, as you avoid the big memory allocations associated with the dense element matrices.

palace/drivers/basesolver.hpp
```diff
@@ -68,12 +67,22 @@ class BaseSolver
...
+                                  const ErrorIndicators &indicators) const;
+
 public:
   BaseSolver(const IoData &iodata, bool root, int size = 0, int num_thread = 0,
              const char *git_tag = nullptr);
   virtual ~BaseSolver() = default;
 
-  virtual void Solve(std::vector<std::unique_ptr<mfem::ParMesh>> &mesh) const = 0;
+  // Performs a solve using the mesh sequence, and report error indicators.
+  virtual ErrorIndicators
```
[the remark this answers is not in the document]
hughcars: I'd rather not, as there's no possibility of this ever being a null state, so pointers adds extra verbosity without a bought feature. It even suggests that any call to `Solve` should be checking for null status before accessing the variable. Returning by value guarantees that the variable is a valid return. `ErrorIndicators` also has move semantics, so even without NRVO, which would almost certainly trigger, this would end up being equivalent to a pointer return.

test/testcase.jl
```diff
@@ -68,7 +92,22 @@ function testcase(
...
 
-            test = isapprox.(data, dataref; rtol=rtol, atol=atol)
+            # Check the number of columns matches, before removing any excluded_columns
+            @test ncol(data) == ncol(dataref)
+            for col ∈ excluded_columns
+                select!(data, Not(Cols(contains(col))))
+                select!(dataref, Not(Cols(contains(col))))
+            end
+
+            fdata = asfloat.(data)
```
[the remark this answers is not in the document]
hughcars: The `error-indicators.csv` file writes a final row with the mean value. This quantity is what is actually being used to drive the adaptation, and combines all the proceeding rows. Rather than apply a special meaning to maybe `-1` or the like, I opted for making machine parsing the csv more complicated as a human reading it would find it much more intuitive.

palace/drivers/basesolver.hpp
```diff
@@ -68,12 +67,22 @@ class BaseSolver
                          double time) const;
   void PostprocessFields(const PostOperator &postop, int step, double time) const;
 
+  // Postprocess granular error indicator to file. The argument normalized indicates if
+  // supplied indicators have already been normalized.
+  void PostprocessErrorIndicators(const std::string &name, int step, double time,
```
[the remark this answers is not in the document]
hughcars: Without assigning a special value meaning to "step", there isn't enough information entering to differentiate a row led by `time` vs a string value `"Mean"`. Another option to the overload is an extra argument to ignore the step value, but I opted against that as it makes both calls more complicated compared to having the compiler pick between the overloads.

I want it to report the granular estimates given we are calculating them anyway (for instance to say which mode or frequency is contributing most to a given adaptation), but also want to report the quantity used to drive the adaptation. In the revision there are only doubles going in, to avoid passing a communicator, so there's even less information.

palace/utils/timer.hpp
```diff
@@ -28,10 +28,12 @@ class Timer
   {
     INIT = 0,
     CONSTRUCT,
-    WAVEPORT,  // Wave port solver
+    WAVEPORT,      // Wave port solver
+    ESTCONSTRUCT,  // Construction of estimate operator
```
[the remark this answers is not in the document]
hughcars: Broken out into an `Estimation` heading for now. AMR PR will add further headings for marking and refinement, at which point estimation can be a subheading.

palace/drivers/drivensolver.cpp
```diff
@@ -93,20 +96,24 @@ void DrivenSolver::Solve(std::vector<std::unique_ptr<mfem::ParMesh>> &mesh) cons
     MFEM_VERIFY(excitations > 0, "No excitation specified for driven simulation!");
   }
   Mpi::Print("\n");
+  // Initalize the error estimation operator.
+  auto estimator = [&]()
+  {
+    BlockTimer bt(Timer::ESTCONSTRUCT);
+    return CurlFluxErrorEstimator(iodata, spaceop.GetMaterialOp(), mesh,
```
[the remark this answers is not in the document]
hughcars: That then means passing `IoData` through into the sweep calls, as it is needed to construct the `CurlFluxErrorEstimate`, either approach is fine, but this way avoids duplicating this construction.

palace/fem/coefficient.hpp
```diff
@@ -477,6 +477,59 @@ inline double DielectricInterfaceCoefficient<DielectricInterfaceType::DEFAULT>::
   return 0.5 * ts * epsilon * (V * V);
 }
 
+// Computes the flux, μ⁻¹ ∇ × X, of a field, X, where X can be the electric field E, or the
+// magnetic vector potential A.
+class CurlFluxCoefficient : public mfem::VectorCoefficient
+{
+private:
+  const mfem::ParGridFunction &X;
```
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/8abe8b52ba5c72f35621d5177241e080114b4402

palace/fem/coefficient.hpp
```diff
@@ -477,6 +477,59 @@ inline double DielectricInterfaceCoefficient<DielectricInterfaceType::DEFAULT>::
...
+  MaterialPropertyCoefficient<MaterialPropertyType::INV_PERMEABILITY> coef;
+  mfem::DenseMatrix muinv;
+  mfem::Vector curl;
+
+public:
+  CurlFluxCoefficient(const mfem::ParGridFunction &pgf, const MaterialOperator &op)
```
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/8abe8b52ba5c72f35621d5177241e080114b4402

palace/fem/coefficient.hpp
```diff
@@ -477,6 +477,59 @@ inline double DielectricInterfaceCoefficient<DielectricInterfaceType::DEFAULT>::
   return 0.5 * ts * epsilon * (V * V);
 }
 
+// Computes the flux, μ⁻¹ ∇ × X, of a field, X, where X can be the electric field E, or the
+// magnetic vector potential A.
+class CurlFluxCoefficient : public mfem::VectorCoefficient
+{
+private:
+  const mfem::ParGridFunction &X;
+  MaterialPropertyCoefficient<MaterialPropertyType::INV_PERMEABILITY> coef;
```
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/8abe8b52ba5c72f35621d5177241e080114b4402

palace/drivers/basesolver.hpp
```diff
@@ -11,15 +11,14 @@
 
 namespace mfem
 {
-
```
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/8abe8b52ba5c72f35621d5177241e080114b4402

palace/drivers/basesolver.hpp
```diff
@@ -68,12 +67,22 @@ class BaseSolver
...
+  void PostprocessErrorIndicators(const std::string &name,
+                                  const ErrorIndicators &indicators) const;
+
 public:
   BaseSolver(const IoData &iodata, bool root, int size = 0, int num_thread = 0,
              const char *git_tag = nullptr);
   virtual ~BaseSolver() = default;
 
-  virtual void Solve(std::vector<std::unique_ptr<mfem::ParMesh>> &mesh) const = 0;
+  // Performs a solve using the mesh sequence, and report error indicators.
```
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/8abe8b52ba5c72f35621d5177241e080114b4402

palace/utils/errorindicators.hpp
```diff
@@ -0,0 +1,95 @@
...
+  MPI_Comm comm;
+  // Statistics, updated simultaneously with the global error indicator.
+  double min, max, mean;
+  // Mean normalization constant.
+  double normalization;
+};
+
+// Operator for performing reduction of a vector of local indicators into a
+// global running total for use in adaptation.
+class ErrorReductionOperator
```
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/3051448c5f36492ae25ba2d991043c44acdd0cf2

palace/linalg/fluxprojector.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "linalg/iterative.hpp"
+#include "linalg/rap.hpp"
+#include "models/materialoperator.hpp"
+
+namespace palace
+{
+
+// Given a finite element space hierarchy, construct a vector of mass matrix
+// operators corresponding to each level.
+std::unique_ptr<Operator> BuildMassMatrixOperator(mfem::ParFiniteElementSpaceHierarchy &h,
```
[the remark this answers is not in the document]
hughcars: I changed the constructor body to better match, but the `BuildMassMatrixOperator` helper function I felt improves understanding. The function acts to compartmentalize the construction, the developer knows exactly what's needed to build it, without having to check each line to check for other variables, or the state of `this`. 

https://github.com/awslabs/palace/pull/85/commits/aed614d73e2e175a0b48c3fa0137eaa391f898a2

palace/linalg/fluxprojector.cpp
```diff
@@ -0,0 +1,83 @@
...
+                                                  int pa_order_threshold)
+{
+  constexpr int skip_zeros = 0;
+  const bool is_scalar_FE_space =
+      h.GetFESpaceAtLevel(0).GetFE(0)->GetRangeType() == mfem::FiniteElement::SCALAR;
```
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/a33bd0d93d0d91a1ca0db36a5455f2c40311ebdf

test/testcase.jl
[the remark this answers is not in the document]
sebastiangrimberg: I do not think we should make changes which change the data type of the output CSV files. To me the simplest solution is to remove this row, since it's easily computable from the other data in the file.

palace/drivers/basesolver.hpp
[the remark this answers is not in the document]
sebastiangrimberg: Seems like the simplification suggested above, of just not reporting the mean, keeps the structure of the output files the same and resolves this issue.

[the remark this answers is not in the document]
hughcars: It also means that all the log files never actually contain information about the actual combined quantity being used to drive the adaptation. I can make the change, but I strongly disagree with it if the only goal is avoiding that overload.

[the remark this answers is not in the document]
sebastiangrimberg: It can still be printed to the log, but doesn't need to be in the postprocessing files where the user is going to have access to the same abilities to compute it as a derived quantity.

For the log output, I'd recommend not including it in this PR and just managing the printing as part of the SEMR loop.

test/testcase.jl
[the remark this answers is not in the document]
hughcars: done https://github.com/awslabs/palace/pull/85/commits/fc5000601a6bd4ede4cfe331884a67c37bcb7054

palace/drivers/basesolver.hpp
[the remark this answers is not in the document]
hughcars: https://github.com/awslabs/palace/pull/85/commits/fc5000601a6bd4ede4cfe331884a67c37bcb7054

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,332 @@
...
+      {
+        T.SetIntPoint(&ip);
+        double smooth_val = smooth_flux_gf.GetValue(e, ip);
+        double coarse_val = coef.Eval(T, ip);
+        const double w_i = ip.weight * T.Weight();
+        estimates[e] += w_i * std::pow(smooth_val - coarse_val, 2.0);
+        normalization += w_i * std::pow(coarse_val, 2.0);
+      }
+    }
+    if constexpr (false)
```
sebastiangrimberg, line 299: Isn't this no longer needed? This code is already in `PostOperator`?

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,332 @@
...
+  for (auto &e : estimates)
+  {
+    e = std::sqrt(e);
+  }
+
+  Mpi::GlobalSum(1, &normalization, nd_fespace.GetComm());
+  normalization = std::sqrt(normalization);
+  if (normalization > 0)
+  {
+    std::for_each(estimates.begin(), estimates.end(),
```
sebastiangrimberg, line 173: Not the way to loop over a `Vector`, see comments below.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,332 @@
...
+      }
+    }
+    estimates[e] = std::sqrt(estimates[e]);
+  }
+
+  Mpi::GlobalSum(1, &normalization, nd_fespace.GetComm());
+  normalization = std::sqrt(normalization);
+  if (normalization > 0)
+  {
+    std::for_each(estimates.begin(), estimates.end(),
```
sebastiangrimberg, line 235: Not the way to loop over a `Vector`, see comments below.

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,332 @@
...
+        {
+          estimates[e] += w_i * std::pow(smooth_vec[c] - coarse_vec[c], 2.0);
+          normalization += w_i * coarse_vec[c] * coarse_vec[c];
+        }
+      }
+    }
+  }
+  for (auto &e : estimates)
+  {
+    e = std::sqrt(e);
```
sebastiangrimberg, line 166: Can move into the loop above?

palace/linalg/errorestimator.cpp
```diff
@@ -0,0 +1,332 @@
...
+      mfem::ParGridFunction est_field(&est_fespace);
+      est_field.SetFromTrueDofs(estimates);
+
+      paraview.RegisterField("ErrorIndicator", &est_field);
+      paraview.Save();
+    }
+  }
+  for (auto &e : estimates)
+  {
+    e = std::sqrt(e);
```
sebastiangrimberg, line 318: Can move into the loop above?

palace/main.cpp
```diff
@@ -186,7 +185,7 @@ int main(int argc, char *argv[])
   mesh::RefineMesh(iodata, mesh);
 
   // Run the problem driver.
-  solver->Solve(mesh);
+  auto solver_output = solver->Solve(mesh);
```
sebastiangrimberg, line 188: `indicator` seems more descriptive here.

palace/utils/errorindicators.hpp
```diff
@@ -0,0 +1,87 @@
...
+  {
+    n = 0;
+    local = Vector();
+    normalization = 0;
+  }
+
+protected:
+  // Elemental localized error indicators. Used for marking elements for
+  // refinement and coarsening.
+  Vector local;
```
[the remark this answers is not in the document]
sebastiangrimberg: Also, another comment that will be helpful. If you construct these `Vector`s as `ParGridFunction`s over an order-0 L2 FE space, you won't need to add the stuff inside of `PostOperator`. This FE space is best constructed by `SpaceOperator` and the like, probably. It can actually be a standard `GridFunction`, since we never need the neighbor data.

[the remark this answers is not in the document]
sebastiangrimberg: Lastly, I'd move this into `fem/`. Seems more appropriate there since it isn't really just a "utility" class anymore.

sebastiangrimberg (changes requested):
I'm working my way up from the bottom, so have added some (final) comments on `ErrorEstimator` and `ErrorIndicator`. I think there is something to be rethought with the `ErrorIndicator` class which will clarify the normalization and make this class more useful.

I haven't looked at anything inside of `drivers/` since I think this changes should be finalized first before I look at the "user" code.

palace/utils/errorindicators.hpp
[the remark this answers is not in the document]
hughcars: The estimator was originally dumb in the first round, just returning the vector and normalization, but I rolled that into `ErrorIndicator` based on your comments about combining it all together. The downside of `ErrorIndicators` storing the vectors internally is now all of that information is going to leak out into the SEMR loop, which was previously just ingesting the single reduced quantity, hence the return type of the `Solve`. Having rolled all the classes in `errorindicators.hpp` into one, I now think that's a much better interface, as it can be used as just the dumb output of error estimator, as it currently is, but is also the reduced form. A reduced form should be indistinguishable from a a single sample from the outside, in terms of data required, and the methods on `ErrorIndicators` enable this. The constructor from a `vector<ErrorIndicators>` is a compromise to your request, of allowing for externally collecting indicators before constructing a reduced form.

Using a `GridFunction` for the local indicators has the effect of making all the SEMR loop now have to accept all the crud associated with the `GridFunction` interface, and breaks even further when considering the finite element spaces now defined in spaceoperator etc. will have been destructed by that point, meaning most of the grid functions methods will break due to the dangling pointer. An `ErrorIndicator` needs to stand in isolation without any references, as it's the one thing that'll be departing the `Solve` scope. So using `GridFunction` here is a hard no.

palace/utils/errorindicators.hpp
```diff
@@ -0,0 +1,87 @@
...
+// communicator for use in adaptation.
+class ErrorIndicators
+{
+public:
+  ErrorIndicators(Vector local, double normalization)
+    : local(std::move(local)), normalization(normalization), n(1)
+  {
+  }
+
+  ErrorIndicators() = default;
```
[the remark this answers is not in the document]
hughcars: The `ErrorIndicators() = default;` is needed, as once a constructor is defined, that is implicitly deleted. I had thought that meant defining all, hence the other 4, but turns out it only needs this one default.

palace/utils/errorindicators.hpp
```diff
@@ -0,0 +1,102 @@
...
+    auto sum = std::accumulate(local.begin(), local.end(), 0.0);
+    Mpi::GlobalSum(1, &sum, comm);
+    return sum / size;
+  }
+  // Return the normalization constant for the absolute error.
+  inline auto GetNormalization() const { return normalization; }
+  // Add a set of indicators to the running totals.
+  void AddIndicators(const ErrorIndicators &indicators);
+  // Reset a running total of error indicators ready for computing a new running average.
+  inline void Reset()
```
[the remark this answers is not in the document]
hughcars: No longer, was originally in treatment of the transient problem, but it's no longer needed.

palace/linalg/errorestimator.cpp
hughcars: Dumping the fluxes to file is not in `PostOperator`, the indicator field yes can be deleted.

palace/linalg/errorestimator.cpp
hughcars: Each element is visited once for each component, accumulating the square, so this has to be done afterwards.

palace/linalg/errorestimator.cpp
hughcars: Each element is visited twice, once for real and imaginary, accumulating the square, so this has to be done afterwards.

hughcars: Introduce `CwiseSqrt`
https://github.com/awslabs/palace/pull/85/commits/eca57eb96d4216cf5d5ac1ed3cc1c9532f069225

docs/src/config/solver.md
```diff
@@ -420,6 +420,12 @@ eigenmode simulation type.
 `"DivFreeMaxIts" [100]` :  Maximum number of iterations for divergence-free cleaning use in
 the eigenmode simulation type.
 
+`"EstimatorTol" [1e-6]` :  Relative tolerance for flux projection used in the
```
sebastiangrimberg, line 423: Nice :)

test/testcase.jl
```diff
@@ -66,7 +67,18 @@ function testcase(
         for file in filesref
             dataref = CSV.File(joinpath(refpostprodir, file); header=1) |> DataFrame
             data    = CSV.File(joinpath(postprodir, file); header=1) |> DataFrame
-            data    = data[1:size(dataref, 1), :]
+            if (size(data, 1) < size(dataref, 1))
```
sebastiangrimberg, line 70: Let's replace this by another argument to `testcase`: A list of files for which to exclude the check for equality in the number of rows. If the file matches the expression given, we can skip the `@test` on line 77 and just test on the first `min(size(data, 1), size(dataref, 1)` rows of `data` and `data_ref`. This way we aren't doing this slightly shady data copying to match rows. This parameter will default to being empty (check for all files) but should include usually the `estimator.csv` output as well as all files (support wildcard?) for eigenmode simulations. What do you think?

palace/drivers/basesolver.cpp
```diff
@@ -565,6 +566,44 @@ void BaseSolver::PostprocessFields(const PostOperator &postop, int step, double
...
+  if (root)
+  {
+    std::string path = post_dir + "error-indicators.csv";
+    auto output = OutputFile(path, (step > 0));
+    if (step == 0)
+    {
+      // clang-format off
+      output.print("{:>{}s},{:>{}s},{:>{}s},{:>{}s},{:>{}s},{:>{}s}\n",
+                   name, table.w1,
+                   "Sum Rel.", table.w,
```
sebastiangrimberg, line 587: For the column headings, I propose removing the `"Elem"` and the `"Rel"`. Thus, it will just be: `"Sum", "Min", "Max", "Mean", "Normalization"`.

The meaning of the quantities as sum/min/max/mean over all elements of the relative quantity already divided by the normalization can be clarified in the documentation (to come in a later PR).

palace/utils/configfile.hpp
```diff
@@ -819,6 +819,12 @@ struct LinearSolverData
   // Maximum number of iterations for solving linear systems in divergence-free projector.
   int divfree_max_it = 1000;
 
+  // Relative tolerance for solving linear systems in the error estimator.
+  double estimator_tol = tol;
+
+  // Maximum number of iterations for solving linear systems in the error estimator.
+  int estimator_max_it = max_it;
```
sebastiangrimberg, line 826: Let's give this a default not tied to `max_it`. 100 according to the docs. Also this points out a discrepancy for `div_free_max_it` -> The default in the docs should be 1000, not 100. Can you fix that?

palace/utils/configfile.cpp
```diff
@@ -1669,6 +1669,10 @@ void LinearSolverData::SetUp(json &solver)
   divfree_max_it = linear->value("DivFreeMaxIts", divfree_max_it);
   gs_orthog_type = linear->value("GSOrthogonalization", gs_orthog_type);
 
+  // Options related to the estimator.
+  estimator_tol = linear->value("EstimatorTol", estimator_tol);
```
sebastiangrimberg, line 1673: If we're breaking out this section, `divfree_tol`, `divfree_max_it`, and `gs_orthog_type` should go here too since they aren't explicitly "preconditioner-specific". Maybe just have the comment read: `"Other linear solver options."`

palace/utils/configfile.cpp
```diff
@@ -1702,6 +1706,9 @@ void LinearSolverData::SetUp(json &solver)
   linear->erase("DivFreeTol");
   linear->erase("DivFreeMaxIts");
   linear->erase("GSOrthogonalization");
+
+  linear->erase("EstimatorTol");
```
sebastiangrimberg, line 1710: Given where these are parsed, move up to between "DivFreeMaxIts" and "GSOrthogonalization" (style nit).

palace/linalg/vector.hpp
```diff
@@ -139,6 +139,9 @@ void SetRandomReal(MPI_Comm comm, VecType &x, int seed = 0);
 template <typename VecType>
 void SetRandomSign(MPI_Comm comm, VecType &x, int seed = 0);
 
+// Component-wise square root.
+void CwiseSqrt(Vector &x);
```
sebastiangrimberg, line 143: I think `Sqrt` is better, in line with the similar method from PETSc: https://petsc.org/main/manualpages/Vec/VecSqrtAbs/.

palace/fem/errorindicators.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "linalg/vector.hpp"
+#include "utils/communication.hpp"
+
+namespace palace
+{
+// Storage for error estimation results from the solve. Required in the AMR loop. This is
+// richer than the IndicatorsAndNormalization because it stores derived quantities, and a
+// communicator for use in adaptation.
+class ErrorIndicators
```
sebastiangrimberg, line 15: OK, I agree with your prior comments and am starting to like the design more. I have a small suggested change here that I think will actually do a lot to make this more clear. Let's rename this to `ErrorIndicator`, since an object of type `ErrorIndicator` is really a single indicator over the mesh of the estimate. I get that an error indicator can be thought of as plural (where the singular would be for one element), but as a function over the mesh elements it is singular. Then we can take the min/max/sum etc. of that single indicator function. I think part of my confusion was that sometimes we had plural indicators and sometimes singular which didn't make sense on the first read. This class is good now.

With the above renaming, you could just have the following: `ErrorIndicator::GetLocal`, `ErrorIndicator::Min/Max/Mean/Norml2`. Do you think this would help?

palace/drivers/eigensolver.hpp
```diff
@@ -40,13 +41,10 @@ class EigenSolver : public BaseSolver
...
-  EigenSolver(const IoData &iodata, bool root, int size = 0, int num_thread = 0,
-              const char *git_tag = nullptr)
-    : BaseSolver(iodata, root, size, num_thread, git_tag)
-  {
-  }
+  using BaseSolver::BaseSolver;
 
-  void Solve(std::vector<std::unique_ptr<mfem::ParMesh>> &mesh) const override;
+  ErrorIndicators
+  Solve(const std::vector<std::unique_ptr<mfem::ParMesh>> &mesh) const final;
```
sebastiangrimberg, line 47: We don't use `final` anywhere else in the code, and it doesn't really serve a purpose here. Let's keep this as `override`.

sebastiangrimberg:
I think we should squash these commits before merging #103 in, and then merging to `main`. What do you think? If that sounds good feel free to go ahead and do so and rebase and merge #103 here.

sebastiangrimberg (approved):

PR #87, opened 2023-08-22
Complex-valued operators for multigrid preconditioning

sebastiangrimberg:
This PR adds the capability, based on #47, to precondition complex-valued linear systems using multigrid based on the actual complex-valued system matrix, rather than its real-valued approximation as is currently implemented. This is possible for the fine level smoothers, not the coarse-level solve, which still always uses the real-valued approximation. Reverting to the previous behavior is controlled with the `PCMatReal` configuration file parameter.

This has positive effects on convergence in my testing, reducing iteration counts by ~2x in some cases. This is not for free: The smoother application is now twice as expensive roughly since we are performing matrix-vector products with a complex-valued operator instead of a real one. However, especially in the case where the coarse solver dominates the multigrid V-cycle cost, this can lead to big cost savings.

Another positive benefit is this preconditioning strategy appears to lead to much more accurate repeated linear solves. For the cavity example with impedance BC, here are the first 5 modes with a real-valued preconditioner matrix:

```
 m     Re{ω}/2π (GHz)     Im{ω}/2π (GHz)        Bkwd. Error         Abs. Error
==============================================================================
 1      +2.904507e+00      +7.086023e-04      +2.453159e-10      +1.766090e-05
 2      +2.922516e+00      +7.051646e-04      +3.229538e-10      +2.339414e-05
 3      +2.922529e+00      +7.051718e-04      +4.048642e-10      +2.932770e-05
 4      +3.468922e+00      +8.640178e-04      +4.039384e-10      +3.472130e-05
 5      +4.147609e+00      +9.784805e-04      +3.396857e-10      +3.490230e-05
```

And with this PR:

```
 m     Re{ω}/2π (GHz)     Im{ω}/2π (GHz)        Bkwd. Error         Abs. Error
==============================================================================
 1      +2.904507e+00      +7.086037e-04      +2.048618e-13      +1.474851e-08
 2      +2.922516e+00      +7.051671e-04      +2.043630e-13      +1.480366e-08
 3      +2.922529e+00      +7.051734e-04      +2.160061e-13      +1.564713e-08
 4      +3.468922e+00      +8.640203e-04      +3.551412e-13      +3.052685e-08
 5      +4.147609e+00      +9.784829e-04      +3.932118e-11      +4.040205e-06
```

The total number of linear solver iterations decreased from 1276 to 745.

Another feature added inthis PR is control over the Chebyshev polynomial smoother implementation, allowing users to pick from the standard smoother based on Chebyshev polynomials for the 1st-kind versus the newer 4th-kind variant (see [Phillips and Fischer, Optimal Chebyshev smoothers and one-sided V-cycles (2022)](https://arxiv.org/abs/2210.03179)). Eigenvalue estimates are also now controlled by parameters which can be modified in the configuration file, which is helpful for debugging.


palace/models/spaceoperator.cpp
```diff
@@ -422,9 +400,19 @@ SpaceOperator::GetComplexExtraSystemMatrix(double omega,
     ai = BuildOperator(GetNDSpace(), (SumCoefficient *)nullptr, (SumCoefficient *)nullptr,
                        &dfbi, &fbi, pa_order_threshold, skip_zeros);
   }
-  auto A = std::make_unique<ComplexParOperator>(std::move(ar), std::move(ai), GetNDSpace());
-  A->SetEssentialTrueDofs(nd_dbc_tdof_lists.back(), diag_policy);
-  return A;
+  if constexpr (std::is_same<OperType, ComplexOperator>::value)
```
hughcars, line 403: `complex` declared above, can reuse

hughcars (approved):

palace/models/spaceoperator.cpp
sebastiangrimberg: Fixed by rearranging, in [1df5506](https://github.com/awslabs/palace/pull/87/commits/1df550654489e49c55948409f16f90aeaad921d9).

PR #88, opened 2023-08-22
Add option to remove high-order mesh curvature information

sebastiangrimberg:
For debugging purposes, it can be useful to strip the nodes associated high-order curved meshes and deal with only the underlying straight-sided mesh defined by the element vertices. This PR adds a `RemoveCurvature` option to the configuration file which performs this operation after loading the mesh from disk.

### DIFF
--- a/palace/utils/geodata.cpp
+++ b/palace/utils/geodata.cpp
@@ -30,7 +30,7 @@ namespace
 const auto MSH_FLT_PRECISION = std::numeric_limits<double>::max_digits10;
 
 // Load the serial mesh from disk.
-std::unique_ptr<mfem::Mesh> LoadMesh(const std::string &);
+std::unique_ptr<mfem::Mesh> LoadMesh(const std::string &, bool);
 
 // Optionally reorder mesh elements based on MFEM's internal reordeing tools for improved
 // cache usage.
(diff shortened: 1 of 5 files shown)


palace/utils/geodata.cpp
```diff
@@ -1096,16 +1098,32 @@ std::unique_ptr<mfem::Mesh> LoadMesh(const std::string &path)
...
+  }
+  if (mesh->GetNodes() && remove_curvature)
+  {
+    mfem::GridFunction *nodes = nullptr;
+    int own_nodes = true;
+    mesh->SwapNodes(nodes, own_nodes);
+    if (own_nodes)
+    {
+      delete nodes;
+    }
```
hughcars, line 1121: This works fine if the mesh is never marked as nonconformal, but I've been doing some further testing and instead of using `DistributeMesh`
```
  smesh->EnsureNCMesh(true);
  std::unique_ptr<mfem::ParMesh> mesh =
      std::make_unique<mfem::ParMesh>(comm, *smesh, partitioning.get());
```
having made sure all ranks build the mesh (even running with 1 rank), causes the eigen linear solve to fail on cavity example. Noting this here, but this is arguably downstream of this pr, so going to approve.

hughcars (approved):
This works for conformal so approving, the issue with NC is a downstream thing to be addressed later.

palace/utils/geodata.cpp
sebastiangrimberg: Definitely worth looking into for `NCMesh` support, thanks for noting. We probably need more details on how the example "fails" with `"RemoveCurvature"` to debug.

PR #96, opened 2023-09-11
Refine parallel tet mesh bug fix from #90

sebastiangrimberg:
Follow-up to #90 after testing with conformal AMR code. Also includes a sequence of minor fixes for `main` encountered during further testing.


hughcars (approved):
Note for later: The `patch_par_tet_mesh_fix.diff` patch incorporates https://github.com/mfem/mfem/pull/3847 in addition to https://github.com/mfem/mfem/pull/3843.

PR #103, opened 2023-10-10
Suggested improvements for #85

sebastiangrimberg:
Summary of changes:
- In `ErrorIndicator`:
  - Use `mfem::forall` for eventual GPU support
  - Remove unused `normalization` member and some member functions which are no longer needed.
- In `ErrorEstimator`:
  - Construct RHS for projection solve using a `BilinearForm` to enable eventual partial assembly and construction of the RHS on GPU (the linear solve already uses this). For now, the `GradientIntegrator` does not support matrix-valued coefficients so the flux for the `GradFluxEstimator` is computed as just `grad(U)` instead of `\epsilon grad(U)`. This will be fixed in https://github.com/awslabs/palace/pull/97.
  - Use `T.OrderW()` in computing the desired quadrature rule (element Jacobian order) instead of `T.Order()` (element order).
  - Call `GetVectorValue(T, ip, ...)` instead of `(e, ip, ...)` to avoid resetting `T` with a redundant call to `SetIntPoint`.
  - Refactor `CurlFluxEstimator` for code reuse between the complex-valued and real-valued variants, with changes to avoid looping over mesh elements twice.
  - Refactor `GradFluxEstimator` to use a vector H1 space where appropriate to take advantage of repeated basis evaluations.
  - Move all timing for estimation into `errorestimation.cpp` to avoid splitting it between the `driver/` classes and the estimator classes.
- Other organization:
  - Remove no longer needed functions and classes in `vector.hpp` and `coefficient.hpp`.
  - Simplify writing of indicator fields by `PostProcessor`, done only once for a solve.
  - Make construction and use of estimator and indicator objects consistent across all drivers.


palace/linalg/errorestimator.cpp
```diff
@@ -91,260 +77,267 @@ FluxProjector<SmoothFluxFiniteElementCollection>::FluxProjector(
...
 
-  Mpi::GlobalSum(1, &normalization, h1_fespace.GetComm());
+  // Finalize the element-wise error estimates.
+  Mpi::GlobalSum(1, &normalization, mesh.GetComm());
   normalization = std::sqrt(normalization);
-  if (normalization > 0)
+  if (normalization > 0.0)
   {
-    estimates /= normalization;
+    estimates *= 1.0 / normalization;
```
hughcars, line 332: If there's a meaningful difference here, this feels like it should be reported to mfem as a bug.

hughcars (approved):

palace/linalg/errorestimator.cpp
sebastiangrimberg: I'm not sure I follow. I have to admit I made this change just because it feels better to me to ensure we are taking care of the 1/x on our end, but looking at the implementation of `/=` it looks like it's identical. Am I missing something?

sebastiangrimberg:
Merged directly into https://github.com/awslabs/palace/pull/85

PR #105, opened 2023-10-12
Remove some unneeded `SetIntPoint` calls for `Coefficient` evaluation

sebastiangrimberg:
An `mfem::ElementTransformation` is configured at the given integration point in the loop over quadrature points inside of the element, at the level of a `Linear/BilinearForm` assembly or `GridFunction` evaluation. Calling `SetIntPoint` again will reset the transformation and require redundant computation of the element Jacobian in cases where it could have been reused.

### DIFF
--- a/palace/fem/coefficient.hpp
+++ b/palace/fem/coefficient.hpp
@@ -196,8 +196,15 @@ public:
   {
   }
 
-  static void GetNormal(mfem::ElementTransformation &T, const mfem::IntegrationPoint &ip,
-                        mfem::Vector &normal);
+  static void GetNormal(mfem::ElementTransformation &T, mfem::Vector &normal)
+  {
+    // Return normal vector to the boundary element at an integration point (it is assumed
+    // that the element transformation has already been configured at the integration point
+    // of interest).
+    normal.SetSize(T.GetSpaceDim());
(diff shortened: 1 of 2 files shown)


palace/fem/coefficient.hpp
```diff
@@ -196,8 +196,15 @@ class BdrGridFunctionCoefficient
   {
   }
 
-  static void GetNormal(mfem::ElementTransformation &T, const mfem::IntegrationPoint &ip,
-                        mfem::Vector &normal);
+  static void GetNormal(mfem::ElementTransformation &T, mfem::Vector &normal)
+  {
+    // Return normal vector to the boundary element at an integration point (it is assumed
+    // that the element transformation has already been configured at the integration point
+    // of interest).
```
hughcars, line 203: Nit: If you move this comment above the method, an IDE ought to display this with a hover over of the method. I don't think there's a nice way of baking a debug check of this assertion, so will have to rely on an informed user, so that might help.

hughcars (approved):

palace/fem/coefficient.hpp
sebastiangrimberg: Fixed in [cf7bc99](https://github.com/awslabs/palace/pull/105/commits/cf7bc99f155748d2aad0019c93e69123f65dd4d9).

PR #107, opened 2023-10-16
Fix default RPATH for installed targets

sebastiangrimberg:
Useful when using `CMAKE_BUILD_SHARED`. With this, there is no need to configure LD_LIBRARY_PATH at runtime.


CMakeLists.txt
```diff
@@ -38,22 +38,42 @@ if(PALACE_WITH_STRUMPACK OR PALACE_WITH_MUMPS OR PALACE_WITH_ARPACK)
...
   message(STATUS "Setting BUILD_SHARED_LIBS to 'OFF' as it was not specified")
-  set(BUILD_SHARED_LIBS OFF CACHE BOOL "Global flag to cause add_library() to create shared libraries if ON")
+  set(BUILD_SHARED_LIBS OFF CACHE BOOL
+    "Global flag to cause add_library() to create shared libraries if ON"
+  )
+endif()
+
+# Configure default RPATH for installed targets if not provided
+if(NOT DEFINED CMAKE_INSTALL_RPATH)
+  message(STATUS "Setting CMAKE_INSTALL_RPATH to '\${CMAKE_INSTALL_PREFIX}/lib;\${CMAKE_INSTALL_PREFIX}/lib64' as it was not specified")
```
hughcars, line 67: Based on the changes for line length elsewhere here, this might be split across lines with [\](https://stackoverflow.com/questions/7637539/how-to-split-strings-across-multiple-lines-in-cmake) to match.

PR #110, opened 2023-10-16
Adaptive Mesh Refinement

hughcars:
* Introduces a Solve-Estimate-Mark-Refinement adaptive mesh refinement loop for all problem types aside from Transient.
* Allows for conformal rebalancing through a gather scatter model. Supports nonconformal adaptation through patches applied to mfem for higher order Nedelec tets, interior boundaries and coarsening.
* Does not currently support WavePorts for nonconformal meshes.
* Employs a Dorfler marking scheme (a.k.a bulk chasing) whereby a minimum number of elements are marked that achieves a given fraction of the total global error. 

Resolves https://github.com/awslabs/palace/issues/2


sebastiangrimberg (changes requested):
I have not yet looked at the larger changes in `basesolver.cpp` and `dorfler.hpp/cpp` or `geodata.hpp/cpp`, but here is a first pass.

Related, but does this comment need addressing for the nonconformal case?

https://github.com/awslabs/palace/blob/2b66e53a12b9fbae56d78560cb2e7ffe0ca77502/palace/fem/coefficient.cpp#L81

palace/utils/timer.hpp
```diff
@@ -54,6 +56,8 @@ class Timer
       "Estimation",
       "  Construction",
       "  Solve",
+      "Adaptation",
```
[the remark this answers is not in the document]
hughcars: The first reason I opted for separate is pragmatic. If running without adaptation, it doesn't appear in the block timer and the estimate section ends up indenting weirdly:
```
Solve
  Preconditioner
  Coarse Solve
  Estimation
    Construction
    Solve
```
Adaptation only shows up in the timing if the max its is greater than one, which makes the indenting of the estimation consistent. The second is more semantic, Adaptation here means the Mark and Refine stages of SEMR. If estimate were placed as a sub timer, then that implies solve should be too, as all four steps are substeps of that loop. Solve being a sub timer feels entirely incorrect, so transitively estimate as a sub timer does too.

Marking and refining themselves ended up being so quick I didn't bother breaking their timers out, but Rebalancing ended up being noticeable, hence the sub timer.

cmake/ExternalMFEM.cmake
```diff
@@ -38,6 +38,7 @@ endif()
 # required
 set(PALACE_MFEM_WITH_ZLIB NO)
 set(PALACE_MFEM_WITH_LIBUNWIND NO)
+set(PALACE_MFEM_WITH_EXCEPTIONS NO)
```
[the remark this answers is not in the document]
hughcars: I've been using this a fair bit with lldb and mpirun, which was pretty helpful especially when errors occurred fairly deep in an adapt and I wanted to get context from the rest of the backtrace. I'll pull it from this PR, but I think it has value in the debug build.

docs/src/config/model.md
```diff
@@ -54,7 +54,19 @@ scale based on the bounding box of the computational domain.
             "Radius": float
         },
         ...
-    ]
+    ],
+    "Adaptation":
```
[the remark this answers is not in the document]
hughcars: Part of the rationale for the substruct was that I had been thinking of Adaptation as a subtype of Refinement. If the idea is that Refinement becomes Adaptation however, then this could change. A lot of these options would then need to be prefixed with `Adaptive` or similar. I thought the substruct was more encapsulating though, hence this choice.

The main advantage of moving to Advanced options is not having to write a doc string for an option, but given I have I'm not sure the extra information is worth removing. Independent of that issue though:
* I could definitely see `"SaveStep"` equal to 0, as that amounts to just overwriting the postpro directory every time a new solve completes, without saving off the substeps. If a user didn't care about storing the intermediate results (or cared about disk space), they might set that option. This could be changed to a boolean instead, of storing results or not, but I originally picked `"SaveStep"` for consistency with other saving off options.
* `"MaximumImbalance"` could be argued as advanced, as adjusting that from 1 ought to be very unusual. 
* `"WriteSerialMesh"` is similar in spirit to the `"SaveStep"` option, if a user didn't care for the extra storage. It's pretty self explanatory though, so again can be made advanced. This option is mainly helpful for cases where a user thinks they might conceivably want to restart on a larger machine, or generate an adapted mesh for use in a parametric sweep.
* `"UseCoarsening"` could also be argued as advanced, given the default to false, and that any reasonable choice of `"UpdateFraction"` makes it practically unneeded. This argument can be extended to deleting coarsening from the main SEMR loop entirely, but I thought it would be likely that users might want to explore it as an option so opted to keep it.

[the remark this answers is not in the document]
sebastiangrimberg: I see that you already updated this in the last commit, but two comments:

1. If it would serve better to rename the top level `Refinement` to something else, I could be on board with that. `config["Model"]["Adaptation"]` or `"MeshAdaptation"` could work too, and the existing `"UniformLevels"` and other parameters wouldn't seem totally out of place in that section. I also do not think it is necessarily confusing to users to have `config["Refinement"]["Tol"]` or `["MaxIts"]` (as currently, but without the `"Adapt"` prefix). There shouldn't be any conflict with any other tolerances associated with refinement. But if you feel it is necessary I guess we can keep it.

2. The main benefit to moving items to the advanced section is not really avoiding doc strings (they should have probably all have doc strings but this makes them harder to change quickly as the code changes, and so for now they are undocumented). It is to be clear to users what items you intend for them to specify and which ones they really never should unless doing development work. So, items in the advanced section are not included in the demo config file above the docstrings.

test/examples/testcase.jl
```diff
@@ -1,6 +1,9 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+using CSV
```
sebastiangrimberg, line 4: These changes should not longer be needed once you rebase.

scripts/schema/config/model.json
```diff
@@ -10,8 +10,8 @@
     "L0": { "type": "number", "exclusiveMinimum": 0.0 },
     "Lc": { "type": "number", "exclusiveMinimum": 0.0 },
     "Partition": { "type": "string" },
-    "ReorientTetMesh": { "type": "boolean" },
     "RemoveCurvature": { "type": "boolean" },
+    "ReorientTetMesh": { "type": "boolean" },
```
sebastiangrimberg, line 14: Ideally the order here mirrors the order in `configfile.cpp` so these are easily to keep track of, so this can be reverted.

palace/utils/iodata.hpp
```diff
@@ -9,6 +9,7 @@
 namespace mfem
 {
 
+class Mesh;
```
sebastiangrimberg, line 12: No longer needed

palace/utils/iodata.hpp
```diff
@@ -44,6 +45,9 @@ class IoData
   // Nondimensionalize input values for use in the solver, including the mesh coordinates.
   void NondimensionalizeInputs(mfem::ParMesh &mesh);
 
+  // Return the mesh scaling factor, (Lc / L0), for mesh IO.
+  double GetMeshScaleFactor() const { return Lc / model.L0; }
```
sebastiangrimberg, line 49: `GetLengthScaleFactor`? Since all units in the config file are expected to by in units of `model.L0 * m`

palace/utils/timer.hpp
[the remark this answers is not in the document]
sebastiangrimberg: I would always have the top level object show up:

```
...
Operator Construction
Solve
  Preconditioner
  Coarse Solve
Mesh Refinement // <- Or just Refinement? Or Adaptation?
  Estimation
  Estimation Solve // <- Lump in construction, probably don't need separate categories?
  Adaptation
  Rebalancing
Postprocessing
...
```

The idea being that we wouldn't really be doing any estimation cost if we weren't going to try to do something with it (adapt the mesh), so grouping them together as the "net cost of AMR" seems like the right way to do it.

palace/utils/dorfler.cpp
```diff
@@ -0,0 +1,168 @@
...
+    return {elems_marked, error_marked};
+  };
+
+  // Each processor will compute a different threshold: if a given processor has lots of low
+  // error elements, their value will be lower and if a processor has high error, their
+  // value will be higher. Thus using the value from the low error processor will give too
+  // many elements, and using the value from the high error processor will give too few. The
+  // correct threshold value will be an intermediate between the min and max over
+  // processors.
+  auto comm = Mpi::World();
```
sebastiangrimberg, line 55: I think the communicator should be an argument to this function alongside the `Vector`. I realize right now everyone uses `MPI_COMM_WORLD` but if someday we want to solve the problems on a subcommunicator, hopefully that shouldn't involve too many deep code changes.

palace/utils/dorfler.cpp
```diff
@@ -0,0 +1,168 @@
...
+  const double total_error = [&]()
+  {
+    double total_error = sum.back();
+    Mpi::GlobalSum(1, &total_error, comm);
```
sebastiangrimberg, line 59: Can we combine the reductions which can be throughout this function? A single `MPI_AllReduce` on > 1 variables is much preferred to many reductions on a single variable.

palace/utils/dorfler.cpp
```diff
@@ -0,0 +1,168 @@
...
+    {
+      Mpi::Print("Threshold: {:e} < {:e} < {:e}, Marked Elems: {} <= {} <= {}\n",
+                 min_threshold, error_threshold, max_threshold, min_elem_marked,
+                 elem_marked, max_elem_marked);
+    }
+
+    // Set the tolerance based off of the largest local indicator value. These tolerance
+    // values extremely tight because this loop is fast, and getting the marking correct is
+    // important.
+    constexpr double frac_tol = 2 * std::numeric_limits<double>::epsilon();
```
sebastiangrimberg, line 114: IWYU `#include <limits>`

palace/utils/geodata.cpp
```diff
@@ -1073,6 +1140,62 @@ void GetSurfaceNormal(mfem::ParMesh &mesh, const mfem::Array<int> &marker,
...
+
+  mesh = DistributeMesh(comm, smesh, partitioning);
+  if (save_serial_mesh)
+  {
+    BlockTimer bt(Timer::IO);
+    if (Mpi::Root(comm))
+    {
+      std::ofstream serial(output_serial_mesh_file);
+      serial.precision(mesh::MSH_FLT_PRECISION);
+      mesh::DimensionalizeMesh(*smesh, iodata.GetMeshScaleFactor());
```
sebastiangrimberg, line 1192: Seems like you can just pass in the length scale instead of hte entire `iodata` object here.

palace/utils/geodata.cpp
```diff
@@ -1073,6 +1140,62 @@ void GetSurfaceNormal(mfem::ParMesh &mesh, const mfem::Array<int> &marker,
...
+      smesh->Print(serial);  // Do not need to nondimensionalize the temporary
+    }
+    Mpi::Barrier(comm);
```
sebastiangrimberg, line 1195: Do we need a barrier here?

palace/utils/geodata.cpp
```diff
@@ -1073,6 +1140,62 @@ void GetSurfaceNormal(mfem::ParMesh &mesh, const mfem::Array<int> &marker,
...
+    partitioning = GetMeshPartitioning(*smesh, Mpi::Size(comm));
+  }
+
+  mesh = DistributeMesh(comm, smesh, partitioning);
+  if (save_serial_mesh)
+  {
+    BlockTimer bt(Timer::IO);
+    if (Mpi::Root(comm))
+    {
+      std::ofstream serial(output_serial_mesh_file);
```
sebastiangrimberg, line 1190: Probably worth adding this (and if done elsewhere) to `Timer::IO`.

palace/utils/geodata.cpp
```diff
@@ -25,38 +25,61 @@ using CVector3dMap = Eigen::Map<const Eigen::Vector3d>;
...
+{
+  bool has_simplices;
+  bool has_tensors;
+  bool has_wedges;
+  bool has_pyramids;
+};
+ElementTypeInfo CheckElements(mfem::Mesh &mesh)
+{
+  auto meshgen = mesh.MeshGenerator();
+  return {bool(meshgen & 1), bool(meshgen & 2), bool(meshgen & 4), bool(meshgen & 8)};
```
sebastiangrimberg, line 80: I believe these are local to the process and not reduced in parallel. I think you need a reduction based on how they are used below.

palace/drivers/basesolver.cpp
```diff
@@ -84,6 +89,282 @@ BaseSolver::BaseSolver(const IoData &iodata, bool root, int size, int num_thread
   }
 }
 
+std::string BaseSolver::IterationPostDir(int iter) const
```
sebastiangrimberg, line 92: Seems like this can be in an anonymous namespace like the `GetPostDir` above just to be consistent.

palace/drivers/basesolver.cpp
```diff
@@ -84,6 +89,282 @@ BaseSolver::BaseSolver(const IoData &iodata, bool root, int size, int num_thread
...
+    {
+      ind.Append(i);
+    }
+  }
+  return ind;
+}
+
+// Helper function responsible for rebalancing the mesh, and optionally writing meshes from
+// the intermediate stages to disk.
+void RebalanceMesh(std::unique_ptr<mfem::ParMesh> &mesh, const IoData &iodata,
```
sebastiangrimberg, line 123: Seems to me like this code belongs in `geodata.cpp`, next to `RebalanceConformalMesh`?

palace/drivers/basesolver.cpp
```diff
@@ -84,6 +89,282 @@ BaseSolver::BaseSolver(const IoData &iodata, bool root, int size, int num_thread
...
+
+  const bool use_amr = param.adapt_max_its > 0;
+  const bool use_coarsening = mesh.back()->Nonconforming() && param.use_coarsening;
+  if (use_amr && mesh.size() > 1)
+  {
+    Mpi::Print("{}\n", "Flattening mesh sequence: AMR will start from the final mesh in "
+                       "the refinement sequence.");
+    mesh.erase(mesh.begin(), mesh.end() - 1);
+    constexpr bool refine = true, fix_orientation = true;
+    mesh.back()->Finalize(refine, fix_orientation);
```
sebastiangrimberg, line 271: Can you not assume this has already been called during the load in `geodata.cpp`?

palace/drivers/basesolver.cpp
```diff
@@ -84,6 +89,282 @@ BaseSolver::BaseSolver(const IoData &iodata, bool root, int size, int num_thread
...
+    if (is_transient)
+    {
+      MFEM_ABORT("AMR is not currently supported for transient simulations!");
+    }
+    else
+    {
+      Mpi::Print("\nAdaptive Mesh Refinement Parameters:\n");
+      Mpi::Print("MaxIter: {}, Tolerance: {:.3e}\n\n", param.adapt_max_its,
+                 param.adapt_tolerance);
+      SavePostProcess(iter);  // Save the initial solution
```
sebastiangrimberg, line 290: As discussed offline, this can become the first step of the `while` loop below so that the final solution is not duplicated on disk.

palace/drivers/basesolver.cpp
```diff
@@ -84,6 +89,282 @@ BaseSolver::BaseSolver(const IoData &iodata, bool root, int size, int num_thread
...
+
+      // Refine.
+      const auto initial_elem_count = mesh.back()->GetGlobalNE();
+      mesh.back()->GeneralRefinement(marked_elements, -1, param.max_nc_levels);
+      const auto final_elem_count = mesh.back()->GetGlobalNE();
+      Mpi::Print("Mesh refinement added {} elements. Initial: {}, Final: {}\n",
+                 final_elem_count - initial_elem_count, initial_elem_count,
+                 final_elem_count);
+    }
+    else if (use_coarsening)
```
sebastiangrimberg, line 326: Feel free to correct me if this is a dumb question or thought here, but it seems like coarsening should be more useful than only after the final solve when `ntdof >= dof_limit`. Every time we solve, we get a new estimate which could in principle expose elements for coarsening, right? Now, maybe we only want to try to coarsen after a few rounds of refinement, for which we could have a config file option, but I don't see the value in coarsening only after you've hit your limit of the maximum size.

sebastiangrimberg (commented):
Second (more thorough) pass done. Everything is looking really great, this is awesome @hughcars! Mostly all minor comments sprinkled throughout.

palace/utils/geodata.cpp
hughcars: The barrier here is to match with the `BlockTimer bt(Timer::IO)` before the branch on` Mpi::Root(comm)`. I copied the pattern in `ReadMesh` where the timer is instantiated for all ranks rather than just root, and kept coordinated by the barrier. Strictly speaking the other ranks don't need to be waiting here, then the timer could be moved inside the `if`, and only time the root processor, and the IO timing would then be desynchronized across processors.

sebastiangrimberg: Whoops sorry about that that, my mistake. Not sure how I missed the `Timer::IO` addition already as well.

palace/drivers/basesolver.cpp
hughcars: The issue is that `UniformRefinement()` of a tet mesh invalidates the finalization. From `mesh.hpp`:
```
   /** @param[in] ref_algo %Refinement algorithm. Currently used only for pure
       tetrahedral meshes. If set to zero (default), a tet mesh will be refined
       using algorithm A, that produces elements with better quality compared to
       algorithm B used when the parameter is non-zero.

       For tetrahedral meshes, after using algorithm A, the mesh cannot be
       refined locally using methods like GeneralRefinement() unless it is
       re-finalized using Finalize() with the parameter @a refine set to true.
       Note that calling Finalize() in this way will generally invalidate any
       FiniteElementSpace%s and GridFunction%s defined on the mesh. */
   void UniformRefinement(int ref_algo = 0);
```
The issue being I believe that the uniform refinement uses another code path that doesn't respect the edge ordering . Hence the finalization only being on the pathway where the vector contains meshes generated from refinement.

sebastiangrimberg: Ah, makes sense to me. Thanks!

palace/drivers/basesolver.cpp
```diff
@@ -235,13 +126,12 @@ void BaseSolver::SolveEstimateMarkRefine(
   // Helper to save off postprocess data.
   auto SavePostProcess = [&](int iter)
   {
-    Mpi::Barrier(comm);
     BlockTimer bt_io(Timer::IO);
+    Mpi::Barrier(comm);
```
sebastiangrimberg, line 130: I think this isn't necessary? If anything it should be before the timer, but I think it makes sense for the timer to include all time waiting.

palace/drivers/basesolver.cpp
```diff
@@ -57,53 +58,20 @@ void WriteMetadata(const std::string &post_dir, const json &meta)
...
   std::string dir = fmt::format("{}Adapt{:0>3d}/", post_dir, iter);
   // Create directory for output.
-  if (root && !std::filesystem::exists(dir))
+  if (!fs::exists(dir))
   {
-    std::filesystem::create_directories(dir);
+    fs::create_directories(dir);
   }
-  return dir;
+  return fs::path(dir);
```
sebastiangrimberg, line 71: Is there an advantage to `fs::path` instead of just `string` as used everywhere else?

palace/drivers/basesolver.cpp
hughcars: I had put it there as I was worried about the root starting to move the post processing directory while another processor is still writing into it. Given that, I viewed that blocking as part of the IO, hence placing it within the barrier.

palace/drivers/basesolver.cpp
hughcars: For this function it could be a string, the one place it gets called, in the `SavePostProcess` lambda, it would then be converted to a path immediately. Using `fs::path` allows for using the `directory_iterator` and `copy` within the lambda, in addition to the `/` operator. So this function could return a string, but `fs::path` would get called on the result immediately, and given the function is only used in this one place, it seemed neater to have it return the finished product. Either is fine with me though.

sebastiangrimberg (approved):

Issue #112, opened 2023-10-17
Increase precision of serialized frequency vector for driven solves


parrangoiz:
Currently frequency-domain driven solves save data to files such as `port-S.csv`, `port-I.csv`, etc and the frequency vector is serialized to 6 digits of precision. Suggest increasing this to 9.

PR #113, opened 2023-10-18
Increase fp precision of first column of postpro files

parrangoiz:
*Description of changes:*

Previously the floating point precision of the first column of all postprocessed data outputs was set to 6. I'm increasing it to 9 to match that of all other columns and to prevent unnecessary rounding (e.g. of frequency vectors for driven solves).

*Issue #, if available:* https://github.com/awslabs/palace/issues/112


sebastiangrimberg:
Thanks @arrangop, looks good. I will address the failing test (looks like the whitespace in a `DataFrame` is being considered when comparing two objects, ideally I'd like to ignore the leading or trailing whitespace).

PR #116, opened 2023-10-19
Fix integration point mapping from boundary element to face element

### LINKED ISSUE #114: Regression with boundary field visualization with ParaView
In `paraview_boundary` we write out a data collection of the computed fields on the boundary elements of the mesh, to go along with the domain data in `paraview`. There seems to be regression with this output where the orientation of some elements is incorrect which creates discontinuities in the results:

- Boundary left, domain right:

image

- Boundary right, domain left:

image

This seems to only show up with Nedelec spaces of order > 1, so could be related to the `DofTransformation` from the `GridFunction` to the boundary element.

sebastiangrimberg:
Resolves #114

The `be_to_bfe` method which originates from `mfem::GridFunction` and this [PR](https://github.com/mfem/mfem/pull/1194) did not account for all possible orientations of the triangle or quadrilateral face. This is likely due to the fact that it was only used for exterior boundaries so the orientations were only those with aligned normals. For interior boundaries, we need to deal with all possible orientations.

Eventually this is a bug fix that belongs in MFEM, but for now we can fix it in Palace.


sebastiangrimberg:
This PR will be replaced by https://github.com/mfem/mfem/pull/3943 in MFEM.

sebastiangrimberg:
Replaced by https://github.com/awslabs/palace/pull/128

PR #118, opened 2023-10-24
Refactor construction of discrete interpolator matrices for reuse

sebastiangrimberg:
Introduce our own `FiniteElementSpaceHierarchy` class to replace `mfem::ParFiniteElementSpaceHierarchy` which lazily constructs the prolongation operators as needed. Also introduces `AuxiliaryFiniteElementSpace` and the corresponding `AuxiliaryFiniteElementSpaceHierarchy` which will lazily construct and store discrete interpolation matrices between spaces on the de Rham complex, for example the discrete gradient, so we are not reconstructing it in many different places. Previously, for an eigenmode simulation, it was constructed on the fine level at least 2 times: distributive relaxation smoothing and divergence free projection.


palace/fem/fespace.hpp
```diff
@@ -35,6 +38,167 @@ class FiniteElementSpace : public mfem::ParFiniteElementSpace
...
+    return P_;
+  }
+};
+
+//
+// A special type of FiniteElementSpaceHierarchy where all members are auxiliary finite
+// element spaces.
+//
+class AuxiliaryFiniteElementSpaceHierarchy : public FiniteElementSpaceHierarchy
+{
```
hughcars, line 146: Not a fan of the slightly tautological interface this ends up with, (ksp.cpp with `aux_fespace->GetFinestAuxiliarySpace()` and then the `AddLevel` override effectively being deleted means this violates the "is a" rule of inheritance. 

I've an alternative suggestion on `hughcars/discrete-interpolator-refactor` that breaks the inheritance of these two spaces whilst reusing the interface, and means the `GetFESpace` methods return auxilliarys directly if needed. Additionally, given the break from mfem already, I'd suggest dropping the `FESpace` from the getter methods, so `GetLevel()` `GetFinestLevel()` etc, as `AddLevel` has already defined what level refers to.

palace/fem/fespace.hpp
```diff
@@ -35,6 +38,167 @@ class FiniteElementSpace : public mfem::ParFiniteElementSpace
...
+  }
+  const AuxiliaryFiniteElementSpace &GetFinestAuxiliaryFESpace() const
+  {
+    return *static_cast<const AuxiliaryFiniteElementSpace *>(&GetFinestFESpace());
+  }
+
+  const Operator &GetDiscreteInterpolatorAtLevel(std::size_t l) const
+  {
+    return GetAuxiliaryFESpaceAtLevel(l).GetDiscreteInterpolator();
+  }
```
hughcars, line 189: `GetDiscreteInterpolatorAtFinestLevel` for consistency with the other getters.

palace/fem/fespace.hpp
```diff
@@ -35,6 +38,167 @@ class FiniteElementSpace : public mfem::ParFiniteElementSpace
...
+public:
+  FiniteElementSpaceHierarchy() = default;
+  FiniteElementSpaceHierarchy(std::unique_ptr<FiniteElementSpace> &&fespace)
+  {
+    AddLevel(std::move(fespace));
+  }
+
+  auto GetNumLevels() const { return fespaces.size(); }
+  auto size() const { return GetNumLevels(); }
+  bool empty() const { return GetNumLevels() == 0; }
```
hughcars, line 90: I get the consistency with stl, but for consistency with all other methods and to avoid inconsistency if other "stl-like" methods appear, probably worth going with Camel. 

The only exception i can think of for this is `begin()` and `end()` to work with range loops, which actually should probably be added, given this is basically a thin wrapper for a std::vector. I would probably go for it referring to the fespaces, but could refer to a pair of space and prolongation if wanted. This would be a bit of a deviation from mfem, but would allow for range loops in a few places elsewhere.

palace/fem/multigrid.hpp
```diff
@@ -70,17 +67,15 @@ std::vector<std::unique_ptr<FECollection>> inline ConstructFECollections(
     }
   }
   std::reverse(fecs.begin(), fecs.end());
+
```
hughcars, line 70: I thought the rule was no white space unless comment comes after?

palace/fem/multigrid.hpp
```diff
@@ -70,17 +67,15 @@ std::vector<std::unique_ptr<FECollection>> inline ConstructFECollections(
...
-// marked.
+// Construct a hierarchy of finite element spaces given a sequence of meshes and finite
+// element collections. Additionally, Dirichlet boundary conditions are marked.
 template <typename FECollection>
-inline std::unique_ptr<mfem::ParFiniteElementSpaceHierarchy>
-ConstructFiniteElementSpaceHierarchy(
-    int mg_max_levels, bool mg_legacy_transfer, int pa_order_threshold,
-    bool pa_discrete_interp, const std::vector<std::unique_ptr<mfem::ParMesh>> &mesh,
+inline FiniteElementSpaceHierarchy ConstructFiniteElementSpaceHierarchy(
+    int mg_max_levels, const std::vector<std::unique_ptr<mfem::ParMesh>> &mesh,
```
hughcars, line 78: Given we've now taken control of space hierarchies, these should probably be constructors for the relevant classes rather than free functions.

palace/fem/multigrid.hpp
```diff
@@ -90,54 +85,95 @@ ConstructFiniteElementSpaceHierarchy(
...
+
+// Similar to ConstructFiniteElementSpaceHierarchy above, but in this case the finite
+// element space at each level is an auxiliary space associated with the coresponding level
+// of the provided finite element space objects.
+template <typename FECollection>
+inline AuxiliaryFiniteElementSpaceHierarchy ConstructAuxiliaryFiniteElementSpaceHierarchy(
+    const FiniteElementSpaceHierarchy &primal_fespaces,
+    const std::vector<std::unique_ptr<FECollection>> &fecs,
+    const mfem::Array<int> *dbc_marker = nullptr,
+    std::vector<mfem::Array<int>> *dbc_tdof_lists = nullptr)
```
hughcars, line 130: As above, should be a constructor given we now have total control.

hughcars (commented):
Getting some unused private field warnings:
In file included from /Users/hughcars/AWS/palace/palace/models/curlcurloperator.cpp:4:
/Users/hughcars/AWS/palace/palace/models/curlcurloperator.hpp:28:14: warning: private field 'pa_discrete_interp' is not used [-Wunused-private-field]
  const bool pa_discrete_interp;  // Use partial assembly for discrete interpolators
             ^
In file included from /Users/hughcars/AWS/palace/palace/models/laplaceoperator.cpp:4:
/Users/hughcars/AWS/palace/palace/models/laplaceoperator.hpp:28:14: warning: private field 'pa_discrete_interp' is not used [-Wunused-private-field]
  const bool pa_discrete_interp;  // Use partial assembly for discrete interpolators

palace/fem/fespace.hpp
sebastiangrimberg: I like your approach and have cherry-picked it, in [7df6a19](https://github.com/awslabs/palace/pull/118/commits/7df6a19cd2dfda85a1821beaabc814b79952de3c) with a few minor amendments. Thanks!

palace/fem/fespace.hpp
```diff
@@ -35,6 +38,167 @@ class FiniteElementSpace : public mfem::ParFiniteElementSpace
   std::size_t GetId() const;
 };
 
+//
+// An AuxiliaryFiniteElement space is just a FiniteElementSpace which allows for lazy
```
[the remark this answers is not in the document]
sebastiangrimberg: Fixed.

palace/fem/fespace.hpp
sebastiangrimberg: There is no `GetProlongationAtFinestLevel` so I have chosen to not include a `GetDiscreteInterpolatorAtFinestLevel`.

palace/fem/fespace.hpp
sebastiangrimberg: Sounds good, fixed.

palace/linalg/divfree.cpp
```diff
@@ -49,14 +49,7 @@ DivFreeSolver::DivFreeSolver(const MaterialOperator &mat_op,
...
   }
-  {
-    constexpr bool skip_zeros_interp = true;
-    DiscreteLinearOperator grad(h1_fespaces.GetFinestFESpace(), nd_fespace);
-    grad.AddDomainInterpolator<GradientInterpolator>();
-    Grad = std::make_unique<ParOperator>(
-        grad.Assemble(pa_discrete_interp ? pa_order_threshold : 99, skip_zeros_interp),
-        h1_fespaces.GetFinestFESpace(), nd_fespace, true);
-  }
+  Grad = &h1_fespaces.GetFinestAuxiliaryFESpace().GetDiscreteInterpolator();
```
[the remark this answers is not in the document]
sebastiangrimberg: Commented on this above, but I don't think we need to introduce a `GetFinestDiscreteInterpolator` method, in the same way that prolongation operators are defined for each level without helpers for the finest/coarsest levels.

palace/fem/multigrid.hpp
sebastiangrimberg: As discussed offline, I prefer to keep the constructors for `FiniteElementSpaceHierarchy` simple and include these helpers as free functions.

palace/fem/multigrid.hpp
sebastiangrimberg: Added this to match `ConstructFiniteElementSpaceHierarchy` and `ConstructAuxiliaryFiniteElementSpaceHierarchy`. I don't think there's a rule here, just what looks best (subjective probably).

palace/linalg/errorestimator.cpp
```diff
@@ -23,14 +22,13 @@ namespace palace
 namespace
 {
 
-std::unique_ptr<Operator>
-GetMassMatrix(const mfem::ParFiniteElementSpaceHierarchy &fespaces, int pa_order_threshold,
-              int skip_zeros)
+std::unique_ptr<Operator> GetMassMatrix(const FiniteElementSpaceHierarchy &fespaces,
+                                        int pa_order_threshold, int skip_zeros)
```
hughcars, line 26: Nit: `int skip_zeros` can be bool no?

hughcars (approved):

palace/linalg/errorestimator.cpp
sebastiangrimberg: Whoops, #120 fixed these but rebasing messed things up. Fixed in [8c05eaf](https://github.com/awslabs/palace/pull/118/commits/8c05eaf6638ac19cd287a7c347987c3f96a8fb54).

PR #120, opened 2023-10-24
Default order for Chebyshev polynomial smoothing

sebastiangrimberg:
Set default smoother order for multigrid based on the solver order, and increase smooth order for mass-matrix solve in error estimation.

### DIFF
--- a/palace/linalg/errorestimator.cpp
+++ b/palace/linalg/errorestimator.cpp
@@ -63,9 +63,15 @@ ConfigureLinearSolver(const mfem::ParFiniteElementSpaceHierarchy &fespaces, doub
   std::unique_ptr<Solver<Operator>> pc;
   if (fespaces.GetNumLevels() > 1)
   {
+    const int dim = fespaces.GetFinestFESpace().GetParMesh()->Dimension();
+    const auto type = fespaces.GetFinestFESpace().FEColl()->GetRangeType(dim);
+    const int mg_smooth_order =
+        (type == mfem::FiniteElement::SCALAR)
+            ? 2
+            : std::max(fespaces.GetFinestFESpace().GetMaxElementOrder(), 2);
     pc = std::make_unique<GeometricMultigridSolver<Operator>>(
-        std::move(amg), fespaces, nullptr, 1, 1, 2, 1.0, 0.0, true, pa_order_threshold,
-        pa_discrete_interp);
+        std::move(amg), fespaces, nullptr, 1, 1, mg_smooth_order, 1.0, 0.0, true,
+        pa_order_threshold, pa_discrete_interp);
   }
   else
   {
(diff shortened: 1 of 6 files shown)


palace/linalg/errorestimator.cpp
```diff
@@ -90,14 +96,13 @@ FluxProjector::FluxProjector(const MaterialOperator &mat_op,
                              bool pa_discrete_interp)
 {
   BlockTimer bt(Timer::CONSTRUCTESTIMATOR);
-  constexpr int skip_zeros = 0;
+  constexpr int skip_zeros = false;
```
hughcars, line 99: Looking at `ParOperator::Assemble`, these are bools, so probably worth just changing the type and being done with it.

palace/linalg/errorestimator.cpp
```diff
@@ -63,9 +63,15 @@ ConfigureLinearSolver(const mfem::ParFiniteElementSpaceHierarchy &fespaces, doub
   std::unique_ptr<Solver<Operator>> pc;
   if (fespaces.GetNumLevels() > 1)
   {
+    const int dim = fespaces.GetFinestFESpace().GetParMesh()->Dimension();
+    const auto type = fespaces.GetFinestFESpace().FEColl()->GetRangeType(dim);
+    const int mg_smooth_order =
+        (type == mfem::FiniteElement::SCALAR)
+            ? 2
+            : std::max(fespaces.GetFinestFESpace().GetMaxElementOrder(), 2);
```
hughcars, line 71: So the smoothing only needs to be strengthened for ND elements, do we know why?

palace/linalg/errorestimator.cpp
sebastiangrimberg: Whoops, I only caught half the issue. Thanks!

palace/linalg/errorestimator.cpp
sebastiangrimberg: Yes, this is based on experimentation and this seems to be a good heuristic tor optimal settings. A few ideas as to why this could be:
- I'm not sure if this is proven, but I believe the element Nedelec mass matrix is more poorly conditioned than the nodal one. Has to do with continuity, etc. so it would make sense the preconditioner should change.
- In the context of Jacobi-preconditioned polynomial smoothing used for multigrid, the diagonal as assembled in parallel in both the nodal and Nedelec cases is approximate when the operator isn't fully assembled. For Nedelec it introduces a few more approximations and thus makes sense why more iterations for the same smoother order might be required.

palace/linalg/errorestimator.cpp
sebastiangrimberg: Fixed by c581aae.

palace/linalg/errorestimator.cpp
sebastiangrimberg: I've just gone ahead and increased the smoother order for all polynomial smoothing in [b1754fd](https://github.com/awslabs/palace/pull/120/commits/b1754fda2f04a321c5945c4b9d14899456c8ea70).  This is nice for consistency and will keep iteration counts similar at least.

hughcars (approved):

PR #143, opened 2023-11-29
Jacobi preconditioning for error estimation

hughcars:
Change the error estimator to use Jacobi preconditioning. The mass matrix is SPD and diagonally dominant, so Jacobi is well suited, and this will reduce pressure on memory whilst allowing the default maximum iteration count to be raised.


sebastiangrimberg (approved):
Thanks @hughcars. I made one small add on to allow runtime specification of a Jacobi preconditioner for the main linear solve. This is not a recommended option in general but could be useful for debugging. Also might be useful for explicit time integration schemes for the transient simulation type.

PR #147, opened 2023-12-13
Timer fixes for linear solves

sebastiangrimberg:
Configure linear solver timing to only count preconditioning and coarse solve time for primary solvers (not error estimation, divergence-free projection, etc.). Also add timing category for divergence-free projection for eigenvalue solves.


sebastiangrimberg:
@kllrak I added you as an optional reviewer as the `Timer` functionality is something you've worked on in the past. Feel free to take a look if you're able!

hughcars (approved):

PR #153, opened 2023-12-13
Return exit code from Palace in wrapper script

sebastiangrimberg:
Previously, even if the binary exited with non-zero exit code the wrapper script would still return 0.


hughcars (approved):
Ran `palace -np 6 cavity_pec.json` changing the exit code in `main.cpp` to 1, then `echo $?`. On main it gives 0, here it gives 1, so looks good to me.

Issue #156, opened 2023-12-18
Bug in computation of currents and charges for driven solves


parrangoiz:
*Description: A clear and concise description of what the bug is.*

I ran a simple driven simulation of a CPW and looked at the results in Paraview; the $E$ and $B$ fields as well as the energy densities $U_E$ and $U_B$ all look exactly as expected, but $J_s$ and $Q_s$ look like gibberish which suggests a bug in the way these are being computed from the fields before being written to disk. See the attached screenshots of $\text{Re}(E_y)$ and $|\text{Re}(J_s)|$.

<img width="1058" alt="image" src="https://github.com/awslabs/palace/assets/60243345/5cfc4f51-258a-4745-b533-3d87f2bf9bb3">

<img width="1060" alt="image" src="https://github.com/awslabs/palace/assets/60243345/f708f419-0cfa-44fe-b2ab-e6eae7ce29a8">

*To reproduce: Please provide minimal example that reproduces the error. For existing
examples, please provide a link.*

N/A but I imagine this should be reproducible using the CPW crosstalk example from the docs.

*Error message: Paste the complete error message or log, if applicable.*

None.

*Environment: Any environment details, such as operating system, compiler, etc.*

PR #158, opened 2023-12-18
Fix bug with mesh nondimensionalization for parallel meshes

sebastiangrimberg:
This results from https://github.com/awslabs/palace/pull/128, where we are rescaling the mesh nodes before writing the VTU output for ParaView visualization. For parallel simulations, the neighbor mesh nodes also need to be scaled so that evaluation using the neighbor `ElementTransformation` objects works.

Also from that PR, this fixes the output of fields with non-identity Piola transformations from the reference to physical space evaluations. H(curl) fields transform with `J^{-T}` and thus need scaling with L when the mesh nodes are scaled by L. Likewise H(div) transform with `J / |J|` and thus need scaling with `L^{d-1}`.

Resolves https://github.com/awslabs/palace/issues/156

### DIFF
--- a/palace/models/postoperator.cpp
+++ b/palace/models/postoperator.cpp
@@ -626,7 +626,42 @@ void PostOperator::WriteFields(int step, double time, const ErrorIndicator *indi
   bool first_save = (paraview.GetCycle() < 0);
   mfem::ParMesh &mesh =
       (E) ? *E->ParFESpace()->GetParMesh() : *B->ParFESpace()->GetParMesh();
+  auto ScaleGridFunctions = [&mesh, this](double L)
+  {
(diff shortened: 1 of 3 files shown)


sebastiangrimberg:
`J_S` from prior to this fix:

Screenshot 2023-12-18 at 3 25 18 PM

After the fix:

Screenshot 2023-12-18 at 3 27 38 PM

hughcars (approved):
Looked through `pgridfunc.hpp` and `pmesh.hpp` for any other neighbor data that might need transforming and not seeing anything. The code for the transforms looks fine and the Piola transforms all match those I looked up, so looks good to me.

PR #164, opened 2024-01-04
Patch for ParNCMesh::Update bug

hughcars:
Proposed changes to `sjg/geodata-updates` to patch the issue with `"SaveAdaptMesh": true` causing segfaults. Tested on the cavity example with
```
    "Refinement":
    {
      "MaxIts": 2,
      "SaveAdaptMesh": true
    }
```


sebastiangrimberg:
For reference: See https://github.com/mfem/mfem/pull/4055.

sebastiangrimberg (approved):

PR #177, opened 2024-01-31
Fix regression for electrostatic and magnetostatic postprocessing

sebastiangrimberg:
The source of this is https://github.com/awslabs/palace/pull/149. The `spheres` and `rings` examples have both been updated to ensure this is caught in the future.


parrangoiz:
Not sure if it's related to this but I think the postprocessed E-field for visualization may be broken too; when I look at voltage in Paraview it looks exactly as expected but `E` looks pretty random.

parrangoiz:
Oh weird. Did you run that in 1 core? Mine ran in 64.

PR #185, opened 2024-02-02
Documentation for GPU support

### LINKED ISSUE #3: Operator partial assembly, GPU support
MFEM's native support for [operator partial assembly](https://mfem.org/performance/) does not extend to non-tensor-product elements like simplices or to mixed meshes. One route for complete partial assembly support is via [libCEED](https://github.com/CEED/libCEED), to which MFEM has an interface but which does not yet support vector finite elements like Nédélec or Raviart-Thomas.

We would like to utilize operator partial assembly for efficient high-order operator construction and application, and improved performance leveraging modern hardware like GPUs. Preconditioning linear systems using p-multigrid is already utilized widely in the code without assuming any assembly of parallel sparse matrices except on the coarsest level for the coarse solve. Thus, the goal is matrix-free p-multigrid leading to efficient implementation of high-order finite element methods in the specific context of electromagnetics.

sebastiangrimberg:
Accompanies #184 and resolves #3

### DIFF
--- a/docs/src/install.md
+++ b/docs/src/install.md
@@ -56,8 +56,8 @@ A build from source requires the following prerequisites installed on your syste
   - C and Fortran (optional) compilers for dependency builds
   - MPI distribution
   - BLAS, LAPACK libraries (described below in [Math libraries](#Math-libraries))
-  - [CUDA Toolkit](https://www.amd.com/en/graphics/servers-solutions-rocm) or [ROCm]
-    (https://developer.nvidia.com/cuda-toolkit) installation (optional, for GPU support
+  - [CUDA Toolkit](https://developer.nvidia.com/cuda-toolkit) or
+    [ROCm](https://rocm.docs.amd.com/en/latest/) installation (optional, for GPU support
     only)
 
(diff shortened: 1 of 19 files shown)


docs/src/install.md
```diff
@@ -56,6 +56,9 @@ A build from source requires the following prerequisites installed on your syste
   - C and Fortran (optional) compilers for dependency builds
   - MPI distribution
   - BLAS, LAPACK libraries (described below in [Math libraries](#Math-libraries))
+  - [CUDA Toolkit](https://www.amd.com/en/graphics/servers-solutions-rocm) or [ROCm]
+    (https://developer.nvidia.com/cuda-toolkit) installation (optional, for GPU support
```
hughcars, line 60: Those hyperlinks are the wrong way round, and the rocm link seems to be broken.

hughcars (changes requested):
Issue with links aside looks good to me. In terms of merging, will be easier for reviewing to merge in sequence once approved, after the target branch auto updates to `main`, rather than into `sjg/gpu-build-system-dev`.

hughcars (approved):

PR #189, opened 2024-02-07
Miscellaneous documentation fixes and improvements

sebastiangrimberg:
NOTE: Waiting on some changes to timer categories and wave ports (see https://github.com/awslabs/palace/pull/191, https://github.com/awslabs/palace/pull/192)

Resolves #155, #187

PR #191, opened 2024-02-08
Updates to timer categories

sebastiangrimberg:
Added a few new categories for timing, including specific to linear solver vs. eigensolver vs. time stepping, and other related changes. This improves the organization and use of the timer classes throughout the code.

### DIFF
--- a/palace/drivers/drivensolver.cpp
+++ b/palace/drivers/drivensolver.cpp
@@ -113,7 +113,6 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &spaceop, PostOperator &
   // Because the Dirichlet BC is always homogenous, no special elimination is required on
   // the RHS. Assemble the linear system for the initial frequency (so we can call
   // KspSolver::SetOperators). Compute everything at the first frequency step.
-  BlockTimer bt0(Timer::CONSTRUCT);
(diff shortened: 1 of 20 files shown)


sebastiangrimberg:
@kllrak, I just requested review from you because you had done the initial work on the timer. Perhaps the more relevant PR for your review is https://github.com/awslabs/palace/pull/189, where the changes here are documented.

kllrak:
Looks like changes are adjustments to timing categories and where the timers are used, rather than the timing mechanism. In that case I'm probably not the most efficient reviewer here, but thanks for the heads up.

palace/drivers/drivensolver.cpp
```diff
@@ -365,11 +355,17 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &spaceop, PostOperator
...
     omega += delta_omega;
   }
+  BlockTimer bt0(Timer::POSTPRO);
+  SaveMetadata(prom.GetLinearSolver());
   return indicator;
 }
 
 int DrivenSolver::GetNumSteps(double start, double end, double delta) const
 {
+  if (end < start)
```
hughcars, line 365: Maybe `<=`? Skips the logic if driving a single frequency.

sebastiangrimberg: My thought was that the logic below works fine for `start = end`, so no need to bypass it.

hughcars (approved):

PR #194, opened 2024-02-09
Use `mfem::Workspace` for temporary `Vector` allocations

sebastiangrimberg:
NOTE: Builds on https://github.com/awslabs/palace/pull/184 (and https://github.com/awslabs/palace/pull/194, and now https://github.com/awslabs/palace/pull/204)

See https://github.com/mfem/mfem/pull/4065

~~TODO: Add patch on MFEM for https://github.com/mfem/mfem/pull/4065 to enable this functionality~~ -> Done as part of https://github.com/awslabs/palace/pull/184

### DIFF
--- a/palace/linalg/iterative.cpp
+++ b/palace/linalg/iterative.cpp
@@ -10,6 +10,7 @@
 #include "linalg/orthog.hpp"
 #include "utils/communication.hpp"
 #include "utils/timer.hpp"
+#include "utils/workspace.hpp"
 
 namespace palace
 {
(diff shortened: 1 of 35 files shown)


palace/linalg/rap.cpp
```diff
@@ -49,21 +50,23 @@ void ParOperator::SetEssentialTrueDofs(const mfem::Array<int> &tdof_list,
...
   }
 
   // Apply the unconstrained operator.
+  auto ly = workspace::NewVector<Vector>(test_fespace.GetVSize());
   A->Mult(lx, ly);
 
-  auto &ty = test_fespace.GetTVector<Vector>();
-  RestrictionMatrixMult(ly, ty);
-  b.Add(-1.0, ty);
+  {
```
hughcars, line 65: `ly` can be moved into the limited scope too.

palace/linalg/iterative.cpp
```diff
@@ -646,12 +644,12 @@ void GmresSolver<OperType>::Mult(const VecType &b, VecType &x) const
     }
     else  // B && pc_side == PrecSide::RIGHT
     {
-      r = 0.0;
+      r.VecType::operator=(0.0);
```
hughcars, line 647: The `using operator=` in the workspace vector classes should mean these method casts aren't necessary.

hughcars (commented):
Only important thing is the `ComplexWorkspaceVector` I think has some very minor issues on the interface/usage of method casting. Otherwise, this is a great change.

palace/utils/workspace.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+  // No copy constructor.
+  ComplexWorkspaceVector(ComplexWorkspaceVector &other) = delete;
+
+  // Copy assignment: copy contents of vector, not metadata.
+  ComplexWorkspaceVector &operator=(ComplexWorkspaceVector &other)
+  {
+    ComplexVector::operator=(other);
+    return *this;
+  }
```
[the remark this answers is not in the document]
sebastiangrimberg: Good catch, indeed this error was copied from the original PR. Fixed in 8e4148ab.

palace/linalg/iterative.cpp
sebastiangrimberg: Resolved above.

palace/linalg/rap.cpp
sebastiangrimberg: This scope block is actually not necessary I think since `ly` and `ty` must always be present in the same scope. So I've removed in f3ef607b to be more similar to the rest of the `ParOperator` functions.

hughcars (approved):

sebastiangrimberg:
Commit [3e5a42b](https://github.com/awslabs/palace/pull/194/commits/3e5a42b9647f5a39ae1cab47ff7be0b6d5b6c985) introduces a patch for MFEM to fix the issues reported here: https://github.com/mfem/mfem/pull/4065#issuecomment-1977653880.

hughcars:
Closing due to the arena allocator not making it into mfem any time soon.

PR #197, opened 2024-02-14
Add `"Active"` config file field for lumped and wave ports

### LINKED ISSUE #171: Wave port boundaries with multiple mode indices
It should be possible to define multiple wave ports to the same set of boundary attributes with different mode indices, in order to simultaneously postprocess scattering parameters into each of the modes or drive systems with a sum of the different incident mode shapes. The impedance boundary condition with coefficient `i k_n \mu^{-1}` would have to be chosen for one of the modes (or the average?). For most engineering cases where you care about multiple modes, the `k_n` should likely be similar for both modes (even-odd, for example).

sebastiangrimberg:
Resolves #171.

In addition, this PR changes field visualization on internal boundaries to always chose the field on the vacuum side (previously, sometimes used vacuum and sometimes used side with lower speed of light).

### DIFF
--- a/docs/src/config/boundaries.md
+++ b/docs/src/config/boundaries.md
@@ -250,6 +250,7 @@ accounts for nonzero metal thickness.
         "Direction": <string> or [<float array>],
         "CoordinateSystem": <string>,
         "Excitation": <bool>,
+        "Active": <bool>,
         "R": <float>,
         "L": <float>,
         "C": <float>,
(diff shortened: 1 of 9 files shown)


hughcars (approved):

hughcars:
Only comment would be to add the documentation for `Active` in the markdown, but the other branch is merging into this with those changes so probably ok.

sebastiangrimberg:
> Only comment would be to add the documentation for `Active` in the markdown, but the other branch is merging into this with those changes so probably ok.

Good catch, missed this when rebasing the PRs to reverse their order.

PR #200, opened 2024-02-20
Fix ParaView output on boundary mesh for shared faces

sebastiangrimberg:
We redimensionalize the mesh coordinates and the local parts of the fields, but also need to redimensionalize the face neighbor data vectors (which have already been exchanged, based on the nondimensional dof data). This resolves issues where the fields for shared faces appear to be very small or very large relative to the rest of the boundary visualization.

Note: This should not have an effect on surface integrals based on boundary coefficients for problems with internal boundaries (like integrated charge, or surface participation). These integrals are computed outside of the field nondimensionalization/redimensionalization, so this bug only affects visualization and also only for parallel simulations which boundary elements coinciding with shared faces.

### DIFF
--- a/palace/models/postoperator.cpp
+++ b/palace/models/postoperator.cpp
@@ -646,24 +646,30 @@ void PostOperator::WriteFields(int step, double time, const ErrorIndicator *indi
     {
       // Piola transform: J^-T
       E->real() *= L;
+      E->real().FaceNbrData() *= L;
       if (has_imaginary)
       {
         E->imag() *= L;
+        E->imag().FaceNbrData() *= L;
       }
     }
     if (B)
     {
       // Piola transform: J / |J|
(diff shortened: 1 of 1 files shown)


sebastiangrimberg:
For example, the `cpw/wave_uniform` case prior to this PR, visualizing electric field energy in `paraview_bdr` on the chip surface, in log scale (pardon the coarse mesh):

Screenshot 2024-02-20 at 9 05 02 AM

After this PR:

Screenshot 2024-02-20 at 9 04 47 AM

One can see that the few triangles on the chip surface which had very small values are resolved. These triangles correspond to subdomain boundary faces where the evaluation in the neighboring subdomain, using the boundary face neighbor data, is incorrect due to the missing redimensionalization.

hughcars (approved):

PR #201, opened 2024-02-22
Fix global synchronization bug for optional material properties

sebastiangrimberg:
Methods like `MaterialOperator::HasLossTangent` are expected to return a value which is consistent globally. This is not the case currently, as material properties are parsed only with respect to local subdomains. This PR fixes this issue and potential hangs for large processor counts or small domains of nonzero loss tangent or conductivity.

### DIFF
--- a/palace/models/materialoperator.cpp
+++ b/palace/models/materialoperator.cpp
@@ -346,6 +346,7 @@ void MaterialOperator::SetUpMaterialProperties(const IoData &iodata,
   mat_invLondon.SetSize(sdim, sdim, nmats);
   mat_c0_min.SetSize(nmats);
   mat_c0_max.SetSize(nmats);
+  has_losstan_attr = has_conductivity_attr = has_london_attr = false;
 
   int count = 0;
   for (std::size_t i = 0; i < iodata.domains.materials.size(); i++)
--- a/palace/models/materialoperator.cpp
+++ b/palace/models/materialoperator.cpp
@@ -461,8 +461,8 @@ void MaterialOperator::SetUpMaterialProperties(const IoData &iodata,
 
     count++;
   }
-  int has_attr[3] = {has_losstan_attr, has_conductivity_attr, has_london_attr};
-  Mpi::GlobalMax(3, has_attr, mesh.GetComm());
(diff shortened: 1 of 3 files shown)


palace/models/materialoperator.cpp
```diff
@@ -461,14 +456,16 @@ void MaterialOperator::SetUpMaterialProperties(const IoData &iodata,
...
+      has_london_attr = true;
     }
 
     count++;
   }
+  int has_attr[3] = {has_losstan_attr, has_conductivity_attr, has_london_attr};
+  Mpi::GlobalMax(3, has_attr, mesh.GetComm());
+  has_losstan_attr = has_attr[0];
+  has_conductivity_attr = has_attr[1];
+  has_london_attr = has_attr[2];
```
hughcars, line 468: Was wondering about why the need for `int` register, and see it doesn't work with `bool` directly. It seems more elegant to have
```
  static void GlobalOr(int len, bool *buff, MPI_Comm comm)
  {
    GlobalOp(len, buff, MPI_LOR, comm);
  }

  static void GlobalAnd(int len, bool *buff, MPI_Comm comm)
  {
    GlobalOp(len, buff, MPI_LAND, comm);
  }
```
in `communicator.hpp` and use it directly. Could also have specializations of `Max` and `Min` for `bool` if really wanted, but I think `Or` and `And` are semantically clearer.

Second to that, can then have `struct {bool losstan, conductivity, london;} has_attr` and send the struct directly with `GlobalOr(3, &has_attr.losstan, mesh.GetComm());`

hughcars (commented):
Looks fine, one minor suggestion to have `GlobalOr` and `GlobalAnd`.

palace/models/materialoperator.cpp
sebastiangrimberg: Resolved in [31329d9](https://github.com/awslabs/palace/pull/201/commits/31329d9cde4d9c105ad53f415ac0bdd03ab0a40e). Not going to go so far as to add a nested `has_attr` struct at this point from a style perspective.

hughcars (approved):

PR #209, opened 2024-03-12
Use RT spaces for electrostatics error estimation

hughcars:
Change the smoothed flux field used for error estimation in electrostatics from H1d to RT. The RT field is stored only in the estimator class similar to the previous H1d due to not being needed otherwise in the laplace operator class. 

The reduction in regularity reduces false positives at material interfaces. 

Before:
Screenshot 2024-03-07 at 10 41 27 AM
After:
Screenshot 2024-03-12 at 11 32 52 AM

These come from slightly different meshes, but the red on the bottom half of the before image corresponds to a silicon air interface that should not cause the estimate to light up.


sebastiangrimberg (approved):
Looks good!

PR #211, opened 2024-03-14
Improve field energy integration for electrostatic and magnetostatic simulations

sebastiangrimberg:
Use the scalar voltage (H1) and vector potential (Nedelec) fields, which typically have smaller space dimension than the corresponding electric field (Nedelec) and magnetic flux density (Raviart-Thomas).

This can actually become a bottleneck for large capacitance or inductance matrix sizes, where `O(N^2)`, with `N` the number of terminals, energy calculations are needed.


hughcars (approved):
LGTM

PR #214, opened 2024-03-19
Fix cryptic error for empty arrays in JSON config file

sebastiangrimberg:
Previously, an (incorrect) configuration file array specified as `"Attributes": []` would lead to an error like:

```
Verification failed: (ec == std::errc()) is false:
 --> Invalid integer conversion in range expansion!
 ... in function: palace::{anonymous}::PreprocessFile(const char*)::<lambda(std::string_view)>
 ... in file: /apps/palace/palace/palace/utils/iodata.cpp:77
```

This case is now handled and checked for in the JSON Schema during `--dry-run`.

### DIFF
--- a/scripts/schema/config/domains.json
+++ b/scripts/schema/config/domains.json
@@ -17,13 +17,7 @@
         "required": ["Attributes"],
         "properties":
         {
-          "Attributes":
-          {
-            "type": "array",
-            "additionalItems": false,
-            "items": { "type": "integer"},
-            "minItems": 1
-          },
+          "Attributes": { "$ref": "#/$defs/Attributes" },
           "Permeability":
           {
             "oneOf" :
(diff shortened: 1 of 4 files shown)


scripts/schema/config/domains.json
```diff
@@ -141,5 +130,15 @@
...
+  "$defs":
+  {
+    "Attributes":
+    {
+      "type": "array",
+      "additionalItems": false,
+      "items": { "type": "integer"},
+      "minItems": 1
+    }
   }
```
hughcars, line 143: If we end up with more of these shared across configs, might be worth bundling into a common/utilities section rather than copy paste. With only one duplicate though, not worth doing yet.

hughcars (approved):
Tested on `spheres.json` changing to an `"Attributes": []`, works as intended.

PR #215, opened 2024-03-20
Collection of minor bug fixes and warning checks

sebastiangrimberg:
From the commit history:
- Fix QFunction bugs caught with Address Sanitizer
- Fix unsigned-signed integer comparison warnings
- Remove mpirun --map-by defaults
- Overflow check for large COO sparse matrix assembly and quadrature data setup
- Fix heap overflow in libCEED quadrature data setup
- Fix possible bug for integrators without a coefficient specified
- Fix std::regex issue observed when stripping comments from large config files


hughcars (approved):
Tested with asan on `cavity_pec.json`. LGTM

Issue #216, opened 2024-03-22
AMS without Multigrid fails


hughcars:
Setting `"MGMaxLevels": 1` in the `cavity_pec.json` example causes an error to throw in `rap.cpp`:

```
Verification failed: (cA) is false:
 --> ParOperator::ParallelAssemble requires A as an hypre::HypreCSRMatrix or ceed::Operator!
 ... in function: mfem::HypreParMatrix &palace::ParOperator::ParallelAssemble(bool) const
 ... in file: /Users/hughcars/AWS/palace/palace/linalg/rap.cpp:92

Backtrace:
0) [0x107eb22df]: mfem::mfem_error(char const*)
1) [0x102d93147]: palace::ParOperator::ParallelAssemble(bool) const
2) [0x102d466fb]: palace::ParOperator::StealParallelAssemble(bool) const
3) [0x102d45c07]: palace::HypreAmsSolver::ConstructAuxiliaryMatrices(palace::FiniteElementSpace&, palace::AuxiliaryFiniteElementSpace&)
4) [0x102d45247]: palace::HypreAmsSolver::HypreAmsSolver(palace::FiniteElementSpace&, palace::AuxiliaryFiniteElementSpace&, int, int, bool, bool, bool, int)
5) [0x102d462b7]: palace::HypreAmsSolver::HypreAmsSolver(palace::FiniteElementSpace&, palace::AuxiliaryFiniteElementSpace&, int, int, bool, bool, bool, int)
6) [0x102d77503]: palace::HypreAmsSolver::HypreAmsSolver(palace::IoData const&, bool, palace::FiniteElementSpace&, palace::AuxiliaryFiniteElementSpace&, int)
7) [0x102d77123]: std::__1::__unique_if<palace::HypreAmsSolver>::__unique_single std::__1::make_unique[abi:ue170006]<palace::HypreAmsSolver, palace::IoData const&, bool, palace::FiniteElementSpace&, palace::AuxiliaryFiniteElementSpace&, int const&>(palace::IoData const&, bool&&, palace::FiniteElementSpace&, palace::AuxiliaryFiniteElementSpace&, int const&)
[...]

PR #217, opened 2024-03-22
Add owning and non constructors from mfem::SparseMatrix

hughcars:
*Description of changes:*
Add owning and non owning constructors for `HypreCSRMatrix` from `mfem::SparseMatrix`.

*Issue #, if available:*
https://github.com/awslabs/palace/issues/216

### DIFF
--- a/palace/linalg/ams.cpp
+++ b/palace/linalg/ams.cpp
@@ -111,8 +111,8 @@ void HypreAmsSolver::ConstructAuxiliaryMatrices(FiniteElementSpace &nd_fespace,
     pi.SetAssemblyLevel(mfem::AssemblyLevel::LEGACY);
     pi.Assemble(skip_zeros_interp);
     pi.Finalize(skip_zeros_interp);
-    ParOperator RAP_Pi(std::make_unique<hypre::HypreCSRMatrix>(*pi.LoseMat()), h1d_fespace,
-                       nd_fespace, true);
+    ParOperator RAP_Pi(std::make_unique<hypre::HypreCSRMatrix>(std::move(*pi.LoseMat())),
+                       h1d_fespace, nd_fespace, true);
     Pi = RAP_Pi.StealParallelAssemble();
     if (cycle_type >= 10)
     {
--- a/palace/linalg/ams.cpp
+++ b/palace/linalg/ams.cpp
@@ -111,8 +111,8 @@ void HypreAmsSolver::ConstructAuxiliaryMatrices(FiniteElementSpace &nd_fespace,
     pi.SetAssemblyLevel(mfem::AssemblyLevel::LEGACY);
     pi.Assemble(skip_zeros_interp);
     pi.Finalize(skip_zeros_interp);
-    ParOperator RAP_Pi(std::make_unique<hypre::HypreCSRMatrix>(std::move(*pi.LoseMat())),
-                       h1d_fespace, nd_fespace, true);
+    ParOperator RAP_Pi(std::make_unique<hypre::HypreCSRMatrix>(pi.SpMat()), h1d_fespace,
+                       nd_fespace, true);
     Pi = RAP_Pi.StealParallelAssemble();
     if (cycle_type >= 10)
     {
--- a/palace/linalg/ams.cpp
+++ b/palace/linalg/ams.cpp
@@ -6,6 +6,7 @@
 #include "fem/bilinearform.hpp"
 #include "fem/fespace.hpp"
 #include "fem/integrator.hpp"
+#include "linalg/hypre.hpp"
 #include "linalg/rap.hpp"
 #include "utils/omp.hpp"
 
@@ -110,7 +111,7 @@ void HypreAmsSolver::ConstructAuxiliaryMatrices(FiniteElementSpace &nd_fespace,
     pi.SetAssemblyLevel(mfem::AssemblyLevel::LEGACY);
     pi.Assemble(skip_zeros_interp);
     pi.Finalize(skip_zeros_interp);
-    ParOperator RAP_Pi(std::unique_ptr<mfem::SparseMatrix>(pi.LoseMat()), h1d_fespace,
+    ParOperator RAP_Pi(std::make_unique<hypre::HypreCSRMatrix>(*pi.LoseMat()), h1d_fespace,
                        nd_fespace, true);
     Pi = RAP_Pi.StealParallelAssemble();
     if (cycle_type >= 10)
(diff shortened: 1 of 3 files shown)


palace/linalg/hypre.cpp
```diff
@@ -47,6 +47,31 @@ HypreCSRMatrix::HypreCSRMatrix(hypre_CSRMatrix *mat) : mat(mat)
   width = hypre_CSRMatrixNumCols(mat);
 }
 
+HypreCSRMatrix::HypreCSRMatrix(mfem::SparseMatrix &m)
+  : palace::Operator(m.Height(), m.Width())
+{
+  // Initialize the internal hypre matrix.
```
sebastiangrimberg, line 53: Hypre's memory location always matches MFEM's (Hypre uses device when MFEM does). So, we can do this without modifying the Hypre matrix's memory location.

```suggestion
  mat = hypre_CSRMatrixCreate(height, width, m.NumNonZeroElems());
  hypre_CSRMatrixI(mat) = m.ReadWriteI();
  hypre_CSRMatrixJ(mat) = m.ReadWriteJ();
  hypre_CSRMatrixData(mat) = m.ReadWriteData();
  hypre_CSRMatrixInitialize(mat);
```

Also, `hypre_CSRMatrixDestroy` always calls 'free` on the I pointer, so I think you either need to set a flag and set it to NULL in the`HypreCSRMatrix` destructor before calling the Hypre free destroy function, or just explicitly copy into it. You need to also set `hypre_CSRMatrixOwnsData(mat) = 0` here.

palace/linalg/ams.cpp
```diff
@@ -110,7 +111,7 @@ void HypreAmsSolver::ConstructAuxiliaryMatrices(FiniteElementSpace &nd_fespace,
     pi.SetAssemblyLevel(mfem::AssemblyLevel::LEGACY);
     pi.Assemble(skip_zeros_interp);
     pi.Finalize(skip_zeros_interp);
-    ParOperator RAP_Pi(std::unique_ptr<mfem::SparseMatrix>(pi.LoseMat()), h1d_fespace,
+    ParOperator RAP_Pi(std::make_unique<hypre::HypreCSRMatrix>(*pi.LoseMat()), h1d_fespace,
```
sebastiangrimberg, line 114: I would prefer to just get rid of the owning `HypreCSRMatrix` constructor altogether, and just use `pi.SpMat()` here.

palace/linalg/hypre.cpp
hughcars: I would rather only have the owning construction then, as that avoids the need for any flag, or creating a new array and explicit copying of the contents of `I`. To construct from a `SparseMatrix` would always require `std::move` and the incoming `SparseMatrix` would then continue to point to the data arrays, but no longer own the data. Works with `*LoseMat()` or with `SpMat()` then, but `LoseMat` seems more expressive. See https://github.com/awslabs/palace/pull/217/commits/3d05f08ff404af4451d49a78d1c022d0200dec4e

sebastiangrimberg: The flag is already there in `hypre_CSRMatrixOwnsData(mat)` isn't it? So it would just involve checking the value of the flag in the destructor and setting ` hypre_CSRMatrixI(mat) = nullptr` if false.

I like and suggested the non-owning approach since it's consistent with the `HypreVector` above. I'm not sure I understand the point about always needing `std::move`. Your non-owning constructor can take a (const) reference to a `mfem::SparseMatrix` and just wrap the data pointers. Then you can call this constructor with `BilinearForm::SpMat()` without any transfer of ownership.

sebastiangrimberg: The transfer of ownership creates issues since MFEM and Hypre may be using different memory allocators. In general we want to have the same library which is allocating memory be the same one which is destroying it to avoid problems down the road. For example, consider MFEM using a managed memory backend but Hypre not, or vice versa. The libraries might be doing something more sophisticated with memory allocation/deallocation and assume that when you are using the libraries' memory deallocation routines, you have used the corresponding allocation routines.

palace/linalg/hypre.cpp
```diff
@@ -47,8 +47,24 @@ HypreCSRMatrix::HypreCSRMatrix(hypre_CSRMatrix *mat) : mat(mat)
...
+
+  hypre_CSRMatrixInitialize(mat);
+}
+
 HypreCSRMatrix::~HypreCSRMatrix()
 {
+  if (hypre_CSRMatrixOwnsData(mat) == false)
+  {
+    hypre_CSRMatrixI(mat) = nullptr;
+    hypre_CSRMatrixRownnz(mat) = nullptr;
```
sebastiangrimberg, line 66: I think this still needs to get free'd, as we aren't allocating or freeing it?

hughcars: I see that my comment on the previous never posted for some reason, but here's the rub on using `hypre_CSRMatrixOwnsData(mat)` as the flag for deleting `I` and `NNZ`.

Then that makes no difference from calling `hypre_CSRMatrixDestroy(mat);`. There's 3 scenarios:
1. Built from `*mat` where `hypre_CSRMatrixOwnsData(mat) == true` -> I, J, Data, NNZ all destroyed in `hypre_CSRMatrixDestroy`.
2. Built from `*mat` where `hypre_CSRMatrixOwnsData(mat) ==false` -> Before this, I and NNZ would be deleted by `hypre_CSRMatrixDestroy`, but not J or Data. After this change, they would not be deleted anywhere. 
3. Built from `&m`, then `hypre_CSRMatrixOwnsData(mat) == false`, and I and NNZ are not deleted here, as they're owned by `m` in the first place.

The use of `hypre_CSRMatrixOwnsData(mat)` as the flag introduces the conflict between 2 and 3. If using 
```
      HYPRE_MemoryLocation memory_location = hypre_CSRMatrixMemoryLocation(mat);
      hypre_TFree(hypre_CSRMatrixI(mat),      memory_location);
      hypre_TFree(hypre_CSRMatrixRownnz(mat), memory_location);
 ```
 as in `HYPRE_Int hypre_CSRMatrixDestroy( hypre_CSRMatrix *matrix )`, then the non-owning of `m`, and that it is a `mfem::SparseMatrix` is being ignored. A flag of `bool hypre_managed_memory;` set to false if entering through the `m` channel would suffice, but I had assumed based on your above that you thought that unnecessary, but have gone ahead and put it in.

sebastiangrimberg:
I just rebased on main and added https://github.com/awslabs/palace/pull/217/commits/35370e759d0dfa46fcb41ee76e5c905e9f9a3b5a, which just fixes in case Hypre is built with 64-bit integers. Let me know if that looks OK and I will approve.

sebastiangrimberg (approved):

PR #220, opened 2024-03-25
Add `"EstimatorMG"` option for estimation mass matrix linear solve

sebastiangrimberg:
For electrostatic problems, PR https://github.com/awslabs/palace/pull/209 made the change from a global vector H1 project to a global RT-space projection, which increased the time spent in the linear solve by a factor of 5 or more. This PR is the result of some experimentation with how to speed up the linear solve associated with error estimation.

The current default is still to use Jacobi-preconditioned CG, but now a runtime option can switch this back to AMG (with matrix-free p-MG when order > 1) for problems where the increased memory overhead is worth the improved convergence rate.

### DIFF
--- a/docs/src/config/solver.md
+++ b/docs/src/config/solver.md
@@ -337,6 +337,9 @@ directory specified by [`config["Problem"]["Output"]`](problem.md#config%5B%22Pr
     "PCSide": <string>,
     "DivFreeTol": <float>,
     "DivFreeMaxIts": <float>,
+    "EstimatorTol": <float>,
+    "EstimatorMaxIts": <float>,
+    "EstimatorMG": <bool>,
     "GSOrthogonalization": <string>
 }
 ```
@@ -447,6 +450,9 @@ error estimate calculation.
 
 `"EstimatorMaxIts" [1000]` :  Maximum number of iterations for flux projection use in the
 error estimate calculation.
+
+`"EstimatorMG" [false]` :  Set to true in order to enable multigrid preconditioner with AMG
+coarse solve for the error estimate linear solver, instead of just Jacobi.
 
 `"GSOrthogonalization" ["MGS"]` :  Gram-Schmidt variant used to explicitly orthogonalize
 vectors in Krylov subspace methods or other parts of the code.
--- a/docs/src/config/solver.md
+++ b/docs/src/config/solver.md
@@ -448,7 +448,7 @@ the eigenmode simulation type.
 `"EstimatorTol" [1e-6]` :  Relative tolerance for flux projection used in the
 error estimate calculation.
 
-`"EstimatorMaxIts" [1000]` :  Maximum number of iterations for flux projection use in the
+`"EstimatorMaxIts" [10000]` :  Maximum number of iterations for flux projection use in the
 error estimate calculation.
 
 `"EstimatorMG" [false]` :  Set to true in order to enable multigrid preconditioner with AMG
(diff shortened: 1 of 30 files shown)


palace/drivers/electrostaticsolver.cpp
```diff
@@ -65,19 +72,24 @@ ElectrostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
                linalg::Norml2(laplaceop.GetComm(), V[step]),
                linalg::Norml2(laplaceop.GetComm(), RHS));
 
+    // Calculate and record the error indicators.
+    Mpi::Print(" Updating solution error estimates\n");
+    estimator.AddErrorIndicator(V[step], indicator);
```
hughcars, line 77: The original design had the estimation as part of the main loop, but I moved it to post processing at your request, why the move back?

hughcars (commented):
Looks fine, my only issue is at the highest level: I think the default is the wrong way round, that the MG solve should be the default. The number of issues with not converging the system giving less trust of the indicator, or the run time going too long seem to have out numbered cases where running out of memory in the estimation solve is an issue.

If a user is running up against memory limitations, they'll get a comprehensible OOM and then they have the low memory, longer runtime option to turn on, but most cases aren't pushing the absolute memory limits, so getting the answer faster, and the estimator more likely to converge is worth it.

sebastiangrimberg:
Unfortunately, using MG uses more memory and in all the cases I've tested is actually slower than Jacobi. Hence the desire to keep the default as Jacobi. Probably for problems where Jacobi does not converge in many iterations there would be a speed benefit to MG but I've got to test that.

palace/drivers/electrostaticsolver.cpp
sebastiangrimberg: When the estimation is fast, it is nice to have at the end I think since the solution vectors are stored. However when it's slow, this leads to the code appearing to "hang" at the end of the simulation when the error estimates are calculated for all vectors which is remedied when they are updated one at a time in the loop. Also, if the error estimation is going to lead to an out of memory error, it's nice to run into that earlier rather than only after all the solves have occurred. That's my thinking.

sebastiangrimberg:
Update here: Even for the blog post example, with p = 3 on a tetrahedral mesh, the error estimation linear solve takes ~200 iterations to convergence in 3.1 s with Jacobi-PCG, vs. 70 iterations in 5.3 s with MG-PCG (5x cost per iteration for 3-level p-MG + AMG coarse solve, vs. Jacobi seems reasonable). There might be more to investigate as to why the iteration count is still 70 vs. something much smaller, but playing around with the fine-level p-multigrid smoothing settings does indeed drop down the iterations at the expense of more cost per iteration.

In all cases Jacobi is the fastest. This also matches what MFEM uses in their time-dependent Maxwell solver (explicit time stepping). I'm inclined based on this to just increase the maximum number of iterations to 5k or 10k so users are not met with annoying error messages. The iterations are (very) cheap. Thoughts?

sebastiangrimberg:
Some final numbers here before merging: For an eigenmode example with `"Order": 6", tetrahedral mesh, the tradeoff between using Jacobi and multigrid preconditioning for the estimator linear solve appears about even. Roughly 2500 iterations for the former vs. 600 for the latter, taking the same amount of time. The number of iterations for the latter can be futher decreased by playing around with the multigrid smoother and AMG settings, but in general the total time is constant. So, we will leave the default as Jacobi, with increased maximum iteration count, for now as planned.

docs/src/config/solver.md
```diff
@@ -448,6 +451,9 @@ error estimate calculation.
 `"EstimatorMaxIts" [1000]` :  Maximum number of iterations for flux projection use in the
```
hughcars, line 451: 10000

hughcars (approved):
LGTM, just missing a change in the `.md`.

PR #224, opened 2024-04-04
Relax requirements for invalid boundary attributes specified for certain boundary conditions

sebastiangrimberg:
Farfield/absorbing boundary conditions and boundary postprocessing can just warn and and ignore invalid boundary attributes. This doesn't pose too much of a risk for users and warnings are sufficiently visible in the log file. This is a usability improvement when copy-pasting large numbers of boundary attributes for models when some domains may not be meshed or may be removed due to no material properties being specified.


sebastiangrimberg:
> Are there other bcs that might have a similar fallback?

Sorry, missed responding to this comment. I don't think so, previous behavior was just for Dirichlet BC and these were adding in the same spirit. Other BCs are more "active" and require parameters which led me to believe that users would/shou;d be more careful in prescribing them to actual boundaries in the model.

PR #225, opened 2024-04-08
Fix timer value of `Timer::TOTAL` for AMR simulations

sebastiangrimberg:
Fix after https://github.com/awslabs/palace/pull/212. The total time value for AMR simulations, when reduction is performed more than once, was previously double counting and printing an incorrect value.


kllrak:
Did not build to try this, but the logic of the changes makes sense to me and should fix what I was seeing.

kllrak (approved):

PR #242, opened 2024-05-02
Snake case consistency

sebastiangrimberg:
We have slowly moved towards consistent use of snake case across all variable naming, especially in the higher level classes and functions where variable names are longer and more descriptive. Though not explicitly enforced in the developer documentation or in PR reviews, this PR goes through some old code and updates variable names to be match the guidance.

The same change is also made for some output variables, namely the field names in the ParaView visualization files.

One name left unchanged is `iodata`, which could/should be renamed to `io_data`. I didn't have particular feelings either way, but if changing it is preferable I can make that change here as well.

Note: Based on https://github.com/awslabs/palace/pull/240

### DIFF
--- a/palace/drivers/magnetostaticsolver.cpp
+++ b/palace/drivers/magnetostaticsolver.cpp
@@ -26,41 +26,41 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
   // handled eliminating the rows and columns of the system matrix for the corresponding
   // dofs.
   BlockTimer bt0(Timer::CONSTRUCT);
-  CurlCurlOperator curlcurlop(iodata, mesh);
-  auto K = curlcurlop.GetStiffnessMatrix();
-  const auto &Curl = curlcurlop.GetCurlMatrix();
-  SaveMetadata(curlcurlop.GetNDSpaces());
+  CurlCurlOperator curlcurl_op(iodata, mesh);
+  auto K = curlcurl_op.GetStiffnessMatrix();
+  const auto &Curl = curlcurl_op.GetCurlMatrix();
+  SaveMetadata(curlcurl_op.GetNDSpaces());
 
   // Set up the linear solver.
-  KspSolver ksp(iodata, curlcurlop.GetNDSpaces(), &curlcurlop.GetH1Spaces());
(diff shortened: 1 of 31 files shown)


palace/drivers/magnetostaticsolver.cpp
```diff
@@ -26,61 +26,61 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+    Mpi::Print("\nIt {:d}/{:d}: Index = {:d} (elapsed time = {:.2e} s)\n", step + 1, n_step,
                idx, Timer::Duration(Timer::Now() - t0).count());
 
     // Form and solve the linear system for a prescribed current on the specified source.
     Mpi::Print("\n");
     A[step].SetSize(RHS.Size());
     A[step].UseDevice(true);
     A[step] = 0.0;
-    curlcurlop.GetExcitationVector(idx, RHS);
+    curlcurl_op.GetExcitationVector(idx, RHS);
```
hughcars, line 71: Not `curl_curl_op`?

palace/models/postoperator.cpp
```diff
@@ -560,16 +562,16 @@ std::complex<double> PostOperator::GetSParameter(const WavePortOperator &wave_po
...
-    Sij.real(Sij.real() - 1.0);
+    S_ij.real(S_ij.real() - 1.0);
   }
   // Port de-embedding: S_demb = S exp(ikₙᵢ dᵢ) exp(ikₙⱼ dⱼ) (distance offset is default 0
   // unless specified).
-  Sij *= std::exp(1i * src_data.kn0 * src_data.d_offset);
-  Sij *= std::exp(1i * data.kn0 * data.d_offset);
-  return Sij;
+  S_ij *= std::exp(1i * src_data.kn0 * src_data.d_offset);
+  S_ij *= std::exp(1i * data.kn0 * data.d_offset);
```
hughcars, line 573: So the convention (which I agree with), is not snake once within a subscript `_`. Namely, not `S_i_j`.

hughcars (commented):
Two tiny clarifications, but LGTM 🐍

palace/drivers/magnetostaticsolver.cpp
sebastiangrimberg: I could go either way on this really but `curlcurl` seemed OK as a variable name. The big gain is separating `op` from the core subject.

hughcars (approved):

PR #245, opened 2024-05-06
Revisiting boundary condition warning messages from #224

sebastiangrimberg:
- Add surface impedance and surface conductivity BCs to allow invalid attributes, following https://github.com/awslabs/palace/pull/224.
- Fix invalid BC warning messages to be more concise.
- Update `Mpi::Warning` color to highlight warnings and increase column width for `PrettyPrint` to 80 characters from 60.

Note: Based on https://github.com/awslabs/palace/pull/242

### DIFF
--- a/palace/utils/communication.hpp
+++ b/palace/utils/communication.hpp
@@ -5,6 +5,7 @@
 #define PALACE_UTILS_COMMUNICATION_HPP
 
 #include <complex>
+#include <fmt/color.h>
 #include <fmt/format.h>
 #include <fmt/printf.h>
 #include <fmt/ranges.h>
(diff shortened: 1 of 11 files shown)


palace/utils/communication.hpp
```diff
@@ -357,7 +358,7 @@ class Mpi
   template <typename... T>
   static void Warning(MPI_Comm comm, fmt::format_string<T...> fmt, T &&...args)
   {
-    Print(comm, "\nWarning!\n");
+    Print(comm, "\n{}\n", fmt::styled("--> Warning!", fmt::fg(fmt::color::yellow)));
```
hughcars, line 361: Nice! I did not appreciate that colouring was something we could do with `fmt`.

hughcars (approved):
LGTM

palace/utils/communication.hpp
sebastiangrimberg: Yeah! I wondered about adding it in more places, for example adding some wrappers for `MFEM_ABORT`/`MFEM_VERIFY`/`MFEM_ASSERT` which print in red. But I think saving that for later is a better idea.

Issue #221, opened 2024-04-03
Add AMR examples to regression tests


sebastiangrimberg:
We should update the examples (some or all) to use AMR in order to test conforming and non-conforming AMR in the regression test suite/CI.

PR #255, opened 2024-05-29
AMR example and test coverage

simbilod:
*Description of changes:*

Adds a new section to the "cavity eigenmodes" example showing convergence behaviour with adaptive mesh refinement and enabling AMR test coverage (as discussed in #221)

* New function in `cavity.jl` to setup the simulation with AMR
* Analysis and plotting code in `amr.jl`; parameters are chosen for quick execution, but still showing the effect
* New section in `cavity.md` documentation showcasing the plots
* Config files, mesh files, new tests, fix bug in testcase to cover AMR

New doc section:

image

*Issue #, if available:*

Addresses #221 

*Disclaimer*


hughcars:
Hi @simbilod, not sure if you're still working on this but it looks like a great contribution to the documentation: it's nice to see amr gaining traction. I think we'll probably want to expand this in a few ways though before any merge:
* It would be good to use amr on the cpw example, as they have interior boundary conditions. These have historically been a bigger stress test for AMR so would be a good place to demonstrate it rather than for the cavity problem where the smoothness of the solution means AMR should refine fairly uniformly.
* The results of the simulation should be baselined and used for a regression example. This would mean deciding on some appropriate tolerance/max iteration for the AMR to ensure the job is small enough to run, but large enough to be interesting.
The figures here already are looking great though, and Palace will be improved by their addition. If you have any questions or need guidance on any of that I'm happy to help.

simbilod:
> Hi @simbilod, not sure if you're still working on this but it looks like a great contribution to the documentation: it's nice to see amr gaining traction. I think we'll probably want to expand this in a few ways though before any merge:
> 
> * It would be good to use amr on the cpw example, as they have interior boundary conditions. These have historically been a bigger stress test for AMR so would be a good place to demonstrate it rather than for the cavity problem where the smoothness of the solution means AMR should refine fairly uniformly.
> * The results of the simulation should be baselined and used for a regression example. This would mean deciding on some appropriate tolerance/max iteration for the AMR to ensure the job is small enough to run, but large enough to be interesting.
>   The figures here already are looking great though, and Palace will be improved by their addition. If you have any questions or need guidance on any of that I'm happy to help.

Hi @hughcars , thanks for the feedback! Yes showcasing a problem that benefits more from AMR is a good idea. I can take another crack in the coming week or so.

simbilod:
Closing to redo with provided feedback down the line

PR #260, opened 2024-06-05
Mesh preprocessing operations

sebastiangrimberg:
- Convert all elements to simplices (enable with `"MakeSimplex"`)
- Convert all elements to hexahedra (enable with `"MakeHexahedral"`, only for tet meshes currently, compatible with `"MakeSimplex"`)
- Serial refinement levels (`"SerialUniformLevels"`)

All mesh conversions work with high-order curved meshes by interpolating the nodal grid function between meshes using GSLIB. Also exposes via configuration file a number of mesh preprocessing options which used to be hard-coded.


palace/utils/geodata.cpp
```diff
@@ -78,9 +83,18 @@ std::unique_ptr<mfem::ParMesh> ReadMesh(MPI_Comm comm, const IoData &iodata, boo
   const auto &refinement = iodata.model.refinement;
   const bool use_amr = refinement.max_it > 0;
   const bool use_mesh_partitioner = !use_amr || !refinement.nonconformal;
+  MPI_Comm node_comm;
+  if (!use_mesh_partitioner)
+  {
+    MPI_Comm_split_type(comm, MPI_COMM_TYPE_SHARED, Mpi::Rank(comm), MPI_INFO_NULL,
+                        &node_comm);
+  }
```
hughcars, line 91: To check I am getting this correctly, the idea is to split the communicator into N communicators, one for each node. Then within each communicator, only the root rank loads the serial, thereby avoiding the memory overloading issue of every rank on a node loading the mesh for preprocessing purposes.

palace/utils/geodata.cpp
```diff
@@ -89,55 +103,114 @@ std::unique_ptr<mfem::ParMesh> ReadMesh(MPI_Comm comm, const IoData &iodata, boo
...
   if (use_mesh_partitioner)
   {
     pmesh = DistributeMesh(comm, smesh, partitioning.get(), iodata.problem.output);
   }
   else
   {
+    // Send the preprocessed serial mesh and partitioning as a byte string.
+    constexpr bool generate_edges = false, refine = true, fix_orientation = false;
+    std::string so;
+    int slen = 0;
```
hughcars, line 176: I think this implicitly is setting a limit of 2.147483647 GB for any mesh (after modifications). I'm not sure if this is a limit that might be practically hit, it seems pretty high. Could consider using `std::size_t` directly, as that would be 4.28. However, the catch between `int` and `std::size_t` for the overflow ought to warn if the size is between 2.14 and 4.28 though, so we'd get a good amount of warning. And if the byte string is going past 4.28 GB then this whole approach would need reconsidering anyway.

palace/utils/geodata.cpp
```diff
@@ -89,55 +103,114 @@ std::unique_ptr<mfem::ParMesh> ReadMesh(MPI_Comm comm, const IoData &iodata, boo
...
+      MFEM_VERIFY(so.size() == (std::size_t)slen, "Overflow in stringbuffer size!");
+    }
+    Mpi::Broadcast(1, &slen, 0, node_comm);
+    if (so.empty())
+    {
+      so.resize(slen);
+    }
+    Mpi::Broadcast(slen, so.data(), 0, node_comm);
+    if (!smesh)
+    {
```
hughcars, line 198: Maybe make this a `MFEM_VERIFY` the comment above about `smesh.reset()` means the logic requires it by empty so might as well enforce and then drop the if branch.

palace/utils/geodata.cpp
```diff
@@ -1571,7 +1615,344 @@ std::unique_ptr<mfem::Mesh> LoadMesh(const std::string &path, bool remove_curvat
...
+      nodes->GetSubVector(vdofs, vals);
+
+      // Find all child elements of this parent in the split mesh and restore their correct
+      // original attribute.
+      int attr = new_mesh.GetAttribute(start);
+      int end = start;
+      while (end < new_mesh.GetNE() && new_mesh.GetAttribute(end) == attr)
+      {
+        new_mesh.SetAttribute(end++, orig_mesh_attr[e]);
+      }
```
hughcars, line 1859: This loop works because the elements are added as `[[new_00, new_01, new_02, ...], [new_10, new_11, new_12,...],...]`, namely all new elements that made up an original element are contiguous. Also recursive, so tet -> hex -> tet will still have this contiguous structure.

palace/utils/geodata.cpp
```diff
@@ -1571,7 +1615,344 @@ std::unique_ptr<mfem::Mesh> LoadMesh(const std::string &path, bool remove_curvat
...
+
+      // Special case: Parent element is unsplit, no interpolation is needed.
+      if (end == start + 1)
+      {
+        new_fespace.GetElementVDofs(start, vdofs);
+        new_nodes.SetSubVector(vdofs, vals);
+        start = end;
+        continue;
+      }
```
hughcars, line 1868: In the MakeSimplex and MakeHex methods, neither allow mixed presently, so is this an edge case that can ever be hit? I.e. all tets get converted, and all hexes get converted. Is this just anticipating mixed meshes?

palace/utils/geodata.cpp
```diff
@@ -1571,7 +1615,344 @@ std::unique_ptr<mfem::Mesh> LoadMesh(const std::string &path, bool remove_curvat
...
+        mfem::ElementTransformation &T = *new_mesh.GetElementTransformation(i);
+        T.Transform(fe.GetNodes(), pointmat);
+        for (int j = 0; j < pointmat.Width(); j++)
+        {
+          for (int d = 0; d < sdim; d++)
+          {
+            // Use default ordering byNODES.
+            xyz(d * npts + offset + j) = pointmat(d, j);
+          }
+        }
```
hughcars, line 1889: For this loop, rather than jumping back and forwards by npts in xyz for each write, it might be more cache coherent to have the d loop on the outer. This loops the slow way through pointmat (assuming column major), but pointmat is smaller than xyz, so the npt jumps about in xyz might be causing more reads etc.

hughcars (commented):
A few minor questions and comments, but LGTM.

palace/fem/interpolator.cpp
```diff
@@ -107,4 +120,184 @@ std::vector<std::complex<double>> InterpolationOperator::ProbeField(const GridFu
...
+  {
+    xyz.MakeRef(*dest_mesh.GetNodes(), 0, dest_mesh.GetNodes()->Size());
+    ordering = dest_mesh.GetNodes()->FESpace()->GetOrdering();
+  }
+  else
+  {
+    int npts = 0, offset = 0;
+    for (int i = 0; i < dest_mesh.GetNE(); i++)
+    {
+      npts += V.FESpace()->GetFE(i)->GetNodes().GetNPoints();
```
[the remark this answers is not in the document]
sebastiangrimberg: Yeah the latter, we just find the duplicate shared points nodes between elements and they are disregarded because we `SetSubVector` later on. Certainly a future optimization could be to remove these duplicates from the GSLIB search and just maintain the indexing separately.

palace/utils/geodata.cpp
```diff
@@ -212,7 +285,7 @@ void RefineMesh(const IoData &iodata, std::vector<std::unique_ptr<mfem::ParMesh>
   // reorient only the coarse mesh so that the refinements are still true refinements of
   // the original mesh (required for geometric multigrid). Otherwise, it happens after
   // refinement.
-  if (iodata.model.reorient_tet && mesh.capacity() > 1)
+  if (iodata.model.reorient_tet_mesh && mesh.capacity() > 1)
```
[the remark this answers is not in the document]
sebastiangrimberg: The check at https://github.com/awslabs/palace/pull/260/files#diff-0642b48eb210cdd7e98a30345d4857eb87c192985610bc00696294583cd09a43R259 should prevent that from ever being the case. The idea here is that if we are storing a multigrid hierarchy of meshes, we want to reorient the coarse mesh, otherwise we will wait until the end to do the reorientation.

palace/utils/geodata.cpp
sebastiangrimberg: Correct. Just one process on each node does the serial mesh preprocessing operations to save memory overhead, and then broadcasts it to all other processes on the node.

palace/utils/geodata.cpp
sebastiangrimberg: The problem is that the MPI interface only allows for `int` message lengths. We have an assertion to check for the overflow but indeed for a big enough mesh we would/will need to segment the message as multiple broadcast messages. Also just an aside but isnt `std::size_t` unsigned 64 bit on a 64 bit machine? If so the maximum value is well beyond 4.28e9.

palace/utils/geodata.cpp
sebastiangrimberg: I'll add a comment for this.

palace/utils/geodata.cpp
sebastiangrimberg: Good catch, this is left over from when the root process was not rebuilding the mesh.

palace/utils/geodata.cpp
sebastiangrimberg: Yeah good call about `pointmat` being small so the jumps not being a big issue. Fixed.

palace/utils/geodata.cpp
hughcars: You're right, I had just been thinking of `std::size_t` as a synonym for unsigned int, but you're right it's vastly larger, so that overflow check should be absolutely plenty.

hughcars (approved):

Issue #251, opened 2024-05-16
PCG solver did NOT converge in 50000 iterations in a coupled microstrip simulation


CosimoMV:
I tried to simulate a coupled microstrip line, but no matter what I do, I always get this error: PCG solver did NOT converge.

I tried to simulate the copper interface with PEC and a metal layer with finite conductivity, I varied the parameters  "Lc", "DivFreeMaxIts", "EstimatorMaxIt" and "MaxIts" , I tried  the solvers "SuperLU", "STRUMPACK" and "MUMPS" with both version 0.12 and 0.13, but I always get this error. I think I am doing something trivial wrong (the system I am simulating is simple), but I do not understand what the problem is.

This is an example of my .json files [simulation.json](https://github.com/awslabs/palace/files/15335726/simulation.json)

This is my mesh: [tutto.zip](https://github.com/awslabs/palace/files/15335802/tutto.zip)

sebastiangrimberg:
Hi @CosimoMV,

Looking at your mesh, it is likely that your problem could be solved by using a nicer initial mesh with more isotropic elements and smoother size grading between elements. This can be handled in Gmsh by applying size fields, see for example [this tutorial](https://gmsh.info/doc/texinfo/gmsh.html#t10). This should help the iterative solvers in Palace to converge.

CosimoMV:
Hi @sebastiangrimberg , thanks for your help. I tried to recreate the mesh as you said, but I was only getting absurdly large files. I therefore reconstructed the geometry of the coupled microstrip directly with gmesh and meshed it with these parameters:
```
Mesh.MeshSizeFromPoints=1;
Mesh.MeshSizeFromCurvature=1;
Mesh.MeshSizeExtendFromBoundary=2;
Mesh.ElementOrder=4;
```
In this way I was able to create files that were quite small and allowed Palace to converge. The problem is that the solution, apart from being unrealistic, is highly dependent on the algorithm chosen to mesh. 
In the graph:

![graph](https://github.com/awslabs/palace/assets/155521906/ac276702-c0ba-45c8-863b-fbf43d27ad13) 

are shown the |S21| (dB) vs frequency (GHz) parameters obtained with various meshing algorithms: the first curve name indicates the algorithm for surfaces, the second the one used for volumes, and the third is the name of Palace solver. The S-parameters values, apart from one case, are very different and in any case unrealistic. In all cases, the Palace solver converged.
[...]

PR #264, opened 2024-06-11
Add conductivity boundaries to Dirichlet BC list for wave ports

sebastiangrimberg:
Wave port mode calculation right now inherits all PEC boundary conditions from the 3D problem. Conductivity boundaries should also be considered as PEC for mode calculation (otherwise it was expected the user adds them with `"WavePortPEC"`).

Also updates the eigenvalue target for the shifted wave port eigemode problem to `EigenvalueSolver::WhichType::SMALLEST_REAL`, which appropriately deprioritizes evanescent modes with large imaginary part and small real part of `k_n`.

### DIFF
--- a/docs/src/config/solver.md
+++ b/docs/src/config/solver.md
@@ -138,20 +138,8 @@ number of eigenmodes of the problem. The available options are:
 
   - `"SLEPc"`
   - `"ARPACK"`
-  - `"FEAST"`
   - `"Default"` :  Use the default eigensolver. Currently, this is the Krylov-Schur
     eigenvalue solver from `"SLEPc"`.
-
-`"ContourTargetUpper" [None]` :  Specifies the upper frequency target of the contour used
-for the FEAST eigenvalue solver, GHz. This option is relevant only for `"Type": "FEAST"`.
-
-`"ContourAspectRatio" [None]` :  Specifies the aspect ratio of the contour used for the
-FEAST eigenvalue solver. This should be greater than zero, where the aspect ratio is the
-ratio of the contour width to the frequency range(`"ContourTargetUpper"` - `"Target"`).
-This option is relevant only for `"Type": "FEAST"`.
-
-`"ContourNPoints" [4]` :  Number of contour integration points used for the FEAST eigenvalue
-solver. This option is relevant only for `"Type": "FEAST"`.
 
 ### Advanced eigenmode solver options
 
--- a/palace/models/waveportoperator.cpp
+++ b/palace/models/waveportoperator.cpp
@@ -1092,7 +1092,8 @@ void WavePortOperator::SetUpBoundaryProperties(const IoData &iodata,
   // are touching and share one or more edges.
   mfem::Array<int> dbc_bcs;
   dbc_bcs.Reserve(static_cast<int>(iodata.boundaries.pec.attributes.size() +
-                                   iodata.boundaries.auxpec.attributes.size()));
+                                   iodata.boundaries.auxpec.attributes.size() +
+                                   iodata.boundaries.conductivity.size()));
   for (auto attr : iodata.boundaries.pec.attributes)
   {
     if (attr <= 0 || attr > bdr_attr_max)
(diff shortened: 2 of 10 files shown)


sebastiangrimberg:
Note: This resolves #251 and #267.

palace/utils/configfile.hpp
```diff
@@ -647,29 +656,13 @@ struct EigenSolverData
...
 
   // Eigenvalue solver type.
-  enum class Type
-  {
-    DEFAULT,
-    SLEPC,
-    ARPACK,
-    FEAST
-  };
+  using Type = WavePortData::EigenSolverType;
```
hughcars, line 659: Seems a little surprising to have the EigenSolverType a subtype within WavePortData rather than freestanding?

palace/models/waveportoperator.cpp
```diff
@@ -599,12 +599,13 @@ WavePortData::WavePortData(const config::WavePortData &data,
...
-                     // associated with shift
-  // const double c_max = mat_op.GetLightSpeedMax().Max();
-  // MFEM_VERIFY(c_max > 0.0 && c_max < mfem::infinity(),
-  //             "Invalid material speed of light detected in WavePortOperator!");
-  // mu_eps_min = 1.0 / (c_max * c_max);
+  const double c_max = mat_op.GetLightSpeedMax().Max();
+  MFEM_VERIFY(c_max > 0.0 && c_max < mfem::infinity(),
+              "Invalid material speed of light detected in WavePortOperator!");
+  mu_eps_min = 1.0 / (c_max * c_max) * 0.5;  // Add a safety factor for minimum propagation
+                                             // constant possible
```
hughcars, line 606: What's the origin of this safety factor? Is it a floating point issue?

docs/src/config/solver.md
```diff
@@ -138,21 +138,9 @@ number of eigenmodes of the problem. The available options are:
 
   - `"SLEPc"`
   - `"ARPACK"`
-  - `"FEAST"`
```
hughcars, line 141: RIP Feast.

hughcars (commented):
Based on the positive response in https://github.com/awslabs/palace/issues/251 and the lack of updates needed for the regression testing I am taking this as functional. Tiny questions but LGTM.

palace/utils/configfile.hpp
sebastiangrimberg: I agree I didn't love this but also wanted to be consistent that everywhere else `enum`s are defined in the scope of a `struct` which uses them. The fact that it is defined here instead of in `EigenSolverData` as it was previously is just because of ordering in the file. Is there a better solution?

docs/src/config/solver.md
sebastiangrimberg: It can come back when needed! Commit c435ace0 has the original code which just needs minor modifications after the linear algebra refactors of last year.

palace/models/waveportoperator.cpp
sebastiangrimberg: Yeah, if the wave port is specified on a waveguide supporting a TEM mode, then setting the safety factor of 1 will lead to the shift being directly on top of the actual eigenvalue which is going to  cause problems. Setting the shift to 0 works fine (and was what we did previously) but having some non-zero shift can speed up the eigenvalue solve convergence. It didn't make a significant difference in my testing but I figured it might as well be some positive value.

palace/utils/configfile.hpp
hughcars: Practically it doesn't make that much of a difference ultimately given this type alias, so not a big deal. It does highlight the difficulty of that convention though, as it introduces this side coupling. I would prefer the enum outside, as at this point the notion of an `EigenSolverType` is "more common", but it's just a style thing so consistency is king here.

hughcars (approved):

palace/utils/configfile.hpp
sebastiangrimberg: Yes I would support the refactor to having all enums be described outside the structs, and we could us `using` statements in the struct scope to avoid needing naming changes to access them. That is a good idea for a later PR to make it all consistent.

PR #269, opened 2024-06-21
Fix divergence cleaning when a model has no real boundaries

sebastiangrimberg:
This avoids solving a singular Laplace problem when calculating the scalar potential for a model with no real boundaries. For example, this can be observed by printing the output of divergence-free projection linear solve for the cavity example with PMC boundaries. After this PR, the divergence-free solves actually converge normally in a few iterations and the eigensolve converges as well.

Correctly handles the case of some subdomains having potentially no true dofs and just enforces a potential = 0 on the processor of lowest rank with some non-zero number of local true dofs.

### DIFF
--- a/palace/linalg/divfree.cpp
+++ b/palace/linalg/divfree.cpp
@@ -50,7 +50,7 @@ DivFreeSolver<VecType>::DivFreeSolver(
 {
   BlockTimer bt(Timer::DIV_FREE);
 
-  // If no boundaries on the mesh have been marked, add a single degree of freedo
+  // If no boundaries on the mesh have been marked, add a single degree of freedom
   // constraint so the system for the projection is not singular. This amounts to enforcing
   // a scalar potential of 0 at a point in space if it is otherwise completely
   // unconstrained.
--- a/palace/linalg/divfree.cpp
+++ b/palace/linalg/divfree.cpp
@@ -64,10 +64,10 @@ DivFreeSolver<VecType>::DivFreeSolver(
     Mpi::GlobalSum(1, &coarse_bdr_tdofs, comm);
     if (coarse_bdr_tdofs == 0)
     {
-      MPI_Comm split_comm = MPI_COMM_NULL;
-      int color = (h1_fespaces.GetFESpaceAtLevel(0).GetTrueVSize() > 0) ? 0 : MPI_UNDEFINED;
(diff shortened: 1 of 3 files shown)


palace/linalg/divfree.cpp
```diff
@@ -49,6 +49,42 @@ DivFreeSolver<VecType>::DivFreeSolver(
     int print)
 {
   BlockTimer bt(Timer::DIV_FREE);
+
+  // If no boundaries on the mesh have been marked, add a single degree of freedo
```
LeilaGhaffari, line 53: Thanks, @sebastiangrimberg ! My test is passing with this change. Just a small typo here. `freedo` -> `freedom`

sebastiangrimberg: Thanks! Resolved in [b69eeba](https://github.com/awslabs/palace/pull/269/commits/b69eebad199ebce073173cd887c69a9d11052217).

palace/linalg/divfree.cpp
```diff
@@ -49,6 +49,42 @@ DivFreeSolver<VecType>::DivFreeSolver(
...
+        "Unexpected empty list of boundary true dofs for finite element space hierarchy!");
+    HYPRE_BigInt coarse_bdr_tdofs = h1_bdr_tdof_lists[0].Size();
+    MPI_Comm comm = h1_fespaces.GetFESpaceAtLevel(0).GetComm();
+    Mpi::GlobalSum(1, &coarse_bdr_tdofs, comm);
+    if (coarse_bdr_tdofs == 0)
+    {
+      MPI_Comm split_comm = MPI_COMM_NULL;
+      int color = (h1_fespaces.GetFESpaceAtLevel(0).GetTrueVSize() > 0) ? 0 : MPI_UNDEFINED;
+      MPI_Comm_split(comm, color, Mpi::Rank(comm), &split_comm);
+      if (split_comm != MPI_COMM_NULL && Mpi::Root(split_comm))
```
hughcars, line 70: So there's only one dof on each processor with more than one dof correct?

sebastiangrimberg: Yeah this was a silly overcompilcation, thanks for pointing this out. Simplified in [ee3dbd7](https://github.com/awslabs/palace/pull/269/commits/ee3dbd73680c1c3109ccfb90569650aca1c60591).

Issue #37, opened 2023-04-17
Run `clang-tidy` in CI


nikosavola:
`clang-tidy` could be run in CI in addition to `clang-format` for enforcung the style provided in `.clang-tidy`.

Roughly the following could work:

```bash
mkdir build && cd build
cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
run-clang-tid
```

sebastiangrimberg:
Hi @nikosavola, yes this is indeed possible and made even easier so by just configuring using `-DANALYZE_SOURCES_CLANG_TIDY=ON` (or `-DANALYZE_SOURCES_CPPCHECK=ON` to do static analysis using `cppcheck`).

The reason why these are not enforced via the CI currently is that the `.clang-tidy` configuration is rather experimental. It is useful as a developer tool, but I am not yet certain that it is worth enforcing over the entire codebase. There was a relatively [recent PR](https://github.com/awslabs/palace/pull/14) which may be of interest to you which looked at the changes required to satisfy the various `clang-tidy` noise, though it was closed since it was deemed this was not a very high priority item at the time. Hope this is helpful!

sebastiangrimberg:
Hi @nikosavola, just following up here. Does my response seem reasonable? If you're interested in making a contribution to run `clang-tidy` checks in the CI that would be great and I would be happy to help in any discussions of source code changes or config file settings in order to get the current `main` to build cleanly. Otherwise, if we should keep it only as a development tool, I can close this issue.

nikosavola:
> Hi @nikosavola, just following up here. Does my response seem reasonable? If you're interested in making a contribution to run `clang-tidy` checks in the CI that would be great and I would be happy to help in any discussions of source code changes or config file settings in order to get the current `main` to build cleanly. Otherwise, if we should keep it only as a development tool, I can close this issue.

Hi, I think I can make a contribution at some point in the near future, so no need to close yet 😁

PR #14, opened 2023-02-21
Clang tidy fixes dev

hughcars:
*Description of changes:*

Fixes to get `clang-tidy` builds to pass again. Was working on another branch and need a passing ground zero to work from. Various miscellaneous changes, most interesting being adding the `NOLINT` suppression patch to mfem. Have sent a [PR upstreaming it](https://github.com/mfem/mfem/pull/3502/files).

*Issue #, if available:*
https://github.com/awslabs/palace/issues/13

PR #270, opened 2024-07-01
Add clang-tidy github action

Yaraslaut:
PR adds clang-tidy check into github action when building with clang compiler
Closes https://github.com/awslabs/palace/issues/37

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -32,5 +32,7 @@ Checks: >
     performance-*,
     mpi-*,
     openmp-*'
+WarningsAsErrors: >-
+  performance-enum-size
 HeaderFilterRegex: 'palace/drivers|palace/fem|palace/linalg|palace/utils'
 FormatStyle: 'file'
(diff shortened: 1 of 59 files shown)


hughcars:
Hi @Yaraslaut thank you for the pr. We've considered adding in clang-tidy in the past, https://github.com/awslabs/palace/issues/37 https://github.com/awslabs/palace/pull/14 as it would be good to have that extra level of linting. However, we found that the number of false positives we were getting was pretty enormous so decided to abandon the project (at least for the time being). If you were interested in taking a look at deciding on what the appropriate settings we'd want for Palace (and probably the corresponding fixes 😅), we'd be very interested in a contribution.

Yaraslaut:
> Hi @Yaraslaut thank you for the pr. We've considered adding in clang-tidy in the past, #37 #14 as it would be good to have that extra level of linting. However, we found that the number of false positives we were getting was pretty enormous so decided to abandon the project (at least for the time being). If you were interested in taking a look at deciding on what the appropriate settings we'd want for Palace (and probably the corresponding fixes 😅), we'd be very interested in a contribution.

Hi @hughcars, thanks for the update, I can fix  some of the checks and update clang-tidy file to ignore noisy one, i will try to include only easy fixes, and can increase number of checks in future PRs, if i will attempt to fix every check it will be too huge of a change for a one go

PR #273, opened 2024-07-11
[WIP] Improve integrals of scalar functions for postprocessing

sebastiangrimberg:
WIP

- [x] Thread-safety of `GridFunction` evaluation and face/boundary coefficients in `coefficient.cpp`
- [ ] Address GPU usage blocker

### DIFF
--- a/palace/fem/coefficient.hpp
+++ b/palace/fem/coefficient.hpp
@@ -121,23 +121,20 @@ public:
 
     // For interior faces, compute Jₛ = n x H = n x μ⁻¹ (B1 - B2), where B1 (B2) is B in
     // element 1 (element 2) and n points into element 1.
-    double W_data[3], VU_data[3];
-    mfem::Vector W(W_data, vdim), VU(VU_data, vdim);
+    StaticVector<3> W, VU;
     B.GetVectorValue(*FET().Elem1, FET().Elem1->GetIntPoint(), W);
     mat_op.GetInvPermeability(FET().Elem1->Attribute).Mult(W, VU);
     if (FET().Elem2)
     {
       // Double-sided, not a true boundary. Add result with opposite normal.
-      double VL_data[3];
-      mfem::Vector VL(VL_data, vdim);
+      StaticVector<3> VL;
       B.GetVectorValue(*FET().Elem2, FET().Elem2->GetIntPoint(), W);
       mat_op.GetInvPermeability(FET().Elem2->Attribute).Mult(W, VL);
       VU -= VL;
     }
 
     // Orient with normal pointing into element 1.
-    double normal_data[3];
-    mfem::Vector normal(normal_data, vdim);
+    StaticVector<3> normal;
     GetNormal(T, normal, ori);
     V.SetSize(vdim);
     linalg::Cross3(normal, VU, V);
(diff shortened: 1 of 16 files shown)


palace/fem/coefficient.hpp
```diff
@@ -34,24 +38,32 @@ namespace palace
...
   }
 
+  mfem::FaceElementTransformations &FET() { return FETs[utils::GetThreadNum()]; }
+
 public:
   BdrGridFunctionCoefficient(const mfem::ParMesh &mesh, double scaling = 1.0)
     : mesh(mesh), scaling(scaling)
   {
+    // Only use multiple entries when not on GPU (matches libCEED initialization).
+    int nt = mfem::Device::Allows(mfem::Backend::DEVICE_MASK) ? 1 : utils::GetMaxThreads();
```
Sbozzolo, line 63: We have a `device.cpp` in utils. Should we just add a function to return whether we are on GPU or not (not sure how that would look like), instead of this obscure `mfem::Device::Allows(mfem::Backend::DEVICE_MASK)` syntax?

palace/fem/integrator.hpp
```diff
@@ -33,6 +34,19 @@ struct DefaultIntegrationOrder
   static int Get(const mfem::Mesh &mesh, mfem::Geometry::Type geom);
 };
 
+double IntegrateFunction(
+    const mfem::ParMesh &mesh, const mfem::Array<int> &marker, bool bdr,
+    mfem::Coefficient &Q,
+    std::function<int(const mfem::ElementTransformation &)> GetQuadratureOrder =
+        [](const mfem::ElementTransformation &T)
+    { return DefaultIntegrationOrder::Get(T); });
+double IntegrateFunctionLocal(
```
Sbozzolo, line 43: Can you add docstrings to these functions?

palace/fem/integrator.hpp
```diff
@@ -310,7 +324,7 @@ class VectorFEBoundaryLFIntegrator : public mfem::LinearFormIntegrator
   mfem::Vector f_loc, f_hat;
 
 public:
-  VectorFEBoundaryLFIntegrator(mfem::VectorCoefficient &QG) : Q(QG) {}
+  VectorFEBoundaryLFIntegrator(mfem::VectorCoefficient &Q) : Q(Q) {}
```
Sbozzolo, line 327: Is this used?

palace/fem/integrator.hpp
```diff
@@ -325,7 +339,7 @@ class BoundaryLFIntegrator : public mfem::LinearFormIntegrator
   mfem::Vector shape;
 
 public:
-  BoundaryLFIntegrator(mfem::Coefficient &QG) : Q(QG) {}
+  BoundaryLFIntegrator(mfem::Coefficient &Q) : Q(Q) {}
```
Sbozzolo, line 342: Is this used?

Sbozzolo (commented):
I started the review and I think that this PR introduces an important regression for GPUs.

My understanding is that code like the following
```
    mfem::LinearForm pr(&nd_fespace);
    pr.AddBoundaryIntegrator(new VectorFEBoundaryLFIntegrator(nxHr_func), attr_marker);
    pr.UseFastAssembly(false);
    pr.UseDevice(false);
    pr.Assemble();
    pr.UseDevice(true);
    dot = -(pr * E.Real()) - 1i * (pr * E.Imag());
```
performs the integration on the device.

The new implementation no longer relies on this approach and performs everything on the CPU.


Unfortunately, I don't think there's an easy way to checking for this type of problems programatically. The only way I know to confirm that a given piece of code is running on GPU is to use Nsight (combined with an NVTX range). Correctness alone cannot be relied upon because the code could still be correct but not running on GPU. Memory transfers would also not be reliable because memory could be moved to the device, but the algorithm could still be running on CPU.

hughcars:
Closing as overtaken by other changes.

PR #293, opened 2024-11-05
JuliaFormatter v2.0 updates

hughcars:
JuliaFormatter 2.0 causes the existing formatting to fail, this update fixes that.


hughcars:
Override merge given the failure is caused by the intel compiler issue, and that issue is being blocked by this CI failure too.

PR #298, opened 2024-11-14
Update Github actions

hughcars:
In some macos github action builds:
```
[build-and-test-macos (gcc, openmpi, openblas, static, int64, openmp, superlu, arpack)](https://github.com/awslabs/palace/actions/runs/11825573260/job/32994205616#step:4:17)
You are using macOS 12.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
It is expected behaviour that Homebrew will be buggy and slow.
Do not create any issues about this on Homebrew's GitHub repositories.
Do not create any issues even if you think this message is unrelated.
Any opened issues will be immediately closed without response.
Do not ask for help from Homebrew or its maintainers on social media.
You may ask for help in Homebrew's discussions but are unlikely to receive a response.
Try to figure out the problem yourself and submit a fix as a pull request.
We will review it but may or may not accept it.
```
Looks like the github action runners `macos-latest-xl` might've been moved to `macos-latest-large`. Should probably add an m1 build to the CI.

### DIFF
--- a/.github/workflows/build-and-test-macos.yml
+++ b/.github/workflows/build-and-test-macos.yml
@@ -88,7 +88,7 @@ jobs:
             export FC=gfortran-12
           fi
           if [[ "${{ matrix.math-libs }}" == 'openblas' ]]; then
-            export OPENBLAS_DIR=/opt/homebrew/opt/openblas
+            export OPENBLAS_DIR=$(brew --prefix openblas)
           fi
           export NUM_PROC_BUILD=$(nproc 2> /dev/null || sysctl -n hw.ncpu)
           if [[ "$NUM_PROC_BUILD" -gt "$NUM_PROC_BUILD_MAX" ]]; then
--- a/.github/workflows/build-and-test-macos.yml
+++ b/.github/workflows/build-and-test-macos.yml
@@ -43,7 +43,7 @@ jobs:
             with-solver: strumpack
             with-eigensolver: slepc
 
-    runs-on: macos-latest-xl
+    runs-on: macos-latest-xlarge
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-test-macos.yml
```diff
@@ -53,6 +53,10 @@ jobs:
         run: |
           brew install pkg-config
 
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: '1'
```
phdum-a, line 58: Is there a recommendation from julia whether to also pin subversion (1.10 / 1.11)? I know these can have some breaking changes, although our usage might be light enough not to worry.

hughcars: I think I'd rather let it run along at 1 until there's a bug to prevent us otherwise. That way we'll naturally incorporate any updates. Our usage is very light anyway, so should be fine I think.

cmake/ExternalBLASLAPACK.cmake
```diff
@@ -163,6 +163,15 @@ else()
   else()
     set(OPENBLAS_DIR)
   endif()
+
+  if(NOT OPENBLAS_DIR STREQUAL "")
+    # If OpenBLAS was found set the vendor to avoid conflict with Accelerate on Darwin
+    set(BLA_VENDOR "OpenBLAS")
+    message(STATUS "Using BLAS/LAPACK from OpenBLAS")
+  else()
+    message(STATUS "Using BLAS/LAPACK located by CMake")
```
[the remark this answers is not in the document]
hughcars: I think this probably worth handling, but I'll do in a follow up PR, as doing it neatly is going to involve a modest amount of refactoring in this file (in particular handling the AOCL section) and then a significant amount of testing for the edges.

PR #326, opened 2025-01-10
PROM Circuit Extraction

phdum-a:
Obtain circuit from PROM operator.

Draft: Pending #302, #309 merges and expecting follow up changes from those.

### DIFF
--- a/palace/drivers/drivensolver.cpp
+++ b/palace/drivers/drivensolver.cpp
@@ -4,6 +4,12 @@
 #include "drivensolver.hpp"
 
 #include <complex>
+#include <iostream>
+#include <Eigen/Core>
+#include <Eigen/Dense>
+#include <Eigen/src/Core/IO.h>
+#include <fmt/core.h>
+#include <fmt/ostream.h>
 #include <mfem.hpp>
 #include "fem/errorindicator.hpp"
 #include "fem/mesh.hpp"
@@ -47,8 +53,9 @@ DrivenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
   double delta_omega = iodata.solver.driven.delta_f;
   double omega0 = iodata.solver.driven.min_f + step0 * delta_omega;
   bool adaptive = (iodata.solver.driven.adaptive_tol > 0.0);
-  if (adaptive && n_step <= 2)
+  if (adaptive && n_step <= 2 && !iodata.solver.driven.adaptive_circuit_synthesis)
   {
+    // If circuit synthesize, it's ok to have <=2 output samples since we resample
     Mpi::Warning("Adaptive frequency sweep requires > 2 total frequency samples!\n"
                  "Reverting to uniform sweep!\n");
     adaptive = false;
@@ -75,7 +82,7 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &space_op,
   // Initialize write directory with default path; will be changed if multiple excitations.
   PostOperator post_op(iodata, space_op, "driven");
   PostprocessPrintResults post_results(root, post_dir, post_op, space_op, excitation_helper,
-                                       n_step, iodata.solver.driven.delta_post);
+                                       n_step, iodata);
 
   // Construct the system matrices defining the linear operator. PEC boundaries are handled
   // simply by setting diagonal entries of the system matrix for the corresponding dofs.
@@ -189,7 +196,7 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &space_op,
     BlockTimer bt0(Timer::POSTPRO);
     SaveMetadata(ksp);
   }
-  post_results.PostprocessFinal(post_op, indicator);
+  post_results.PostprocessFinal(iodata, post_op, indicator);
   return indicator;
 }
 
@@ -202,7 +209,7 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
   // Initialize write directory with default path; will be changed if multiple excitations.
   PostOperator post_op(iodata, space_op, "driven");
   PostprocessPrintResults post_results(root, post_dir, post_op, space_op, excitation_helper,
-                                       n_step, iodata.solver.driven.delta_post);
+                                       n_step, iodata);
 
   // Configure PROM parameters if not specified.
   double offline_tol = iodata.solver.driven.adaptive_tol;
@@ -213,8 +220,6 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
   {
     max_size_per_excitation = 20;  // Default value
   }
-  // Maximum size — no more than nr steps needed
-  max_size_per_excitation = std::min(max_size_per_excitation, (n_step - step0));
   int convergence_memory = iodata.solver.driven.adaptive_memory;
 
   // Allocate negative curl matrix for postprocessing the B-field and vectors for the
@@ -250,15 +255,27 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
   space_op.GetWavePortOp().SetSuppressOutput(
       true);  // Suppress wave port output for offline
 
+  if (iodata.solver.driven.adaptive_circuit_synthesis)
+  {
+    auto &lumped_port_op = space_op.GetLumpedPortOp();
+    for (const auto &[port_idx, port_data] : lumped_port_op)
+    {
+      ComplexVector port_excitation_E;
+      port_excitation_E.UseDevice(true);
+      space_op.GetLumpedPortExcitationVector(port_idx, port_excitation_E, true);
+      prom_op.UpdatePROM(port_excitation_E, fmt::format("port_{:d}", port_idx));
+    }
+  }
+
   // Initialize the basis with samples from the top and bottom of the frequency
   // range of interest. Each call for an HDM solution adds the frequency sample to P_S and
   // removes it from P \ P_S. Timing for the HDM construction and solve is handled inside
   // of the RomOperator.
   int paraview_step = 0;
-  auto UpdatePROM = [&](ExcitationIdx excitation_idx, double omega)
+  auto UpdatePROM = [&](ExcitationIdx excitation_idx, double omega, int sample_idx)
   {
     // Add the HDM solution to the PROM reduced basis.
-    prom_op.UpdatePROM(E);
+    prom_op.UpdatePROM(E, fmt::format("sample_e{:d}_s{:d}", excitation_idx, sample_idx));
     prom_op.UpdateMRI(excitation_idx, omega, E);
 
     // Compute B = -1/(iω) ∇ x E on the true dofs, and set the internal GridFunctions in
@@ -284,18 +301,23 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
 
     prom_op.SetExcitationIndex(excitation_idx);  // Pre-compute RHS1
 
+    int it_sample = 1;  // Note: one-based index for printing
+    // Left freq boundary
     prom_op.SolveHDM(excitation_idx, omega0, E);
-    UpdatePROM(excitation_idx, omega0);
+    UpdatePROM(excitation_idx, omega0, it_sample);
+    it_sample++;
+    // Right freq boundary
     prom_op.SolveHDM(excitation_idx, omega0 + (n_step - step0 - 1) * delta_omega, E);
-    UpdatePROM(excitation_idx, omega0 + (n_step - step0 - 1) * delta_omega);
-    int it0 = 2;  // Nr of initial solutions in PROM
+    UpdatePROM(excitation_idx, omega0 + (n_step - step0 - 1) * delta_omega, it_sample);
+    it_sample++;
+    int it0 = it_sample;  // Nr of initial solutions in PROM
 
     // Greedy procedure for basis construction (offline phase). Basis is initialized with
     // solutions at frequency sweep endpoints.
     int memory = 0;
-    int it = it0;
     std::vector<double> max_errors = {0.0, 0.0};
-    for (; (it < max_size_per_excitation) && (memory < convergence_memory); it++)
+    for (; (it_sample <= max_size_per_excitation) && (memory < convergence_memory);
+         it_sample++)
     {
       // Compute the location of the maximum error in parameter domain (bounded by the
       // previous samples).
@@ -321,15 +343,16 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
       Mpi::Print(
           "\nGreedy iteration {:d} (PROM n = {:d}): ω* = {:.3e} GHz ({:.3e}), error = "
           "{:.3e}, memory = {:d}/{:d}\n",
-          it - it0 + 1, prom_op.GetReducedDimension(), omega_star * unit_GHz, omega_star,
+          it_sample - it0 + 1, prom_op.GetReducedDimension(), omega_star * unit_GHz, omega_star,
           max_errors.back(), memory, convergence_memory);
-      UpdatePROM(excitation_idx, omega_star);
+      UpdatePROM(excitation_idx, omega_star, it_sample);
     }
     Mpi::Print("\nAdaptive sampling{} {:d} frequency samples:\n"
                " n = {:d}, error = {:.3e}, tol = {:.3e}, memory = {:d}/{:d}\n",
-               (it == max_size_per_excitation) ? " reached maximum" : " converged with", it,
-               prom_op.GetReducedDimension(), max_errors.back(), offline_tol, memory,
-               convergence_memory);
+               (it_sample == max_size_per_excitation) ? " reached maximum"
+                                                      : " converged with",
+               it_sample, prom_op.GetReducedDimension(), max_errors.back(), offline_tol,
+               memory, convergence_memory);
     utils::PrettyPrint(prom_op.GetSamplePoints(excitation_idx), unit_GHz,
                        " Sampled frequencies (GHz):");
     utils::PrettyPrint(max_errors, 1.0, " Sample errors:");
@@ -395,7 +418,7 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
     BlockTimer bt0(Timer::POSTPRO);
     SaveMetadata(prom_op.GetLinearSolver());
   }
-  post_results.PostprocessFinal(post_op, indicator);
+  post_results.PostprocessFinal(iodata, post_op, indicator, &prom_op);
   return indicator;
 }
 
@@ -1147,11 +1170,71 @@ void DrivenSolver::SParametersPostPrinter::AddMeasurement(
   port_S.WriteFullTableTrunc();
 }
 
+void DrivenSolver::PROMPostPrinter::PostprocessPROM(const IoData &iodata,
+                                                    const RomOperator &prom_op)
+{
+  if (!do_measurement_ || !root_)
+  {
+    return;
+  }
+  Mpi::Print("Post-process synthesized circuit");
+  using VT = IoData::ValueType;
+
+  // Don't use structured binding for lambda below: captured structured bindings are C++20
+  const auto &prom_mat = prom_op.GetReducedMatrices();
+  const auto &Kr = std::get<0>(prom_mat);
+  const auto &Mr = std::get<1>(prom_mat);
+  const auto &Cr = std::get<2>(prom_mat);
+  const auto &v_norms = std::get<3>(prom_mat);
+  const auto &v_label = std::get<4>(prom_mat);
+
+  auto prom_size = Kr.rows();
+  MFEM_VERIFY((Mr.rows() == prom_size) && ((Cr.rows() == prom_size) || (Cr.rows() == 0)) &&
+                  (v_norms.rows() == prom_size) && (v_label.size() == prom_size),
+              "Inconsisten PROM size!");
+
+
+  auto print_table = [&](const Eigen::MatrixXd &mat, std::string_view filename)
+  {
+    auto out = TableWithCSVFile(post_dir_ / filename);
+    out.table.col_options.float_precision = 17;
+    for (int i = 0; i < prom_size; i++)
+    {
+      out.table.insert_column(v_label[i], v_label[i]);
+      for (int j = 0; j < prom_size; j++)
+      {
+        out.table[v_label[i]] << mat(i, j);
+      }
+    }
+    out.WriteFullTableTrunc();
+  };
+
+  // De-normalize all volltages
+  auto v_d = v_norms.asDiagonal();
+
+  // PROM matrices should be real
+  auto unit_H = iodata.DimensionalizeValue(VT::INDUCTANCE, 1.0);
+  Eigen::MatrixXd m_Linv = (1.0 / unit_H * v_d) * Kr.real() * v_d;
+  print_table(m_Linv, "prom-Linv.csv");
+
+  auto unit_F = iodata.DimensionalizeValue(VT::CAPACITANCE, 1.0);
+  Eigen::MatrixXd m_C = (unit_F * v_d) * Mr.real() * v_d;
+  print_table(m_C, "prom-C.csv");
+
+  // In principle, Cr need not exist
+  if (Cr.size() > 0)
+  {
+    auto unit_ohm = iodata.DimensionalizeValue(VT::IMPEDANCE, 1.0);
+    Eigen::MatrixXd m_Rinv = (1.0 / unit_ohm * v_d) * Cr.real() * v_d;
+    print_table(m_Rinv, "prom-Rinv.csv");
+  }
+}
+
 DrivenSolver::PostprocessPrintResults::PostprocessPrintResults(
     bool root, const fs::path &post_dir, const PostOperator &post_op,
     const SpaceOperator &space_op, const PortExcitationHelper &excitation_helper,
-    int n_expected_rows, int delta_post_)
-  : delta_post{delta_post_}, write_paraview_fields{delta_post_ > 0},
+    int n_expected_rows, const IoData &iodata)
+  : delta_post{iodata.solver.driven.delta_post}, write_paraview_fields{delta_post > 0},
     domains{true, root, post_dir, post_op, excitation_helper, n_expected_rows},
     surfaces{true, root, post_dir, post_op, excitation_helper, n_expected_rows},
     currents{
@@ -1167,7 +1250,8 @@ DrivenSolver::PostprocessPrintResults::PostprocessPrintResults(
                  space_op.GetWavePortOp(),
                  excitation_helper,
                  n_expected_rows},
-    error_indicator{true, root, post_dir}
+    error_indicator{true, root, post_dir},
+    prom{iodata.solver.driven.adaptive_circuit_synthesis, root, post_dir}
 {
 }
 
@@ -1197,9 +1281,14 @@ void DrivenSolver::PostprocessPrintResults::PostprocessStep(const IoData &iodata
 }
 
 void DrivenSolver::PostprocessPrintResults::PostprocessFinal(
-    const PostOperator &post_op, const ErrorIndicator &indicator)
+    const IoData &iodata, const PostOperator &post_op, const ErrorIndicator &indicator,
+    const RomOperator *prom_op)
 {
   BlockTimer bt0(Timer::POSTPRO);
+  if (prom_op != nullptr)
+  {
+    prom.PostprocessPROM(iodata, *prom_op);
+  }
   auto indicator_stats = indicator.GetSummaryStatistics(post_op.GetComm());
   error_indicator.PrintIndicatorStatistics(post_op, indicator_stats);
   if (write_paraview_fields)
--- a/palace/drivers/drivensolver.cpp
+++ b/palace/drivers/drivensolver.cpp
@@ -288,12 +288,14 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
     UpdatePROM(excitation_idx, omega0);
     prom_op.SolveHDM(excitation_idx, omega0 + (n_step - step0 - 1) * delta_omega, E);
     UpdatePROM(excitation_idx, omega0 + (n_step - step0 - 1) * delta_omega);
+    int it0 = 2;  // Nr of initial solutions in PROM
 
     // Greedy procedure for basis construction (offline phase). Basis is initialized with
     // solutions at frequency sweep endpoints.
-    int it = 2, it0 = it, memory = 0;
+    int memory = 0;
+    int it = it0;
     std::vector<double> max_errors = {0.0, 0.0};
-    while (true)
+    for (; (it < max_size_per_excitation) && (memory < convergence_memory); it++)
     {
       // Compute the location of the maximum error in parameter domain (bounded by the
       // previous samples).
@@ -303,34 +305,25 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op,
       prom_op.SolveHDM(excitation_idx, omega_star, E);
       prom_op.SolvePROM(excitation_idx, omega_star, Eh);
       linalg::AXPY(-1.0, E, Eh);
+
       max_errors.push_back(linalg::Norml2(space_op.GetComm(), Eh) /
                            linalg::Norml2(space_op.GetComm(), E));
       if (max_errors.back() < offline_tol)
       {
-        if (++memory == convergence_memory)
-        {
-          break;
-        }
+        ++memory;
       }
       else
       {
         memory = 0;
       }
-      if (it == max_size_per_excitation)
-      {
-        break;
-      }
 
       // Sample HDM and add solution to basis.
-      Mpi::Print("\nGreedy iteration {:d} (n = {:d}): ω* = {:.3e} GHz ({:.3e}), error = "
-                 "{:.3e}{}\n",
-                 it - it0 + 1, prom_op.GetReducedDimension(), omega_star * unit_GHz,
-                 omega_star, max_errors.back(),
-                 (memory == 0)
-                     ? ""
-                     : fmt::format(", memory = {:d}/{:d}", memory, convergence_memory));
+      Mpi::Print(
+          "\nGreedy iteration {:d} (PROM n = {:d}): ω* = {:.3e} GHz ({:.3e}), error = "
+          "{:.3e}, memory = {:d}/{:d}\n",
+          it - it0 + 1, prom_op.GetReducedDimension(), omega_star * unit_GHz, omega_star,
+          max_errors.back(), memory, convergence_memory);
       UpdatePROM(excitation_idx, omega_star);
-      it++;
     }
     Mpi::Print("\nAdaptive sampling{} {:d} frequency samples:\n"
                " n = {:d}, error = {:.3e}, tol = {:.3e}, memory = {:d}/{:d}\n",
--- a/palace/models/romoperator.cpp
+++ b/palace/models/romoperator.cpp
@@ -422,7 +422,7 @@ void RomOperator::SolveHDM(ExcitationIdx excitation_idx, double omega, ComplexVe
   ksp->Mult(r, u);
 }
 
-void RomOperator::UpdatePROM(const ComplexVector &u)
+void RomOperator::UpdatePROM(const ComplexVector &u, std::string_view node_label)
 {
 
   // Update V. The basis is always real (each complex solution adds two basis vectors if it
@@ -436,14 +436,19 @@ void RomOperator::UpdatePROM(const ComplexVector &u)
   MFEM_VERIFY(dim_V + has_real + has_imag <= V.size(),
               "Unable to increase basis storage size, increase maximum number of vectors!");
   const std::size_t dim_V0 = dim_V;
-  std::vector<double> H(dim_V + static_cast<std::size_t>(has_real) +
-                        static_cast<std::size_t>(has_imag));
+  const std::size_t new_dim_V = dim_V + std::size_t(has_real) + std::size_t(has_imag);
+  v_node_label.reserve(new_dim_V);
+  std::vector<double> H(new_dim_V);
+  voltage_norm_H.conservativeResize(new_dim_V);
+
   if (has_real)
   {
     V[dim_V] = u.Real();
     OrthogonalizeColumn(orthog_type, comm, V, V[dim_V], H.data(), dim_V);
     H[dim_V] = linalg::Norml2(comm, V[dim_V]);
     V[dim_V] *= 1.0 / H[dim_V];
+    voltage_norm_H[dim_V] = H[dim_V];
+    v_node_label.emplace_back(fmt::format("{}_re", node_label));
     dim_V++;
   }
   if (has_imag)
@@ -452,6 +457,8 @@ void RomOperator::UpdatePROM(const ComplexVector &u)
     OrthogonalizeColumn(orthog_type, comm, V, V[dim_V], H.data(), dim_V);
     H[dim_V] = linalg::Norml2(comm, V[dim_V]);
     V[dim_V] *= 1.0 / H[dim_V];
+    voltage_norm_H[dim_V] = H[dim_V];
+    v_node_label.emplace_back(fmt::format("{}_im", node_label));
     dim_V++;
   }
 
--- a/palace/models/romoperator.hpp
+++ b/palace/models/romoperator.hpp
@@ -6,6 +6,8 @@
 
 #include <complex>
 #include <memory>
+#include <string>
+#include <tuple>
 #include <vector>
 #include <Eigen/Dense>
 #include "linalg/ksp.hpp"
@@ -72,6 +74,8 @@ private:
   Eigen::MatrixXcd Kr, Mr, Cr, Ar;
   Eigen::VectorXcd RHS1r;
   Eigen::VectorXcd RHSr;
+  Eigen::VectorXd voltage_norm_H;
+  std::vector<std::string> v_node_label;
 
   // PROM reduced-order basis (real-valued) and active dimension.
   std::vector<Vector> V;
@@ -96,6 +100,13 @@ public:
     return mri.at(excitation_idx).GetSamplePoints();
   }
 
+  std::tuple<const Eigen::MatrixXcd &, const Eigen::MatrixXcd &, const Eigen::MatrixXcd &,
+             const Eigen::VectorXd &, const std::vector<std::string> &>
+  GetReducedMatrices() const
+  {
+    return std::tie(Kr, Mr, Cr, voltage_norm_H, v_node_label);
+  }
+
   // Set excitation index to build corresponding RHS vector (linear in frequency part)
   void SetExcitationIndex(ExcitationIdx excitation_idx);
 
@@ -103,7 +114,7 @@ public:
   void SolveHDM(ExcitationIdx excitation_idx, double omega, ComplexVector &u);
 
   // Add field configuration to the reduced-order basis and update the PROM.
-  void UpdatePROM(const ComplexVector &u);
+  void UpdatePROM(const ComplexVector &u, std::string_view node_label);
 
   // Add solution u to the minimal-rational interpolation for error estimation. MRI are
   // separated by excitation index.
--- a/palace/models/spaceoperator.cpp
+++ b/palace/models/spaceoperator.cpp
@@ -917,6 +917,47 @@ bool SpaceOperator::GetExcitationVector(ExcitationIdx excitation_idx, double ome
   return nnz1 || nnz2;
 }
 
+bool SpaceOperator::GetLumpedPortExcitationVector(int port_idx, ComplexVector &RHS,
+                                                  bool zero_metal)
+{
+  RHS.SetSize(GetNDSpace().GetTrueVSize());
+  RHS.UseDevice(true);
+  RHS = 0.0;
+
+  MFEM_VERIFY(RHS.Size() == GetNDSpace().GetTrueVSize(),
+              "Invalid T-vector size for AddExcitationVector1Internal!");
+  SumVectorCoefficient fb(GetMesh().SpaceDimension());
+
+  const auto &data = lumped_port_op.GetPort(port_idx);
+
+  mfem::Array<int> attr_list;
+  mfem::Array<int> attr_marker;
+
+  for (const auto &elem : data.elems)
+  {
+    attr_list.Append(elem->GetAttrList());
+    fb.AddCoefficient(
+        elem->GetModeCoefficient(1.0 / (elem->GetGeometryWidth() * data.elems.size())));
+  }
+  auto &mesh = GetNDSpace().GetParMesh();
+  int bdr_attr_max = mesh.bdr_attributes.Size() ? mesh.bdr_attributes.Max() : 0;
+  mesh::AttrToMarker(bdr_attr_max, attr_list, attr_marker);
+
+  mfem::LinearForm rhs1(&GetNDSpace().Get());
+  rhs1.AddBoundaryIntegrator(new VectorFEBoundaryLFIntegrator(fb), attr_marker);
+  rhs1.UseFastAssembly(false);
+  rhs1.UseDevice(false);
+  rhs1.Assemble();
+  rhs1.UseDevice(true);
+  GetNDSpace().GetProlongationMatrix()->AddMultTranspose(rhs1, RHS.Real());
+
+  if (zero_metal)
+  {
+    linalg::SetSubVector(RHS.Real(), nd_dbc_tdof_lists.back(), 0.0);
+  }
+  return true;
+}
+
 bool SpaceOperator::GetExcitationVector1(ExcitationIdx excitation_idx, ComplexVector &RHS1)
 {
   // Assemble the frequency domain excitation term with linear frequency dependence
--- a/palace/models/spaceoperator.hpp
+++ b/palace/models/spaceoperator.hpp
@@ -200,6 +200,8 @@ public:
   bool GetExcitationVector(ExcitationIdx excitation_idx, Vector &RHS);
   bool GetExcitationVector(ExcitationIdx excitation_idx, double omega, ComplexVector &RHS);
 
+  bool GetLumpedPortExcitationVector(int port_idx, ComplexVector &RHS, bool zero_metal);
+
   // Separate out RHS vector as RHS = iω RHS1 + RHS2(ω). The return value indicates whether
   // or not the excitation is nonzero (and thus is true most of the time).
   bool GetExcitationVector1(ExcitationIdx excitation_idx, ComplexVector &RHS1);
--- a/palace/utils/configfile.cpp
+++ b/palace/utils/configfile.cpp
@@ -1629,6 +1629,8 @@ void DrivenSolverData::SetUp(json &solver)
   adaptive_tol = driven->value("AdaptiveTol", adaptive_tol);
   adaptive_max_size = driven->value("AdaptiveMaxSamples", adaptive_max_size);
   adaptive_memory = driven->value("AdaptiveConvergenceMemory", adaptive_memory);
+  adaptive_circuit_synthesis =
+      driven->value("AdaptiveCircuitSynthesis", adaptive_circuit_synthesis);
 
   // Cleanup
   driven->erase("MinFreq");
@@ -1639,6 +1641,7 @@ void DrivenSolverData::SetUp(json &solver)
   driven->erase("AdaptiveTol");
   driven->erase("AdaptiveMaxSamples");
   driven->erase("AdaptiveConvergenceMemory");
+  driven->erase("AdaptiveCircuitSynthesis");
   MFEM_VERIFY(driven->empty(),
               "Found an unsupported configuration file keyword under \"Driven\"!\n"
                   << driven->dump(2));
--- a/palace/utils/iodata.cpp
+++ b/palace/utils/iodata.cpp
@@ -226,7 +226,17 @@ void IoData::CheckConfiguration()
   // problem type.
   if (problem.type == config::ProblemData::Type::DRIVEN)
   {
-    // No unsupported domain or boundary objects for frequency domain driven simulations.
+    // Driven (frequency) solver itself has no unsupported domain or boundary objects.
+    // Can't synthesize non-LRC circuit (same conditions as eigenmode solver)
+    if (solver.driven.adaptive_circuit_synthesis)
+    {
+      MFEM_VERIFY(
+          !(!boundaries.auxpec.empty() || !boundaries.waveport.empty() ||
+            (!boundaries.farfield.empty() && boundaries.farfield.order > 1) ||
+            !boundaries.conductivity.empty()),
+          "Driven system with circuit synthesis is not supported in systems with any of: "
+          "WavePort, Absorbing (order > 1), or Conductivity boundary conditions!\n");
+    }
   }
   else if (problem.type == config::ProblemData::Type::EIGENMODE)
   {
--- a/palace/drivers/drivensolver.hpp
+++ b/palace/drivers/drivensolver.hpp
@@ -8,6 +8,7 @@
 #include <vector>
 #include "drivers/basesolver.hpp"
 #include "models/portexcitationhelper.hpp"
+#include "utils/filesystem.hpp"
 #include "utils/strongtype.hpp"
 #include "utils/tablecsv.hpp"
 
@@ -21,6 +22,7 @@ class PostOperator;
 class SpaceOperator;
 class SurfaceCurrentOperator;
 class WavePortOperator;
+class RomOperator;
 
 //
 // Driver class for driven terminal simulations.
@@ -154,10 +156,23 @@ private:
                         const WavePortOperator &wave_port_op, const IoData &iodata);
   };
 
+  class PROMPostPrinter
+  {
+    bool root_ = false;
+    bool do_measurement_ = false;
+    fs::path post_dir_;
+
+  public:
+    PROMPostPrinter(bool do_measurement, bool root, fs::path post_dir)
+      : root_(root), do_measurement_(do_measurement), post_dir_(std::move(post_dir)) {};
+
+    void PostprocessPROM(const IoData &iodata, const RomOperator &prom_op);
+  };
+
   struct PostprocessPrintResults
   {
-    bool write_paraview_fields = false;
     int delta_post = 0;
+    bool write_paraview_fields = false;
 
     DomainsPostPrinter domains;
     SurfacesPostPrinter surfaces;
@@ -167,15 +182,18 @@ private:
     SParametersPostPrinter s_parameters;
 
     ErrorIndicatorPostPrinter error_indicator;
+    PROMPostPrinter prom;
 
     PostprocessPrintResults(bool is_mpi_root, const fs::path &post_dir,
                             const PostOperator &post_op, const SpaceOperator &space_op,
                             const PortExcitationHelper &excitation_helper,
-                            int n_expected_rows, int delta_post);
+                            int n_expected_rows, const IoData &iodata);
     void PostprocessStep(const IoData &iodata, const PostOperator &post_op,
                          const SpaceOperator &space_op, int step,
                          ExcitationIdx excitation_idx);
-    void PostprocessFinal(const PostOperator &post_op, const ErrorIndicator &indicator);
+    void PostprocessFinal(const IoData &iodata, const PostOperator &post_op,
+                          const ErrorIndicator &indicator,
+                          const RomOperator *prom_op = nullptr);
   };
 
   ErrorIndicator SweepUniform(SpaceOperator &space_op,
--- a/palace/drivers/eigensolver.cpp
+++ b/palace/drivers/eigensolver.cpp
@@ -613,7 +613,7 @@ void EigenSolver::EPRPostPrinter::AddMeasurementQ(double eigen_print_idx,
   double E_cap = post_op.GetLumpedCapacitorEnergy();
   double E_m = E_elec + E_cap;
 
-  port_EPR.table["idx"] << eigen_print_idx;
+  port_Q.table["idx"] << eigen_print_idx;
   for (const auto idx : ports_with_R)
   {
     double Kl = post_op.GetExternalKappa(idx, E_m);
--- a/palace/drivers/eigensolver.cpp
+++ b/palace/drivers/eigensolver.cpp
@@ -367,7 +367,7 @@ void EigenSolver::EigenPostPrinter::PrintStdoutHeader()
   }
   to("{:s}", eig.table.print_row_separator);
 
-  Mpi::Print("{}{}", std::string{buf.data(), buf.size()},
+  Mpi::Print("{}{}\n", std::string{buf.data(), buf.size()},
              std::string(stdout_int_print_width + 4 * eig.table[1].col_width(), '='));
   eig.table.col_options = save_defaults;
 }
@@ -397,7 +397,7 @@ void EigenSolver::EigenPostPrinter::PrintStdoutRow(size_t j)
     to("{:s}", eig.table[i].format_row(j));
   }
   to("{:s}", eig.table.print_row_separator);
-  Mpi::Print("{}", buf);
+  Mpi::Print("{}", fmt::to_string(buf));
   eig.table.col_options = save_defaults;
 }
 
--- a/palace/utils/configfile.hpp
+++ b/palace/utils/configfile.hpp
@@ -706,6 +706,9 @@ public:
   // Memory required for adaptive sampling convergence.
   int adaptive_memory = 2;
 
+  // Return circuit matrices from port with port excitation vectors
+  bool adaptive_circuit_synthesis = false;
+
   void SetUp(json &solver);
 };
 
--- a/scripts/schema/config/solver.json
+++ b/scripts/schema/config/solver.json
@@ -51,7 +51,8 @@
         "Restart": { "type": "integer", "exclusiveMinimum": 0 },
         "AdaptiveTol": { "type": "number", "minimum": 0.0 },
         "AdaptiveMaxSamples": { "type": "number", "exclusiveMinimum": 0 },
-        "AdaptiveConvergenceMemory": { "type": "integer", "exclusiveMinimum": 0 }
+        "AdaptiveConvergenceMemory": { "type": "integer", "exclusiveMinimum": 0 },
+        "AdaptiveCircuitSynthesis": { "type": "boolean" }
       }
     },
     "Transient":


palace/models/romoperator.cpp
```diff
@@ -353,6 +353,13 @@ RomOperator::RomOperator(const IoData &iodata, SpaceOperator &space_op,
 
 void RomOperator::SetExcitationIndex(int excitation_idx)
 {
+  // Return if cached. Ctor constructs with excitation_idx_cache = 0 which is not a valid
+  // expiation index, so this is triggered the first time it is called in drivensolver.
```
hughcars: expiration? excitation?

palace/models/romoperator.cpp
```diff
@@ -535,4 +544,64 @@ std::vector<std::complex<double>> RomOperator::ComputeEigenvalueEstimates() cons
   return {};
 }
 
+void RomOperator::PrintPROMMatrices(const Units &units, const fs::path &post_dir) const
+{
+  auto print_table = [post_dir, labels = this->v_node_label](const Eigen::MatrixXd &mat,
+                                                             std::string_view filename)
+  {
+    MFEM_VERIFY(labels.size() == mat.cols(), "Insonsistent PROM size!");
```
hughcars: "Inconsistent"

palace/drivers/drivensolver.cpp
```diff
@@ -261,14 +265,30 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op) const
...
+    for (const auto &[port_idx, port_data] : lumped_port_op)
+    {
+      ComplexVector port_excitation_E;
+      port_excitation_E.UseDevice(true);
+      space_op.GetLumpedPortExcitationVector(port_idx, port_excitation_E, true);
+      prom_op.UpdatePROM(port_excitation_E, fmt::format("port_{:d}", port_idx));
+    }
+    const auto &orth_R = prom_op.GetRomOrthogonalityMatrix();
+    MFEM_VERIFY(orth_R.isDiagonal(), "Lumped port modes should have exactly zero overlap");
+    // TODO(phdum): Test this.
```
hughcars: Debris, I assume wip given our inner product discussions.

palace/models/romoperator.hpp
```diff
@@ -50,58 +53,79 @@ class RomOperator
...
   Eigen::VectorXcd RHSr;
 
-  // PROM reduced-order basis (real-valued) and active dimension.
+  // PROM reduced-order basis (real-valued).
   std::vector<Vector> V;
-  std::size_t dim_V = 0;
   Orthogonalization orthog_type;
 
-  // MRIs: one for each excitation index.
+  // Label to distinguish port modes from solution  projection and to print PROM matrices
```
hughcars: Nit: double space `solution projection`

palace/models/romoperator.hpp
```diff
@@ -50,58 +53,79 @@ class RomOperator
...
+  // Return overlap matrix form orthogonalization of vectors in ROM.
+  const auto &GetRomOrthogonalityMatrix() const { return orth_R; }
+
   // Set excitation index to build corresponding RHS vector (linear in frequency part).
   void SetExcitationIndex(int excitation_idx);
 
   // Assemble and solve the HDM at the specified frequency.
   void SolveHDM(int excitation_idx, double omega, ComplexVector &u);
 
   // Add field configuration to the reduced-order basis and update the PROM.
```
hughcars: Describe why the node_label is needed.

palace/models/romoperator.hpp
```diff
@@ -50,58 +53,79 @@ class RomOperator
...
-  auto GetReducedDimension() const { return dim_V; }
+  auto GetReducedDimension() const { return V.size(); }
 
   // Return set of sampled parameter points for basis construction.
   const auto &GetSamplePoints(int excitation_idx) const
   {
     return mri.at(excitation_idx).GetSamplePoints();
   }
 
+  // Return overlap matrix form orthogonalization of vectors in ROM.
```
hughcars: This sentence doesn't really parse for me, is `form` supposed to be `from`?

palace/models/romoperator.hpp
```diff
@@ -113,13 +137,16 @@ class RomOperator
...
-  std::vector<double> FindMaxError(int excitation_idx, int N = 1) const
+  std::vector<double> FindMaxError(int excitation_idx, std::size_t N = 1) const
   {
     return mri.at(excitation_idx).FindMaxError(N);
   }
 
   // Compute eigenvalue estimates for the current PROM system.
   std::vector<std::complex<double>> ComputeEigenvalueEstimates() const;
+
+  // Print PROM matrices to file include in input (SI) units.
```
hughcars, line 262: Doesn't parse, does that mean the input SI units are being included?

test/unit/test-romoperator.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  CHECK(mri_1.GetSamplePoints() == std::vector<double>{});
+  CHECK_THROWS(mri_1.FindMaxError(1));
+
+  for (double x_sample : {-2.0, -1.0, 1.0, 2.0})
+  {
+    auto tan_eval = fn_tan_shift(x_sample) / double(Mpi::Size(comm));
+    std::vector<std::complex<double>> tmp = {x_sample * tan_eval, -tan_eval, 5. * tan_eval,
+                                             10. * x_sample * tan_eval,
+                                             2 * x_sample * x_sample * x_sample * tan_eval};
```
hughcars, line 105: This is a little bit "magical numbers", put a small comment above explaining their source.

test/unit/test-romoperator.cpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  CHECK(mri_1.GetSamplePoints().size() == 4);
+  CHECK(mri_1.GetSamplePoints() == std::vector<double>{-2.0, -1.0, 1.0, 2.0});
+
+  // By symmetry of poles max error should be at zero.
+  auto max_err_1 = mri_1.FindMaxError(5);
+  REQUIRE(max_err_1.size() == 5);
+
+  // By symmetry highest error should be at at zero.
```
hughcars: at at

palace/utils/configfile.cpp
```diff
@@ -1909,6 +1911,7 @@ void DrivenSolverData::SetUp(json &solver)
   driven->erase("AdaptiveTol");
   driven->erase("AdaptiveMaxSamples");
   driven->erase("AdaptiveConvergenceMemory");
+  driven->erase("AdaptiveCircuitSynthesis");
```
hughcars: Add a corresponding print step to the `if constexpr` branch above.

palace/models/romoperator.cpp
```diff
@@ -279,40 +279,45 @@ std::vector<double> MinimalRationalInterpolation::FindMaxError(int N) const
   // }
 
   // Fall back to sampling Q on discrete points if no roots exist in [start, end].
-  if (std::abs(z_star[0]) == 0.0)
+  // TODO: currently we always us this. Consider other optimization above again.
```
hughcars: us to use

palace/models/romoperator.cpp
```diff
@@ -279,40 +279,45 @@ std::vector<double> MinimalRationalInterpolation::FindMaxError(int N) const
...
+
+  // We could use priority queue here to keep the N lowest values. However, we don't use
+  // std::priority_queue class since we want to have access to the vector and also binary
+  // tree structure of heap class as rebalancing is excessive overhead for tiny size N.
+  using q_t = std::pair<std::complex<double>, double>;
+  std::vector<q_t> queue{};
+  queue.reserve(N);
+
+  const std::size_t nr_sample = 1.0e6;  // must be >= N
```
hughcars, line 309: Add a verify of that.

palace/models/romoperator.cpp
```diff
@@ -335,24 +340,34 @@ RomOperator::RomOperator(const IoData &iodata, SpaceOperator &space_op,
...
 
-  // The initial PROM basis is empty. The provided maximum dimension is the number of sample
-  // points (2 basis vectors per point). Basis orthogonalization method is configured using
-  // GMRES/FGMRES settings.
-  MFEM_VERIFY(max_size_per_excitation * excitation_helper.Size() > 0,
-              "Reduced order basis storage must have > 0 columns!");
-  V.resize(2 * max_size_per_excitation * excitation_helper.Size(), Vector());
+  // Reserve empty vectors but don't pre-allocate actual memory size due to overhead.
+  V.reserve(max_prom_size);
+  v_node_label.reserve(max_prom_size);
```
hughcars: reserve does allocate the memory, it just keeps the "end" pointer where it was. Still good and convenient to do so.

palace/models/romoperator.cpp
```diff
@@ -409,78 +423,88 @@ void RomOperator::SolveHDM(int excitation_idx, double omega, ComplexVector &u)
...
+  const std::size_t dim_V_old = V.size();
+  const std::size_t dim_V_new =
+      V.size() + static_cast<std::size_t>(has_real) + static_cast<std::size_t>(has_imag);
+
+  orth_R.conservativeResizeLike(Eigen::MatrixXd::Zero(dim_V_new, dim_V_new));
+
+  auto add_real_vector_to_basis = [this](const Vector &vector, std::string_view full_label)
+  {
+    auto dim_V = V.size();
+    V.push_back(vector);
```
hughcars: Nit: `auto &v = V.emplace_back(vector);` then `v` instead of `v.back()` everywhere.

palace/models/romoperator.cpp
```diff
@@ -409,78 +423,88 @@ void RomOperator::SolveHDM(int excitation_idx, double omega, ComplexVector &u)
...
+  const bool has_real = (norm_re > norm_tol);
+  const bool has_imag = (norm_im > norm_tol);
+
+  const std::size_t dim_V_old = V.size();
+  const std::size_t dim_V_new =
+      V.size() + static_cast<std::size_t>(has_real) + static_cast<std::size_t>(has_imag);
+
+  orth_R.conservativeResizeLike(Eigen::MatrixXd::Zero(dim_V_new, dim_V_new));
+
+  auto add_real_vector_to_basis = [this](const Vector &vector, std::string_view full_label)
```
hughcars: NIt: This is doing 2 things, despite the name. Change to only take the `Vector`, then the name matches, and do the `v_node_label.emplace_back` directly after each call directly rather than pass it in. It's cleaner and doesn't mix the linear algebra and accountancy.

palace/models/romoperator.cpp
```diff
@@ -409,78 +423,88 @@ void RomOperator::SolveHDM(int excitation_idx, double omega, ComplexVector &u)
...
-    OrthogonalizeColumn(orthog_type, comm, V, V[dim_V], H.data(), dim_V);
-    H[dim_V] = linalg::Norml2(comm, V[dim_V]);
-    V[dim_V] *= 1.0 / H[dim_V];
-    dim_V++;
+    add_real_vector_to_basis(u.Imag(), fmt::format("{}_im", node_label));
   }
 
-  // Update reduced-order operators. Resize preserves the upper dim0 x dim0 block of each
-  // matrix and first dim0 entries of each vector and the projection uses the values
-  // computed for the unchanged basis vectors.
```
hughcars, line 443: This comment is pretty helpful, why delete it?

palace/models/romoperator.cpp
```diff
@@ -535,4 +559,64 @@ std::vector<std::complex<double>> RomOperator::ComputeEigenvalueEstimates() cons
...
+
+  // De-normalize PROM matrices voltages (both port and sampled). Define so that 1.0 on port
+  // i corresponds to full (un-normalized solution), so you can use Linv, Rinv, C directly.
+  auto v_d = orth_R.diagonal().cwiseInverse().asDiagonal();
+
+  // Note: When checking for imaginary parts, it is better to do this for K,C,M as this is a
+  // nullptr check. Kr, Mr, Cr would require a numerical check on imag elements.
+
+  auto unit_henry = units.GetScaleFactor<Units::ValueType::INDUCTANCE>();
+  auto m_Linv = ((1.0 / unit_henry) * v_d) * Kr * v_d;
```
hughcars: `m_Linv` isn't a great unit name, is this the Linv for a mode?

palace/models/spaceoperator.cpp
```diff
@@ -841,6 +841,47 @@ bool SpaceOperator::GetExcitationVector(int excitation_idx, double omega,
...
+  rhs1.UseDevice(false);
+  rhs1.Assemble();
+  rhs1.UseDevice(true);
+  GetNDSpace().GetProlongationMatrix()->AddMultTranspose(rhs1, RHS.Real());
+
+  if (zero_metal)
+  {
+    linalg::SetSubVector(RHS.Real(), nd_dbc_tdof_lists.back(), 0.0);
+  }
+  return true;
```
hughcars: I'm not looking at this rn based on our discussions, I believe you're fixing this up based on your new understanding.

palace/models/spaceoperator.hpp
```diff
@@ -215,6 +215,8 @@ class SpaceOperator
   bool GetExcitationVector(int excitation_idx, Vector &RHS);
   bool GetExcitationVector(int excitation_idx, double omega, ComplexVector &RHS);
 
+  bool GetLumpedPortExcitationVector(int port_idx, ComplexVector &RHS, bool zero_metal);
```
hughcars: Doc string.

palace/drivers/drivensolver.cpp
```diff
@@ -362,6 +380,16 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op) const
...
 
+  if (iodata.solver.driven.adaptive_circuit_synthesis)
+  {
+    BlockTimer bt0(Timer::POSTPRO);
+    Mpi::Print(" Printing PROM Matrices to disk.\n");
+    if (root)
+    {
+      prom_op.PrintPROMMatrices(iodata.units, iodata.problem.output);
+    }
+  }
```
hughcars, line 381: Nit: Move the root early return inside the method, try to reduce the clutter in the driver

palace/models/romoperator.cpp
phdum-a: Put back.

I had a version which pre-allocated the matrices up to their maximum size and then used views to project. Since these matrices are small, that would have avoided the penalty of moving memory every time you resize. But the Eigen view does not play nicely with the raw point MPI reduction done in `ProjectMatInternal` so I reverted and squashed. But the comment not coming back was a reverting mistake.

palace/models/romoperator.cpp
phdum-a: excitation. Fixed

palace/models/romoperator.hpp
phdum-a: Sorry, yes. This is the upper triangular R matrix out of the QR orthonormalization matrix of the PROM vectors. But "R matrix" is an overloaded term. I improved the comments.

palace/models/romoperator.hpp
phdum-a: Yes. This just means that e.g. the capacitance matrix is in farad, etc. like all measurement rather than printing the internal matrix. I will rework the comment also when the normalization and inner product is fixed.

palace/models/romoperator.cpp
phdum-a: Done. Also improved this comment.

palace/models/romoperator.cpp
phdum-a: Done

palace/models/romoperator.hpp
phdum-a: Done. We want it for the circuit matrix csv printing as headers. Specifically, we want to distinguish port from solve field configurations. Also, we since split up complex vectors into real and imaginary components (pure real or pure imaginary get 1 column, complex gets 2). We can't really reconstruct what vector is what without labels.

palace/models/romoperator.cpp
phdum-a: Yes, it reserve the size of an `mfem::Vector()` class, but we don't pre-allocate heap memory by setting the size of these vectors. I removed that part of the comment.

palace/drivers/drivensolver.cpp
```diff
@@ -4,6 +4,12 @@
 #include "drivensolver.hpp"
 
 #include <complex>
+#include <iostream>
+#include <Eigen/Core>
+#include <Eigen/Dense>
+#include <Eigen/src/Core/IO.h>
+#include <fmt/core.h>
+#include <fmt/ostream.h>
```
hughcars: Are all these includes still needed?

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+  std::complex<double> InnerProduct(const ComplexVector &x, const ComplexVector &y) const
+  {
+    return LocalDot(x, y);
+  }
+
+  std::complex<double> InnerProduct(MPI_Comm comm, const ComplexVector &x,
+                                    const ComplexVector &y) const
+  {
+    return Dot(comm, x, y);
+  }
```
hughcars: I believe these reduce to
```
  template <typename VecType>
  auto InnerProduct(const VecType &x, const VecType &y) const
  {
    return LocalDot(x, y);
  }

  template <typename VecType>
  auto InnerProduct(MPI_Comm comm, const VecType &x, const VecType &y) const
  {
    return Dot(comm, x, y);
  }
```

palace/models/lumpedportoperator.hpp
```diff
@@ -55,11 +55,40 @@ class LumpedPortData
...
                  const mfem::ParMesh &mesh);
 
-  double GetToSquare(const LumpedElementData &elem) const
+  [[nodiscard]] double GetToSquare(const LumpedElementData &elem) const
   {
     return elem.GetGeometryWidth() / elem.GetGeometryLength() * elems.size();
   }
 
+  // Normalization of tangential electric field of port corresponding \int \vert E_t \vert^2
+  // dS = \vert Z_0 \vert \sum W_e / L_e. In this function we set the impedance magnitude
```
hughcars: Though nowadays many can parse latex by reading, we try to avoid it as much as possible in comments as it can result in unnecessary confusion. For example here: 
```
∫|E|²ds = |Z₀| ∑Wₑ/Lₑ 
```
the easiest way of getting these I find is using julia and tab completion, but I'm sure there are equivalents.

palace/models/spaceoperator.hpp
```diff
@@ -220,6 +220,20 @@ class SpaceOperator
   bool GetExcitationVector1(int excitation_idx, ComplexVector &RHS1);
   bool GetExcitationVector2(int excitation_idx, double omega, ComplexVector &RHS2);
 
+  // Fill vector corresponding to the tangential electric field E_t on a lumped port, with
+  // overall normalization such that the reference impedance \vert Z_R \vert = 1 in internal
+  // units.
+  bool GetLumpedPortExcitationVectorPrimaryEt(int port_idx, ComplexVector &Et_primary,
+                                              bool zero_metal);
+
+  // Fill vector corresponding to the tangential electric field H_t x n on a lumped port,
```
hughcars: electric or H_t?

palace/utils/iodata.cpp
```diff
@@ -149,56 +150,9 @@ std::stringstream PreprocessFile(const char *filename)
 
 using json = nlohmann::json;
 
-IoData::IoData(const char *filename, bool print) : units(1.0, 1.0), init(false)
+IoData::IoData(nlohmann::json &&config_) : units(1.0, 1.0), init(false)
```
hughcars: Why not just pass by value? If it's fed by an rvalue, it'll thieve it, if not it can do RVO etc. Using `&&` seems unneeded.

palace/utils/iodata.cpp
```diff
@@ -219,13 +173,79 @@ IoData::IoData(const char *filename, bool print) : units(1.0, 1.0), init(false)
...
+    // Can't synthesize non-LRC circuit (same conditions as eigenmode solver)
+    if (solver.driven.adaptive_circuit_synthesis)
+    {
+      MFEM_VERIFY(
+          !(!boundaries.auxpec.empty() || !boundaries.waveport.empty() ||
+            (!boundaries.farfield.empty() && boundaries.farfield.order > 1) ||
+            !boundaries.conductivity.empty()),
+          "Driven system with circuit synthesis is not supported in systems with any of: "
+          "WavePort, Absorbing (order > 1), or Conductivity boundary conditions!\n");
+    }
```
hughcars: I don't think you need so many double negatives? You want either, no adaptive synthesis OR all those BC are empty. Also I don't think you need to check for waveportPEC empty in addition to waveport empty. 

```
    MFEM_VERIFY(!solver.driven.adaptive_circuit_synthesis ||
        (boundaries.waveport.empty() && (boundaries.farfield.empty() || boundaries.farfield.order == 1) && boundaries.conductivity.empty()),
        "Driven system with circuit synthesis is not supported in systems with any of: "
        "WavePort, Absorbing (order > 1), or Conductivity boundary conditions!\n");
```

palace/utils/iodata.hpp
```diff
@@ -38,11 +39,15 @@ class IoData
 private:
   bool init;
 
-  // Check configuration file options and compatibility with requested problem type.
+public:
+  // Check configuration file options and compatibility with requested problem type. Should
+  // not be called by user, but temporarily made public for testing.
```
hughcars: I guess this is debris?

palace/utils/iodata.hpp
```diff
@@ -38,11 +39,15 @@ class IoData
...
   void CheckConfiguration();
 
-public:
-  IoData(const Units &units) : units(units), init(false) {}
+  explicit IoData(const Units &units) : units(units), init(false) {}
+
+  // Take parsed json and override options defaults.
+  explicit IoData(nlohmann::json &&config);
```
hughcars: As above, just use value, then it will initialize from a `&&` with move ctor if it thinks necessary.

test/unit/test-lumpedportintegration.cpp
```diff
@@ -0,0 +1,802 @@
...
+    mesh::RefineMesh(iodata, mfem_mesh);
+    for (auto &m : mfem_mesh)
+    {
+      mesh_.push_back(std::make_unique<Mesh>(std::move(m)));
+    }
+  }
+  return mesh_;
+}
+
+// TODO: Should work in parallel but fails due to MFEM parallel issues.
```
hughcars: What issue?

test/unit/test-lumpedportintegration.cpp
```diff
@@ -0,0 +1,802 @@
...
+  // mesh this label is a single square, for the tet mesh this is two triangles. The bulk
+  // domain is a trivial material (ε=1.0, µ=1.0, no loss).
+  //
+  // This test iterates: (a) without any PEC boundary, (b) with a metal (PEC) boundary
+  // neighbor next to the port (SIDE), and (c) with a metal boundary neighbor at the end of
+  // the port (END). In (b), we set the boundary square with diagonal (1,0,0)-(2,1,0) to be
+  // PEC. This will cause the electric field of port on the shared edged to be set to zero
+  // and change integrals and normalizations as the effective width of the port is changed.
+  // In (c), we set the boundary square with diagonal (0, 1, 0)-(1, 2, 0) to be PEC. This
+  // should not alter the field with are perpendicularly incident on the metal so should
```
hughcars: with are

test/unit/test-lumpedportintegration.cpp
```diff
@@ -0,0 +1,802 @@
...
+    // std::array<std::array<double, 3>, 3> bbox_axes_out = {
+    //     std::array<double, 3>{1.0, 0.0, 0.0}, std::array<double, 3>{0.0, 1.0, 0.0},
+    //     std::array<double, 3>{0.0, 0.0, 0.0}};
+    // CHECK(bbox.axes == bbox_axes_out);
+
+    // CHECK(bbox.Normals().size() == 3);
+    // std::array<std::array<double, 3>, 3> bbox_normals_out = {
+    //     std::array<double, 3>{0.0, 0.0, 0.0}, std::array<double, 3>{0.0, 0.0, 0.0},
+    //     std::array<double, 3>{0.0, 0.0, 0.0}};
+    // CHECK(bbox.Normals() == bbox_normals_out);
```
hughcars: I believe tests on these are in geodata.cpp, as this is really testing `mesh::` namespace functions.

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+    v_workspace.UseDevice(blueprint.UseDevice());
+  }
+
+public:
+  template <typename OpType>
+  explicit InnerProductRealWeight(const std::shared_ptr<OpType> &weight_op_)
+    : weight_op(weight_op_)
+  {
+  }
+  // Follow same conventions as Dot:  yᴴ x or yᵀ x (not y comes second in the arguments).
```
hughcars: note

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+};
+
+class InnerProductRealWeight
+{
+  // Choose generic operator, although can improve by refining for specialized type.
+  std::shared_ptr<Operator> weight_op;
+
+  // Don't have inner product wrapper / implemented in Operator, so need to allocate a
+  // vector as a workspace. (TODO: Optimize this away).
+  mutable Vector v_workspace = {};
```
hughcars: An alternative would be that this buffer is passed in to the `InnerProduct` method as an optional third arg. Similar to the `Mult2` methods in the solver classes.
```
mutable std::unique_ptr<Vector> v_workspace;
double InnerProduct(const Vector &x, const Vector &y, Vector *buff = v_workspace.get()) const
```
and then the size of the buffer checked with an assert and then used. 

Upside: if a callsite already has a buffer declared, it can use that, but if it doesn't then the internal workspace vector will be sized and used. 
Downside: it makes this interface more cluttered, and the memory persists even if a new buffer is introduced. 

Having the 3 form combined with the `operator()` change would allow for binding into a functor based on the local scope as needed, something like
```
Vector buff;
OrthogonalizeColumCGS(...., [&buff](const Vector &x, const Vector &y){ return op(x, y, buff); });
```

The rough philosophy being callsites control their own memory, which should be as closed to fixed as possible before a call like this. Ultimately, this functor is only being used for testing, so it's moot.

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+
+  void SetWorkspace(const Vector &blueprint) const
+  {
+    v_workspace.SetSize(blueprint.Size());
+    v_workspace.UseDevice(blueprint.UseDevice());
+  }
```
hughcars: The workspace vector is fixed in size once the weight_op is, so this can be done in the ctor.

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+
+public:
+  template <typename OpType>
+  explicit InnerProductRealWeight(const std::shared_ptr<OpType> &weight_op_)
+    : weight_op(weight_op_)
+  {
+  }
```
hughcars: MFEM_VERIFY is square and initialize the workspace. If the inner prod has been declared, assume it is about to be used and finish the initialization.

palace/models/romoperator.cpp
```diff
@@ -3,13 +3,29 @@
...
 
-#include <Eigen/SVD>
+#include <algorithm>
+#include <cmath>
+#include <cstddef>
+#include <memory>
+#include <string_view>
+#include <tuple>
+#include <utility>
+// #include <Eigen/SVD>
```
hughcars: Debris

palace/models/romoperator.cpp
```diff
@@ -31,26 +47,43 @@ namespace
...
+    return linalg::Norml2(comm, x);
+  }
+  else
+  {
+    VecType x_tmp = x;
+    weight->Mult(x, x_tmp);
+    return std::sqrt(std::abs(linalg::Dot(comm, x, x_tmp)));
+  }
+}
+
```
hughcars: This doesn't seem to be used anymore? I presume this is because it was superseded by the InnerProduct approach.

palace/models/romoperator.cpp
```diff
@@ -279,42 +311,61 @@ std::vector<double> MinimalRationalInterpolation::FindMaxError(int N) const
...
 RomOperator::RomOperator(const IoData &iodata, SpaceOperator &space_op,
-                         int max_size_per_excitation)
+                         std::size_t max_size_per_excitation)
   : space_op(space_op), orthog_type(iodata.solver.linear.gs_orthog)
 {
+  // Always use refined CGS with adaptive synthesis for higher accuracy on print-out.
+  if (iodata.solver.driven.adaptive_circuit_synthesis)
+  {
+    orthog_type = Orthogonalization::CGS2;
+  }
```
hughcars: Overriding options like this isn't ideal. Better to have this be a warning, as this effectively hardcodes some branches away. If you are suggesting the default should be changed based on the presence of the adaptive circuit synthesis, that should be done in the `configfile.cpp`, and be included in the documentation.

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+
+  std::complex<double> InnerProduct(MPI_Comm comm, const ComplexVector &x,
+                                    const ComplexVector &y) const
+  {
+    return Dot(comm, x, y);
+  }
+};
+
+class InnerProductRealWeight
```
hughcars: This class only appears to be used in tests? If it's been overtaken by the hybrid one in the rom operator, should it just be removed/moved into the test folder?

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+    return dot;
+  }
+};
+
+template <typename VecType, typename ScalarType,
+          typename InnerProductW = InnerProductStandard>
 inline void OrthogonalizeColumnMGS(MPI_Comm comm, const std::vector<VecType> &V, VecType &w,
-                                   ScalarType *H, int m)
+                                   ScalarType *H, std::size_t m,
+                                   const InnerProductW &dot_op = {})
```
hughcars, line 43: Cute trick to avoid runtime polymorphism. I tried a few variations with a base `InnerProduct` class, but I think this ends up being neatest, as you avoid the need for `MFEM_ABORT` within virtual methods for vec types you don't want to support, and instead get compilation errors.

palace/models/lumpedportoperator.hpp
```diff
@@ -55,11 +55,40 @@ class LumpedPortData
   LumpedPortData(const config::LumpedPortData &data, const MaterialOperator &mat_op,
                  const mfem::ParMesh &mesh);
 
-  double GetToSquare(const LumpedElementData &elem) const
+  [[nodiscard]] double GetToSquare(const LumpedElementData &elem) const
```
hughcars: Nit: Stylistically we've been avoiding these attributes proliferating too much. This is a cheap function to evaluate, so there's no pressing need to avoid it (and an unused call should be caught otherwise). So rather than putting these everywhere and being prescriptive, can just ditch them.

palace/models/lumpedportoperator.hpp
```diff
@@ -55,11 +55,40 @@ class LumpedPortData
...
+  // later as needed.
+  [[nodiscard]] double GetExcitationFieldEtNormSqWithUnityZR() const
+  {
+    double norm_et = 0.0;
+    for (const auto &el : elems)
+    {
+      norm_et += el->GetGeometryWidth() / el->GetGeometryLength();
+    }
+    return norm_et;
+  }
```
hughcars: The names of these two methods are confusing to me, and they really only appear to be used in `test-lumpedportintegration` and `test-romoperator`. If the point is they are operating on `elems`, `elems` is public so this test calculation could be done in those unit tests as needed. Delete them and perform these calculations in the unit tests as required instead.

palace/models/romoperator.cpp
```diff
@@ -31,26 +47,43 @@ namespace
...
     case Orthogonalization::CGS:
-      linalg::OrthogonalizeColumnCGS(comm, V, w, Rj, j);
+      linalg::OrthogonalizeColumnCGS(comm, V, w, Rj, j, false, dot_op);
       break;
     case Orthogonalization::CGS2:
-      linalg::OrthogonalizeColumnCGS(comm, V, w, Rj, j, true);
+      linalg::OrthogonalizeColumnCGS(comm, V, w, Rj, j, true, dot_op);
       break;
   }
 }
```
hughcars, line 68: This function seems like a natural fit for `orthog.hpp`, given it's only a selector wrapper around the methods there. In fact this is identical to `OrthogonalizeIteration` in iterative.cpp. Instead move this function to `orthog.hpp`, and use the same wrapper in both places. This also suggests plausibly could use an alternative innerproduct in future for iterative schemes which could be interesting.

palace/models/romoperator.cpp
```diff
@@ -335,26 +386,112 @@ RomOperator::RomOperator(const IoData &iodata, SpaceOperator &space_op,
...
+  auto max_prom_size = 2 * max_size_per_excitation * space_op.GetPortExcitations().Size();
+  if (iodata.solver.driven.adaptive_circuit_synthesis)
+  {
+    max_prom_size += space_op.GetLumpedPortOp().Size();  // Lumped ports are real fields
+
+    // Build inner-product weight matrix. This is made from MassIntegrator of the domain and
+    // port boundaries summed together. However:
+    // - We zero out the domain mass matrix on the dof of the boundary, leaving on the
+    //   boundary mass matrix.
+    // - We weight the mass matrix by 1 / \eta with reference impedance \vert Z_R \vert = 1,
```
hughcars: Nit: I'm assuming weight the boundary mass matrix, make it non-ambiguous. 

I know we've discussed it so I understand the reasons for this construction but I think the statements in these bullets can be tightened up. Something like 
1. `(E_i, E_i)_M = 1 \forall i \in I` where I are the ports and E_i is the projection of the port field onto the basis. 
2. (E, E)_M >= 0 \forall E, (E,E)_M=0 iff ||E|| = 0
You'll want to smith the equation for 1. a little to have the precise RHS you want. But it's that requirement that explains why you have to go to lengths to remove those dofs from the volume product. After that why those rows and cols get zeroed out makes much more sense. You're effectively making a mesh dependent norm that is asymptotic to measuring the field fully separated into that on the surface and that in the volume.

palace/models/romoperator.cpp
```diff
@@ -335,26 +386,112 @@ RomOperator::RomOperator(const IoData &iodata, SpaceOperator &space_op,
...
+    //   so that power orthogonality of modes is enforced.
+    // - We don't weight by material coefficients so that is fully real and corresponds to
+    //   to full space overlap (except excised bulk part).
```
hughcars: You could use the `PermittivityReal`, as that then turns this effectively into an electric energy measurement (for the non-lossy piece of eps). I don't know how that relates to your statements on how you want the normalization to work out, but something to consider.

palace/models/romoperator.cpp
```diff
@@ -335,26 +386,112 @@ RomOperator::RomOperator(const IoData &iodata, SpaceOperator &space_op,
...
+    auto port_attr_marker = mesh::AttrToMarker(bdr_attr_max, port_attr_list);
+    space_op.GetNDSpace().Get().GetEssentialTrueDofs(port_attr_marker, port_tdof_list);
+
+    // Bulk:
+    BilinearForm w_bulk(space_op.GetNDSpace());
+
+    // Use Palace existing palace machinery, but make a trivial bulk material.
+    MaterialPropertyCoefficient f_bulk(mat_op.MaxCeedAttribute());
+    const auto &eps_ref = mat_op.GetPermittivityReal();
+    mfem::DenseTensor eps_id(eps_ref.SizeI(), eps_ref.SizeJ(), eps_ref.SizeK());
```
hughcars: This is roughly what I was saying above, you'd basically be moving from working on `||E||` to `(E, D)`, which is the electric energies. Places more weight on non-vacuum materials. Given this is just for excluding the bulk, probably doesn't matter much, but seems neat compared to constructing another attribute here.

palace/models/romoperator.cpp
```diff
@@ -409,78 +545,207 @@ void RomOperator::SolveHDM(int excitation_idx, double omega, ComplexVector &u)
...
+  Mr.conservativeResize(dim_V_new, dim_V_new);
+  ProjectMatInternal(comm, V, *M, Mr, r, dim_V_old);
+  if (RHS1.Size())
+  {
+    RHS1r.conservativeResize(dim_V_new);
+    ProjectVecInternal(comm, V, RHS1, RHS1r, dim_V_old);
+  }
+}
+
+void RomOperator::ReorthogonalizePROM()
```
hughcars: This is only used in a test, this seems more like something to be used to construct an test value against which to compare. This can be moved into `RomOperatorTest` or just be a method in that cpp file for generating a comparison.

palace/models/romoperator.cpp
```diff
@@ -523,9 +788,9 @@ void RomOperator::SolvePROM(int excitation_idx, double omega, ComplexVector &u)
   else
   {
     // LU solve.
-    RHSr = Ar.partialPivLu().solve(RHSr);
+    RHSr = Ar.fullPivLu().solve(RHSr);
```
hughcars: 1) is the loss of precision really that important here to warrant jumping to full? 
2) If the precision is an issue, why not use a different solve than LU like `Eigen::HouseholderQR`? The downside of householder (non-parallelizable) doesn't really apply, but for small and dense systems it is I believe the most stable.

palace/models/romoperator.cpp
```diff
@@ -535,4 +800,113 @@ std::vector<std::complex<double>> RomOperator::ComputeEigenvalueEstimates() cons
...
+{
+  using mat_t = decltype(Kr);
+  std::unique_ptr<mat_t> m_Linv = {};
+  std::unique_ptr<mat_t> m_Rinv = {};
+  std::unique_ptr<mat_t> m_C = {};
+
+  Eigen::VectorXd v_conc = Eigen::VectorXd::Ones(GetReducedDimension());
+
+  // Lumped ports are real, added at the beginning and in order.
+  for (long j = 0; j < space_op.GetLumpedPortOp().Size(); j++)
```
hughcars, line 792: This lights up on `-Wextra -Wall`, use `std::size_t`. There's a few others in these changes, I suggest turning those two flags on and chasing them down.

palace/models/romoperator.cpp
```diff
@@ -409,78 +545,207 @@ void RomOperator::SolveHDM(int excitation_idx, double omega, ComplexVector &u)
...
+              "Lumped port fields on the mesh should have exactly zero overlap. This may "
+              "be non-zero if attributes share edges.");
+
+  // // Debug Print
+  // if constexpr (false)
+  // {
+  //   fs::path folder_tmp = fs::path(iodata.problem.output) / "prom_port_debug";
+  //   fs::create_directories(folder_tmp);
+  //   PrintPROMMatrices(iodata.units, folder_tmp);
+  // }
```
hughcars: `if constexpr (false)` is used to ensure it will continue to compile, leave it uncommented.

palace/models/romoperator.cpp
```diff
@@ -409,78 +545,207 @@ void RomOperator::SolveHDM(int excitation_idx, double omega, ComplexVector &u)
...
+
+  const auto norm_re = linalg::Norml2(comm, u.Real());
+  const auto norm_im = linalg::Norml2(comm, u.Imag());
+  const auto norm_tol = ORTHOG_TOL * std::sqrt(norm_re * norm_re + norm_im * norm_im);
+  const bool has_real = (norm_re > norm_tol);
+  const bool has_imag = (norm_im > norm_tol);
+
+  const std::size_t dim_V_old = V.size();
+  std::size_t dim_V_new =
+      V.size() + static_cast<std::size_t>(has_real) + static_cast<std::size_t>(has_imag);
```
hughcars: Use `std::size_t{has_real}` I believe that avoids the conversion warning as it knows a single bit can be embedded losslessly.

palace/models/romoperator.cpp
```diff
@@ -535,4 +800,113 @@ std::vector<std::complex<double>> RomOperator::ComputeEigenvalueEstimates() cons
   return {};
 }
 
+std::tuple<std::unique_ptr<Eigen::MatrixXcd>, std::unique_ptr<Eigen::MatrixXcd>,
+           std::unique_ptr<Eigen::MatrixXcd>>
+RomOperator::CalculateNormalizedPROMMatrices(const Units &units) const
+{
+  using mat_t = decltype(Kr);
```
hughcars: `Kr` appears concrete now (`Eigen::MatrixXcd`), prefer that over `decltype` usage.

palace/models/romoperator.hpp
```diff
@@ -48,45 +123,77 @@ class MinimalRationalInterpolation
...
   SpaceOperator &space_op;
 
   // Used for constructing & reuse of RHS1.
   int excitation_idx_cache = 0;
 
-  // HDM system matrices and excitation RHS.
+  // HDM system matrices and excitation:
+  // - System matrix is: A(ω) = K + iω C - ω² M + A2(ω).
+  // - Excitation / drive: = iω RHS1 + RHS2(ω).
+  // - Vector r is internal vector workspace of size RHS
```
hughcars, line 160: This vector `r` would be a good candidate for a buffer passed into the InnerProduct as above.

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
 {
-  MFEM_ASSERT(static_cast<std::size_t>(m) <= V.size(),
-              "Out of bounds number of columns for MGS orthogonalization!");
-  for (int j = 0; j < m; j++)
+  MFEM_ASSERT(m <= V.size(), "Out of bounds number of columns for MGS orthogonalization!");
+  for (std::size_t j = 0; j < m; j++)
   {
-    H[j] = linalg::Dot(comm, w, V[j]);  // Global inner product
+    // Global inner product: Note order is important for complex vectors.
+    H[j] = dot_op.InnerProduct(comm, w, V[j]);
```
hughcars: If the global reduction here is done afterwards, akin to the `comm` usage below, then the requirements on `dot_op` drop to being a definition of an inner product processor local. Which when combined with the type erasure trick you're using here, allows for passing in lambdas directly provided they satisfy `VecType x VecType -> ScalarType` as a single callable event. 

```
  for (std::size_t j = 0; j < m; j++)
  {
    // Note order is important for complex vectors.
    H[j] = dot_op(w, V[j]);
    Mpi::GlobalSum(1, H[j], comm);
    w.Add(-H[j], V[j]);
  }
```

palace/linalg/orthog.hpp
```diff
@@ -15,36 +19,131 @@ namespace palace::linalg
...
+// Assumes that the input vectors are normalized, but does not normalize the output vectors!
+// If done in a loop, normalization has to be managed by hand! (TODO: Reconsider).
+//
+
+// Concept: InnerProductHelper has function InnerProduct(VecType, VecType) -> ScalarType,
+// acting on local degrees of freedom. Also add MPI reduction.
+
+// Simplest case is canonical inner product on R & C.
+
+class InnerProductStandard
```
hughcars: This file can be reorganized and the type erasure trick you're doing expanded considerably, if you drop the MPI requirement, see below. This also plays into the naming.
```
class InnerProduct
{
public:
  template <typename VecType>
  auto operator()(const VecType &x, const VecType &y) const
  {
    return LocalDot(x, y);
  }
};
```
Then `RealWeightedInnerProduct` would be the other class. The really key thing is by dropping MPI, and moving to `operator()` the only requirement becomes being a bilinear functor on `R` or `C` or both, the purest definition of an inner product. This then means arbitrary inner products can be defined locally to any call site through lambdas (and can use captures, which becomes pretty powerful).

palace/models/romoperator.cpp
```diff
@@ -335,26 +386,112 @@ RomOperator::RomOperator(const IoData &iodata, SpaceOperator &space_op,
...
 
-  // The initial PROM basis is empty. The provided maximum dimension is the number of sample
-  // points (2 basis vectors per point). Basis orthogonalization method is configured using
-  // GMRES/FGMRES settings.
-  MFEM_VERIFY(max_size_per_excitation * excitation_helper.Size() > 0,
-              "Reduced order basis storage must have > 0 columns!");
-  V.resize(2 * max_size_per_excitation * excitation_helper.Size(), Vector());
+    // Zero out port dofs of bulk in ctor.
+    weight_op_W = {std::move(port_tdof_list), std::move(w_bulk_assemble_parop),
+                   std::move(w_port_assemble_parop)};
```
hughcars: This is an example of how an external buffer for an inner product could be helpful. weight_op_W has access to an external buffer in the form of `r` in the `RomOperator` class, which I believe is the correct size. So it could be passed into an `operator()` based `Orthogonalize` routine with a lambda with a reference into the 3 arg version.

palace/models/romoperator.hpp
```diff
@@ -6,18 +6,93 @@
...
+struct InnerProductHybridBulkBoundary
+{
+  // Need this since SetEssentialTrueDofs makes pointer to these values, even though it
+  // looks like it is passing by references.
+  mfem::Array<int> port_tdof_list;
+
+  std::unique_ptr<ParOperator> W_inner_product_weight_bulk = {};
+  std::unique_ptr<ParOperator> W_inner_product_weight_port = {};
+
+  mutable Vector v_workspace = {};
```
hughcars: Get rid of this workspace and change the `InnerProduct` (soon `operator()`) to a 3 arg and then pass through with `r` from `RomOperator` below.

palace/models/romoperator.hpp
```diff
@@ -6,18 +6,93 @@
...
+  {
+    W_inner_product_weight_bulk->Mult(x, y);
+    W_inner_product_weight_port->AddMult(x, y);
+  }
+
+  inline double InnerProduct(const Vector &x, const Vector &y) const
+  {
+    v_workspace.SetSize(x.Size());
+    v_workspace.UseDevice(x.UseDevice());
+    //
```
hughcars: Debris.

palace/models/romoperator.hpp
```diff
@@ -100,8 +207,24 @@ class RomOperator
...
+  // to reconstruct if field configuration are pure real, imaginary or complex.
+  void UpdatePROM(const ComplexVector &u, std::string_view node_label,
+                  double drop_degenerate_vector_norm_tol = 0.0);
+
+  // Re-orthogonalize PROM basis. Take current PROM vectors V and perform additional V =
+  // Q'R' decomposition, then combine R'R. Brute force approach to numerical precision
+  // issues due to almost orthogonality of basis (from convergence or otherwise). Contrast
+  // with `drop_degenerate_vector_norm_tol` in `UpdatePROM` above. Requires re-projecting
+  // high-dimensional matrices.
+  void ReorthogonalizePROM();
```
hughcars: This appears to be a testing method not used in any source, only in tests. Move it there or remove.

palace/models/spaceoperator.cpp
```diff
@@ -841,6 +841,78 @@ bool SpaceOperator::GetExcitationVector(int excitation_idx, double omega,
...
+
+  GridFunction rhs(GetNDSpace());
+  rhs = 0.0;
+  rhs.Real().ProjectBdrCoefficientTangent(fb, attr_marker);
+  GetNDSpace().GetRestrictionMatrix()->Mult(rhs.Real(), Et_primary.Real());
+  if (zero_metal)
+  {
+    linalg::SetSubVector(Et_primary.Real(), GetNDDbcTDofLists().back(), 0.0);
+  }
+  return true;
```
hughcars: What is the point of these returning `bool`? The other excitation are indicating if there is a non-zero excitation, and thus an early exit possible. It seems these will never do such an early exit, might you do something similar. 

Additionally, `GetLumpedPortExcitationVectorPrimaryHtcn` appears to be only used in testing, move there or remove.

palace/models/romoperator.hpp
```diff
@@ -6,18 +6,93 @@
...
+    MFEM_VERIFY((W_inner_product_weight_bulk->NumCols() ==
+                 W_inner_product_weight_bulk->NumCols()) &&
+                    (W_inner_product_weight_bulk->NumCols() ==
+                     W_inner_product_weight_bulk->NumCols()),
+                "Mismatch sizes of inner product matrices!")
+  }
+
+  inline auto NumCols() const { return W_inner_product_weight_bulk->NumCols(); }
+
+  inline auto NumRows() const { return W_inner_product_weight_bulk->NumRows(); }
```
hughcars: Given this is a struct, and these methods are only used in testing, can consider deleting these and the `Mult`.

test/unit/test-romoperator.cpp
```diff
@@ -0,0 +1,712 @@
...
+  // Check element-wise sum is ok.
+  for (int i = 0; i < weight_op_eigen.rows(); i++)
+  {
+    for (int j = 0; j < weight_op_eigen.cols(); j++)
+    {
+      CHECK_THAT(weight_op_eigen(i, j),
+                 WithinRel(W_bulk_eigen(i, j) + W_port_eigen(i, j)) ||
+                     WithinAbs(0.0, 1e-18));
+    }
+  }
```
hughcars, line 280: This test amounts to checking that `Mult` and `AddMult` on `ParOperator` works, which is beside the point of the tests in this file. Testing that the sub operators behave above as expected is good and valid, but if you drop this unnecessary bit, you can drop `Mult` and `NumCols` and `NumRows` from the HybridInnerProduct struct, which aren't necessary in the source code.

test/unit/test-romoperator.cpp
```diff
@@ -0,0 +1,712 @@
...
+
+  // Debug Print.
+  // if constexpr (false)
+  // {
+  //   Eigen::IOFormat HeavyFmt(4, 0, ", ", ";\n", "[", "]", "[", "]");
+  //   std::cout << W_bulk_eigen.format(HeavyFmt) << "\n";
+  //   std::cout << W_port_eigen.format(HeavyFmt) << "\n";
+  //   std::cout << weight_op_eigen.format(HeavyFmt) << "\n";
+  // }
```
hughcars: `if constexpr (false)` shouldn't be commented, as it ensures can compile in future.

test/unit/test-romoperator.cpp
```diff
@@ -0,0 +1,712 @@
...
+  RomOperatorTest prom_op(iodata, space_op, max_size_per_excitation);
+
+  // Test hybrid weight operator works as expected.
+  const auto &weight_op = prom_op.GetWeightOp();
+  REQUIRE(weight_op.has_value());
+
+  const auto &W_bulk = weight_op->W_inner_product_weight_bulk;
+  const auto &W_port = weight_op->W_inner_product_weight_port;
+
+  // If serial — concertise as an Eigen matrix on a thread and manually check zeros of row
```
hughcars: concretize

test/unit/test-romoperator.cpp
```diff
@@ -0,0 +1,712 @@
...
+        {
+          mat(j, i) = v(j);
+        }
+      }
+      return mat;
+    };
+
+    auto W_port_eigen = toEigenMatrix(*W_port);
+    auto W_bulk_eigen = toEigenMatrix(*W_bulk);
+    auto weight_op_eigen = toEigenMatrix(*weight_op);
```
hughcars: Unused.

palace/drivers/drivensolver.cpp
phdum-a: Done in commit below. Went through all changed cpp files on this branch using clang-tidy misc-include-cleaner UnusedIncludes.

test/unit/test-romoperator.cpp
phdum-a: Done.

test/unit/test-romoperator.cpp
phdum-a: Ok. The way debug-printing is handled in general with `if constexpr (false)` could probably use work throughout Palace, but happy to follow the other cases.

test/unit/test-romoperator.cpp
phdum-a: Done.

palace/utils/iodata.hpp
phdum-a: No, it is important for now. This is the "quick-fix" that allowed dealing with global state issue in the tests. Currently both Stratton-Chu, and LumpedPort tests need to call this function, and also this is needed in #556.

We should refactor IOData to make it better, but I would say that is beyond the scope of this PR.

test/unit/test-lumpedportintegration.cpp
phdum-a: Fixed

test/unit/test-lumpedportintegration.cpp
phdum-a: Ok, happy to remove them from here.

palace/models/romoperator.cpp
phdum-a: Yes – You are right. This slipped through the rebase from an earlier implementation of the weighted inner product.

palace/models/romoperator.cpp
phdum-a: Done

palace/linalg/orthog.hpp
phdum-a: Done

palace/models/lumpedportoperator.hpp
phdum-a: Ok. I'm happy to adjust to "standard" Palace, although I have thoughts about this for another time. (e.g. `nₑₗₑₘₛ²`)

palace/utils/iodata.cpp
phdum-a: Done for the double neagtives.

 For the `auxpec` — I just followed the already existing checks from below which check the two independently. E.g. 

```
    if (!boundaries.auxpec.empty() || !boundaries.waveport.empty())
    {
      Mpi::Warning(
          "Transient problem type does not support wave port boundary conditions!\n");
    }
```

So would leave that as is to match that style.

test/unit/test-lumpedportintegration.cpp
phdum-a: Oh, this is just a placeholder to understand what MFEM does here. The failure comes when calling `InnerProduct` on the `mfem::ParBilinearForm`. For example in test-lumpedportintegration.cpp:260 we have

```
  double port_power =
      g_boundary_id.InnerProduct(port_primary_gf_ht_cn.Real(), port_primary_gf_et.Real());
  if (boundary_pec_type != PEC::SIDE)
  {
    CHECK_THAT(port_power, WithinRel(1.0));
  }
```

which passes on 1 MPI rank the error on 2 MPI ranks

```
/test-lumpedportintegration.cpp:265: FAILED:
  CHECK_THAT( port_power, WithinRel(1.0) )
with expansion:
  0.0 and 1.0 are within 2.22045e-12% of each other
```

The reason I left this open is I did not have the time to figure out which of `InnerProduct`, `ParInnerProduct` and `TrueInnerProduct` is correct here. Or if I need to do a manual MPI reduction additionally. Or if I need to add Prolongation or Restriction matrices somewhere. 

I looked at the MFEM docs and the corresponding code, which did not help. I think it is `ParInnerProduct`, but if I do that I still get test errors.

palace/models/lumpedportoperator.hpp
phdum-a: Done

palace/models/romoperator.cpp
phdum-a: Yes. I just followed the same structure as already existed. As far as I can tell the difference between OrthogonalizeColumn & OrthogonalizeIteration is a j / j+1 indexing. But I must admin that I did not want to change the `iterative.cpp` code. There seems to be a lot of generic linear algebra routines that could be pulled out and refactored. Maybe I can open an issue for this?

palace/models/romoperator.cpp
phdum-a: It's a good question. The reason I made this change is because of the issues in the high-precision benchmarks against uniform driven solves (which we should still address). It did not help but I left it.

My feeling here is that we are doing so much work in the adaptive approach to solve the HDM system and that anything on the ROM is cheap. So why not use a much more stable algorithm here? I'm very happy to switch to pivoted QR just on that basis.

However there are more serious stability points why we should worry about this more.

- As with other iterative methods (like Lanczos or similar) convergence of a good basis is by construction related to loss of orthogonality. Similarly here as we add more vectors to the PROM they will become more and more orthogonal. (See also the CG2 comment). 
- We do not check for this problem at all in the PROM logic, probably because we assume that the PROM error is so much bigger than the double limit. But given some of the interpolation / circuit things we want to do that may not be a good assumption. We have definitely observed poor conditioning of the PROM system in postprocessing.
- I don't know this for certain, but I would expect the port modes also make conditioning much worse since the "natural excitation" of the port modes by themselves are at very high energy.

palace/models/romoperator.cpp
phdum-a: This is useful for debugging but more on the side of debris. Removed.

palace/models/lumpedportoperator.hpp
hughcars: There's a bit of a mixture, where a whole word subscript is being used, then _{} is fine. It's an indicator that maybe the whole word subscript ought to be dropped though and the formula made more compact given the limited context. `nₑ²` for instance.

phdum-a: - `nₑ` would not be ideal since the index e is already used for individual elements `e`.
- I would have thought that people who read these comments, would know how to read the latex directly or can find a way to render it in their IDE?

palace/models/romoperator.cpp
phdum-a: I agree that this is bad to do. The problem is that `orthog_type` is used for both for orthogonality of the actual GMRES & SLEPC solvers as well as for the PROM projection orthogonality, which are related but different. Since the PROM is a rational interpolation, things should be more accurate and we should avoid losing precision since this will introduce bigger errors once interpolating.

We could add a `adaptive_solver_orthog_type` config option?

palace/linalg/orthog.hpp
phdum-a: Yes, this should all be static polymorphism since you know that operation needs what inner product in advance.

palace/drivers/drivensolver.cpp
phdum-a: Outdated — no longer an issue.

palace/models/spaceoperator.hpp
phdum-a: Outdated — now addressed.

palace/models/spaceoperator.cpp
phdum-a: See new code below.

palace/models/romoperator.cpp
phdum-a: See #617

palace/models/romoperator.cpp
phdum-a: Throughout the code (and all of palace) integer lengths and convergence are inconsistent and light up. For example, Eigen matrices use `long` as their `Index` type, MFEM generally uses `int` and `std::vector` uses `size_t`. These need a more systematic approach if we want to deal with these warnings.

In this case if I use size_t in the loop, the `orth_R(j, j)` call lights up. If I just clang-tidy with (clang-diagnostic-sign-conversion,cppcoreguidelines-narrowing-conversions) checks on all of palace I get >150 hits. Granted many are in the rom code, but that was because it uses Eigen more than other places.

I did a quick once over to see if there are any easy things to fix to reduce some easy cases, but I don't think sprinkling 100s of static-casts around the place the best solution.

palace/models/romoperator.cpp
phdum-a: Fixed.

palace/models/romoperator.cpp
phdum-a: Done.

palace/models/romoperator.hpp
phdum-a: Done.

palace/models/romoperator.cpp
phdum-a: I don't understand what this means:

>  is this the Linv for a mode?

K,C,M is the spring-system / "mechanics" labels for this. In electrical circuits we would normally the matrix admittance equation is $Y(\omega) = (j\omega L)^{-1} + R^{-1} + j \omega C$ where $L^{-1}$ is the matrix inverse of the inductance matrix, $R^{-1}$ is the inverse of the resistance, and $C$ is the (Maxwell) capacitance matrix.

The name `m_Linv` is just `matrix_L_inverse`. I'm ok to drop the `m_` although then you just have the variable `C` floating around for the capacitance.

test/unit/test-romoperator.cpp
phdum-a: Added comment.

palace/linalg/orthog.hpp
phdum-a: Sure. You can even make them static...

palace/utils/iodata.cpp
phdum-a: Might be over optimization — pass by value is ok, but it is a copy of an object we then don't need anymore. In the tests, I have the case where I am constructing an `json` object rather than passing it to the ctor. In that case config is an lvalue and I want to move. Of course, in the long term I don't really need to build a json object in the test. It is just a quick-fix way to run the current `IoData` ctor with all side-effects but without loading from a file.

palace/linalg/orthog.hpp
phdum-a: Yes, moved to tests. Originally I thought that I would be able to make the hybrid operator into a type  `Operator` and then use this. But I did not find an easy way to do that, even though I think that makes it much nicer structurally.

Even some of the tests may now be obsolete, but I think having an orthogonalisation with and without inner product makes sense structurally to test. Am just going to move the class into the tests.

palace/linalg/orthog.hpp
phdum-a: (See longer comment below)

palace/models/romoperator.hpp
phdum-a: ~~Done.~~

No, I want to keep this for now. (see general discussion below)

palace/models/romoperator.hpp
phdum-a: To be discussed as part of the stability matters. 

Re-orthogonalization is pretty standard in Krylov subspace techniques, since convergence and loss of orthogonality go hand-in-hand. This is a function that should exist somewhere (maybe in the `linalg` section with othogonalize columns). I added this since I thought stability was a concern — it did not have a large impact in the test cases, but we should discuss if it might.

palace/models/spaceoperator.cpp
phdum-a: It was basically to mirror the other function signature, even thought the bool here contained no interesting value. Removed.

palace/models/romoperator.hpp
phdum-a: See below. The ideal scenario would be if InnerProduct did not need a buffer.

palace/models/romoperator.cpp
phdum-a: See comment below

palace/utils/iodata.hpp
phdum-a: (See above comment)

palace/linalg/orthog.hpp
phdum-a: (Now this code is only used in tests)

palace/linalg/orthog.hpp
phdum-a: Used. (Now this code is only used in tests)

phdum-a:
I think the practical question here is what to fix for the `v_workspace` in `InnerProductHybridBulkBoundary`. Without wrapping InnerProduct, maybe we could split this class into an Operator and a InnerProductOp class that passes the buffer vector `r` at the call site?

palace/models/romoperator.cpp
phdum-a: Right. I would not pass a pointer here since that will be really bad for race conditions on `r`, so passing at the call site is better.

test/unit/test-romoperator.cpp
phdum-a: Adjusted in lates PR since HybridInnerProduct is now a Operator not an innerproduct. Keep this as it is cheap and a complementary test then previous column ones.

palace/models/romoperator.hpp
phdum-a: Done, in rework.

palace/linalg/orthog.hpp
phdum-a: Done. See below and rewok.

palace/linalg/orthog.hpp
phdum-a: Used. See below and rework.

palace/linalg/orthog.hpp
phdum-a: Now Obsolete.

palace/models/spaceoperator.hpp
phdum-a: Remove "electric". The combination $H_t \times n$ is with $Z_R = 1$ is not quite either. See below and port notes.

palace/utils/iodata.cpp
phdum-a: Ignore given refactors.

palace/utils/iodata.hpp
phdum-a: Ignore given refactors.

palace/models/romoperator.cpp
phdum-a: (Done with variable renames to `inductance_L_inv` and similar).

test/unit/test-lumpedportintegration.cpp
phdum-a: See #626. Will resolve for the purposes of this PR.

palace/models/romoperator.cpp
phdum-a: Done with new `AdaptiveGSOrthogonalization` config option.

palace/models/lumpedportoperator.hpp
phdum-a: Right. Originally, I thought that I would want to use them to normalise the circuit matrices explicitly, although the new inner product should take care of that. That said, I do think they are important functions (basically they are the norm of the analytic port vectors) in the same way the `LumpedPortData::GetExcitationPower()`. This normalisation and choices is implicitly baked into all the the functions of `LumpedPortData` and these norm function help document and check consistency. If you prefer, we can move them to the unit tests.

palace/models/romoperator.cpp
phdum-a: Switched to fully pivoted Householder QR for max stability.

palace/models/romoperator.hpp
phdum-a: Ok. Removed `ReorthogonalizePROM`. See also below for more detailed discussion.

palace/models/romoperator.cpp
phdum-a: Added as option

palace/models/romoperator.cpp
phdum-a: Added as option

palace/models/romoperator.cpp
phdum-a: Added as options: Now have Space overlap, (twice) real energy density, and FE space overlap, ∫dxE(x)∗E(x), ∫dxD(x)∗E(x), ∑_a E^aE^a

phdum-a:
Happy to merge as is, my convergence studies on this look good.

test/unit/test-lumpedportintegration.cpp
```diff
@@ -0,0 +1,780 @@
...
+  //
+  // This tests also iterates over solver orders 1-3.
+
+  using VT = palace::Units::ValueType;
+  MPI_Comm world_comm = Mpi::World();
+
+  // Generate 3 x 2 x 3 different test configuration.
+  auto solver_order = GENERATE(1, 2, 3);
+  auto [mesh_is_hex, mesh_path] =
+      GENERATE(std::make_tuple(true, fs::path(PALACE_TEST_DIR) /
```
Sbozzolo:
```suggestion
      GENERATE(std::make_tuple(true, fs::path(PALACE_TEST_DATA_DIR) /
```
and move your lampedport_mesh inside `data`

phdum-a: Thanks! There were also some output folder I used the fixture for, although it is unclear if I strictly needed.

PR #335, opened 2025-03-03
Export mesh after cracking pre-processing

simlapointe:
Palace does some pre-processing of the mesh to enable mesh cracking. To enable comparative studies on the same mesh whether mesh cracking is used or not, users may want an option to export this mesh after pre-processing, before cracking, for use with other configurations.


hughcars (commented):
This mesh doesn't include all the various preprocessing, so I think it's probably better to change the name. "ExportPrerefinedMesh" because the it's not removing the unused domain attributes etc which a user might think happens before cracking. Having considered if it's possible to export the whole preprocessed mesh (with removed attributes etc.) but I think that would make this way more complicated. Otherwise LGTM.

hughcars (approved):

PR #353, opened 2025-03-17
Change MUMPS solver type when using complex coarse preconditioner

simlapointe:
Set the MUMPS solver type to UNSYMMETRIC when using the complex coarse preconditioner to ensure it is exact.


hughcars (approved):
Was able to recreate the issue, then applying this fixed it.

Issue #343, opened 2025-03-09
NVRTC_ERROR_COMPILATION when running palace


WEIXUEP:
Hi, I come arcoss a strange error in the running step of palace using GPU. However, if I comment the **#include <math.h>** in palace/build/include/palace/qfunctions/33/utils_33_qf.h, everything will be ok.

The following output is generated when an error occurs:
```bash
>> /usr/bin/mpirun -n 1 /home/palace_gpu/palace/build/bin/palace-x86_64.bin cpw_lumped_uniform.json

_____________     _______
_____   __   \____ __   /____ ____________
____   /_/  /  __ ` /  /  __ ` /  ___/  _ \
___   _____/  /_/  /  /  /_/  /  /__/  ___/
  /__/     \___,__/__/\___,__/\_____\_____/

Git changeset ID: v0.13.0-345-gbee499c6-dirty
Running with 1 MPI process
Detected 1 CUDA device
Device configuration: cuda,cpu
Memory configuration: host-std,cuda
libCEED backend: /gpu/cuda/magma

Added 1796 elements in 2 iterations of local bisection for under-resolved interior boundaries
Added 980 duplicate vertices for interior boundaries in the mesh
Added 2752 duplicate boundary elements for interior boundaries in the mesh
Added 466 boundary elements for material interfaces to the mesh
Finished partitioning mesh into 1 subdomain

Characteristic length and time scales:
 L₀ = 4.500e-03 m, t₀ = 1.501e-02 ns

Mesh curvature order: 1
Mesh bounding box:
 (Xmin, Ymin, Zmin) = (-2.500e-04, -5.000e-04, -1.000e-03) m
 (Xmax, Ymax, Zmax) = (+4.250e-03, +2.432e-03, +1.000e-03) m

Parallel Mesh Stats:

                minimum     average     maximum       total
 vertices          4171        4171        4171        4171
 edges            24081       24081       24081       24081
 faces            36718       36718       36718       36718
 elements         16810       16810       16810       16810
 neighbors            0           0           0

            minimum     maximum
 h       0.00281235    0.167882
 kappa      1.06818      14.748

Configuring Robin absorbing BC (order 1) at attributes:
 4

Configuring Robin impedance BC for lumped ports at attributes:
 5: Rs = 2.241e+02 Ω/sq, n = (+0.0, +0.0, +1.0)
 9: Rs = 2.241e+02 Ω/sq, n = (-0.0, +0.0, +1.0)
[...]

PR #354, opened 2025-03-19
Fix linker error for cuda shared builds

hughcars:
When building with cuda on a shared build would fail with
```
[ 97%] Building CXX object CMakeFiles/palace.dir/main.cpp.o
[100%] Linking CUDA device code CMakeFiles/palace.dir/cmake_device_link.o
[100%] Linking CXX executable palace-x86_64.bin
/opt/gcc/lib/gcc/x86_64-redhat-linux/12.3.0/../../../../x86_64-redhat-linux/bin/ld: CMakeFiles/palace.dir/cmake_device_link.o: undefined reference to symbol '__cudaUnregisterFatBinary@@libcudart.so.12'
/opt/gcc/lib/gcc/x86_64-redhat-linux/12.3.0/../../../../x86_64-redhat-linux/bin/ld: /usr/local/cuda-12.4/targets/x86_64-linux/lib/libcudart.so.12: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
gmake[5]: *** [CMakeFiles/palace.dir/build.make:179: palace-x86_64.bin] Error 1
gmake[4]: *** [CMakeFiles/Makefile2:250: CMakeFiles/palace.dir/all] Error 2
gmake[3]: *** [Makefile:136: all] Error 2
gmake[2]: *** [CMakeFiles/palace-install.dir/build.make:83: palace-cmake/src/palace-stamp/palace-build] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:292: CMakeFiles/palace-install.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
```

Additionally address libceed issue with nvidia run time compiler identified in https://github.com/awslabs/palace/issues/343


cameronrutherford:
I know a review wasn't requested from me, but LGTM. This is standard CMake for linking CUDA libs. There are other CMake targets available if you want to link against more / different libraries.

https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#cuda-runtime-library

phdum-a:
In general, this looks fine. 

I was going through the GPU build and I'm not sure I understand everything, but here some things I wanted to flag. (They may or many not be actual issues).

1. The [CMAKE_CUDA_STANDARD](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_STANDARD.html), [CMAKE_CUDA_STANDARD_REQUIRED](https://cmake.org/cmake/help/v3.31/variable/CMAKE_CUDA_STANDARD_REQUIRED.html), and [CUDA_EXTENSIONS](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_EXTENSIONS.html)  are not set in the `./palace/CMakeLists.txt`. As far as I can tell these are not automatically defaulted to their corresponding C++ versions? That said, using a target_compile_feature to specify the standard apparently works for both. (See here: https://developer.nvidia.com/blog/building-cuda-applications-cmake/#c++_language_level). I don't know if we want to specify these flags as well?
2. I'm a bit confused by the nvidia compiler vs toolkit and how it ensure that the correct versions of each or used (in the case of multiple versions). And also how Palace deals with this information. The only concern I have is that the libraries you find and link to with `FindCUDAToolkit` are actually the same ones that are passed to all the other dependancies (ceed, mfem, etc). Here what I found:
- In `FindCUDAToolkit` searches for the toolkit in a specific order (first the search directory of `nvcc` if the CUDA language is enabled, then from [CMAKE_CUDA_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER) env variable, then `CUDAToolkit_ROOT`, then `CUDA_PATH `, etc.
- However, in the superbuild `./CMakteLists.txt` cuda is found by searching the `CMAKE_CUDA_COMPILER` variable and then during truncations of directory. That variable then gets repackaged as `BIN_NVCC_DIR` and `CUDA_DIR` and passed onto CEED and HYPER. The other deps MFEM, SuperLU still use `CMAKE_CUDA_COMPILER`.
- I'm not 100% sure, but the Palace set-up seems to make some extra assumptions here. For example, you could use `clang` as a cuda compiler not `nvcc` and specify a custom `CMAKE_CUDA_COMPILER_TOOLKIT_ROOT`. There is a warning about this in [FindCUDAToolkit](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html) that `CUDAToolkit_NVCC_EXECUTABLE` could be different from `CMAKE_CUDA_COMPILER`.
- Maybe it's better to call `FindCUDAToolkit` in `./CMakteLists.txt` as well  and then pass the specific executables down to ceed & hyper? Unsure.
3. Is linking the runtime the only thing you need here? Are there others components that are needed?
4. Are there other cuda interactions/settings that need checking. For example, I know Eigen can use CUDA kernels,and this is not explicitly turned off anywhere in Palace (`EIGEN_NO_CUDA`). No idea if this "does the right thing", but I guess you've not seen any errors...

phdum-a (approved):

phdum-a:
`check_language` is fine, but `enable_language` will gives the same behavior right now. cmake will just throw an error if it fails the check.

cameronrutherford:
> `check_language` is fine, but `enable_language` will gives the same behavior right now. cmake will just throw an error if it fails the check.

Right, and don't we want that error in case something is wrong? I suppose if something's wrong, the error will happen at compile time anyway, not CMake configure time.

hughcars:
> In general, this looks fine.
> 
> I was going through the GPU build and I'm not sure I understand everything, but here some things I wanted to flag. (They may or many not be actual issues).
> 
> 1. The [CMAKE_CUDA_STANDARD](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_STANDARD.html), [CMAKE_CUDA_STANDARD_REQUIRED](https://cmake.org/cmake/help/v3.31/variable/CMAKE_CUDA_STANDARD_REQUIRED.html), and [CUDA_EXTENSIONS](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_EXTENSIONS.html)  are not set in the `./palace/CMakeLists.txt`. As far as I can tell these are not automatically defaulted to their corresponding C++ versions? That said, using a target_compile_feature to specify the standard apparently works for both. (See here: https://developer.nvidia.com/blog/building-cuda-applications-cmake/#c++_language_level). I don't know if we want to specify these flags as well?
> 2. I'm a bit confused by the nvidia compiler vs toolkit and how it ensure that the correct versions of each or used (in the case of multiple versions). And also how Palace deals with this information. The only concern I have is that the libraries you find and link to with `FindCUDAToolkit` are actually the same ones that are passed to all the other dependancies (ceed, mfem, etc). Here what I found:
> 
> * In `FindCUDAToolkit` searches for the toolkit in a specific order (first the search directory of `nvcc` if the CUDA language is enabled, then from [CMAKE_CUDA_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER) env variable, then `CUDAToolkit_ROOT`, then `CUDA_PATH `, etc.
> * However, in the superbuild `./CMakteLists.txt` cuda is found by searching the `CMAKE_CUDA_COMPILER` variable and then during truncations of directory. That variable then gets repackaged as `BIN_NVCC_DIR` and `CUDA_DIR` and passed onto CEED and HYPER. The other deps MFEM, SuperLU still use `CMAKE_CUDA_COMPILER`.
> * I'm not 100% sure, but the Palace set-up seems to make some extra assumptions here. For example, you could use `clang` as a cuda compiler not `nvcc` and specify a custom `CMAKE_CUDA_COMPILER_TOOLKIT_ROOT`. There is a warning about this in [FindCUDAToolkit](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html) that `CUDAToolkit_NVCC_EXECUTABLE` could be different from `CMAKE_CUDA_COMPILER`.
> * Maybe it's better to call `FindCUDAToolkit` in `./CMakteLists.txt` as well  and then pass the specific executables down to ceed & hyper? Unsure.
> 
> 3. Is linking the runtime the only thing you need here? Are there others components that are needed?
> 4. Are there other cuda interactions/settings that need checking. For example, I know Eigen can use CUDA kernels,and this is not explicitly turned off anywhere in Palace (`EIGEN_NO_CUDA`). No idea if this "does the right thing", but I guess you've not seen any errors...

1. I'm not sure we need those terms right now, but there's a reasonable chance we might expose issues as more gpu builds are done. We can cross that bridge when we get to it.
2. I cleaned up the hypre and libceed builds that use CUDA_DIR, did a few builds and no other issues presently.
3. I believe it is, some dependencies might use other pieces, but thats hypre and libceed which are using the cuda root dir to find the pieces they need as far as I understand it. I've not had issues, but as we find I can add in the extra linking steps fairly easily.
4. We're not doing and device transfers with eigen presently, so I don't think we need to toggle that unless we encounter issues. 

I also added in a fix for some libceed/cuda includes that were causing issues elsewhere.

PR #358, opened 2025-03-26
Fix compiler spec in GitHub workflow to work with `spack@v1.x`

cameronrutherford:
Currently main pipelines are failing to build palace due to changes that were made in `spack@v1.x`. This moved compilers into a package instead of a compiler explicitly, and so hence this also changed how the spec syntax handles packages.

### DIFF
--- a/.github/workflows/spack.yml
+++ b/.github/workflows/spack.yml
@@ -82,7 +82,7 @@ jobs:
           fi
 
           # Build and install
-          PALACE_SPEC="local.palace@develop%${{ matrix.compiler }}$GPU_VARIANT ^${{ matrix.mpi }}"
+          PALACE_SPEC="local.palace@develop$GPU_VARIANT ^${{ matrix.mpi }} %${{ matrix.compiler }}"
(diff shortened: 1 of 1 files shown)


cameronrutherford:
Spack build is now at least concretizing! Hopefully nothing else broke.

cameronrutherford:
@hughcars spack builds are failing with this error:

```
 ...
     277              CC build/backends/xsmm/ceed-xsmm-tensor.o
     278              CC build/backends/ceed-backend-weak.o
     279              CC build/gallery/ceed-gallery-weak.o
     280            LINK build/libceed.so
     281    /usr/bin/ld: build/interface/ceed.o: warning: relocation against `C
            eedMemTypes' in read-only section `.text'
     282    /usr/bin/ld: build/interface/ceed-basis.o: relocation R_X86_64_PC32
             against symbol `CeedFESpaces' can not be used when making a shared
             object; recompile with -fPIC
  >> 283    /usr/bin/ld: final link failed: bad value
  >> 284    collect2: error: ld returned 1 exit status
```

Looks to be something todo with shared libs and how libceed is compiled.

PR #359, opened 2025-03-27
Add GHCR cache to speed up spack builds

cameronrutherford:
Will take over #358 as the build ended up failing.

This should hopefully speed up builds substantially before we work on debugging the underlying cause of the failing pipelines.

PR #362, opened 2025-04-02
Draft: Modify package.py locally to get spack builds to pass

cameronrutherford:
Based on #359, this attempts to debug the issues in our local `package.py` that was causing errors in building / concretization.

For now this is a draft MR as a convenience to run CI tests, but eventually will be made into it's own PR with isolated / smaller changes.


spack/local/packages/palace/package.py
cameronrutherford, line 1: Seems like I have successfully refactored / modified this in order to get the concretization to almost match previously, but with new dependencies added. I still need to figure out why libceed / mfem needed to be commented out.

cameronrutherford:
I won't do this now, but it seems like it would be a good idea to maintain spack pipelines for develop and the latest tagged spack release separately. That way we:

- are insulated against breaking spack changes, but know when things are broken
- ensure we have stable builds with a stable spack for regular CI

.github/workflows/spack.yml
```diff
@@ -81,25 +76,124 @@ jobs:
...
+      - name: Load Blas
+        run: eval $(spack -e ${SPACK_ENV} load --sh ${{ matrix.blas }}) 
+
+      - name: Build Dependencies
+        run: spack -e ${SPACK_ENV} install --only-concrete --no-check-signature --fail-fast --show-log-on-error --only dependencies
+
+      - name: Build Palace
+        # Build palace from source using this current directory
+        run: spack -e ${SPACK_ENV} install --verbose --only-concrete --keep-stage --show-log-on-error --only package --no-cache
+
```
cameronrutherford, line 164:
```suggestion
      - name: Build Dependencies
        run: |
          eval $(spack -e ${SPACK_ENV} load --sh ${{ matrix.blas }})
          spack -e ${SPACK_ENV} install --only-concrete --no-check-signature --fail-fast --show-log-on-error --only dependencies

      - name: Build Palace
        # Build palace from source using this current directory
        run: spack -e ${SPACK_ENV} install --verbose --only-concrete --keep-stage --show-log-on-error --only package --no-cache

```

cameronrutherford:
Okay I am not really convinced quite yet, but it appears like palace spack build went through. I was able to reproduce the build issues running in Ubuntu on WSL though. Next step is going to be pointing the CMake to the external libCEED instead, and then getting the build to work both locally and in CI.

If that works and tests pass, then we can work on moving other dependencies into having them optionally spack managed.

cameronrutherford:
Spack CPU builds are passing. CUDA builds are failing because of libCEED (again...), but ROCm builds are failing because of mesa. I will try and reproduce the CUDA errors outside of CI and debug further, and same with ROCm builds.

I am tempted to just disable the libCEED check, but it seems to be failing for good reason, and so I will keep debugging.

IIRC I think for the ROCm failures, they get fixed by building / loading the llvm toolchain as a compiler before installing everything else. It's a little clunky, but it's the way compilers "should be installed with spack" anyway (i.e. `spack install llvm-amdgpu && spack external find --path=$(spack location -i llvm-amdgpu) llvm-amdgpu` or something similar).

I will at least add testing support in for the non-GPU build, and then if I can't get the GPU builds to pass, we can always address in another PR.

cameronrutherford:
Closing in favor of #366

PR #364, opened 2025-04-08
Remove function up-streamed in mfem 4.8

phdum-a:
Small tweak to remove helper functions for interpolator. These that upstreamed to mfem and are now released in mfem 4.8.


hughcars:
Cherry picked commit into https://github.com/awslabs/palace/pull/363

PR #366, opened 2025-04-13
Spack CPU Build Fix. Add spack built libCEED support.

cameronrutherford:
Using this in favor of #362 and #359.

I have modified a few things in order to include an external libCEED here. I will need to upstream the libCEED changes that I have made here, and I have already up streamed the gslib changes.

The next steps after this PR gets merged would be to:

- [ ] Get GPU builds working for CUDA
- [ ] Get GPU builds working form ROCm
- [ ] Expanding build configurations tested of palace in CI (more variants, compilers. perhaps replacing existing linux tests...)
- [ ] Move other packages to be internally includable with spack. Currently I want to start with `mfem`, but ideally that will be all packages moving forward

I have some scripts in `spack/scripts` that I was using for local testing. If it would be good to keep those in the repo, I can clean them up a bit. In the end there was really only one script that was useful here. While I did have another script that I used for internal testing, it's almost identical to what is in `spack/scripts/local-build.sh`.

PR #367, opened 2025-04-15
Draft: External MFEM and libxsmm on mac support

cameronrutherford:
Will help remove our constraints on mac not supporting libxsmm, as well as being able to actually use a spack built mfem.

While I could just hack together some `find_package` magic, I instead just want to use the exported config - so I will be testing changes from https://github.com/spack/spack/pull/49790

This is targeting #366 so I can make more incremental changes.


cameronrutherford:
Going to create a new branch / PR with more accurate naming as I took care of MacOS builds in #366

Issue #263, opened 2024-06-10
Allow frequency domain driven simulations at list of desired frequency samples


sebastiangrimberg:
In addition to the linear frequency sweep currently implemented, configured with `"MinFreq"`, `"MaxFreq"`, and `"FreqStep"`, allow user to override with a list of frequencies, perhaps using the `"FreqSamples"` keyword. This is incompatible with adaptive fast frequency sweep.

WEIXUEP:
Hi, I believe that the Advanced Design System (ADS) by Keysight can serve as a valuable reference. 

![Image](https://github.com/user-attachments/assets/57259eff-89fa-4c5c-97a8-40d6ea49c1f4)

In my code, I run palace multiple times and combine the results of different frequency range. This only requires packaging of the code, without changing the core algorithm.

`Freq1 = {"Type": "Adaptive",
         "MinFreq": 1.8,
         "MaxFreq": 2.2,
         "FreqStep": 0.05}

Freq2 = {"Type": "Linear",
         "MinFreq": 1.0,
         "MaxFreq": 1.2,
         "FreqStep": 0.1}

user_config_data["Freq"].append(Freq1)
user_config_data["Freq"].append(Freq2)`

Issue #373, opened 2025-05-23
Support ranges in `"SaveStep"`


hughcars:
`"SaveStep"` right now supports fixed interval samples, every N-th sample. Users sometimes are most interested in a few critical frequencies or times, and these aren't distributed uniformly in the sampling space. It would be good to be able to support explicitly provided samples and times, through `"SaveStep": [2.0, 5.0, 5.5, 10.0]` for instance in a frequency sample. 

Would need to come up with a mechanism for checking if these samples are actually being sampled however. Samples could be requested at times that were not part of the time integration, and at frequencies not part of the sampling. Could also support indexes in the case of eigenmode and statics simulations, "save only the 1st, 4th and 5th solutions".

PR #374, opened 2025-05-27
Support arbitrary frequencies in driven simulations

hughcars:
Introduce `"FreqSamples"` to support arbitrary frequencies in driven simulations, these samples are added into those given by the linear space sampling, for a "uniform" sampling these are simply added into the set of frequencies, interleaved in the appropriate location. In the adaptive sampling these points are explicitly added into the basis, thereby ensuring the model accuracy in the vicinity of these points, and providing a mechanism to manually reconstruct a given PROM from one simulation to another. These points are used in calculating the termination criteria, but are not exited from early, thus if the final samples in a sequence of manual samples satisfies the memory convergence criteria, no additional samples will be taken. 

Closes https://github.com/awslabs/palace/issues/263, https://github.com/awslabs/palace/issues/373


palace/cmake/PkgConfigHelpers.cmake
```diff
@@ -133,12 +133,8 @@ int main()
     RUN_OUTPUT_VARIABLE PETSC_TEST_OUTPUT
   )
   if(PETSC_TEST_COMPILED AND PETSC_TEST_EXITCODE EQUAL 0)
-    # message(STATUS "PETSc test program - Successful")
     set(${_petsc_test_success} TRUE PARENT_SCOPE)
   else()
-    message(STATUS "PETSc test program - Failed")
-    message(STATUS "PETSC_TEST_COMPILE_OUTPUT: ${PETSC_TEST_COMPILE_OUTPUT}")
-    message(STATUS "PETSC_TEST_OUTPUT: ${PETSC_TEST_OUTPUT}")
```
phdum-a, line 141: Is this an intended cleanup as a part of this PR?

palace/drivers/drivensolver.cpp
```diff
@@ -38,31 +38,44 @@ DrivenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+  std::for_each(omega.begin(), omega.end(), [=](double &x)
+                { x = iodata.solver.driven.min_f + x * iodata.solver.driven.delta_f; });
+  omega.insert(omega.end(), iodata.solver.driven.sample_f.begin(),
+               iodata.solver.driven.sample_f.end());
+  std::sort(omega.begin(), omega.end());
+  constexpr double tol = 1e-12;
+  omega.erase(std::unique(omega.begin(), omega.end(),
+                          [=](auto x, auto y) { return std::abs(x - y) < tol; }),
+              omega.end());
+
```
phdum-a, line 57: Depending on what we decided `Restart` means in this case, `step0` should either truncate the linspace in line 47, something like
```
std::vector<double> omega(n_step-step0);
std::iota(omega.begin(), omega.end(), step0);
```
or (more likely) truncate the beginning of omega after it has been assembled.

palace/drivers/drivensolver.cpp
```diff
@@ -227,11 +249,9 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op, int n_step,
   const double unit_GHz = iodata.units.Dimensionalize<Units::ValueType::FREQUENCY>(1.0);
   Mpi::Print("\nBeginning PROM construction offline phase:\n"
              " {:d} points for frequency sweep over [{:.3e}, {:.3e}] GHz\n",
-             n_step - step0, omega0 * unit_GHz,
-             (omega0 + (n_step - step0 - 1) * delta_omega) * unit_GHz);
+             omega_sample.size() - step0, omega_hdm[0] * unit_GHz, omega_hdm[1] * unit_GHz);
```
phdum-a, line 252: What is some of the manually specified frequencies are outside [f_min, f_max]?

palace/drivers/drivensolver.cpp
```diff
@@ -177,21 +190,30 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &space_op, int n_step, i
...
 {
   // Initialize postprocessing for measurement and printers.
   PostOperator<config::ProblemData::Type::DRIVEN> post_op(iodata, space_op);
   auto excitation_helper = space_op.GetPortExcitations();
 
+  // Configure HDM sample array first with end points then any explicit samples.
+  std::vector<double> omega_hdm = {iodata.solver.driven.min_f, iodata.solver.driven.max_f};
+  omega_hdm.insert(omega_hdm.end(), iodata.solver.driven.sample_f.begin(),
+                   iodata.solver.driven.sample_f.end());
+
```
phdum-a, line 205: See comment text.

phdum-a: Side node: this is different than the current behavior even if there is no `sample_f`. Currently, if we have a restart the PROM will be computed with the new omega0 as minimum of the PROM domain, meaning we have a changing PROM for different restarts. Honestly, restart should probably not be used for adaptive and if it were your sampling is more consistant....

palace/models/romoperator.cpp
```diff
@@ -491,7 +491,7 @@ void RomOperator::SolvePROM(int excitation_idx, double omega, ComplexVector &u)
   // the matrix Aᵣ(ω) = Kᵣ + iω Cᵣ - ω² Mᵣ + Vᴴ A2 V(ω) and source vector RHSᵣ(ω) =
   // iω RHS1ᵣ + Vᴴ RHS2(ω). A2(ω) and RHS2(ω) are constructed only if required and are
   // only nonzero on boundaries, will be empty if not needed.
-  if (has_A2)
+  if (has_A2 && Ar.rows() > 0)
   {
```
phdum-a, line 495: Why are these changes needed? The PROM should never be evaluated without some HDM solves first and, we should throw / abort earlier?

phdum-a (commented):
Hi Hugh,

The idea of having a more flexible frequency sampling interface is great. However I would argue that this PR needs a little more time baking.

(Below I will call the normal linear grid `LinSpace` for brevity).

Big picture thoughts:
1. How does `SaveStep` interact with the added frequency samples? This is what determines when the Paraview fields get added. Since the `LinSpace` and `FreqSamples` get merged and sorted, doing anything except saving everything (`SaveStep = 1`) is fiddly. Let's say I want a normal linear grid and the visualize *only* the few `FreqSamples` I added by hand. Right now you can't do that since `SaveStep` applies to the merged list.
2. How does this interact with `Restart`? Currently, I think (?) there is a bug and this information (`step0` in `drivensolver.cpp`) does not truncate the beginning of your frequency range anymore. The restart is a slightly odd option that only really makes sense in a specific situation (for the non-adaptive solve with one excitation only). But should `Restart` act on the combined `LinSpace` and `FreqSamples` or just `LinSpace`? I can make arguments for either.
3. I would argue against the idea of automatically adding the frequencies `FreqSamples` as HDM solution points for the PROM. For me, `FreqSamples` are frequencies to samples the output (the online phase). We could add an option to force frequencies in the offline phase (in addition to the end-points), but this should be a separate flag.
	1. Coupling these things reduces the usefulness of the adaptive solver to begin with as compared to just the direct solver. I expect most people will sample at a whole bunch of output frequencies, in which case you are doing way more HDM solves than needed.
	2. It also introduces the following annoying behavior. Say you have a problem in your PROM model and want to re-run to check the Paraview output at specific frequencies `FreqSamples`. Then the PROM does a new HDM solve there and that region will be super-accurate — not telling you anything about the PROM before you put the sample point there.
	3. Lastly, I would be careful about including a non-opt-in strategy for PROM sampling into the API since we may changes the default at a later date if we have better ways of PROM sampling. If we want to revert this decision, people rerunning old jobs might not get the accuracy they expect anymore since they way oversampled by adding manual pivots...
4. There are not constraints on the frequencies in `FreqSamples` — e.g. check that zero frequency is not in our sample. Although we don't do this right now either. But for example should FreqSamples be constrained between fmin,fmax? Something to consider at least, even if we decided not to.

Details:
1. Your tolerance for frequency deduplication (1e-12) is different from the tolerance used in `GetNumSteps` (1e-9). Probably want to make this the same and maybe even make this a user facing option with default?
2. I do like your change of not dropping the last HDM solve from the PROM loop. This was actually a commit I had on my PROM branch.

---

I don't know how extensive you are open to changing the interface for this. But a maximalist solution could be in the driven solver to have a list of ranges that will then be concretized and merged. E.g.

```json
"FrequencySamples" : [
	{"Type": "Linear", "MinFreq" : 1.0, "MaxFreq": 5.0, "FreqStep": 0.2, "SaveFreqStep": 0.4},
	{"Type": "Linear", "MinFreq" : 6.0, "MaxFreq": 10.0, "NrSteps": 10, "SaveStep": 0},
	{"Type": "Points", "Freq" : [3.11, 3.22, 3.33], "SaveStep": 1},
	{"Type": "Points", "Freq" : [7.11, 7.22, 7.33], "SaveStep": 0},
	{"Type": "Log", "MinFreq" : 10.0, "MaxFreq": 50.0, "NrSteps": 0.4},
	...
]
```

We could determine `MinFreq` and `MaxFreq` globally from those samples or still demand that the user adds them (as global filters / cut-offs). If we want to go there we could also add filters to the above list, although that is probably too much for this interface:

```json
	...
	{"Type": "Filter", "MinStep": 5}
	{"Type": "Filter", "MaxFreq": 2.0}
```

If you want to bring in PROM sampling into it, we could add an extra keyword like 

```json
{"Type": "Points", "Freq" : [3.11, 3.22, 3.33], "SaveStep": 1, "SolveHDMInAdaptive": True}
```

That way, online and offline phases can be separated.

Also – frequency point sampling is analogous to time checkpointing specification for the adaptive solve. @simlapointe — I would be curious if you have opinions on these types of interfaces?

Ok, I hope this helps. Couple of minor points inline.

hughcars (commented):
Expanded the PR to handle the different selection of ranges suggested, used slightly different wording for some entries, but the broad capability is there. Additionally the `"Save"` keyword is introduced for driven which allows for picking specific frequencies (provided they are in the list of sampled frequencies). This is joined onto those selected by `"SaveStep"`. Should consequently partially address https://github.com/awslabs/palace/issues/373 which can also be expanded to other simulation types with `"Save"` supporting a vector.

palace/utils/configfile.cpp
```diff
@@ -1589,31 +1589,268 @@ void BoundaryData::SetUp(json &config)
...
+      if (it_sample == sample_f.end())
+      {
+        std::cout << "sample_f: ";
+        for (auto x : sample_f)
+          std::cout << x << ' ';
+        std::cout << '\n';
+        MFEM_ABORT("Save frequency " << *it_save << " not found in sample frequencies!");
+      }
+    }
+    save_step.emplace_back(std::distance(sample_f.begin(), it_sample) + 1);  // 1-based
```
simlapointe, line 1806: I don't think this should be 1-based since it will be used with a 0-based step in postoperator.cpp

palace/drivers/drivensolver.cpp
```diff
@@ -123,14 +127,9 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &space_op, int n_step, i
     post_op.InitializeParaviewDataCollection(excitation_idx);
 
     // Frequency loop.
-    double omega = omega0;
-    for (int step = step0; step < n_step; step++, omega += delta_omega)
+    int step0 = std::max(iodata.solver.driven.rst - 1, 0), step = step0;
```
phdum-a, line 130: Readability: split initialization statements onto two rows.

palace/drivers/drivensolver.cpp
```diff
@@ -163,7 +167,7 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &space_op, int n_step, i
       }
 
       auto total_domain_energy =
-          post_op.MeasureAndPrintAll(excitation_idx, step, E, B, omega);
+          post_op.MeasureAndPrintAll(excitation_idx, step++, E, B, omega);
```
phdum-a: Readability: pass `step` here  and step++ at end of loop.

palace/drivers/drivensolver.cpp
```diff
@@ -377,7 +381,7 @@ ErrorIndicator DrivenSolver::SweepAdaptive(SpaceOperator &space_op, int n_step,
         // B = -1/(iω) ∇ x E + 1/ω kp x E
         floquet_corr->AddMult(E, B, 1.0 / omega);
       }
-      post_op.MeasureAndPrintAll(excitation_idx, step, E, B, omega);
+      post_op.MeasureAndPrintAll(excitation_idx, step++, E, B, omega);
```
phdum-a: As above on readability

palace/models/postoperator.cpp
```diff
@@ -108,7 +108,7 @@ PostOperator<solver_t>::PostOperator(const IoData &iodata, fem_op_t<solver_t> &f
   // initialize if needed.
   if (solver_t == config::ProblemData::Type::DRIVEN)
   {
-    paraview_delta_post = iodata.solver.driven.delta_post;
+    paraview_save_indices = iodata.solver.driven.save_indices;
```
phdum-a, line 111: I thought about suggesting a `std::move` of the vector here since `save_indices` could be large. But this will leave the entry in iodata empty state and I'm not sure we want that. Maybe a reference? Ok for now, but the use or non-use of iodata after ctor is all done might be worth pondering.

palace/models/postoperator.hpp
```diff
@@ -393,8 +398,7 @@ class PostOperator
 public:
   // Ctor.
   // - nr_expected_measurement_rows is helper int to reserve space in csv printer tables.
```
phdum-a: Remove comment about `nr_expected_measurement_rows`

docs/src/config/solver.md
```diff
@@ -198,6 +211,48 @@ sampling algorithm for constructing the reduced-order model for adaptive fast fr
...
+{
+    "Type": <string>,
+    "MinFreq": <float>,
+    "MaxFreq": <float>,
+    "FreqStep": <float>,
+    "NSample": <float>,
+    "Freq": [<int float>],
+    "SaveStep": <int>,
+    "AddToPROM": <bool>
+}
```
phdum-a: "Samples" should be an array containing dictionaries of this, not just a single dictionary. Not sure how this is written out in JSON.... Maybe

```
"Samples": [ <range_spec>]
```
where
```
<ranges_spec> = { ...}
```

docs/src/config/solver.md
```diff
@@ -178,9 +180,20 @@ fields to disk for [visualization with ParaView](../guide/postprocessing.md#Visu
 Files are saved in the `paraview/` directory under the directory specified by
 [`config["Problem"]["Output"]`](problem.md#config%5B%22Problem%22%5D).
 
+`"Samples" [None]` : Array of sample specifications that specify how to construct frequency
+samples. These are all combined to form a sorted and unique collection of samples. These
+samples can be instead of, or in addition to, the interface provided by `"MinFreq"`,
+`"MaxFreq"`, `"FreqStep"` and `"SaveStep"`. See
+[`solver["Driven"]["Samples"]`](solver.md#solver%5B%22Driven%22%5D%5B%22Samples%22%5D).
+
+`"Save" [None]` : Array of frequencies to save computed fields to disk for [visualization
```
phdum-a, line 191: "Additional array of samples..." / in addition to one specified through SaveStep of ranges.

palace/models/postoperator.cpp
hughcars: I considered this; binding a reference is possible, but if the save_indices is introduced for other drivers it becomes complicated. Either it means binding all post operators onto the driven piece of iodata, or having some `? :` logic in the initializer list, or an iile with a switch like the `dom_post_op`. Given the number of save samples is in general going to be <100, copying 100 doubles is 800 bytes (+std::vector overhead), for a significant reduction in the complexity of the source. So I opted for the copy.

docs/src/config/solver.md
hughcars: This approach is mirroring the definition of substructs elsewhere, for instance in the `"Solver"` definition:
```
"Solver":
{
    "Eigenmode":
    {
        ...
    },
    "Driven":
    {
        ...
    },
}
```
or in the boundaries definition
```
"Boundaries":
{
    "PEC":
    {
        ...
    },
    "PMC":
    {
        ...
    },
    "Impedance":
    [
        ...
    ],
```
the idea being that any `...` means "A definition of this subsection is about to appear, before we resolve the next option at this current depth in the json", which is then addressed by the `"Samples"` section that immediately follows, and is linked to in the comments about that entry. Separating the `Samples` specification out of the main body of the `"Driven"` given the length of both `"Driven"` and the relative complexity of the discussion of `"Samples"` was worthwhile. However the entry should be `"Samples" : [ ... ]` to match the style of the `"Impedance"` entry.

palace/drivers/drivensolver.cpp
hughcars: In general yes, but I think initialization like this is ok for such a copy, especially given they're both values, so no possible `& *` confusion.

docs/src/config/solver.md
phdum-a: Yeah, I was just worried about the square brackets, since it currently looks like "Samples" is a single dictionary.  "Impedance" currently looks like

 ```json
  "Impedance":
  [
      {
          "Attributes": [<int array>],
          "Rs": <float>,
          "Ls": <float>,
          "Cs": <float>
      },
      ...
  ]
```

So this could be

```
"Samples":
[
  {
      "Type": <string>,
      "MinFreq": <float>,
      "MaxFreq": <float>,
      "FreqStep": <float>,
      "NSample": <float>,
      "Freq": [<float array>],
      "SaveStep": <int>,
      "AddToPROM": <bool>
  },
  ...
]
```

PR #380, opened 2025-06-12
Fix eigenvalue errors printed in wrong columns

simlapointe:
Fixes a small typo causing the absolute and backward eigenvalue errors to get printed under the wrong column headers.


phdum-a:
Thanks for catching things — sorry for the bug. 
Is there way we could have caught this with the CI or the tolerances to unsteady?

PR #384, opened 2025-06-17
Move enums from config into labels.hpp

hughcars:
Various enums from config are used more widely in the code base, or are shadowed in applications. This PR hoists them out and does some renaming of them and instances of them. This was part of the PostOperator refactor PR, but doing it completely got too large, so is easier to move out as it should change no functionality.


simlapointe (approved):
Looks good to me!

PR #395, opened 2025-07-01
Added --version command line flag for version information

### LINKED ISSUE #394: Implement a `--version` flag
There is no `--version` CLI flag. I think you have to run Palace and inspect log output to determine that, given only some Palace executable. Suggest not to use `-v` shorthand given potential for confusion (e.g. `--verbose`).

sanjayy-gowdaa:
## Summary
Adds a new `--version` command line flag to Palace that displays version information without requiring a full simulation run.

## Motivation
Previously, users had to run Palace and inspect log output to determine the version. This change provides a standard CLI pattern for quickly checking version information.

## Changes
- ✅ Added `--version` command line option to argument parsing
- ✅ Displays Git changeset ID when available from build system
- ✅ Falls back to "UNKNOWN" when version information is not available
- ✅ Exits immediately after printing version (consistent with `--help` behavior)
- ✅ Updated help text to document the new `--version` option
- ✅ No short form (`-v`) to avoid confusion with potential `--verbose` flag

## Usage
```bash
# Show version information
palace --version

# Example outputs:
# Palace version: v0.13.0-123-g1a2b3c4d  (when git info available)
# Palace version: UNKNOWN                 (when git info not available)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format of this changelog is based on
 
 ## In progress
 
+  - Added `--version` command line flag for displaying Palace version information.
   - Fixed a small regression bug for boundary postprocessing when specifying
     `"Side": "LargerRefractiveIndex"`, introduced as part of v0.13.0.
   - Added an improvement to numeric wave ports to avoid targetting evanescent modes at

### CODE REVIEW

File: palace/main.cpp [hunk 1]
```diff
@@ -231,6 +232,18 @@ int main(int argc, char *argv[])
...
+      if (std::strcmp(GetPalaceGitTag(), "UNKNOWN"))
+      {
+        Mpi::Print(world_comm, "Palace version: {}\n", GetPalaceGitTag());
+      }
+      else
+      {
+        Mpi::Print(world_comm, "Palace version: UNKNOWN\n");
+      }
+      return 0;
+    }
```
hughcars: Given `GetPalaceGitTag()` already returns UNKNOWN as required, no need for the extra logic:
```
    if (argv_i == "--version")
    {
      Mpi::Print(world_comm, "Palace version: {}\n", GetPalaceGitTag());
      return 0;
    }
```

hughcars (changes requested):
Works as described, a minor simplification is possible.

palace/main.cpp
sanjayy-gowdaa: Thanks @hughcars, I've updated the code as suggested to simplify the version check. The logic now directly prints the output from GetPalaceGitTag()

sanjayy-gowdaa:
Thanks! @hughcars,  I've made the suggested changes.

hughcars (approved):

hughcars:
I almost forgot, please add a short entry to the `CHANGELOG.md` file.

PR #396, opened 2025-07-03
Pin MUMPS to older version

hughcars:
Current MUMPS CI has an MPI hang, this pr pins to the version previous to https://github.com/awslabs/palace/pull/387 as an immediate solution


zhanggy2004:
Not sure if it is related to this issue, but I noticed MUMPS has changed some cmake option names [between version 5.7 and 5.8](https://github.com/scivision/mumps/compare/v5.7.2...v5.8.0.0). In particular, `openmp->MUMPS_openmp`, `scalapack->MUMPS_scalapack`, `metis->MUMPS_metis`, etc. 

image

Just FYI. Thanks.

PR #398, opened 2025-07-07
Add flags to generate coverage reports

phdum-a:
Adds the cmake variable `PALACE_BUILD_WITH_COVERAGE` which adds compilation and linking flags to libpalace and palace unit tests. This can generate coverage reports. Flags are available for both  LLVM-based compilers (Clang, AppleClang, IntelLLVM) by using the "[source-based code coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html)" and for gcc using [`gcov`](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html).

In order to get a proper coverage report of palace, we have to link the full static library libpalace.a to the unit tests. This is apparently system and linker dependent. I ran into problems with this using the mac linker compiling with the external clang. Cmake now provides an easy way to do this [link](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_LINK_LIBRARY_USING_FEATURE.html#loading-a-whole-static-library), but it requires cmake 3.24 released in Aug. 2022.

Also, this PR does not contain any helper functions or helper cmake targets to actually run the tests with profiling or post-process the output (`llvm-profdata` and `llvm-cov` for clang / `gcov` for gcc).

It also bumps the Catch2 version to a more recent version.

### DIFF
--- a/docs/src/developer/testing.md
+++ b/docs/src/developer/testing.md
@@ -22,14 +22,14 @@ with `CMake`. From the *Palace* root directory:
 
 ```bash
 mkdir build && cd build
-cmake -DPALACE_MFEM_USE_EXCEPTIONS=yes -DPALACE_BUILD_WITH_COVERAGE=yes ..
+cmake -DPALACE_MFEM_USE_EXCEPTIONS=yes ..
 make -j palace-tests
 ```
 
 The `PALACE_MFEM_USE_EXCEPTIONS` option is necessary to properly capture and
-test assertions. The `PALACE_BUILD_WITH_COVERAGE` flag is used to enable
-measuring test coverage. See [Unit test coverage](#Unit-test-coverage) for more
-details on this.
+test assertions. If you want to also measure test coverage, turn
+`PALACE_BUILD_WITH_COVERAGE` on. See [Unit test coverage](#Unit-test-coverage)
+for more details on this.
 
 Once the build completes, run the tests with:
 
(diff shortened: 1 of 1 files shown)


docs/src/developer/notes.md
```diff
@@ -38,6 +38,13 @@ and `ANALYZE_SOURCES_CPPCHECK` to `ON` will turn on static analysis using
 [`cppcheck`](https://cppcheck.sourceforge.io/), respectively, during the build step. This
 requires the executables to be installed and findable by CMake on your system.
 
+## Coverage
+
+During the `cmake` configuration step, setting the variable `PALACE_BUILD_WITH_COVERAGE` to `ON`
+will turn on coverage reporting. For LLVM-based compilers (Clang, AppleClang, IntelLLVM) this uses
+the "[source-based code coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html)". For
+gcc, it uses [`gcov`](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html).
+
```
hughcars: The links here are technically sufficient, in that following the llvm one was enough to get going, but the gcov one took annoyingly long. I would suggest adding a link to gcovr, and the incantation we arrived at in the end `gcovr --gcov-executable $(which gcov-15) -r ../palace .` (for a brew installed `gcc-15`), and possibly an example call for mac and linux development. Unless the barrier to usage is pretty low, adoption in the development workflow is unlikely to be significant.

The ultimate version would be a target `palace-test-coverage` which would do the relevant incantation to generate the resulting reports, but this is a step in the right direction. We can add such a target later if developers are getting benefit from this.

hughcars (changes requested):
LGTM, small fix to the documentation. Basically to add some example command sequences to generate the reports, rather than requiring each user to grep through the reference material (which is technically sufficient, but also a barrier to uptake).

hughcars:
Also, add an entry to `CHANGELOG`.

hughcars (approved):
Do not actually merge. Handing over review approval to @Sbozzolo.

PR #401, opened 2025-07-08
Move binary cache push to end of Spack pipeline

cameronrutherford:
In #394 and #397, @hughcars observed that the Spack pipelines are failing due to the inability for external fork based PRs to push to the GHCR cache.

This PR attempts to resolve that issue.

### DIFF
--- a/.github/workflows/spack.yml
+++ b/.github/workflows/spack.yml
@@ -163,5 +163,7 @@ jobs:
       # Push built binaries, even if the build fails
       # NOTE: This might fail as external fork PRs can't push to GHCR.
       - name: Push to GHCR cache
-        if: "!cancelled()"
+        if: |
+          !cancelled() &&
+          (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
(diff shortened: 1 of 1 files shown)


.github/workflows/spack.yml
```diff
@@ -160,3 +159,9 @@ jobs:
           # Run tests
           julia --project=test/examples -e 'using Pkg; Pkg.instantiate()'
           julia --project=test/examples --color=yes test/examples/runtests.jl
+
+      # Push built binaries, even if the build fails
+      # NOTE: This might fail as external fork PRs can't push to GHCR.
+      - name: Push to GHCR cache
+        if: "!cancelled()"
+        run: spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache
```
cameronrutherford, line 169: In order to always have the green checkmark for external fork PRs, we would need something like this:

```suggestion
        run: spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache || true
```

Not that I am explicitly against something like this, but I think that seeing legitimate failures here is important, and so hopefully this just means that someone is forced to look at the spack pipeline, see that the final step is what failed, and then decide to merge...

PR #406, opened 2025-07-11
Fix bug in bounding box algo for planar boxes.

hughcars:
The bounding box algorithm for non-axis aligned planar ports will fail presently, as the change in https://github.com/awslabs/palace/pull/268 could check vertices against themselves. 

This adds a test using a point cloud from a failing case, and also some more printing.


palace/fem/lumpedelement.cpp
```diff
@@ -50,14 +50,16 @@ UniformElementData::UniformElementData(const std::array<double, 3> &input_dir,
   direction /= direction.Norml2();
 
   // Compute the length from the most aligned normal direction.
-  constexpr double rel_tol = 1.0e-6;
+  constexpr double rel_tol = 1.0e-3;
```
simlapointe: What is the rationale for this fairly loose tolerance?

simlapointe (approved):
LGTM! Just one question

palace/fem/lumpedelement.cpp
hughcars: There was a failing test from Sophie, where the projected length was mismatching even after addressing the bounding box error. But I think we figured out her port geometry was being mangled by gds anyway, so I think I'll revert this.

PR #410, opened 2025-07-17
Set OMP_NUM_THREADS in CI

hughcars:
The OpenMP CI jobs were not actually testing OpenMP because OMP_NUM_THREADS was not being set.


simlapointe (commented):
LGTM!

PR #413, opened 2025-07-25
Fix typos

Sbozzolo:
As I was going through palace, I found a few typos here and here. This PR fixes the ones I've seen so far.

The only noteworthy one is that the example was mentioning the `surface-C.csv` file. I don't know if this is an old file, but the capacitance matrix is now in the `terminal-C.csv`.


hughcars (approved):
LGTM

PR #415, opened 2025-07-25
Add "Quick Start" tutorial

Sbozzolo:
This PR adds a "Quick Start" tutorial that walks new users through their first simulation by providing a different view on the "two sphere" example. 

The aim of the tutorial is to expose new users to the basic concepts involved with installing palace and running and analyzing simulations. `quick.md` leverages Documenter.jl's ability to execute code upon doc generation, so that many sections in the tutorial are populated automatically. This has two consequences:
- the "quick start" experience itself is somewhat tested,
- many updates to the two sphere examples or to palace are automatically reflected to the rendered tutorial.
For this to happen, I updated the docs.yml to install palace with spack. In order to avoid a 30+ minute build, the workflow has to be configured to ensure that it hits the GHCR cache.

The tutorial does not cover meshing. Instead, this PR also adds comments to the `examples/spheres/mesh.jl` file, with the hope to be more accessible to users with no prior gmsh experience.

Read it here:
https://awslabs.github.io/palace/previews/PR415/quick/

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,18 +7,84 @@ on:
     tags: '*'
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# We need spack so that we can install palace for the tutorials
+env:
+  SPACK_VERSION: develop
+  REGISTRY: ghcr.io/awslabs/palace
+  GITHUB_USER: ${{ github.actor }}
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 jobs:
   build-docs:
     permissions:
+      packages: write
       contents: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
+      - name: Setup Spack
+        uses: spack/setup-spack@v2
+        with:
+          ref: develop
+          buildcache: true
+          color: true
+
+      - name: Setup Environment
+        run: |
+          # Spack.yaml with most / all settings configured
+          cat << EOF > spack.yaml
+          spack:
+            specs:
+              - palace
+            view: false
+            config:
+              install_tree:
+                root: /opt/spack
+                padded_length: False
+            concretizer:
+              reuse: false
+              unify: true
+              targets:
+                granularity: generic
+            packages:
+              petsc:
+                require: ~hdf5
+            mirrors:
+              spack:
+                binary: true
+                url: https://binaries.spack.io/${SPACK_VERSION}
+              local-buildcache:
+                binary: true
+                url: oci://${{ env.REGISTRY }}-${SPACK_VERSION}
+                signed: false
+                access_pair:
+                  id_variable: GITHUB_USER
+                  secret_variable: GITHUB_TOKEN
+          EOF
+
+      - name: Configure Binary Mirror Keys
+        run: spack -e . buildcache keys --install --trust
+
+      - name: Bootstrap Spack
+        run: spack  -e . bootstrap now
+
+      - name: Concretize
+        run: spack -e . concretize -f
+
+        # Relies on cache-hit
+      - name: Install Palace
+        run: spack -e . install --only-concrete --keep-stage --show-log-on-error
+
       - name: Build and deploy
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          eval $(spack -e . load --sh palace)
           julia --project=docs -e 'using Pkg; Pkg.instantiate()'
           julia --project=docs --color=yes docs/make.jl
 
--- a/docs/src/quick.md
+++ b/docs/src/quick.md
@@ -1,0 +1,548 @@
+```@raw html
+<!--- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved. --->
+<!--- SPDX-License-Identifier: Apache-2.0 --->
+```
+
+```@setup json
+using JSON
+using OrderedCollections # Needed to preserve printing order as in the original file
+
+# JSON.jl doesn't support // comments
+strip_json_comments(json_str) = replace(json_str, r"//.*?($|\n)" => s"\1")
+
+# JSON.jl expands lists on multiple lines, but we prefer preserving the attributes on a single line
+fix_attributes_format(json_str) = replace(replace(json_str, r"\"Attributes\":\s*\[\s*(\d+)\s*\]" => s"\"Attributes\": [\1]"))
+
+# Print just one section of the entire JSON file and prepend the correct header
+print_section(json, section) = println("\"$section\":", fix_attributes_format(JSON.json(json[section], 4)))
+
+# Move mesh and json file to build directory, so that we don't pollute the example directory
+spheres_basepath = joinpath(@__DIR__, "..", "..", "examples", "spheres")
+json_name = "spheres.json"
+mesh_name = "spheres.msh"
+json_basepath = joinpath(spheres_basepath, json_name)
+
+Base.cp(json_basepath, json_name, force = true)
+Base.mkpath("mesh")
+Base.cp(joinpath(spheres_basepath, "mesh", mesh_name), joinpath("mesh", mesh_name), force = true)
+
+json_text = strip_json_comments(read(json_basepath, String))
+spheres_json = JSON.parse(json_text, dicttype=OrderedDict)
+
+function logrun(cmd)
+    try
+        run(cmd)
+    catch e
+        println(e)
+    end
+    return nothing
+end
+
+```
+
+# Run your first simulation with Palace
+
+Welcome to your first tutorial with *Palace*!
+
+In this tutorial, we will:
+
+ 1. [Install *Palace* using Spack](#installing-palace)
+ 2. [Set up a simulation using a provided mesh](#the-config-file)
+ 3. [Run the simulation and visualize results with ParaView](#running-the-simulation-and-inspecting-the-output)
+
+By the end of this page, you'll understand the basic workflow of electromagnetic
+simulations with *Palace*. You will be able to follow the
+[examples](examples/examples.md) and start experimenting with setting up and
+running your own simulations.
+
+As a motivating example, we'll solve for the electrostatic potential and
+capacitance matrix of two conducting spheres inside a larger grounded sphere.
+Our system consists of:
+
+  - Two inner spherical conductors
+  - A larger outer sphere at zero potential
+  - Vacuum in between
+
+This is [one of the
+examples](examples/spheres.md#capacitance-matrix-for-two-spheres) included with
+*Palace*. For more details on the physics and comparisons against analytical
+results, see the [full example
+page](examples/spheres.md#capacitance-matrix-for-two-spheres).
+
+## A bird's-eye view of *Palace*
+
+*Palace* is a finite-element code for **PA**rallel **LA**rge-scale
+**C**omputational **E**lectromagnetics. It is a command-line executable that
+runs on single or multiple nodes, supporting both [CPUs and
+GPUs](guide/parallelism.md).
+
+A *Palace* simulation requires two main inputs:
+
+  - **Mesh file**: The mesh file describes the target geometry. *Palace* does not
+    construct meshes, so you must supply one. A large number of formats are
+    supported, allowing you to use your preferred CAD or meshing software. See the
+    [supported mesh formats](guide/model.md#supported-mesh-formats) for more
+    details.
+
+  - **Config file**: A JSON file that defines what problem to solve and how.
+    It specifies material properties, boundary conditions, problem type,
+    solver parameters, and output settings. See the [configuration
+    documentation](config/config.md) for complete details.
+
+*Palace* can solve Maxwell's equations under a few different set of assumptions,
+leading to different electromagnetic [problem types](guide/problem.md):
+
+  - **Electrostatic**: Computes static electric fields and capacitance matrices (used in this tutorial)
+  - **Magnetostatic**: Computes static magnetic fields and inductance matrices
+  - **Eigenmode**: Computes eigenmodes and eigenvalues in the frequency domain
+  - **Driven**: Computes frequency-domain response to harmonic excitations
+  - **Transient**: Computes time-domain response to boundary excitations
+
+*Palace* produces two types of primary output:
+
+ 1. CSV files with post-processed quantities (capacitance matrices, field values
+    at probe points, etc.)
+ 2. PVD files for visualizing fields with [ParaView](https://www. paraview.org)
+    or compatible software
+
+The full list of problem types and their outputs is available in the
+[problem configuration guide](config/problem.md).
+
+In this tutorial, we'll use a mesh generated with [Gmsh](https://gmsh.info/),
+create a configuration file for an `Electrostatic` problem, and visualize the
+resulting electric field with ParaView.
+
+## Installing Palace
+
+As a user, the simplest way to install *Palace* is with
+[Spack](https://spack.readthedocs.io), a package manager designed for
+high-performance computing applications.
+
+Follow the [instructions on the official
+website](https://spack.readthedocs.io/en/latest/getting_started.html) to install
+Spack. This involves cloning a repository and sourcing a `setup-env` shell
+script. Come back here once you are done with that.
+
+Let's check that Spack is correctly installed on your system. This can be
+accomplished by running:
+
+```bash
+spack --version
+```
+
+You should see your Spack version (yours may differ, that's okay):
+
+```@example json
+logrun(`spack --version`); # hide
+nothing # hide
+```
+
+!!! note "spack command not found"
+    
+    If you get a `command not found` error, revisit the [Spack instructions](https://spack.readthedocs.io/en/latest/getting_started.html)
+    and ensure you've completed all steps, including sourcing the setup script (the command starting with `.`).
+    Consider adding this to your shell initialization file (typically, `.bashrc` or `.zshrc`).
+
+With Spack installed, we can now move to *Palace*:
+
+```bash
+spack install palace
+```
+
+This will download and compile the latest release of *Palace* and its
+dependencies. This step may take tens of minutes depending on your system.
+
+Once installed, load *Palace* and verify it works:
+
+```bash
+spack load palace
+palace --help
+```
+
+You should see:
+
+```@example json
+logrun(`palace --help`); # hide
+nothing # hide
+```
+
+!!! tip "Loading Palace"
+    
+    You need to load *Palace* with `spack load palace` in each new shell session.
+    For convenience, add this command to your shell initialization file if you are a frequent *Palace* user.
+
+## (Optional) Install ParaView
+
+*Palace* optionally saves electromagnetic field data in the PVD format, which is
+immediately accessible by ParaView or ParaView-compatible software. You can
+download ParaView from the [official
+website](https://www.
+paraview.org/download/) or using your package manager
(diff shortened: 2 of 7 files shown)


.github/workflows/docs.yml
```diff
@@ -15,6 +26,66 @@ jobs:
...
+          spack:
+            specs:
+              - ${PALACE_SPEC}
+            view: false
+            config:
+              install_tree:
+                root: /opt/spack
+                padded_length: False
+            concretizer:
+              reuse: false
```
cameronrutherford, line 50:
```suggestion
              reuse: true
```

.github/workflows/docs.yml
```diff
@@ -15,6 +26,66 @@ jobs:
...
+              unify: true
+              targets:
+                granularity: generic
+            packages:
+              petsc:
+                require: ~hdf5
+              rocblas:
+                require: ~tensile
+            repos:
+            - spack/local
```
cameronrutherford:
```suggestion
```

As discussed, not needed for this workflow

.github/workflows/docs.yml
```diff
@@ -15,6 +26,66 @@ jobs:
...
+            mirrors:
```
cameronrutherford, line 57: These are evaluated in-order FYI

.github/workflows/docs.yml
```diff
@@ -15,6 +26,66 @@ jobs:
...
+        run: |
+          # If we cached these, that would be faster and safer
+          spack -e . buildcache keys --install --trust
+
+      - name: Bootstrap Spack
+        run: spack  -e . bootstrap now
+
+      - name: Install Palace
+        run: spack  -e . install palace
+
```
cameronrutherford: For reference

```
      # Push built binaries, even if the build fails
      # NOTE: This might fail as external fork PRs can't push to GHCR.
      - name: Push to GHCR cache
        if: |
          !cancelled() &&
          (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
        run: spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache
```

cameronrutherford (commented):
Small comments per offline discussion

.github/workflows/docs.yml
```diff
@@ -4,9 +4,20 @@ on:
...
+# We need spack so that we can install palace for the tutorials
+env:
+  SPACK_VERSION: develop
+  REGISTRY: ghcr.io/awslabs/palace
+  GITHUB_USER: ${{ github.actor }}
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
 jobs:
   build-docs:
     permissions:
```
cameronrutherford, line 23:
```suggestion
    permissions:
      packages: write
```

This is the only other thing that I can think of from the slack thread... It seems like we are doing everything right.

The other alternatives I can think of are:
- Just to abandon GHCR for this and use an S3 bucket
- Use different GHCR caches for each workflow

I maintain that this should just work....

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+    It specifies material properties, boundary conditions, problem type,
+    solver parameters, and output settings. See the [configuration
+    documentation](config/config.md) for complete details.
+
+*Palace* can solve Maxwell's equations under a few different set of assumptions,
+leading to different electromagnetic [problem types](guide/problem.md):
+
+  - **Electrostatic**: Computes static electric fields and capacitance matrices (used in this tutorial)
+  - **Magnetostatic**: Computes static magnetic fields and inductance matrices
+  - **Eigenmode**: Computes eigenmodes and eigenvalues in the frequency-domain
```
simlapointe: Nit: This one should be "frequency domain", no hyphen.

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+
+```@example json
+print_section(spheres_json, "Problem") # hide
+```
+
+#### Section 2: `Model` specification
+
+The `Model` section specifies the desired geometry. In addition to defining the
+mesh, it specifies how to convert mesh units to physical units using the `L0`
+parameter. For example, `L0` of `1e-2` means that one unit on the mesh is one
```
simlapointe: Nit: I'd instead write something like "one mesh unit corresponds to one centimeter"

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+
+#### Section 4: `Boundary` conditions
+
+The `Boundaries` section maps 2D surfaces in your mesh to their physical
+boundary conditions. *Palace* offers numerous boundary condition types, all
+documented in the [boundaries reference](config/boundaries.md).
+
+Unlike 3D regions which all require `Material` specifications, 2D surfaces have
+a default behavior: any surface without an explicit boundary condition is
+treated as a Perfect Magnetic Conductor (PMC), where the normal component of the
```
simlapointe: I believe PMC sets the _tangential_ component of the magnetic field to zero.

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+print_section(spheres_json, "Boundaries") # hide
+```
+
+#### Section 5: Solver settings
+
+Finally, the `Solver` section prescribes properties of the problem and the
+numerical algorithm, what device to use for the solution, and how much to save
+as PVD files. For this problem, we run on CPU, specify third-order finite
+elements, and save the fields for both terminal activations. The details of the
+algorithm parameters in `"Linear"` are not essential for this tutorial.
```
simlapointe: Nit: Instead of "algorithm parameters" I'd write "linear solver parameters".

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+nothing # hide
+```
+
+Finally, we are ready to run the simulation:
+
+```bash
+palace -np 1 spheres.json
+```
+
+`-np 1` instruct *Palace* to run on with single MPI process.
```
simlapointe: Typo/missing word? I'd write "... to run with a single MPI process."

.github/workflows/docs.yml
```diff
@@ -7,18 +7,87 @@ on:
...
 
+      - name: Setup Spack
+        uses: spack/setup-spack@v2
+        with:
+          ref: develop
+          buildcache: true
+          color: true
+          # This seems unnecessary, but otherwise we clone into ./spack
+          # This then would override the spack local repo dir, and break CI
+          path: spack-path
```
cameronrutherford: With #427, we no longer need this with our internal spack packages repo changing it's pathname.

In addition to that, this workflow doesn't use that in the first place, so we don't need this.

```suggestion
          color: true
```

docs/src/quick.md
Sbozzolo: Thank you!

docs/src/quick.md
Sbozzolo: Yes, typo!

.github/workflows/docs.yml
Sbozzolo: Thank you!

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+If you wish to skip the explanation and jump directly to running your
+simulations, go to [Running the simulation and inspecting the
+output](#running-the-simulation-and-inspecting-the-output).
+
+#### Section 1: `Problem` definition
+
+The `Problem` section identifies the problem type and the output directory. In
+this case, we choose `Electrostatic`. This means that *Palace* solves Poisson's
+equation for electric potential sequentially activating all the `Terminals` on
+the mesh. All simulation types in *Palace* have some form of iteration (over
```
hughcars: , whilst setting the non-activated terminals to ground.

that the other terminals are held at ground whilst this is occurring is important.

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+specifying its physical properties. In our mesh, we have just one 3D region (the
+vacuum between the spheres and outer boundary) identified by attribute 1. While
+vacuum properties are applied by default, you can specify various material
+characteristics as detailed in the [domains
+documentation](config/domains.md#domains%5B%22Materials%22%5D).
+
+The `Domains` section also includes a `Postprocessing` subsection for
+calculating specific quantities. In this example, we add:
+
+  - `Energy` activates integrated energy calculations in the 3D domain
```
hughcars: `Energy` which activates.

The grammar is a bit weird without the connective.

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+  - `Probe` requests field values at specific coordinates defined by a `Center`
```
hughcars: `Probe` which requests

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+When configuring `Postprocessing`, you must specify an `Index` that determines
+the suffix for column headers in the output CSV files. For example, with `Index: 1`, the probe output will show headers like `E_x[1]`.
+
+!!! note "What is the difference between `Attributes` and `Index`?"
+    
+    `Attributes` identify mesh regions and come from the mesh file. In our example,
+    attributes 1-4 identify the vacuum region, outer boundary, and two spheres.
+    
+    `Index` is used only for postprocessing and defines a notation used in the output CSV files. It has no relation to mesh attributes and can
+    be any positive integer.
```
hughcars, line 316: Maybe note that there can be multiple attributes in one index, in the example there is only a single for each, but you can have an array in general. I.e. if there were two different 3D volumes of silicon, there could be a single index referring to the combined volume of both regions.

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+#### Section 4: `Boundary` conditions
+
+The `Boundaries` section maps 2D surfaces in your mesh to their physical
+boundary conditions. *Palace* offers numerous boundary condition types, all
+documented in the [boundaries reference](config/boundaries.md).
+
+Unlike 3D regions which all require `Material` specifications, 2D surfaces have
+a default behavior: any surface without an explicit boundary condition is
+treated as a Perfect Magnetic Conductor (PMC), where the tangential component of
+the magnetic field is zero.
```
hughcars: This isn't exactly correct, any *external* boundary with no condition set is marked as PMC (whatever the natural boundary condition is strictly speaking (`n . grad phi = 0` or `n x curl E = 0` or `n x curl A = 0`)), an internal boundary has no condition applied. This is because the terms from left and right will cancel on such a boundary, so not explictly adding something is equivalent to doing nothing.

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+
+For our electrostatic problem, we define:
+
+  - The outer boundary as `Ground` (zero potential)
+  - Two `Terminal` surfaces (one for each sphere)
+
+`Terminals` are particularly important for `Electrostatic` simulations. *Palace*
+activates each terminal sequentially, applying a unit of potential and solving
```
hughcars: , when a terminal is not activated, it is held at ground (V = 0).

docs/src/quick.md
```diff
@@ -0,0 +1,542 @@
...
+problem types will have different iteration patterns (e.g., `Driven` iterating
+over frequencies or `Transient` over time steps) and many of the output CSV
+files are organized along these iterations.
+
+### Understanding the output
+
+Once the simulation is completed, you'll find a `postpro` directory containing:
+
+```@example json
+logrun(`tree postpro`); # hide
```
hughcars: I get an error for this section after having all the previous (including the run) work
```
Base.IOError("could not spawn `tree postpro`: no such file or directory (ENOENT)", -2)
```
I assume this is because I ran with `cd docs && julia ` then `] activate .` followed by `include("make.jl)`. I presume this can be done more neatly to pull this data correctly.

hughcars (changes requested):
Generally looks pretty good to me. Some minor wordsmithing and one possible issue with generating the html output of the postpro tree view, but basically ready.

docs/src/quick.md
Sbozzolo: Ah! I found out that macOS does not come with `tree` apparently. I'll get this in a different way

Sbozzolo: I changed to just use `ls`

examples/spheres/mesh/mesh.jl
```diff
@@ -1,76 +1,132 @@
...
 )
+    # Boilerplate
     kernel = gmsh.model.occ
-
     gmsh.initialize()
     gmsh.option.setNumber("General.Verbosity", verbose)
 
-    # Add model
+    # Create a new model. The name spheres is not important. If a model was already added,
+    # remove it first (this useful when interactively the body of this function in the REPL)
```
hughcars: this is

examples/spheres/mesh/mesh.jl
```diff
@@ -1,76 +1,132 @@
...
         gmsh.model.setCurrent("spheres")
         gmsh.model.remove()
     end
     gmsh.model.add("spheres")
 
-    # Geometry parameters (in cm)
-    radius_farfield = 15.0 * center_d
-
-    # Mesh parameters
+    # n_sphere: Controls how many elements around sphere circumference,
```
hughcars: the sphere's

examples/spheres/mesh/mesh.jl
```diff
@@ -1,76 +1,132 @@
...
+    #           higher values = finer mesh on spheres
+    # l_farfield: Maximum element size at the outer boundary,
+    #             larger values = coarser mesh at boundary
     n_sphere = 16
     l_farfield = 20.0
 
-    # Geometry
+    # Create three spheres: two inner conductors along the x axis and one outer boundary
```
hughcars, line 99: two inner conductors along the x axis and one outer boundary with centers along the x axis

examples/spheres/spheres.json
```diff
@@ -15,8 +14,7 @@
     "Materials":
     [
       {
-        "Attributes": [1],
-        "Permittivity": 1.0
```
hughcars: I don't think the permittivity should be removed here. It will default to 1.0 yes, but in general people might not have that, so it serves as a visual example of where it would be set.

Sbozzolo:
> Generally looks pretty good to me. Some minor wordsmithing and one possible issue with generating the html output of the postpro tree view, but basically ready.

I rebased against main, and addressed the comments.

examples/spheres/spheres.json
```diff
@@ -2,7 +2,6 @@
   "Problem":
   {
     "Type": "Electrostatic",
-    "Verbose": 2,
```
hughcars: Don't change this, the verbose 2 level is common to all the examples, and is a good balance. Additionally it provides most of the information that is needed in debugging, so we want to encourage it as a default for those who will just copy and paste.

examples/spheres/spheres.json
```diff
@@ -15,8 +14,7 @@
     "Materials":
     [
       {
-        "Attributes": [1],
-        "Permittivity": 1.0
+        "Attributes": [1]
```
hughcars: Don't remove this. The default will be vacuum yes, but we want to encourage the specification of permittivity in general, as many users will just copy paste these files without really trying to understand and then be confused about how to modify the permittivity.

docs/src/quick.md
```diff
@@ -0,0 +1,548 @@
...
+  - **Magnetostatic**: Computes static magnetic fields and inductance matrices
+  - **Eigenmode**: Computes eigenmodes and eigenvalues in the frequency domain
+  - **Driven**: Computes frequency-domain response to harmonic excitations
+  - **Transient**: Computes time-domain response to boundary excitations
+
+*Palace* produces two types of primary output:
+
+ 1. CSV files with post-processed quantities (capacitance matrices, field values
+    at probe points, etc.)
+ 2. PVD files for visualizing fields with [ParaView](https://www. paraview.org)
```
hughcars: space after `.` in link, think that's broken?

docs/src/quick.md
```diff
@@ -0,0 +1,548 @@
...
+
+```
+
+# Run your first simulation with Palace
+
+Welcome to your first tutorial with *Palace*!
+
+In this tutorial, we will:
+
+ 1. [Install *Palace* using Spack](#installing-palace)
```
hughcars: Have you tried these links? When I click I get `file:///Users/hughcars/AWS/palace/docs/build/quick/index.html#installing-palace` (on my local build, from chrome) it doesn't actually move to the correct section below. The links are case sensitive despite what the vscode markdown plug in says `#Installing-Palace` works. 

This applies to all the markdown links I believe.

docs/src/quick.md
```diff
@@ -0,0 +1,548 @@
...
+output](#running-the-simulation-and-inspecting-the-output).
+
+#### Section 1: `Problem` definition
+
+The `Problem` section identifies the problem type and the output directory. In
+this case, we choose `Electrostatic`. This means that *Palace* solves Poisson's
+equation for electric potential sequentially activating all the `Terminals` on
+the mesh while setting the non-activated terminals to ground. All simulation types
+in *Palace* have some form of iteration (over frequencies, times, mode numbers,
+or terminals). We save the output in the `postpro` folder, which will be
```
hughcars: Save in the `"Output"` folder from the `"Problem"` section, in this example `"postpro"`. Basically want to make sure this is an obviously configurable thing

docs/src/quick.md
```diff
@@ -0,0 +1,548 @@
...
+overwritten if it already exists. See the [problem types
+documentation](config/problem.md#problem-types) for details on all available
```
hughcars: In the rest of the documentation we are linking to pieces within the configuration file with their nested format
`config["Boundaries"]["PEC"]["Attributes"]` for example, to highlight in each link the location within the containing tree. I think most of the links to sections of the `config/` directory could be upgraded to links of that form (and also checked, I think most of these links aren't working due to the case sensitivity).

docs/src/quick.md
```diff
@@ -0,0 +1,548 @@
...
+print_section(spheres_json, "Problem") # hide
+```
+
+#### Section 2: `Model` specification
+
+The `Model` section specifies the desired geometry. In addition to defining the
+mesh, it specifies how to convert mesh units to physical units using the `L0`
+parameter. For example, `L0` of `1e-2` means that one mesh unit corresponds to
+one centimeter. The `Model` section can also include settings for adaptive mesh
+refinement. See [documentation](config/model.md) for more information.
```
hughcars: Gramatically this is "See the documentation", which is done in most other places here. This should be a link to the actual `config["Boundaries"]["PEC"]["Attributes"]` style section however, which would sidestep this need for the article.

docs/src/quick.md
```diff
@@ -0,0 +1,548 @@
...
+
+The `Boundaries` section maps 2D surfaces in your mesh to their physical
+boundary conditions. *Palace* offers numerous boundary condition types, all
+documented in the [boundaries reference](config/boundaries.md).
+
+Unlike 3D regions, which all require `Material` specifications, 2D surfaces have
+default behavior: any external surface without an explicit boundary condition is
+treated as a Perfect Magnetic Conductor (PMC), where the tangential component of
+the magnetic field is zero, and no conditions are imposed on internal surfaces
+(since terms from the left and right cancel out on such boundaries).
```
hughcars: Reading this left and right is unnecessarily ambiguous, `terms from either side cancel out on such boundaries` avoids confusion of why it's left or right rather than top/bottom, inside/out.

hughcars (changes requested):
Took the chance to read through again and found a few more small issues: most links are broken and there's a few minor grammar issues. Additionally the `spheres.json` file shouldn't be modified, as we don't want to encourage reducing the verbosity and eliding material properties.

docs/src/quick.md
Sbozzolo: I didn't notice that! Good to know.

PR #425, opened 2025-08-04
Add README for regression tests

Sbozzolo:
I wanted to add a new regression test and found that I needed to go through the code just to understand how to run the tests, so I added a short readme.


test/examples/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+This directory contains regression tests for the Palace examples. The tests
+validate the correctness of Palace simulations by comparing output results
+against reference data.
+
+## Running Tests
+
+### Prerequisites
+
+  - Julia with the project environment instantiated
```
simlapointe: Very much nitpicking here, but since you explain how to instantiate the Julia environment below, then having the project environment instantiated is not really a prerequisite, the prerequisite is just having Julia.

test/examples/runtests.jl
```diff
@@ -7,8 +7,17 @@ include("testcase.jl")
 
 if "PALACE_TEST" in keys(ENV)
     palace = String.(split(ENV["PALACE_TEST"], ' '))
+    palace_exec = first(palace)
```
simlapointe: Is there a way to avoid having both `palace` and `palace_exec` here? Can we just do the `which()` call below using `palace` or `first(palace)`?

simlapointe (changes requested):
Looks good! I think we should link/refer to this in the Palace docs to increase visibility, probably in the Developer Notes/Testing section.

Very minor comments below.

test/examples/runtests.jl
Sbozzolo: We can do that by ensuring that `palace` is always a Vector (it is not currently because the else branch sets `palace = "palace"`, a String, and `first(palace) = 'p'`.

Is this what you mean? (See current version)

test/examples/README.md
Sbozzolo: Good point

test/examples/runtests.jl
simlapointe: Yes, that's good.

simlapointe:
I’d add something like this at the end of the Developer Notes / Testing section of the docs: “Instructions to run and update the regression tests can be found in `test/examples/README.md`”.

simlapointe (approved):
LGTM!

PR #427, opened 2025-08-05
Update with latest Spack compatible changes.

cameronrutherford:
I am asking why we didn't see failures in our tests since these changes are necessary to work with latest spack. I think it's part of how we are specifying which package API we are using - see [slack thread](https://spackpm.slack.com/archives/C08Q62S7XEX/p1754419018258109).

Here is the corresponding Spack PR I am iterating on - https://github.com/spack/spack-packages/pull/961

I ran `spack style --fix` locally as well, so not sure why that isn't passing there, and maybe we should check that internally.


cameronrutherford:
I don't think we should merge this until we are 100% sure that the newest package API is what we are using and testing locally.

spack/local/repo.yaml
```diff
@@ -1,2 +1,3 @@
 repo:
-  namespace: 'local'
+  namespace: local
+  api: v2.0
```
cameronrutherford: I am hoping that these changes will ensure we use the latest package API, but I think we will need to manually update this every so often moving forward. I have asked if there's a way to always use the latest version, and will make another PR if needed to add that.

cameronrutherford: https://github.com/spack/spack/issues/51121

spack/local/repo.yaml
```diff
@@ -1,2 +1,3 @@
 repo:
-  namespace: 'local'
+  namespace: "local"
```
cameronrutherford: I was experimenting without the quotes here, since the docs don't have quotes, but that doesn't seem to work according to our CI - https://spack.readthedocs.io/en/latest/repositories.html

cameronrutherford:
Alright I thought the quoting issue on the repo name was why CI is failing, but clearly I was wrong. I also learned that there's API version `v2.1`.

I will run some of my local scripts to debug this further and see what I can find...

spack_repo/local/repo.yaml
cameronrutherford, line 1:
```
==> Warning: package repository /Users/rcamruzz/projects/palace/darwin-sequoia-armv8.4a/../spack/local is disabled due to: Invalid repository path '/Users/rcamruzz/projects/palace/spack/local'. Path must contain 'spack_repo/
```

This local warning isn't shown in CI, and was the answer to the failures. I think this should pass now.

cameronrutherford:
Rebased and addressed Spack PR feedback. It's a little annoying to iterate on two PRs, but could be worse. I find that some variant of this is handy:

`cp ../palace/spack_repo/local/packages/palace/package.py $(spack location -p palace)/package.py`

Issue #424, opened 2025-08-01
Eigenmodes exhibit run-to-run phase variations


simlapointe:
Eigenmodes sometime exhibit run-to-run variations. The eigenvalues and the magnitudes of the complex eigenvectors are stable, but the phases of the complex eigenvectors often differ.

PR #433, opened 2025-08-06
Normalize the phase of eigenvectors

simlapointe:
The mean phases of eigenvectors sometimes varies run-to-run. This change normalizes eigenvectors so their mean phase is on the positive real axis. The original mean phase (before normalization) is saved in eig.csv for potential use in post-processing.

Closes https://github.com/awslabs/palace/issues/424.


palace/models/postoperatorcsv.cpp
```diff
@@ -23,6 +23,7 @@ Measurement Measurement::Dimensionalize(const Units &units,
   measurement_cache.eigenmode_Q = nondim_measurement_cache.eigenmode_Q;              // NONE
   measurement_cache.error_abs = nondim_measurement_cache.error_abs;                  // NONE
   measurement_cache.error_bkwd = nondim_measurement_cache.error_bkwd;                // NONE
+  measurement_cache.phase = nondim_measurement_cache.phase;                          // NONE
```
hughcars: I know as an angle this is dimensionless, and thus doesn't need a transform, it does have units, of radians (or degrees?). I think it's worth indicating this, particular for writing the mean phase to the csv.

test/examples/ref/cylinder/cavity_impedance/eig.csv
```diff
@@ -1,16 +1,16 @@
...
- 9.000000000e+00,        +4.626848224e+00,        +1.052954226e-03,        +2.197079527e+03,        +1.172918862e-10,        +7.298205445e-06
- 1.000000000e+01,        +4.626848229e+00,        +1.052954206e-03,        +2.197079570e+03,        +2.399181808e-11,        +1.492833163e-06
- 1.100000000e+01,        +4.777130953e+00,        +1.125535002e-03,        +2.122160162e+03,        +2.338416481e-10,        +1.502228306e-05
- 1.200000000e+01,        +5.000482916e+00,        +1.085153291e-03,        +2.304044542e+03,        +3.623840461e-11,        +2.436722705e-06
- 1.300000000e+01,        +5.000486074e+00,        +1.170207011e-03,        +2.136581889e+03,        +6.138256645e-11,        +4.127455111e-06
- 1.400000000e+01,        +5.000486075e+00,        +1.170207811e-03,        +2.136580428e+03,        +1.309370044e-10,        +8.804399025e-06
- 1.500000000e+01,        +5.290573375e+00,        +1.207021636e-03,        +2.191581888e+03,        +1.802688600e-10,        +1.282398664e-05
+        m,                Re{f} (GHz),                Im{f} (GHz),                          Q,              Error (Bkwd.),               Error (Abs.),                 Mean Phase
```
hughcars: A user might be confused what the mean phase is expressed in, I suspect most people would default to radians, but worth being specific.

As a unit magnitude transform, I think the absolute and backwards error aren't changed by the rotation, is that your expectation?

PR #434, opened 2025-08-07
Various improvements to documentation

Sbozzolo:
This PR collects a few fixes/improvements that I am applying as I go through Palace.

More specifically:
- typos
- fix inaccurate instruction on how to build the documentation in the README

PR #435, opened 2025-08-07
Revert "Skip expensive CI actions when only modifying mds"

Sbozzolo:
Yesterday I opened a PR (#432) to skip CI workflows when only markdown files are modified. This is to avoid unnecessary >500 billable minutes. However, I didn't realize that we require those workflows to pass as required condition to merge a PR (through the GitHub UI).

We can fix this in four ways:
- Don't require those workflows to pass; this means that a developer could in theory click on "merge PR" even if some of those workflows are not green
- Add creating complementary workflow files to just post a green status (as officially recommended by [GitHub](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks))
- Implement path filtering at the job level so that a status is always posted
- Revert to the previous behavior

Given that PR that only modify md files are not very frequent, I feel that the added complexity is not warranted, so I opted for just reverting the change.

Note, apparently this feature is being considered for implementation natively in GitHub (https://github.com/orgs/community/discussions/44490).

This reverts commit e40a3a3a73473dba370f4d5e44f5708007723ae6.

(Found in #434)

### DIFF
--- a/.github/workflows/build-and-test-linux.yml
+++ b/.github/workflows/build-and-test-linux.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - main
   pull_request:
-    paths-ignore:
-      - '**.md'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/build-and-test-macos.yml
+++ b/.github/workflows/build-and-test-macos.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - main
   pull_request:
-    paths-ignore:
-      - '**.md'
 
 concurrency:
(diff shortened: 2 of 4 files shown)


hughcars:
I actually think we should go for the full solution, and apply the same approach to modifications of singularity and spack recipes. I.e. to not trigger the full suite when only the spack recipe is modified. 

So if only `docs` changes, don't run the full suite, only the style and docs ones, if only `spack` changes, only run spack, if only `singularity` changes, only run singularity etc. This would be pretty helpful for making small changes to the spack suite in future too, and the dummy jobs would get re-used.

PR #440, opened 2025-08-08
Add path filter to only run CI as necessary

hughcars:
Adds skips for CI runs if only docs, spack or singularity are updated. Should speed up doc CRs and tweaks to spack and singularity.


.github/workflows/build-and-test-linux.yml
Sbozzolo: You missed one here

Sbozzolo (approved):

Issue #423, opened 2025-08-01
Wave port eigensolver fails to converge in simple waveguides


simlapointe:
In some simple waveguide cases, the wave port eigensolver fails to converge and returns an error message such as:

> Verification failed: (num_conv >= mode_idx) is false:
 --> Wave port eigensolver did not converge!
 ... in function: void palace::WavePortData::Initialize(double)

This can be reproduced in the cylinder example, by changing the problem type to `Driven` and using `WavePort` boundaries instead of periodic boundaries, e.g. [waveguide_waveport.json](https://github.com/user-attachments/files/21552624/waveguide_waveport.json). The issue is likely related to changes made in https://github.com/awslabs/palace/pull/264 to avoid evanescent modes (switching back to `EigenvalueSolver::WhichType::LARGEST_MAGNITUDE` resolves the current issue, but might not be a robust solution).

Issue #437, opened 2025-08-08
WavePort excites different modes for different frequencies


vladimir-yu-fedorov:
I simulate a simple rectangular waveguide and have set the following parameters for the frequency range: `MinFreq: 0.02`, `MaxFreq: 0.025`, `FreqStep: 0.005`. For the input `WavePort`, I have specified the excitation of the lowest mode: `Mode: 1`.

The results for the 20 MHz frequency appear to be correct; however, at the 25 MHz frequency, I observe that a higher mode is being excited:

Palace v0.13:
| 20 MHz | 25 MHz |
| ----------- | ----------- |
| <img width="1458" height="855" alt="Image" src="https://github.com/user-attachments/assets/2821bd4a-d8e3-4858-99ab-132607f39fe9" /> | <img width="1458" height="855" alt="Image" src="https://github.com/user-attachments/assets/198fba75-7cca-4e74-b599-9a5e5feeef9f" /> |

In contrast, when using an earlier version of Palace, I obtain the expected result, with the same mode being excited for both frequencies:

Palace v0.12:
| 20 MHz | 25 MHz |
| ----------- | ----------- |
| <img width="1458" height="855" alt="Image" src="https://github.com/user-attachments/assets/ca0e0926-0eae-4931-9834-73fd6ad11828" /> | <img width="1458" height="855" alt="Image" src="https://github.com/user-attachments/assets/6dae2b65-950b-4efa-aff1-5e091f47f039" /> |

I have attached the mesh files and the Palace input file for your reference:
[mesh_and_input.zip](https://github.com/user-attachments/files/21676382/mesh_and_input.zip)

I am currently using Ubuntu 24.04.3 LTS, and the Palace was build using Apptainer/Singularity.

P.S. Additionally, I noticed a minor issue is that in the Paraview output, where the frequencies of 20 MHz and 25 MHz are displayed alongside an unexpected entry corresponding to a frequency of 99 GHz.

vladimir-yu-fedorov:
Thank you for your response.

I have some thoughts based on theory. If I understand correctly, the propagation constant $k_n$ is defined as:

$$ k_n = \sqrt{k_0^2 - k_c^2}, $$
 
where $k_0=w_0 / c_0$ is the wavenumber, with $w_0 = 2\pi f_0$ being an angular frequency, corresponding to the central frequency $f_0$, and $c_0$ is the speed of light in vacuum. The cut-off wavenumber $k_c$ is defined as

[...]

PR #448, opened 2025-08-15
Change wave port eigenproblem shift and ordering

simlapointe:
This PR changes the shift and eigenvalue ordering used in the wave port eigenvalue problem to resolve some ongoing issues where:
1.  No modes are found when the largest mode has $k_n < 0.5 \omega^2/c^2$.
2. The first mode is not always the largest $k_n$.
4. The modes are not ordered by descending wave number. 

A driven simulation of a cylindrical waveguide with wave ports at both ends is added to the examples/regression tests. This was previously failing because the wave port solver could not find any modes.

I tested that the change does not lead to evanescent modes in the CPW example or in the previously reported https://github.com/awslabs/palace/issues/251.

This will likely change the results of some wave port cases where the mode ordering will change.

Closes https://github.com/awslabs/palace/issues/423 and also potentially addresses https://github.com/awslabs/palace/issues/437.


palace/models/waveportoperator.cpp
```diff
@@ -598,13 +598,11 @@ WavePortData::WavePortData(const config::WavePortData &data,
...
-  const double c_max = mat_op.GetLightSpeedMax().Max();
-  MFEM_VERIFY(c_max > 0.0 && c_max < mfem::infinity(),
+  const double c_min = mat_op.GetLightSpeedMax().Min();
+  MFEM_VERIFY(c_min > 0.0 && c_min < mfem::infinity(),
               "Invalid material speed of light detected in WavePortOperator!");
-  mu_eps_min = 1.0 / (c_max * c_max) * 0.5;  // Add a safety factor for minimum propagation
-                                             // constant possible
-  // mu_eps_min = 0.0;  // Use standard inverse transformation to avoid conditioning issues
-  //                    // associated with shift
+  mu_eps_max = 1.0 / (c_min * c_min) * 0.999;  // Add a safety factor for maximum
```
sebastiangrimberg: Hi @simlapointe, sorry to chime in but I just saw this and was curious. Please correct me if I'm mistaken, but shouldn't the safety factor be > 1? If you have a TEM mode for a homogeneous waveguide with `k^2 = mu * epsilon`, as it is now wouldn't you get `1/(-0.001 * mu * epsilon)` as the eigenvalue which ends up being negative?

I am realizing now a very simple mistake in my reasoning is that I had always thought of `k_n` as increasing with mode index, when in reality it is the cutoff `k_n,c` that is increasing and the propagation constant `k_n^2 = k_n,max^2 - k_n,c^2` will decrease with increasing mode index at a fixed operating frequency, exactly as you've correctly identified. Thanks for fixing my error!

simlapointe: Hi @sebastiangrimberg, thanks for chiming in, I appreciate it! You’re right, the safety factor should indeed be greater than 1. I got scared by the singular matrix at exactly 1, but I just tried with 1.1 and it all works fine and will avoid the issue you mention.

CHANGELOG.md
```diff
@@ -118,6 +118,8 @@ The format of this changelog is based on
   - Added `scnlib` as a dependency to Palace.
   - Added parsing ability of existing csv files into the `Table` class. Fix `"Restart"`
     behaviour with the multi-excitation feature of driven solver.
+  - Change wave port eigenproblem shift and sorting so the modes are ranked in descending
+    order from the TEM limit.
```
hughcars: Might be worth expanding this a little (also move it into a new "In progress" section).

examples/cylinder/driven_wave.json
```diff
@@ -0,0 +1,70 @@
...
+      "Driven":
+      {
+          "MinFreq": 2.5,
+          "MaxFreq": 5.0,
+          "FreqStep": 0.5,
+          "SaveStep": 1
+      },
+    "Linear":
+    {
+        "Type": "Default",
```
hughcars: Nit: indent

test/examples/runtests.jl
```diff
@@ -146,6 +147,20 @@ end
 reltol = 2.0e-2
 abstol = 2.0e-12
```
hughcars: Maybe move this below, given the estimator on this problem, I suspect that the tolerances for this solve can be a lot tighter.

examples/cylinder/driven_wave.json
```diff
@@ -0,0 +1,70 @@
...
+  "Boundaries":
+  {
+      "WavePort":
+      [
+      {
+        "Index": 1,
+        "Attributes": [2],
+        "Mode": 1,
+        "Excitation": true,
+      },
```
hughcars: Nit: indent

hughcars (approved):
I fixed some nit things, but otherwise LGTM, excellent fix. Consider expanding the changelog entry a little bit.

PR #454, opened 2025-08-19
Remove assignment to `mean_phase`

Sbozzolo:
Fixes compiler warning:
```
  320 |     double mean_phase = linalg::NormalizePhase(space_op.GetComm(), E);
      |            ^~~~~~~~~~
1 warning generated.
```


simlapointe (commented):
While the output `mean_phase` is currently not used, `NormalizePhase(...)` modifies the `E` vector so we certainly don't want to remove this line. If we really want to get rid of the unused variable warning, we could remove the `double mean_phase =` part while keeping the `NormalizePhase(...)` part. I chose to keep the unused output so that if someone ever needs to print/use the mean phase prior to normalization, it is readily available (we debated printing it to the log file but chose not to for now). My (weak) opinion is to leave this as is for now.

simlapointe:
I'm ok removing the warning, either by removing `double mean_phase` or using this `maybe_unused`. 

Also agree that we should add a test, I'll do that.

simlapointe (approved):

PR #458, opened 2025-08-21
Add DeviceLayout transmon example

Sbozzolo:
This PR adds a new example: the single transmon case from [DeviceLayout.jl](https://aws-cqc.github.io/DeviceLayout.jl/stable/examples/singletransmon/). 

At this stage, the simulation takes ~ 8 minutes on 8 x86 cores, so it is run in parallel with the other regression tests and without AMR. If/when more computational resources become available for CI, we can add AMR to this example.

The example uses msh/json files that are saved to the repo to increase reproducibility (the test is already not super robust across architectures/variants) and to avoid instantiating a new chunky julia environment to produce the files on the fly.

I added a new documentation page to discuss this example. The page includes an interactive GLVis visualization and two plots. In an effort to make things more reproducible, I added the files that I used to generate those plots (we might end up using them if we re-generate the documentation nightly).

I am not entirely sure what are the most interesting topics to discuss in the documentation page, so I am open to suggestions.

https://awslabs.github.io/palace/previews/PR458/examples/transmon/

### DIFF
--- a/.github/actions/palace-ci/action.yml
+++ b/.github/actions/palace-ci/action.yml
@@ -279,7 +279,7 @@ runs:
 
     - name: Push to GHCR cache
       shell: bash
-      run: spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache
+      run: spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache || true
 
     - name: Run Regression Tests
       if: inputs.run-regression-tests == 'true'
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -164,4 +164,4 @@ jobs:
         if: |
           !cancelled() &&
           (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
-        run: spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache
+        run: spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache || true


examples/transmon/transmon.json
```diff
@@ -0,0 +1,126 @@
...
+      "Attributes": [
+        3
+      ]
+    }
+  },
+  "Model": {
+    "Refinement": {
+      "MaxIts": 0
+    },
+    "Mesh": "/Users/gbozzola/repos/palace/examples/transmon/mesh/transmon.msh2",
```
hughcars: This absolute path to the mesh means that the actual mesh is going to be different, and I suspect you're regenerating it?

docs/src/assets/js/glvis/index.js
```diff
@@ -15,17 +15,17 @@
   // node
   if (typeof exports === "object" && typeof module === "object") {
     console.log("mod type 1");
-    module.exports = factory(require("./glvis"));
+    module.exports = factory(require("./glvis.js"));
```
Sbozzolo, line 18: It seems that deploying GLVis is more robust if I add `.js`.

docs/src/quick.md
```diff
@@ -562,7 +562,7 @@ list of keys.
 
 ```@raw html
 <div id="glvis-container" style="width: 100%; height: 500px;">
-  <div id="glvis-div" style="width: 100%; height: 100%;" tabindex="0"></div>
+  <div id="glvis-div" style="width: 100%; height: 100%;"></div>
```
Sbozzolo, line 568: `tabindex="0"` is needed to make the frame focusable, but if I add it immediately, GLVis focuses upon page load (scrolling the page down). Instead, I add this attribute only if a user click on the GLVis. See below.

test/examples/runtests.jl
```diff
@@ -89,6 +90,44 @@ if "rings" in cases
...
+            "Re{V[2]} (V)",
+            "Im{V[2]} (V)",
+            "Re{V[3]} (V)",
+            "Im{V[3]} (V)",
+            "Re{I[1]} (A)",
+            "Im{I[1]} (A)",
+            "Re{I[2]} (A)",
+            "Im{I[2]} (A)",
+            "Re{I[3]} (A)",
+            "Im{I[3]} (A)"
```
Sbozzolo, line 162: I following the `cpw_lumped_eigen` example below. Some of these were prone to failing

docs/src/examples/transmon.md
```diff
@@ -0,0 +1,245 @@
...
+The `error-indicators.csv` file provides quantitative measures of numerical
+accuracy (see, the reference section on [error
+estimation](../reference.md#Error-estimation-and-adaptive-mesh-refinement-(AMR))).
+For a second-order solution:
+
+```@example include_example
+include_example_file("transmon", "error-indicators.csv") #hide
+```
+
+These error indicators provide metrics for estimating numerical accuracy across
```
hughcars, line 239: "Norm" here doesn't really mean that. The estimator is $`\|F - F_{numeric}\|`$ yes, but those are approximations to the flux terms, which are $`\epsilon \nabla V`$, thus the norm is more like a measure of the H1 or Hcurl seminorms of the solution (and the semi there belies the greatest shortcoming of the estimate ultimately).

test/examples/runtests.jl
hughcars: I wonder if this is a case for checking the magnitude again, I would expect that to be constant even if the phase is rotating arbitrarily.

.github/workflows/spack.yml
```diff
@@ -89,11 +90,17 @@ jobs:
...
+          if [[ "${{ matrix.transmon }}" == "true" ]]; then
+            TEST_VARIANT=""
+          else
+            TEST_VARIANT="+tests"
+          fi
+
           # Develop requires ^compiler, not %compiler
           if [[ "${{ matrix.spack_version }}" == 'v0.23.1' ]]; then
-            PALACE_SPEC="local.palace@develop+strumpack+arpack+mumps+tests %${{ matrix.compiler }} ${GPU_VARIANT} ^${{ matrix.mpi }} ^${{ matrix.blas }}"
+            PALACE_SPEC="local.palace@develop+strumpack+arpack+mumps ${TEST_VARIANT} %${{ matrix.compiler }} ${GPU_VARIANT} ^${{ matrix.mpi }} ^${{ matrix.blas }}"
```
hughcars: Nit: space before `${TEST_VARIANT}`

.github/workflows/spack.yml
```diff
@@ -182,7 +189,7 @@ jobs:
         run: spack -e . install --only-concrete --keep-stage --show-log-on-error --only package --no-cache
 
       - name: Run Unit Tests
-        if: needs.filter.outputs.test == 'true' && matrix.gpu == 'none' # Skip tests for GPU builds
+        if: needs.filter.outputs.test == 'true' && matrix.gpu == 'none'  && matrix.transmon != true # Skip tests for GPU builds
```
hughcars: and for transmon test builds.

docs/src/assets/examples/transmon-2.png
hughcars, line 1: Given there's a negative here, and it's energy density, I'm assuming this is a log plot?

docs/src/examples/transmon.md
```diff
@@ -0,0 +1,245 @@
...
+automated design and optimization of quantum devices. For a detailed tutorial on
+using DeviceLayout.jl to generate this geometry, see [this
+example](https://aws-cqc.github.io/DeviceLayout.jl/stable/examples/singletransmon/).
+
+### Overview and Configuration
+
+The system consists of three main components: a transmon qubit, a quarter-wave
+coplanar waveguide resonator, and a feedline for signal input and output. The
+transmon is capacitively coupled to the resonator through a "claw" structure
+that extends around the transmon islands, while the resonator couples to the
```
hughcars: I believe it's the transmon cutout and the transmon island within it, the claw is around the cutout piece.

docs/src/examples/transmon.md
```diff
@@ -0,0 +1,245 @@
...
+three lumped ports defining the electromagnetic boundary conditions. Two
+resistive ports at the feedline ends represent 50-Ohm terminations to the
+external world, while a third port models the Josephson junction as a linearized
+LC circuit with both inductance and capacitance. The substrate is sapphire,
+chosen for its low dielectric loss and well-characterized anisotropic material
+properties (see,
+[`config["Domains"]["Materials"]`](../config/domains.md#domains%5B%22Materials%22%5D)
+for more information on how to configure anisotropic materials).
+[`config["Boundaries"]["Absorbing"]`](../config/boundaries.md#boundaries%5B%22Absorbing%22%5D)
+boundary conditions at the simulation box edges capture electromagnetic
```
hughcars: Edges are lines, prefer boundaries
```suggestion
boundary conditions on the external boundaries of the simulation box capture electromagnetic
```

docs/src/examples/transmon.md
```diff
@@ -0,0 +1,245 @@
...
+    n_meander_turns=5,
+    hanger_length=500μm,
+    bend_radius=50μm,
+    save_mesh::Bool=false,
+    save_gds::Bool=false
+)
+```
+
+Beyond the physical dimensions, the `save_mesh` and `save_gds` options enable
+output of the mesh file for simulation and a [GDS
```
hughcars: Nit: mixing definite and indefinite article
```suggestion
output of a mesh file for simulation and a [GDS
```

examples/transmon/transmon.json
```diff
@@ -0,0 +1,129 @@
...
+        ]
+      }
+    ]
+  },
+  "Solver": {
+    "Eigenmode": {
+      "N": 2,
+      "Save": 2,
+      "Tol": 1.0e-8,
+      "Target": 3
```
hughcars: Part of the reason p=1 can be so much faster, is that the target here is closer to the p=1 modes. If a target of 4 were used, the order 2 solve might be sped up.

examples/transmon/transmon.jl
```diff
@@ -0,0 +1,289 @@
...
+    # Update paths for local directory structure.
+    config["Model"]["Mesh"] = joinpath("mesh", mesh_filename)
+    config["Problem"]["Output"] = "postpro"
+
+    # Set tight tolerances for reproducible results (important for CI/testing).
+    config["Solver"]["Eigenmode"]["Tol"] = 1e-8
+    config["Solver"]["Linear"]["Tol"] = 1e-12
+
+    # Get closer to expected values to speed up convergence.
+    config["Solver"]["Eigenmode"]["Target"] = 3
```
hughcars: This target is a lot better for p=1 than for p=2, which will explain some of the 20x difference in runtime.

docs/src/assets/examples/transmon-2.png
Sbozzolo: Yes, added in the label

examples/transmon/transmon.json
Sbozzolo: I changed it to 4.

test/examples/runtests.jl
Sbozzolo: I tried and I still had failing cases

PR #449, opened 2025-08-15
Add far-field extraction

Sbozzolo:
This PR introduces a new feature in Palace, extracting far-fields from near field solutions, and a new example, a dipole antenna.

## Far-field extraction

This feature is based upon Stratton-Chu's transformations in the limit for kr >> 1 (with k wave number and r observation distance). One can show that, in this limit [1],

$$
r \mathbf{E}_p = \frac{jk}{4\pi} \mathbf{r}_0 \times \int [\mathbf{n} \times \mathbf{E} - \eta \mathbf{r}_0 \times (\mathbf{n} \times \mathbf{H})] \exp(jk\mathbf{r} \cdot \mathbf{r}_0) dS 
$$

where:
- $E_p$ is the electric field at the observation point
- $k$ is the wave number
- $r₀$ is the unit vector from source to observation point, parameterized by $(\theta, \phi)$
- $n$ is the surface normal (to $S$)
- $E, H$ are the tangential fields on the surface
- $η$ is the impedance

The integral is over the surface $S$.

Note, we obtain $`r \mathbf{E}_p`$ because the electric field decays with $exp(jkr)/r$, so multiplying it by $r$ ensures that the quantity is finite. Note also that the solution is defined up to a global phase factor.

This feature is meant to be used as follows:
- users define a surface (typically a sphere or a box) that encloses the system
- users set `Absorbing` boundary conditions on this surface
- users enable the `FarField` `PostProcessing` option on this surface and define the sampling resolution for the far fields

Palace will perform the integrals (one per observation option) and produce CSV file with the results.

The section in the config file will look like this
```json
"Boundaries":
    {
        "Absorbing":
        {
            "Attributes": [2]
        },
        "Postprocessing":
        {
            "FarField":
            {
                "Attributes": [2],
                "NSample": 100
            }
        }
    }
```

The output CSV file looks like
```
        f (GHz),               theta (deg.),                 phi (deg.),              r*Re{E_x} (V),              r*Im{E_x} (V),              r*Re{E_y} (V),              r*Im{E_y} (V),              r*Re{E_z} (V),              r*Im{E_z} (V)
 7.49000000e-02,         0.000000000000e+00,         0.000000000000e+00,        +1.030863711196e-03,        +4.419657726714e-04,        +2.360040259865e-03,        -1.471646095969e-04,        -0.000000000000e+00,        +0.000000000000e+00
[...]

PR #459, opened 2025-08-21
Regenerate meshes as binary fines, re-baseline tests

Sbozzolo:
This PR regenerates all the meshes as binary files and changes the defaults of the julia scripts to make sure that they produce the actual meshes used. 

OLD PR DESCRIPTION:

My #449 PR adds 360,888 lines. Of course, this does not really reflect the change in this PR. The reason for so many lines is that it includes msh files.

msh files are text files so git handles them as source code. However, we never really want to look at diff of msh files.

I have locally this change to my gitattributes so that I don't accidentally print a 100k loc git diff, but I think that this might be useful more broadly.

This change tells Git to treat all files with the .msh extension as binary files. This means:

1. No diff display: Git won't show textual diffs for these files
2. No merge attempts: Git won't try to merge conflicts in these files
3. Obscure, but did you know that in some cases git modifies line endings (depending on the operating system and git settings)? Setting msh files are binary means that git won't attempt line ending conversions

In addition to this, this PR regenerates all the meshes to save them as binary files.


cameronrutherford:
> However, we never really want to look at diff of msh files.

So long as this assumption holds true, I think these changes are fine.

Is it still easy to inspect the diff if I wanted to?

Sbozzolo:
> So long as this assumption holds true, I think these changes are fine.

If the number of total element in a mesh changes, every line of the file will be different, and pretty much any change in the mesh generation leads to a different number of elements.

> Is it still easy to inspect the diff if I wanted to?

Yes, you can `git diff --text` if you need to.

Sbozzolo:
Note that an alternative is to directly output binary mesh files. That would reduce storage requirement for meshes (and this won't be needed)

hughcars:
I think this is a good idea, but should probably be accompanied by converting the meshes to binary format anyway. This would however require rerunning all the mesh generation files, which will change the meshes a little bit. So in the spheres example, this means changing `mesh.jl` with
```

**Comment by Sbozzolo** [comment]
I went through the exercise of updating the examples and found quite a few things.

1. There were no instructions on how to generate the meshes. I assumed the defaults values were used to everything, but for the `cpw` examples we have two variants of each mesh. I had to open the mesh with Gmsh and visually inspect it to reverse engineer the parameter used.
2. Updating the documentation is a chore and doing it properly (ie, regenerating the scaling plots, recomputing certain numbers that are mentioned in the text) requires hours of busywork
3. A couple of filenames were wrong (`surface-M` -> `terminal-M`)
4. The `baseline` script was out of sync with the `runtest.jl` so it did not work out of the box

I tried improving what I could along the way:
- I fixed the sync issue with `baseline` (but we need a more automatic solution)
- I updated the documentation to directly include the files from the `ref` folder instead of manually copy&pasting it.
- I removed references to specific numbers (e.g., 0.7%, 1.2%) and used a more generic "percent-level" 

I could not understand how to reproduce this result:
> However, the bulk dielectric loss postprocessing results, computed from the energies written
to `postpro/cavity_impedance/domain-E.csv`, still give ``Q_d = 1/0.004 = 2.50\times 10^3`` for
every mode as expected.

When I look at `domain-E.csv` I see this
```csv
        m,                 E_elec (J),                  E_mag (J),                  E_cap (J),                  E_ind (J),              E_elec[1] (J),                  p_elec[1],               E_mag[1] (J),                   p_mag[1]
 1.00e+00,        +9.139656208419e-02,        +9.139656939516e-02,        +0.000000000000e+00,        +0.000000000000e+00,        +9.139656208419e-02,        +1.000000000000e+00,        +9.139656939516e-02,        +1.000000000000e+00
 2.00e+00,        +9.139656208419e-02,        +9.139656940175e-02,        +0.000000000000e+00,        +0.000000000000e+00,        +9.139656208419e-02,        +1.000000000000e+00,        +9.139656940175e-02,        +1.000000000000e+00
 ```
 So, I don't know where the 0.004 comes from.

**Comment by Sbozzolo** [comment]
I take it back, not all the meshes were created with the default function arguments. I tried reproducing them by tweaking `order` and `refinement`, but I could not get quite the same meshes, so I went with ones that have similar number of degrees of freedom.

**Comment by hughcars** [comment]
> I went through the exercise of updating the examples and found a few things.
> 
> 1. There were no instructions on how to generate the meshes. I assumed the defaults values were used to everything, but for the `cpw` examples we have two variants of each mesh. I had to open the mesh with Gmsh and visually inspect it to reverse engineer the parameter used. (And I later on learned that some meshes had not been generated with default values, see post below)
> 2. Updating the documentation is a chore and doing it properly (ie, regenerating the scaling plots, recomputing certain numbers that are mentioned in the text) requires hours of busywork
> 3. A couple of filenames were wrong (`surface-M` -> `terminal-M`)
> 4. The `baseline` script was out of sync with the `runtest.jl` so it did not work out of the box
> 
> I tried improving what I could along the way:
> 
> * I added instructions on how to reproduce the example mesh in each file
> * I fixed the sync issue with `baseline` (but we need a more automatic solution)
> * I updated the documentation to directly include the files from the `ref` folder instead of manually copy&pasting it.
> * I removed references to specific numbers (e.g., 0.7%, 1.2%) and used a more generic "percent-level"
> 
> I could not understand how to reproduce this result:
> 
> > However, the bulk dielectric loss postprocessing results, computed from the energies written
> > to `postpro/cavity_impedance/domain-E.csv`, still give `Q_d = 1/0.004 = 2.50\times 10^3` for
> > every mode as expected.
> 
> When I look at `domain-E.csv` I see this
> 
> ```
>         m,                 E_elec (J),                  E_mag (J),                  E_cap (J),                  E_ind (J),              E_elec[1] (J),                  p_elec[1],               E_mag[1] (J),                   p_mag[1]
>  1.00e+00,        +9.139656208419e-02,        +9.139656939516e-02,        +0.000000000000e+00,        +0.000000000000e+00,        +9.139656208419e-02,        +1.000000000000e+00,        +9.139656939516e-02,        +1.000000000000e+00
>  2.00e+00,        +9.139656208419e-02,        +9.139656940175e-02,        +0.000000000000e+00,        +0.000000000000e+00,        +9.139656208419e-02,        +1.000000000000e+00,        +9.139656940175e-02,        +1.000000000000e+00
> ```
> 
> So, I don't know where the 0.004 comes from.

1. This seems like tech debt, the meshes were not historically being regenerated because of HXT, but moving away from that we ought to have that the args needed to generate them are at least documented in the corresponding `mesh.jl`
2. Agreed, this shouldn't be so complicated. The convergence study for example ought to be automated by loading the config and UniformLevels to generate the runs, which can be dispatched. This would still take time to run, but the data generation and then wrangling afterwards should be automated (already or we should fix)
3. Good example of failures relating to these not be automatically updated too. 
4. Similar to above. 

> I could not understand how to reproduce this result:
> 
> > However, the bulk dielectric loss postprocessing results, computed from the energies written
> > to `postpro/cavity_impedance/domain-E.csv`, still give `Q_d = 1/0.004 = 2.50\times 10^3` for
> > every mode as expected.

This is a bit opaque, but if you read through Pozar the Qd (lossless exterior, lossy dielectric) is a function of omega, electric energy and magnetic energy (eq 6.48 in the textbook). So if those 3 don't change, and looking also at 6.47, P_d is a rescaling of the electric energy in the dielectric, thus doesn't change. So the Qd will be identical if energy and frequencies don't change, hence `2.50\times 10^3`. So the extra loss, which is reducing Q via the formula `Q^-1 = (Qc^-1 + Qd^-1)` makes sense. Then the text that comes afterwards with the specific calculated values from the example that gets reference is putting in Qc and Qd manually calculated and checking the answer.

File: docs/src/examples/spheres.md [hunk 1]
```diff
@@ -93,26 +95,6 @@ configuration file for this example contains this information under
...
-<br/><p align="center">
-  <img src="../../assets/examples/spheres-3.png" width="45%" />
-  <img src="../../assets/examples/spheres-4.png" width="45%" />
-</p>
-```
-
-## References
-
-[1] J. Lekner, Capacitance coefficients of two spheres, _Journal of Electrostatics_ 69
-(2011) 11-14.
```
hughcars, line 118: These fields ought to not change very much, and don't include a visualization of the mesh, so should be kept. Additionally the reference should be kept.

examples/spheres/mesh/mesh.jl
```diff
@@ -27,6 +27,8 @@ From this directory, run:
 ```bash
 julia -e 'include("mesh.jl"); generate_spheres_mesh(; filename="spheres.msh")'
 ```
+This generates mesh used in the example.
```
hughcars: Nit: "the mesh"

test/examples/ref/cpw/lumped_uniform/probe-E.csv
```diff
@@ -1,8 +1,8 @@
...
+ 3.20000000e+01,        +1.676590417379e+04,        -2.936105208754e+03,        -1.105271286457e+04,        +3.224783330005e+03,        +4.288024860420e+04,        -6.651309576978e+03,        -1.149144121237e+03,        +5.790116114172e+02,        +3.060019572519e+03,        +8.343827668965e+01,        -9.384795545300e+04,        +1.540446486396e+04,        +1.757413854131e+04,        -1.988027783710e+03,        -7.975800807180e+03,        +4.266168365511e+03,        +4.160461123005e+04,        -9.424367096400e+03,        -1.364860560785e+03,        -3.043219223216e+02,        +6.519557141141e+03,        +2.740374214503e+03,        -9.320010959657e+04,        +1.753028110349e+04
```
hughcars, line 8: These values look like they're changing pretty significantly?

test/examples/ref/rings/probe-B.csv
```diff
@@ -1,3 +1,3 @@
-               i,          B_x[1] (Wb/m²),          B_y[1] (Wb/m²),          B_z[1] (Wb/m²)
- 1.000000000e+00,        +2.349559482e-05,        +6.218888487e-05,        +3.270184717e-03
- 2.000000000e+00,        +7.248344643e-08,        +1.884734325e-07,        +3.258199125e-04
+        i,             B_x[1] (Wb/m²),             B_y[1] (Wb/m²),             B_z[1] (Wb/m²)
+ 1.00e+00,        -4.670591481526e-05,        -1.186659609674e-07,        +3.276020100865e-03
+ 2.00e+00,        -1.353599037008e-07,        -5.220635297932e-09,        +3.259800587101e-04
```
hughcars: Fairly large change here, is this a linear solve tolerancing issue?

test/examples/ref/cylinder/waveguide/eig.csv
```diff
@@ -1,16 +1,20 @@
...
+ 1.30e+01,        +5.982999856731e+00,        +1.196600007969e-03,        +2.499999973486e+03,        +7.904471866565e-12,        +2.479341141372e-09
+ 1.40e+01,        +6.201001663971e+00,        +1.240200116879e-03,        +2.500000485243e+03,        +1.884414922021e-11,        +5.950376528092e-09
+ 1.50e+01,        +6.334084161795e+00,        +1.266816772481e-03,        +2.500000168165e+03,        +1.321478741148e-11,        +4.190273700451e-09
+ 1.60e+01,        +6.334381201579e+00,        +1.266876146552e-03,        +2.500000235031e+03,        +1.546414432989e-11,        +4.903567846424e-09
+ 1.70e+01,        +6.420730119035e+00,        +1.284145977807e-03,        +2.500000139554e+03,        +5.762333621916e-12,        +1.832223243212e-09
+ 1.80e+01,        +6.437632559281e+00,        +1.287526462921e-03,        +2.500000145018e+03,        +2.363043807984e-11,        +7.517733521279e-09
+ 1.90e+01,        +6.665130922016e+00,        +1.333026264038e-03,        +2.499999900651e+03,        +1.056782676355e-11,        +3.386986157681e-09
```
hughcars: This is part of the issue with that CI failure you were seeing, the reference is only supposed to check for 15 modes (the number asked for), the solver might generate more, but cannot generate less. So we only copy over the rows for the first 15. This tweak is done manually, as we don't really regenerate these files often.

test/examples/ref/cylinder/floquet/eig.csv
```diff
@@ -1,16 +1,18 @@
...
+ 9.00e+00,        +5.115701278077e+00,        +1.023140245241e-03,        +2.500000075350e+03,        +5.363905237755e-11,        +1.641593583746e-08
+ 1.00e+01,        +5.317690333664e+00,        +1.063539035910e-03,        +2.499997771811e+03,        +3.265220844914e-11,        +1.004755580956e-08
+ 1.10e+01,        +5.586886788605e+00,        +1.117376824815e-03,        +2.500001242316e+03,        +4.231610477299e-11,        +1.311970789534e-08
+ 1.20e+01,        +5.783323643106e+00,        +1.156662039041e-03,        +2.500005863236e+03,        +2.610518409518e-10,        +8.139868242903e-08
+ 1.30e+01,        +5.961600495798e+00,        +1.192322373909e-03,        +2.499995280423e+03,        +4.111042606063e-11,        +1.288686898543e-08
+ 1.40e+01,        +6.236202074607e+00,        +1.247240516445e-03,        +2.499999846502e+03,        +3.651023364588e-11,        +1.154175443328e-08
+ 1.50e+01,        +6.422834972967e+00,        +1.284568486076e-03,        +2.499997147307e+03,        +4.947391386541e-11,        +1.573250497155e-08
```
hughcars: These values are very different, something is wrong. The first mode is fine, but the 15th is not. I believe it's possible the cylinder mesh was generated with an extra refinement level in gmsh, to ensure there was sufficient number of elements in the periodic direction.

test/examples/ref/cylinder/cavity_pec/eig.csv
```diff
@@ -1,16 +1,17 @@
...
+ 1.00e+01,        +5.417728924691e+00,        +1.083545726667e-03,        +2.500000184445e+03,        +2.476254120596e-11,        +2.619216637312e-09
+ 1.10e+01,        +5.984116193613e+00,        +1.196823190657e-03,        +2.500000150403e+03,        +2.618478395125e-12,        +2.798781312098e-10
+ 1.20e+01,        +6.109109254530e+00,        +1.221822329789e-03,        +2.499999070147e+03,        +1.409996917967e-10,        +1.510757801515e-08
+ 1.30e+01,        +6.203306538565e+00,        +1.240661258166e-03,        +2.500000149840e+03,        +1.949829334437e-13,        +2.093063169808e-11
+ 1.40e+01,        +6.335867394227e+00,        +1.267173451858e-03,        +2.500000103243e+03,        +1.694350289177e-11,        +1.823668260884e-09
+ 1.50e+01,        +6.388501883061e+00,        +1.277700342408e-03,        +2.500000116926e+03,        +6.450773962702e-12,        +6.950559035399e-10
+ 1.60e+01,        +6.486767181405e+00,        +1.297353408238e-03,        +2.500000104039e+03,        +9.536572117342e-11,        +1.029621942352e-08
```
hughcars: These modes have changed quite significantly. I believe this is related to my point about the refinement level necessary for the periodic checks.

test/examples/ref/cylinder/cavity_impedance/eig.csv
```diff
@@ -1,16 +1,17 @@
...
+ 9.00e+00,        +5.290437180008e+00,        +1.206987831018e-03,        +2.191586850229e+03,        +9.020804275681e-12,        +8.409928216485e-07
+ 1.00e+01,        +5.415691234390e+00,        +1.256685370261e-03,        +2.154752298529e+03,        +2.621802995318e-12,        +2.502069327027e-07
+ 1.10e+01,        +5.982808714750e+00,        +1.366666587850e-03,        +2.188832639955e+03,        +3.594658509845e-11,        +3.789379176515e-06
+ 1.20e+01,        +6.108814329819e+00,        +1.342414136730e-03,        +2.275309202345e+03,        +1.470154220739e-11,        +1.582403711918e-06
+ 1.30e+01,        +6.201053832008e+00,        +1.367797575901e-03,        +2.266802519655e+03,        +6.033933374362e-12,        +6.592618807723e-07
+ 1.40e+01,        +6.334125080382e+00,        +1.440371760038e-03,        +2.198781390988e+03,        +2.162045348130e-11,        +2.412880026463e-06
+ 1.50e+01,        +6.388019130983e+00,        +1.447724026348e-03,        +2.206228251646e+03,        +5.223606635379e-12,        +5.879196321476e-07
```
hughcars: See others too, these modes are significantly changed.

test/examples/ref/cpw/wave_uniform/port-S.csv
```diff
@@ -1,8 +1,8 @@
...
+ 2.60000000e+01,        -1.666857053872e+01,        +1.450202306938e+02,        -1.554834376819e-01,        +5.476377339055e+01,        -4.380524309615e+01,        -2.924935652818e+01,        -4.666320843387e+01,        -1.046381960012e+02,        -1.554833404623e-01,        +5.476374795487e+01,        -1.665417611768e+01,        +1.436610295289e+02,        -4.712660945709e+01,        -1.069337196645e+02,        -4.357252482744e+01,        -2.604919634636e+01
+ 3.20000000e+01,        -2.784324110881e+01,        -1.257074675959e+02,        -5.986160174635e-02,        -1.637318810533e+01,        -5.104789815483e+01,        -3.286456129192e+01,        -4.581734026175e+01,        -1.783858501957e+02,        -5.986175775304e-02,        -1.637322373880e+01,        -2.735830629742e+01,        -7.911261046332e+01,        -4.618509159414e+01,        -1.775181669117e+02,        -5.062003384433e+01,        -3.582574808621e+01
```
hughcars, line 8: The last value here for the phases for instance, looks like a pretty big change. It does appear to be just a change of sign though.

test/examples/ref/cpw/wave_adaptive/surface-F.csv
```diff
@@ -1,302 +1,302 @@
...
+ 3.14000000e+01,        -1.032949863314e-14,        +9.919007965529e-14,        +1.289800059198e-03,        +9.976024818211e-01,        +9.788759562090e-01
+ 3.15000000e+01,        -5.700923150937e-15,        +9.854423108429e-14,        +1.289720564238e-03,        +9.975228545016e-01,        +9.787851987951e-01
+ 3.16000000e+01,        -1.113003297396e-15,        +9.779319019535e-14,        +1.289386979640e-03,        +9.974156076920e-01,        +9.786684209569e-01
+ 3.17000000e+01,        +3.431695046943e-15,        +9.693841562337e-14,        +1.288804851828e-03,        +9.972809335007e-01,        +9.785258489234e-01
+ 3.18000000e+01,        +7.930659411908e-15,        +9.598143996925e-14,        +1.287979709669e-03,        +9.971190860313e-01,        +9.783577585966e-01
+ 3.19000000e+01,        +1.238143320651e-14,        +9.492386765002e-14,        +1.286917049312e-03,        +9.969303809209e-01,        +9.781644748330e-01
+ 3.20000000e+01,        +1.678161810864e-14,        +9.376737265214e-14,        +1.285622322665e-03,        +9.967151946577e-01,        +9.779463706155e-01
```
hughcars, line 302: The first column final value is changed sign, pretty close to zero, but that is a surprise.

docs/src/examples/rings.md
```diff
@@ -77,22 +81,19 @@ M_{bb} &= 707.2050\text{ pH}
...
+in the self inductances versus the analytic solutions.
 
 The typical approach used by *Palace* for lumped parameter extraction uses the computed
 field energies, but one can also compute the inductance by explicitly integrating the
 magnetic flux through a surface and dividing by the excitation current. This is configured
 under
 [`config["Boundaries"]["Postprocessing"]["Inductance"]`](../config/boundaries.md#boundaries%5B%22Postprocessing%22%5D%5B%22Inductance%22%5D)
 in the configuration file. The resulting postprocessed values are written to
-`postpro/surface-M.csv`:
+`postpro/terminal-M.csv`:
```
hughcars: The text here is somewhat confusing. What's supposed to be output is `surface-F`, the actual magnetic flux, which can effectively be used to calculate M by dividing by values from `terminal-I.csv`, but isn't symmetric by construction, unlike the `terminal-M.csv` etc which come from energy and are naturally symmetric.

examples/cylinder/mesh/mesh.jl
```diff
@@ -1,6 +1,11 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
+# Generate example meshes with:
+# julia -e 'include("mesh/mesh.jl"); generate_cylindrical_mesh(filename="cylinder_tet.msh", refinement=1, order=2, mesh_type=0)'
+# julia -e 'include("mesh/mesh.jl"); generate_cylindrical_mesh(filename="cylinder_prism.msh", refinement=1, order=2, mesh_type=1)'
+# julia -e 'include("mesh/mesh.jl"); generate_cylindrical_mesh(filename="cylinder_hex.msh", refinement=1, order=2, mesh_type=2)'
+
 using Gmsh: gmsh
 
 """
```
hughcars, line 11: I think the default refinement level should probably be higher, as from memory that's the source of the issues with the periodic datasets.

hughcars (changes requested):
Read through, the cylinder mesh there's a big issue with the default params, check the frequencies of the output meshes. There's a few other spots I noticed quickly in the changed files too, maybe they can be tightened with tolerances, but I'm very wary about adjusting the cpw tolerances to tighter.

docs/src/examples/spheres.md
Sbozzolo: Yes, this is some rebasing gone wrong

docs/src/examples/rings.md
Sbozzolo: Good to know! I fixed this to show the correct result obtained by the procedure described.

docs/src/examples/rings.md
```diff
@@ -77,22 +81,46 @@ M_{bb} &= 707.2050\text{ pH}
...
+in the configuration file. The postprocessed magnetic flux values are written to `postpro/surface-F.csv`:
 
+```@example include_example
+include_example_file("rings", "surface-F.csv") # hide
 ```
-               i,                M[1] (H),                M[2] (H)
- 1.000000000e+00,        +4.260888932e-11,        +1.890068508e-12
- 2.000000000e+00,        +1.955578103e-12,        +7.130510940e-10
+
+Diving this with the values from `postpro/terminal-I.csv`,
```
hughcars: Combining with the values in `postpro/terminal-I.csv` we can compute the inductance matrix in this alternative fashion,

....

This method of calculating the inductance matrix directly from flux values is in general less accurate than using the energy method, due to convergence properties of finite element functional outputs, but serves as a validation of the energy calculation.

examples/coaxial/mesh/mesh.jl
```diff
@@ -146,12 +149,15 @@ function generate_coaxial_mesh(;
     gmsh.option.setNumber("Mesh.MeshSizeFromCurvature", 0)
     gmsh.option.setNumber("Mesh.MeshSizeExtendFromBoundary", 0)
 
+    gmsh.option.setNumber("Mesh.Algorithm", 10)
```
hughcars: There isn't a 2D algorithm number 10 from the gmsh docs:
```
Mesh.Algorithm
2D mesh algorithm (1: MeshAdapt, 2: Automatic, 3: Initial mesh only, 5: Delaunay, 6: Frontal-Delaunay, 7: BAMG, 8: Frontal-Delaunay for Quads, 9: Packing of Parallelograms, 11: Quasi-structured Quad)
```
Did you mean the default 6?

test/examples/testcase.jl
```diff
@@ -97,7 +97,11 @@ function testcase(
             if !skip_rowcount
                 @test nrow(data) == nrow(dataref) || logdump(data, dataref)
             end
+            # For Eigenmode simulations, some solvers find more modes than requested. This
+            # is very sensitive to details, so only compare the first nrows that both data
+            # and dataref have.
             data = data[1:min(nrow(data), nrow(dataref)), :]
+            dataref = dataref[1:min(nrow(data), nrow(dataref)), :]
```
hughcars: The problem with this is that it will pass even if the test case generated an empty file. As then the nrow will be 0.

An alternative would be that the "regenerate" script could remove the unobserved data, or include a check that the number of rows is greater than or equal the requested number in the source. I.e. eigensolves assert that the returned number of eigenvalues is greater than or equal to the request. So just before the `return` in `EigenSolver::Solve` in `eigensolver.cpp` put

```diff
diff --git a/palace/drivers/eigensolver.cpp b/palace/drivers/eigensolver.cpp
index 661114a3..ea7c41af 100644
--- a/palace/drivers/eigensolver.cpp
+++ b/palace/drivers/eigensolver.cpp

+  MFEM_VERIFY(num_conv >= iodata.solver.eigenmode.n, "Eigenmode solve only found "
+                                                         << num_conv << " modes when "
+                                                         << iodata.solver.eigenmode.n
+                                                         << " were requested!");
   return {indicator, space_op.GlobalTrueVSize()};
 }
 
```
That way we can be assured it's always checking at a minimum the right number of entries.

test/examples/ref/rings/probe-B.csv
```diff
@@ -1,3 +1,3 @@
-               i,          B_x[1] (Wb/m²),          B_y[1] (Wb/m²),          B_z[1] (Wb/m²)
- 1.000000000e+00,        +2.349559482e-05,        +6.218888487e-05,        +3.270184717e-03
- 2.000000000e+00,        +7.248344643e-08,        +1.884734325e-07,        +3.258199125e-04
+        i,             B_x[1] (Wb/m²),             B_y[1] (Wb/m²),             B_z[1] (Wb/m²)
+ 1.00e+00,        -4.670591481522e-05,        -1.186659609706e-07,        +3.276020100864e-03
+ 2.00e+00,        -1.353599037098e-07,        -5.220635297645e-09,        +3.259800587101e-04
```
hughcars, line 3: Due to symmetry the x,y values here are mostly noise, and should be close to 0, so the z-value remaining pretty constant shows it's ok.

examples/coaxial/mesh/mesh.jl
Sbozzolo: I think I confused `Mesh.Algorithm` with `Mesh.Algorithm3D`

test/examples/testcase.jl
Sbozzolo: The reason I had to add that line was to avoid manually removing the additional lines in the reference solution when additional eigenmodes are found. What I was getting is that my local `eig.csv` had one addition eigenmode, so the reference solution had one additional "spurious line", so that the test would fail for those cases that find the correct number of eigenmodes. The `MFEM_VERIFY` would not solve this problem.

I see that my current solution is not correct. I'll try to see if there's any easy way to automate trimming the csv files to the correct number of lines in `testcase.jl`

Sbozzolo: I reverted this change and added a short note in the README.md of the test folder about this.

PR #462, opened 2025-08-22
Update clang to 18 and gcc to 15 inlight of update macos runners

hughcars:
The default macos image updated, breaking the runners. This updates to using llvm18 and gcc14 which are on the osx15 image, from the old 15 and 12.


Sbozzolo:
I ended up on this [page](https://discourse.paraview.org/t/issues-with-superbuild-with-5-9/6363) and tried updating `cmake` in the workflow. It fixed the issue.

Issue #422, opened 2025-08-01
Nonlinear eigensolver


simlapointe:
Eigenmode simulations cannot currently be performed when using `WavePort`, `Conductivity`, or second order `Absorbing` boundary conditions. These boundary conditions are frequency-dependent, making the eigenvalue problem nonlinear. Adding a nonlinear eigensolver to Palace would alleviate this restriction.

PR #467, opened 2025-08-27
Nonlinear eigensolver

simlapointe:
Adding nonlinear eigensolver capability to Palace to enable eigenmode simulations with frequency-dependent boundary conditions such as `WavePort`, `Conductivity`, or second-order `Absorbing`.

Specifically, two nonlinear eigensolver options are introduced:
- Hybrid (default): a hybrid approach where the nonlinear term is (coarsely) approximated with a second-order polynomial, the quadratic EVP is solved with the existing SLEPc or ARPACK polynomial eigensolvers, and the resulting eigenmodes are used as initial guesses for a (deflated) quasi-Newton eigenvalue solver.
- SLP: SLEPc's Successive Linear Problem (SLP) solver from the NEP interface. 

Three eigenmode regression tests are added:
- CPW with lumped ports and second-order absorbing BCs.
- CPW with wave ports.
- Adapter waveguide (rectangle at one end, ellipse at the other)

Closes https://github.com/awslabs/palace/issues/422.


docs/src/config/solver.md
```diff
@@ -141,13 +139,26 @@ number of eigenmodes of the problem. The available options are:
   - `"Default"` :  Use the default eigensolver. Currently, this is the Krylov-Schur
     eigenvalue solver from `"SLEPc"`.
 
+`"NonlinearType" ["Default"]` : Specifies the nonlinear eigenvalue solver to be used for nonlinear problems (e.g. frequency-dependent boundary conditions). The available options are:
+
+  - `"Hybrid"` : Hybrid algorithm where a (quadratic) polynomial approximation of the nonlinear problem is first solved and the eigenmodes are then refined with a quasi-Newton nonlinear eigensolver.
+  - `"SLP"` : SLEPc's Successive Linear Problem nonlinear eigensolver.
+  - `"Default"` : Use the default nonlinear eigensolver. Currently, this is the hybrid algorithm.
```
hughcars: If the default value doesn't change as a function of build conditions, i.e. that for ARPACK the hybrid is also the default, you can drop the indirection and have `"Hybrid"` directly as the default, and similarly in the enum. Can then only allow SLP if slepc is built.

Is the nonlinear setting always defaulting on if necessary? I'm basically wondering if it would be better to have a user error out if they don't set the nonlinear solver type, or to have it pick this default and then potentially be computationally more expensive than they thought.

docs/src/config/solver.md
```diff
@@ -141,13 +139,26 @@ number of eigenmodes of the problem. The available options are:
...
+
+`"TargetUpper" [3 * Target]` : Upper end of the frequency target range in which to search for eigenvalues, GHz. Only used in nonlinear problems.
```
hughcars: Maybe expand a little that the looser this upper bound the slower the overall convergence.

examples/cpw/cpw_lumped_eigen.json
```diff
@@ -0,0 +1,143 @@
...
+    {
+      "Attributes": [4],
+      "Order": 2
+    },
+    "LumpedPort":
+    [
+      {
+        "Index": 1,
+        "R": 56.02,  // Ω, 2-element uniform
+        //"Excitation": true,
```
hughcars: Debris?

examples/cpw/cpw_wave_eigen.json
```diff
@@ -0,0 +1,177 @@
...
+    "Verbose": 2,
+    "Output": "postpro/wave_eigen"
+  },
+  "Model":
+  {
+    "Mesh": "mesh/cpw_wave_0.msh",
+    "L0": 1.0e-6,  // μm
+      "Refinement": {
+          "UniformLevels": 0
+      },
```
hughcars: Nit: Indent, and `{}` are new lines:
```
{ 
  UniformLevesl": 0
}
```

examples/cpw/cpw_lumped_eigen.json
```diff
@@ -0,0 +1,143 @@
...
+    "Verbose": 2,
+    "Output": "postpro/lumped_eigen"
+  },
+  "Model":
+  {
+    "Mesh": "mesh/cpw_lumped_0.msh",
+    "L0": 1.0e-6,  // μm
+      "Refinement": {
+          "UniformLevels": 0
+      },
```
hughcars: Nit: Indent. If you highlight all the file contents in vscode, then `ctrl+shift+p` and type reindent, it should fix them all up for you.

examples/cpw/cpw_wave_eigen.json
```diff
@@ -0,0 +1,177 @@
...
+    [
+      {
+        "Index": 1,
+        "Attributes": [4],
+        "Mode": 1,
+        "Offset": 0.0,
+        "MaxIts": 100,
+        "KSPTol": 1e-8,
+        "EigenTol": 1e-6,
+        "Verbose": 0
```
hughcars: Are all these options necessary? If they are, it might be worth actually changing the defaults to these if you think these more aggressive values are worth having.

palace/drivers/drivensolver.cpp
```diff
@@ -149,8 +149,9 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &space_op) const
       auto A = space_op.GetSystemMatrix(std::complex<double>(1.0, 0.0), 1i * omega,
                                         std::complex<double>(-omega * omega, 0.0), K.get(),
                                         C.get(), M.get(), A2.get());
-      auto P = space_op.GetPreconditionerMatrix<ComplexOperator>(1.0, omega, -omega * omega,
-                                                                 omega);
+      auto P = space_op.GetPreconditionerMatrix<ComplexOperator>(
+          std::complex<double>(1.0, 0.0), 1i * omega,
+          std::complex<double>(-omega * omega, 0.0), omega);
```
hughcars: Maybe `1.0 + 0.0i` and `-omega * omega + 0.0i`? I would have thought implicit conversion of doubles would be fine though.

palace/drivers/eigensolver.cpp
```diff
@@ -39,6 +40,20 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
 
+  // Check if there are nonlinear terms and, if so, setup interpolation operator.
+  const double target = iodata.solver.eigenmode.target;
+  auto A2 = space_op.GetExtraSystemMatrix<ComplexOperator>(target, Operator::DIAG_ZERO);
+  bool has_A2 = (A2 != nullptr);
+  NonlinearEigenSolver nonlinear_type = iodata.solver.eigenmode.nonlinear_type;
+  std::unique_ptr<Interpolation> interp_op;
+  if (has_A2 && nonlinear_type != NonlinearEigenSolver::SLP)
+  {
+    const int npoints = 3;  // Always use second order interpolation for now
```
hughcars: Nit: `constexpr`

palace/drivers/eigensolver.cpp
```diff
@@ -124,16 +155,27 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
   else
   {
     eigen->SetOperators(*K, *M, scale);
   }
   eigen->SetNumModes(iodata.solver.eigenmode.n, iodata.solver.eigenmode.max_size);
-  eigen->SetTol(iodata.solver.eigenmode.tol);
+  const double tol = (has_A2 && nonlinear_type != NonlinearEigenSolver::SLP)
+                         ? iodata.solver.eigenmode.linear_tol
+                         : iodata.solver.eigenmode.tol;
+  eigen->SetTol(tol);
```
hughcars: This is a little surprising, in that adding in the nonlinear solver basically changes the definition of the tolerance variables in the config, what's the rationale behind this? Is it necessary?

palace/linalg/arpack.cpp
```diff
@@ -825,8 +877,17 @@ double ArpackPEPSolver::GetResidualNorm(std::complex<double> l, const ComplexVec
...
+  if (opC)
+  {
+    opC->AddMult(x, r, l);
+  }
   opM->AddMult(x, r, l * l);
+  if (has_A2)
+  {
+    auto A2 = space_op->GetExtraSystemMatrix<ComplexOperator>(std::abs(l.imag()),
+                                                              Operator::DIAG_ZERO);
+    A2->AddMult(x, r, std::complex<double>(1.0, 0.0));
```
hughcars: Can a pointer to `A2` be passed in directly?

palace/linalg/arpack.cpp
```diff
@@ -646,19 +672,23 @@ ArpackPEPSolver::ArpackPEPSolver(MPI_Comm comm, int print)
...
+void ArpackPEPSolver::SetOperators(SpaceOperator &space_op_ref, const ComplexOperator &K,
+                                   const ComplexOperator &C, const ComplexOperator &M,
                                    EigenvalueSolver::ScaleType type)
 {
   MFEM_VERIFY(!opK || K.Height() == n, "Invalid modification of eigenvalue problem size!");
   bool first = (opK == nullptr);
   opK = &K;
   opC = &C;
   opM = &M;
+  space_op = &space_op_ref;
```
hughcars: Why the need for the indirection? Can't you use `space_op` directly below?

palace/linalg/vector.hpp
```diff
@@ -198,9 +198,9 @@ inline auto Dot(MPI_Comm comm, const VecType &x, const VecType &y)
 
 // Calculate the vector 2-norm.
 template <typename VecType>
-inline auto Norml2(MPI_Comm comm, const VecType &x)
+inline auto Norml2(MPI_Comm comm, const VecType &x, bool squared = false)
 {
-  return std::sqrt(std::abs(Dot(comm, x, x)));
+  return squared ? std::abs(Dot(comm, x, x)) : std::sqrt(std::abs(Dot(comm, x, x)));
```
hughcars: Is there a reason not to use `Dot(x,x)` directly at the call site?

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ The format of this changelog is based on
   - Change wave port eigenproblem shift and sorting to fix an issue with the mode ordering.
     The first mode now has the largest propagation constant, closest to the TEM limit, and
     subsequent modes are ordered by decreasing propagation constant.
+  - Added support for nonlinear eigenvalue problems arising from frequency-dependent boundary
+    conditions. Two nonlinear eigensolvers are now available and can be specified by setting the
+    `config["Solver"]["Eigenmode"]["NonlinearType"]` option to `"Hybrid"` (default) or `"SLP"`.
```
hughcars, line 25: Maybe add some text that this will autoenable if a nonlinear bc is chosen?

examples/adapter/mesh/mesh.jl
```diff
@@ -0,0 +1,258 @@
...
+    refinement::Integer  = 0,
+    order::Integer       = 2,
+    ellipse_a::Real      = 1.5,
+    ellipse_b::Real      = 0.9,
+    rect_a::Real         = 2.3,
+    rect_b::Real         = 1.0,
+    adapter_length::Real = 14.0,
+    verbose::Integer     = 5,
+    gui::Bool            = false
+)
```
hughcars, line 49: Given this extruded geometry type, it's plausible this might make for a good hex/prism mesh option, similar to the cavity case. However, I don't think it's important for validating this feature, so only worth it if you think there's some extra value in it, otherwise just leave it as is.

examples/adapter/mesh/mesh.jl
```diff
@@ -0,0 +1,258 @@
...
+    # Generate mesh
+    gmsh.option.setNumber(
+        "Mesh.MeshSizeMax",
+        minimum([ellipse_a, ellipse_b, rect_a/2, rect_b/2, adapter_length])
+    )
+    gmsh.option.setNumber("Mesh.MeshSizeFromPoints", 0)
+    gmsh.option.setNumber("Mesh.MeshSizeFromCurvature", 0)
+    gmsh.option.setNumber("Mesh.MeshSizeExtendFromBoundary", 0)
+    gmsh.option.setNumber("Mesh.Algorithm", 6)
+    gmsh.option.setNumber("Mesh.Algorithm3D", 10)
```
hughcars: For these small meshes we're opting more towards Delaunay, on the basis that it is more easily repeatable than HXT (10).

test/examples/runtests.jl
```diff
@@ -247,3 +250,93 @@ if "cpw/wave_adaptive" in cases
...
+            "Re{I[4]} (A)",
+            "Im{I[4]} (A)",
+            "Q_ext[1]",
+            "κ_ext[1] (GHz)",
+            "Q_ext[2]",
+            "κ_ext[2] (GHz)",
+            "Q_ext[3]",
+            "κ_ext[3] (GHz)",
+            "Q_ext[4]",
+            "κ_ext[4] (GHz)"
```
hughcars, line 293: Are these values sufficiently unstable that we really can't check them? If you tighten eigen/linear tolerances does it improve?

palace/utils/configfile.hpp
```diff
@@ -702,6 +702,30 @@ struct EigenSolverData
...
+  NonlinearEigenSolver nonlinear_type = NonlinearEigenSolver::DEFAULT;
+
+  // For nonlinear problems, refine the linearized solution with a nonlinear eigensolver.
+  bool refine_nonlinear = true;
+
+  // For nonlinear problems using the hybrid approach, relative tolerance of the linear
+  // eigenvalue solver used to generate the initial guess.
+  double linear_tol = 1e-3;
+
+  // Upper end of the target range for nonlinear eigenvalue solver [GHz].
```
hughcars: Add a comment that `<0` gives the default condition.

palace/models/spaceoperator.cpp
```diff
@@ -730,12 +731,120 @@ auto BuildLevelParOperator<ComplexOperator>(std::unique_ptr<Operator> &&br,
...
+      dfbr(mat_op.MaxCeedBdrAttribute()), fbr(mat_op.MaxCeedBdrAttribute()),
+      fpr(mat_op.MaxCeedAttribute());
+  AddStiffnessCoefficients(a0.real(), dfr, fr);
+  AddStiffnessBdrCoefficients(a0.real(), fbr);
+  AddDampingCoefficients(a1.imag(), fr);
+  AddDampingBdrCoefficients(a1.imag(), fbr);
+  AddAbsMassCoefficients(pc_mat_shifted ? std::abs(a2.real()) : a2.real(), fr);
+  AddRealMassBdrCoefficients(pc_mat_shifted ? std::abs(a2.real()) : a2.real(), fbr);
+  AddExtraSystemBdrCoefficients(a3, dfbr, dfbr, fbr, fbr);
+  AddRealPeriodicCoefficients(a0.real(), fr);
```
hughcars, line 604: Should this be `fpr`? If not, fpr isn't being used here and in the other 2 operator method.

palace/linalg/slepc.hpp
```diff
@@ -27,6 +27,7 @@ typedef struct _p_PEP *PEP;
 typedef struct _p_BV *BV;
 typedef struct _p_ST *ST;
 typedef struct _p_RG *RG;
+typedef struct _p_NEP *NEP;
```
hughcars: Consider adding `GetNEP()` like the `GetRG()` methods on the base class.

palace/linalg/slepc.hpp
```diff
@@ -410,6 +445,124 @@ class SlepcPEPSolver : public SlepcPEPSolverBase
...
+  operator PetscObject() const override { return reinterpret_cast<PetscObject>(nep); };
+};
+
+// Nonlinear eigenvalue problem solver: T(λ) x = (K + λ C + λ² M + A2(λ)) x = 0.
+class SlepcNEPSolver : public SlepcNEPSolverBase
+{
+public:
+  using SlepcEigenvalueSolver::delta;
+  using SlepcEigenvalueSolver::gamma;
+  using SlepcEigenvalueSolver::has_A2;
```
hughcars: has_A2 and opInterp appear unneeded here, I presume because once you're in NEPSolver, you know they are needed, and only need them within class scope.

palace/models/spaceoperator.cpp
```diff
@@ -859,6 +916,29 @@ std::unique_ptr<OperType> SpaceOperator::GetPreconditionerMatrix(double a0, doub
...
+      typename std::conditional<std::is_same<OperType, ComplexOperator>::value,
+                                ComplexParOperator, ParOperator>::type;
+  const auto *PtAP_A = (A) ? dynamic_cast<const ParOperType *>(A) : nullptr;
+  const auto *PtAP_B = (B) ? dynamic_cast<const ParOperType *>(B) : nullptr;
+  MFEM_VERIFY((!A || PtAP_A) && (!B || PtAP_B),
+              "SpaceOperator requires ParOperator or ComplexParOperator for system matrix "
+              "construction!");
+
+  int height = PtAP_A->LocalOperator().Height();
+  int width = PtAP_A->LocalOperator().Width();
```
hughcars: If A was null, this will segfault, can fix with a guard to check for `PtAP_A` or `PtAP_B`.

palace/models/spaceoperator.cpp
```diff
@@ -859,6 +916,29 @@ std::unique_ptr<OperType> SpaceOperator::GetPreconditionerMatrix(double a0, doub
   return B;
 }
 
+template <typename OperType, typename ScalarType>
+std::unique_ptr<OperType>
+SpaceOperator::GetDividedDifferenceMatrix(ScalarType eps, const OperType *A,
+                                          const OperType *B,
+                                          Operator::DiagonalPolicy diag_policy)
```
hughcars: This function seems independent of `SpaceOperator`, aside from `BuildParSumOperator` (which is actually only a local scope function, not a member) and `nd_dbc_tdof_lists`. If this were broken out, and available as free functions similar to `multigrid.hpp`, then could avoid passing `SpaceOperator` directly into the eigensolvers, and use `std::function` instead. That way the abstraction of the eigensolver is kept disjoint from the spaceoperator class.

palace/linalg/slepc.cpp
```diff
@@ -1611,4 +2403,117 @@ PetscErrorCode __pc_apply_PEP(PC pc, Vec x, Vec y)
...
+      std::abs(lambda.imag()), palace::Operator::DIAG_ZERO);
+  const auto eps = std::sqrt(std::numeric_limits<double>::epsilon());
+  ctxF->opA2p = ctxF->space_op->GetExtraSystemMatrix<palace::ComplexOperator>(
+      std::abs(lambda.imag()) * (1.0 + eps), palace::Operator::DIAG_ZERO);
+  std::complex<double> denom = std::complex<double>(0.0, eps * std::abs(lambda.imag()));
+  ctxF->opAJ = ctxF->space_op->GetDividedDifferenceMatrix<palace::ComplexOperator>(
+      denom, ctxF->opA2p.get(), ctxF->opA2.get(), palace::Operator::DIAG_ZERO);
+  ctxF->opJ = ctxF->space_op->GetSystemMatrix(
+      std::complex<double>(0.0, 0.0), std::complex<double>(1.0, 0.0), 2.0 * lambda,
+      ctxF->opK, ctxF->opC, ctxF->opM, ctxF->opAJ.get());
```
hughcars: Both of these operator constructors, are effectively assembled from the `K`, `C` and `M` matrices already provided, and similarly for the divided difference. If the `BuildParSumOperator` free functions were not in `spaceoperator.cpp` you could side step the space_op pointer here. In fact, the `BuildSumOperator` functions can be broken apart I think, as ultimately they could work for any collection of coefficients and operators, then you wouldn't need checking for K,C,M directly, as the utility function would actually be more general.

palace/drivers/eigensolver.cpp
```diff
@@ -255,9 +308,10 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
   // to the complex system matrix.
   auto A = space_op.GetSystemMatrix(std::complex<double>(1.0, 0.0), 1i * target,
                                     std::complex<double>(-target * target, 0.0), K.get(),
-                                    C.get(), M.get());
-  auto P = space_op.GetPreconditionerMatrix<ComplexOperator>(1.0, target, -target * target,
-                                                             target);
+                                    C.get(), M.get(), A2.get());
+  auto P = space_op.GetPreconditionerMatrix<ComplexOperator>(
+      std::complex<double>(1.0, 0.0), 1i * target,
```
hughcars: Nit: can use `1.0 + 0i` here to match the `1i` usage in the other args.

palace/linalg/nleps.cpp
```diff
@@ -0,0 +1,876 @@
...
+
+  // Delta used in to compute divided difference Jacobian.
+  const auto delta = std::sqrt(std::numeric_limits<double>::epsilon());
+
+  // Suppress wave port output during Newton iterations.
+  space_op->GetWavePortOp().SetSuppressOutput(true);
+
+  // Set a seed and distribution for random Eigen vectors to ensure the same values on all
+  // ranks.
+  unsigned int seed = nev;
```
hughcars: Probably just pick a fixed number here, otherwise technically you'll get different results if you adjust the config file to request an extra eigenvector, any constant should do. Or is the idea that this gives you a secret lever for twiddling the randomness if necessary? An alternative is `Mpi::Size(comm)`, as changing the number of ranks is going to subtly kick up noise anyway.

palace/linalg/nleps.cpp
```diff
@@ -0,0 +1,876 @@
...
+
+    // Normalize eigenvector estimate.
+    double norm_v = std::sqrt(linalg::Norml2(GetComm(), v, true) + v2.squaredNorm());
+    v *= 1.0 / norm_v;
+    v2 *= 1.0 / norm_v;
+
+    // Set the linear solver operators.
+    opA2 = space_op->GetExtraSystemMatrix<ComplexOperator>(std::abs(eig.imag()),
+                                                           Operator::DIAG_ZERO);
+    opA = space_op->GetSystemMatrix(std::complex<double>(1.0, 0.0), eig, eig * eig, opK,
```
hughcars: Nit: can use `using namespace std::complex_literals;` at top of file and `1.0 + 0i`.

palace/drivers/eigensolver.cpp
```diff
@@ -282,6 +336,49 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
                    ? fmt::format(" (first = {:.3e}{:+.3e}i)", lambda.real(), lambda.imag())
                    : "");
   }
+
+  if (has_A2 && iodata.solver.eigenmode.refine_nonlinear &&
+      nonlinear_type != NonlinearEigenSolver::SLP)
+  {
+    Mpi::Print("\n Refining eigenvalues with Quasi-Newton solver\n");
+    std::unique_ptr<NonLinearEigenvalueSolver> qn;
+    qn = std::make_unique<QuasiNewtonSolver>(space_op.GetComm(), iodata.problem.verbose);
```
hughcars: `auto qn = std::make_unique....` then it's possible to remove the `SetPreconditionerLag` type methods which all have MFEM_ABORT on the various eigensolvers, and just have it be a pointer method directly within `QuasiNewton` which seems to be the only call. Same thing should apply to a few more of the methods introduced for the other solvers but only used in the nonlinear setting.

Could actually have a convenience ctor from an existing eigensolver (the eigen above), as mostly you're copying in the same settings that were set previously.

palace/drivers/eigensolver.cpp
```diff
@@ -282,6 +336,49 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+    else
+    {
+      qn->SetOperators(space_op, *K, *M, EigenvalueSolver::ScaleType::NONE);
+    }
+    qn->SetNumModes(iodata.solver.eigenmode.n, iodata.solver.eigenmode.max_size);
+    qn->SetPreconditionerLag(iodata.solver.eigenmode.preconditioner_lag,
+                             iodata.solver.eigenmode.preconditioner_lag_tol);
+    qn->SetMaxRestart(iodata.solver.eigenmode.max_restart);
+    qn->SetLinearSolver(*ksp);
+    qn->SetShiftInvert(1i * target);
```
hughcars, line 375: This is only used for the shift and inversion for eigenvalues without linear initial guesses correct? Looking at `sigma` in `nleps.cpp`.

palace/linalg/eps.hpp
```diff
@@ -84,6 +92,13 @@ class EigenvalueSolver
   // Set maximum number of Arnoldi update iterations.
   virtual void SetMaxIter(int max_it) = 0;
 
+  // Set the update frequency and tolerance of the preconditioner.
+  virtual void SetPreconditionerLag(int preconditioner_update_freq,
+                                    double preconditioner_update_tol) = 0;
+
+  // Set the maximum number of restarts with the same initial guess.
+  virtual void SetMaxRestart(int max_num_restart) = 0;
+
```
hughcars: If I'm understanding correctly, the hybrid approach using the quasi newton scheme doesn't need this interface, as the eigensolve doesn't make use of it, and it's only used from within the nonlinear solver scope where a new pointer to `QuasiNewton` is created then moved to `eigen` eventually. So move these methods to `NonlinearEigenvalueProblem`?

palace/linalg/eps.hpp
```diff
@@ -52,17 +54,23 @@ class EigenvalueSolver
...
   virtual void SetOperators(const ComplexOperator &K, const ComplexOperator &M,
                             ScaleType type) = 0;
   virtual void SetOperators(const ComplexOperator &K, const ComplexOperator &C,
                             const ComplexOperator &M, ScaleType type) = 0;
+  virtual void SetOperators(SpaceOperator &space_op, const ComplexOperator &K,
+                            const ComplexOperator &M, ScaleType type) = 0;
+  virtual void SetOperators(SpaceOperator &space_op, const ComplexOperator &K,
+                            const ComplexOperator &C, const ComplexOperator &M,
+                            ScaleType type) = 0;
+  virtual void SetNLInterpolation(const Interpolation &interp) = 0;
```
hughcars: Given the number of these operators which are later being overrode just with an abort one class down, I wonder if this should be refactored to have an `ABORT` default method here for `SetOperators`. Each class then only overrides those it needs rather than needing an intermediate "Base" for ARPACK and slepc that does this. The abort mesages in each "Base" class aren't much more informative than just a general abort message here, and reduces the indirection.

Secondly `SetNLInterpolation` appears to be necessary here as it has to be accessed via a base class pointer to this class. As discussed offline, I think this might be better named as `PolynomialExpansion` than `Interpolation`, as each of the coefficient terms in the expansion are being used 1 at a time, rather than combined into an interpolation evaluated over the range of omega. 

`SetPreconditionerLag` and `SetMaxRestart` are only ever used via the QuasiNewton pointer in eigensolver, and should probably just be `NonlinearEigenSolver` methods. I don't think they make sense in the general "linear allowed" setting.

palace/linalg/nleps.cpp
```diff
@@ -0,0 +1,876 @@
...
+  }
+  if (k > n || k < 0 || n == 0)
+  {
+    return 0.0;
+  }
+  return elementarySymmetric(points, k, n - 1) +
+         points[n - 1] * elementarySymmetric(points, k - 1, n - 1);
+}
+
+void NewtonInterpolationOperator::Interpolate(const int order,
```
hughcars: Nit: no need for `const` if passing by value.

palace/linalg/eps.hpp
```diff
@@ -52,17 +54,23 @@ class EigenvalueSolver
...
-  // Set operators for the generalized eigenvalue problem or for the quadratic polynomial
-  // eigenvalue problem.
+  // Set operators for the generalized eigenvalue problem, quadratic polynomial
+  // eigenvalue problem, or nonlinear eigenvalue problem.
   virtual void SetOperators(const ComplexOperator &K, const ComplexOperator &M,
                             ScaleType type) = 0;
   virtual void SetOperators(const ComplexOperator &K, const ComplexOperator &C,
                             const ComplexOperator &M, ScaleType type) = 0;
+  virtual void SetOperators(SpaceOperator &space_op, const ComplexOperator &K,
+                            const ComplexOperator &M, ScaleType type) = 0;
```
hughcars: I'm not a fan of `SpaceOperator` as an argument here, as it breaks the abstraction of the eigensolves acting on linear algebra objects. The 4 methods I can see that really need SpaceOperator rn are `GetExtraSystemMatrix`, `GetSystemMatrix`, `GetDividedDifferenceMatrix`, and `GetPreconditionerMatrix`. The middle two there are convenience wrappers around `BuildSumOperator` which we could break out of SpaceOperator then. Then the other two are really maps from `omega -> unique_ptr<ComplexOperator>` which could be passed in as function pointers. 

Something like `SetPreconditionerUpdate(std::function<std::unique_ptr<ComplexOperator>(std::complex<double>)>);` which would be passed a lambda wrapping the spaceoperator call. The same for the nonlinear system term, which has the same map structure. This means moving lots of things out of `spaceoperator.cpp`

palace/linalg/nleps.hpp
```diff
@@ -0,0 +1,282 @@
...
+  bool sinvert;
+
+  // Storage for computed eigenvalues and eigenvectors.
+  std::vector<std::complex<double>> eigenvalues;
+  std::vector<ComplexVector> eigenvectors;
+  std::unique_ptr<int[]> perm;
+
+  // Storage for eigenpairs initial guesses.
+  std::vector<std::complex<double>> init_eigenvalues;
+  std::vector<ComplexVector> init_eigenvectors;
```
hughcars: Rather than storing these initial values separately, can the eigenvalues and eigenvectors be initialized directly? This could be combined with a consuming constructor from `EigenvalueSolver&&`, which would move the `eigenvalues` and `eigenvectors` into these variables. That way on initializing a `NonlinearEigenSolver` from an initial guess from a linear solver, no extra working variable allocation needs to be made.

palace/linalg/nleps.cpp
```diff
@@ -0,0 +1,876 @@
...
+
+  // Storage for eigenpairs.
+  std::vector<ComplexVector> X;
+  std::vector<std::complex<double>> eigs;
+
+  // Reset any previously computed eigenpairs.
+  eigenvalues.clear();
+  eigenvectors.clear();
+  eigenvectors.reserve(nev);
+  X.reserve(nev);
```
hughcars, line 395: This will be constructing and destructing a lot of `ComplexVector`, why can't the class member variables be used, and those variables be initialized directly with the initial values. Essentially lift the initialization of the member variables from
```
    // Set the eigenpair estimate to the initial guess.
    std::complex<double> eig, eig_opInv;
    if (guess_idx < num_init_guess)
    {
      eig = init_eigenvalues[guess_idx];
      v = init_eigenvectors[guess_idx];
    }
    else
    {
      eig = sigma;
      space_op->GetRandomInitialVector(v);
    }
    eig_opInv = eig;  // eigenvalue estimate used in the (lagged) preconditioner
```
below and put this into the ctor where all the information is known. The ctor can then be passed a `mfem::Array<int>` for the essential dofs, which are used to initialize the random op case, or instead can use an average of the other eigenvectors? 

All initial guess eigenvectors are assumed to respect the bcs, so their average would too, and be pseudo random, so you could `AXPBYPCZ` a random selection of them together to have a vector that satisfies the bcs without ever knowing which dofs they are.

palace/linalg/nleps.cpp
```diff
@@ -0,0 +1,876 @@
...
+  if (nleps_it <= 0)
+  {
+    nleps_it = 100;
+  }
+
+  // Delta used in to compute divided difference Jacobian.
+  const auto delta = std::sqrt(std::numeric_limits<double>::epsilon());
+
+  // Suppress wave port output during Newton iterations.
+  space_op->GetWavePortOp().SetSuppressOutput(true);
```
hughcars: Cna hoist this setting and unsetting into `eigensolver.cpp`, similar to how it is done in `drivensolver.cpp`, one fewer `space_op` necessary.

palace/linalg/nleps.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+  // Storage for computed residual norms and eigenvector scalings.
+  std::unique_ptr<double[]> res, xscale;
```
hughcars, line 58: Why use `unique_ptr` to a buffer over a `std::vector` aside from consistency to `arpack.hpp` etc? It does marginally reduce the overhead, but makes using them a bit weirder, and doesn't allow reuse on resizing etc.

palace/linalg/slepc.hpp
```diff
@@ -410,6 +445,124 @@ class SlepcPEPSolver : public SlepcPEPSolverBase
...
+                    const ComplexOperator &M, ScaleType type) override;
+  void SetOperators(SpaceOperator &space_op, const ComplexOperator &K,
+                    const ComplexOperator &C, const ComplexOperator &M,
+                    ScaleType type) override;
+  void SetBMat(const Operator &B) override;
+};
+
+// Base class for SLEPc's NEP problem type
+class SlepcNEPSolverBase : public SlepcEigenvalueSolver
+{
```
hughcars, line 417: This is a bit weird, but do you think this should probably inherit from both `SlepcEigenvalueSolver` and `NonlinearEigenvalueSolver`. This does create a diamond inheritance pattern, from `EigenvalueSolver`, but I believe this is ok because `EigenvalueSolver` is a pure virtual class with no member variables, only defining interface. Basically wondering if this might help move some of the nonlinear pieces out of the standard eigensolver interface (which ideally should be basically unchanged).

palace/drivers/eigensolver.cpp
```diff
@@ -39,6 +40,20 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
   auto C = space_op.GetDampingMatrix<ComplexOperator>(Operator::DIAG_ZERO);
   auto M = space_op.GetMassMatrix<ComplexOperator>(Operator::DIAG_ZERO);
 
+  // Check if there are nonlinear terms and, if so, setup interpolation operator.
+  const double target = iodata.solver.eigenmode.target;
+  auto A2 = space_op.GetExtraSystemMatrix<ComplexOperator>(target, Operator::DIAG_ZERO);
+  bool has_A2 = (A2 != nullptr);
+  NonlinearEigenSolver nonlinear_type = iodata.solver.eigenmode.nonlinear_type;
+  std::unique_ptr<Interpolation> interp_op;
```
hughcars, line 58: Rather than creating this operator and passing it into the slepc and arpack below, can we fold the interpolation directly into KCM immediately use `BuildParSumOperator`?
```
std::unique_ptr<ComplexOperator> Ko,Co,Mo;
if (has_A2 && nonlinear_type != NonlinearEigenSolver::SLP)
{
  Interpolation interp_op(space_op);
  const int npoints = 3; // Always use second order interpolation for now
  const double target_max = iodata.solver.eigenmode.target_upper;
  interp_op.Interpolate(npoints - 1, 1i * target, 1i * target_max);
  Ko = std::move(K); Co = std::move(C); Mo = std::move(M);
  K = BuildParSumOperator<2>({1.0+0i, 1.0+0i}, {interp_op.GetOrderMat(0), Ko});
  C = BuildParSumOperator<2>({1.0+0i, 1.0+0i}, {interp_op.GetOrderMat(1), Co});
  M = BuildParSumOperator<2>({1.0+0i, 1.0+0i}, {interp_op.GetOrderMat(2), Mo});
}
```
Then slepc and arpack never need to know anything about the interpolation at all, they're just working on slightly different looking KCM matrices. This requires `BuildParSumOperator` being broken out, which I have worked out.

palace/linalg/nleps.hpp
```diff
@@ -0,0 +1,282 @@
...
+  void SetOperators(const ComplexOperator &K, const ComplexOperator &M,
+                    ScaleType type) override;
+  void SetOperators(const ComplexOperator &K, const ComplexOperator &C,
+                    const ComplexOperator &M, ScaleType type) override;
+  void SetOperators(SpaceOperator &space_op, const ComplexOperator &K,
+                    const ComplexOperator &M, ScaleType type) override;
+  void SetOperators(SpaceOperator &space_op, const ComplexOperator &K,
+                    const ComplexOperator &C, const ComplexOperator &M,
+                    ScaleType type) override;
+  void SetNLInterpolation(const Interpolation &interp) override;
```
hughcars: Once `K,C,M` have folded in the interpolation operator components, can go back to having only the 2 SetOperators on the original `EigenvalueSolver`, and then add a third option here with the `std::function` handle input, and another method for setting the preconditioner update function, both of which are anonymous function wrappers of space operator.

Something like this in `eigensolver.cpp`
```
[&space_op](std::complex<double> omega) -> std::unique_ptr<ComplexOperator>
{
  return space_op.GetExtraSystemMatrix<ComplexOperator>(0*omega, omega, omega*omega);
};
```

hughcars (changes requested):
Based on our discussion, here's the rough feedback. A few small things here and there, two larger ones:
1) I think we can remove `SpaceOperator` from the interface for nonlinear eigen solver, function pointers for the extra system matrix, and one for the preconditioner update. This needs some things broken out of spaceop, I'll do that and make another pr you can use
2) I think the InterpolationOperator can be used as a polynomial expansion, with the resulting variables folded into KCM in eigensolver, then EigenvalueSolver doesn't need to know anything nonlinear. To report the full nonlinear residual, then the `Solve()` on `QuasiNewton` can exit early, and the resulting `eigen` pointer will evaluate the full residual rather than the quadratic approximation system. 

The very hacky stuff is on `hughcars/nonlinear-eigen-wip` for any inspiration. The template stuff I'll break out into another pr as discussed.

palace/models/spaceoperator.cpp
simlapointe: No, it should be `fr`. AddRealPeriodicCoefficients adds the mass term, so `fr` is the right place. `fpr`/`fpi` are for the mixed curl terms. But you're right that `fpr` is unused here and in the similar method below so I will remove them (but it IS used in the fully complex `AssemblePreconditioner` method above).

test/examples/runtests.jl
simlapointe: They are very unstable, changing `np` leads to large variations.

docs/src/config/solver.md
simlapointe: Yeah hybrid is the default NLE for both SLEPc and ARPACK so I can drop "Default".

Right now it is always defaulting on if necessary. I slightly prefer that over erroring out and forcing the user to set a nonlinear solver, but can be convinced otherwise.

palace/drivers/eigensolver.cpp
simlapointe: The full eigensolve (linear or nonlinear) will always use `config["Solver"]["Eigenmode"]["Tol"]` as the desired tolerance. When using the hybrid approach, the polynomial eigensolve used to get the initial guess for the Newton can use a looser tolerance `config["Solver"]["Eigenmode"]["LinearTol"]`  which is an optional advanced config param that only affects the hybrid nonlinear eigensolver.

palace/linalg/vector.hpp
simlapointe: No _good_ reason, will use dot and remove this squared option.

palace/linalg/slepc.hpp
simlapointe: Not sure this is needed since `NEP` is like `PEP` and `EPS` and we don't have/need `GetPEP()`.

palace/drivers/eigensolver.cpp
simlapointe: For the Newton solver this is only used to pass `sigma` which is then used for filtering (discard eigenvalues < sigma), so the `ShiftInvert` loses its meaning a bit, but I didn't think it was worth creating a new function.

palace/linalg/arpack.cpp
simlapointe: Removed in latest, no need now that we're using extended K', C', M' operators

palace/linalg/arpack.cpp
simlapointe: Removed in latest.

palace/models/spaceoperator.cpp
simlapointe: Removed in latest, using `BuildParSumOperator`.

palace/linalg/slepc.cpp
simlapointe: Using `BuildParSumOperator` everywhere.

palace/linalg/eps.hpp
simlapointe: Agreed, changed in latest.

palace/linalg/nleps.hpp
simlapointe: Agreed, did something along these lines in the latest.

palace/linalg/nleps.hpp
simlapointe: It was just for consistency with both `arpack.hpp` and `slepc.hpp`.

palace/drivers/eigensolver.cpp
simlapointe: Now doing something like that in the latest.

palace/linalg/nleps.hpp
simlapointe: Done in the latest.

palace/linalg/eps.hpp
simlapointe: `SpaceOperator` is now no longer used in the eigensolvers.

palace/linalg/nleps.cpp
simlapointe: Now using `AXPBYPCZ` on a random selection of initial guess, or a purely random guess if there aren't enough initial guesses.

palace/linalg/slepc.hpp
simlapointe: This is a bit painful and requires moving quite a few things around. See if you still think it's necessary given all the other changes.

palace/linalg/nleps.hpp
```diff
@@ -0,0 +1,294 @@
...
+  // Workspace vector for operator applications.
+  mutable ComplexVector x1, y1;
+
+  // // Function to compute the A2 operator.
+  // std::optional<std::function<std::unique_ptr<ComplexOperator>(double)>> funcA2;
+
+  // // Function to compute the preconditioner matrix.
+  // std::optional<std::function<std::unique_ptr<ComplexOperator>(
+  //     std::complex<double>, std::complex<double>, std::complex<double>, double)>>
+  //     funcP;
```
hughcars: Debris?

PR #469, opened 2025-08-29
Fix macOS CI (again)

Sbozzolo:
The macos image changed and CI fails with

```
Error: cmake was installed from the local/pinned tap
but you are trying to install it from the homebrew/core tap.
Formulae with the same name from different taps cannot be installed at the same time.

To install this version, you must first uninstall the existing formula:
  brew uninstall cmake
Then you can install the desired version:
  brew install cmake
```
If I naively upgrade `cmake`, the build fails.

I found that:
- Default CMake has problems finding MPI
- CMake 4.1 has problems finding some std headers with llvm 18
- CMake 4.1 + llvm > 18 has problems with pthreads

So, I ended up pinning CMake to 3.31.8 (the latest in the 3.x series).

### DIFF
--- a/.github/workflows/build-and-test-macos.yml
+++ b/.github/workflows/build-and-test-macos.yml
@@ -144,21 +144,24 @@ jobs:
           echo "DYLD_LIBRARY_PATH=/opt/arm/armpl_24.10_flang-new_clang_19/lib:${DYLD_LIBRARY_PATH:+${DYLD_LIBRARY_PATH}:}" >> $GITHUB_ENV
           hdiutil detach "/Volumes/armpl_24.10_flang-new_clang_19_installer" -quiet
 
+        # We have to upgrade CMake because the bundled one has a bug with
+        # FindMPI that prevents it from finding MPI when compiling with GNU
+        # compilers.
+      - name: Setup cmake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '3.31.8'
+
       - name: Build Palace
         env:
           CMAKE_BUILD_TYPE: Release
           NUM_PROC_BUILD_MAX: '32'
         run: |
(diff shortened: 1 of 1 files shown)


cameronrutherford:
Maybe the CMake incompatibility is just a problem with those older LLVM versions? The spack package [here](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/llvm/package.py#L57) has version `20.1.8` that is two major versions ahead, and doesn't have any clear restrictions on CMake.

IMO all of our pipelines should be in Spack to avoid having to maintain such nightmares, but independent of that sentiment, PR looks good assuming tests pass.

Sbozzolo:
Just to clarify, I tried CMake 4.1 and LLVM 20, and CMake couldn't find the headers for pthread. So, even using recent versions, there's still something wrong.

cameronrutherford:
> Just to clarify, I tried CMake 4.1 and LLVM 20, and CMake couldn't find the headers for pthread. So, even using recent versions, there's still something wrong.

Hmmm.... Maybe this is a bug that isn't represented in Spack yet, or Spack somehow is better with the headers?

I know that @phdum-a and I have had many a difficulty with pthreads, and we often needed to change some things to get the standard library config to work...

hughcars:
> Just to clarify, I tried CMake 4.1 and LLVM 20, and CMake couldn't find the headers for pthread. So, even using recent versions, there's still something wrong.

Is this `pthread.h` miss in building libxsmm in particular?

hughcars:
We've had issues with this in the past https://github.com/libxsmm/libxsmm/issues/921, it had disappeared and I'd thought maybe cmake/llvm/apple had decided to just get along. It would appear that was an interlude. From memory the source of the issue was erratically defined `C_INCLUDE_PATH`, possibly from libxsmm not looking in the correct place.

Issue #468, opened 2025-08-28
The `floquet` example produces non-convergence warnings


Sbozzolo:
Running the provided `cylinder/floquet` example results in non-convergence warnings towards the end of the simulation:
```
--> Warning!
Linear solver did not converge, norm(Ax-b)/norm(b) = 1.562e-06 (norm(b) = 5.619e-01)!

     m,        Re{f} (GHz),        Im{f} (GHz),                  Q,      Error (Bkwd.),       Error (Abs.)
     1,      +2.319514e+00,      +4.639028e-04,      +2.500000e+03,      +7.363488e-12,      +2.134403e-09
 Wrote mode 1 to disk
PCG solver did NOT converge in 100 iterations (avg. reduction factor: 8.779e-01)

--> Warning!
Linear solver did not converge, norm(Ax-b)/norm(b) = 2.215e-06 (norm(b) = 1.134e-01)!

     2,      +2.978141e+00,      +5.956283e-04,      +2.499999e+03,      +1.634190e-11,      +4.782096e-09
 Wrote mode 2 to disk
PCG solver did NOT converge in 100 iterations (avg. reduction factor: 8.772e-01)

...
```
I understand that this has to do with error estimates and is harmless (from internal discussion). That said, this is an example Palace provides, so it would be nice if it run cleanly. It would also be good to discuss this warning somewhere in the docs (the only references to PCG in the entire Palace codebase are in printing the error messages).

Issue #447, opened 2025-08-14
FloquetWaveVector interface


kllrak:
When using a mesh that is not periodic, from the docs I understand you are supposed to use:

```
"Boundaries": {
    "Periodic": [
        {
            "DonorAttributes": [<int array>],
            "ReceiverAttributes": [<int array>],
            "Translation": [<float array>],
            "AffineTransformation": [<float array>],
            "FloquetWaveVector": [<float array>]
        },
        ...
    ],
    ...
}
```

Palace is incorporating phase delays directly into the PDE, so the association of a `FloquetWaveVector` to each boundary pair is somewhat misleading to me. There should be a single `FloquetWaveVector` entry rather than summing over whatever `FloquetWaveVector` entries are found in the config file. 

A minimal change would be to drop this option and instead only allow users to configure `["Boundaries"]["FloquetWaveVector"]`, as is currently possible. I suggest the following interface instead:

```
"Boundaries": {
    "Periodic": {
        "FloquetWaveVector": [<float array>],
        "Attributes": [  // could also be called BoundaryPairs or something
            {
                "DonorAttributes": [<int array>],
                "ReceiverAttributes": [<int array>],
                "Translation": [<float array>],
                "AffineTransformation": [<float array>],
            },
            ...
        ]
    },
    ...
}
```

When working with a periodic mesh you would simply drop the `Attributes` array since the periodicity is baked into the mesh already. Specifying `["Boundaries"]["Periodic"]["FloquetWaveVector"]` makes it clear where this is intended to be used even if strictly speaking the use of `FloquetWaveVector` modifies the PDE being solved within a domain.

laylagi:
> One more thing I'm not clear on is why `DonorAttributes` and `ReceiverAttributes` accept arrays. Typical application is for one boundary face to map onto another boundary face to establish periodic boundaries, which would suggest a single int for each instead of arrays. You would have multiple boundary pairs by adding entries to `Attributes`, not adding to `DonorAttributes` and `ReceiverAttributes`, if I understand correctly.

This was confusing to me as well when I was first adding these attributes. Apparently, we can have a "boundary" in our model which is actually a union of several mesh attributes. I don't remember setting up such a model myself though.

hughcars:
[...]

PR #471, opened 2025-08-29
Fix built-in periodic mesh issue and update Floquet wave vector interface

simlapointe:
Fixed an issue where meshes with built-in periodicity were failing since https://github.com/awslabs/palace/pull/316. Now all periodic boundary pairs should be specified in `config["Boundaries"]["Periodic"]["BoundaryPairs"]` whether or not the mesh has built-in periodicity.

Also updated the `FloquetWaveVector` interface as suggested in https://github.com/awslabs/palace/issues/447. Now a single global `FloquetWaveVector` is specified in `Periodic`.

Closes https://github.com/awslabs/palace/issues/447 and increased `MaxIts` on the `cylinder/floquet` example to close https://github.com/awslabs/palace/issues/468.

### DIFF
--- a/palace/utils/geodata_impl.cpp
+++ b/palace/utils/geodata_impl.cpp
@@ -953,5 +953,4 @@ DeterminePeriodicVertexMapping(std::unique_ptr<mfem::Mesh> &mesh,
                                      mesh_tol);
 }
 
-
 }  // namespace palace::mesh
--- a/test/unit/test-geodata.cpp
+++ b/test/unit/test-geodata.cpp
@@ -1,8 +1,8 @@
+#include <vector>
 #include <catch2/catch_approx.hpp>
 #include <catch2/catch_test_macros.hpp>
-#include <catch2/generators/catch_generators_all.hpp>
 #include <nlohmann/json.hpp>
-#include <vector>
+#include <catch2/generators/catch_generators_all.hpp>
 
 #include "utils/geodata.hpp"
 #include "utils/geodata_impl.hpp"


palace/utils/configfile.hpp
```diff
@@ -634,7 +628,7 @@ struct BoundaryData
   WavePortBoundaryData waveport = {};
   SurfaceCurrentBoundaryData current = {};
   PeriodicBoundaryData periodic = {};
-  FloquetData floquet;
+  // FloquetData floquet;
```
hughcars: Debris, delete. The `FloquetData` should also be removed as no-longer needed I guess.

palace/utils/geodata.cpp
```diff
@@ -1722,6 +1725,15 @@ DeterminePeriodicVertexMapping(std::unique_ptr<mfem::Mesh> &mesh,
     auto donor_normal = ComputeNormal(mesh, bdr_e_donor, true);
     auto receiver_normal = ComputeNormal(mesh, bdr_e_receiver, false);
 
+    // Return empty mapping if centroids and normal vectors are the same (up to a sign).
+    mfem::Vector diff = donor_centroid;
+    diff -= receiver_centroid;
+    double dot = donor_normal * receiver_normal;
+    if (diff.Norml2() < mesh_tol && std::abs(std::abs(dot) - 1.0) < mesh_tol)
+    {
+      return {};
```
hughcars: What test meshes did you use to check for the breakage? Are they standard mfem meshes? If so, would be good to have a small unit test to guard against this. We can pull from the `mfem/data` folder now for unit testing purposes, without needing to add the mesh into the repo.

PR #476, opened 2025-09-15
Drop small entries in matrix used in sparse direct solver

simlapointe:
Adding an option (true by default) to drop the small entries in the matrix used in the sparse direct solver. 

The sparse matrix used in the (coarse) direct solver sometimes contains a significant fraction of very small values (below machine epsilon). The fraction of small values differs from case to case, e.g. it is 20% in the cylinder example and 28% in the CPW example. These small values likely originate in part from using `skip_zeros=false` in the operator assembly and in part from the matrix multiplications with the restriction and prolongation operators `op = R A P`. Dropping these small values reduces the computational cost and memory usage of the sparse direct solver without any noticeable increase in Krylov iterations or overall computational time.

### DIFF
--- a/examples/cpw/cpw_lumped_adaptive.json
+++ b/examples/cpw/cpw_lumped_adaptive.json
@@ -199,8 +199,8 @@
       "Type": "Default",
       "KSPType": "GMRES",
       "Tol": 1.0e-8,
-      "MaxIts": 200
+      "MaxIts": 200,
+      "ReorderingReuse": false
     }
   }
 }
-


simlapointe:
> Symmetry of the sparsity pattern is sometimes important especially for column permutations from Metis/ParMetis if I recall correctly, just something to be wary of. I do think that MFEM or Hypre had an option to zero dropping which only did so if it preserved symmetry in the sparsity pattern, might be something to check out or write yourself if it's relevant.

Thanks for the heads up! I did check in my tests that the sparsity pattern was not modified when dropping zeros and it was always the case as long as the threshold was small so I don't think it's an issue. I

palace/models/waveportoperator.cpp
```diff
@@ -718,6 +718,7 @@ WavePortData::WavePortData(const config::WavePortData &data,
           return {};
         }());
     pc->SetSaveAssembled(false);
+    pc->SetDropSmallEntries(false);
```
Sbozzolo, line 724: Why is this set to `false`? It might be good to explain.

hughcars:
> I looked at this PR. It is reasonable to me. I verified that it compiles and that `cavity_pec` runs with and without this option.
> 
> My main reservation is on whether this should be a user-controllable option or not, and its documentation if we want to keep it controllable. My guess is that one would consider turning this off when they find weird results that cannot be explained/debugged by setting other options. This makes this a rather advanced feature that is probably only needed in corner cases?
> 
> If we want to keep it, I suggest improving the documentation. The current documentation is
> 
> > When set to `true`, entries smaller than the double precision machine epsilon are dropped from the system matrix used in the sparse direct solver.
> 
> This line describes the setting briefly without telling users why they should care about it or under which conditions they should consider setting it to `false`. As a result, this feature is not really accessible to non-experts.
> 
> Finally, thinking about testing, setting `true` as default means that there are currently no tests for the case with `false` (unless we consider the `SetDropSmallEntries(false)` in `waveportoperator`) . It would be good to have at least one case to make sure that the go through those branches of code.

A feature like this is mostly for developer usage, there are many "advanced" options that a normal user should never really touch, which are useful to have as options for us in debugging an issue, without having to hack and recompile. Examples of this include `"RemoveCurvature"` which converts a higher order mesh to lower order mesh, and is something no sane user will generally use, but is incredibly helpful for us in bisecting to find a bug. I'm perfectly ok with a feature exposed that should never be changed by a normal user, as there are power users such as us who will plausibly want the option to toggle it at some point, and debugging scenarios where we might not have access to the instance or build in question and be asking users to toggle these options and report results to us. A more expansive doc string is fine, but I think the current text is sufficient for a developer user, and the terseness if anything helpfully raises the barrier to entry to using it.

I am more concerned about Sebastian's point to do with symmetry in the sparsity pattern, but looking through the hypre implementation, it doesn't seem trivial to add such symmetry preservation. I think adding this first, and then if we see issues for symmetric systems, to investigate that as a hypre feature is a compromise (I suspect they would appreciate such a feature, we can't be the first to think it might help).

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ The format of this changelog is based on
     use in a non-amr simulation. [Issue 444](https://github.com/awslabs/palace/issues/444).
   - Fixed bug where `"CrackInternalBoundaryElements"` would result in incorrect
     results for some lumped port boundary conditions. [Issue 501](https://github.com/awslabs/palace/issues/501).
+  - Added an option to drop small entries (below machine epsilon) from the matrix used in the sparse
+    direct solver. This can be specified with `config["Solver"]["Linear"]["DropSmallEntries"]`.
```
hughcars, line 37: Add a PR link. (Will be changing all the Issue links above to PR ones in the near future.

palace/linalg/solver.cpp
```diff
@@ -81,6 +81,15 @@ void MfemWrapperSolver<ComplexOperator>::SetOperator(const ComplexOperator &op)
...
+    if (drop_small_entries)
+    {
+      const auto nnz_before = A->NNZ();
+      A->DropSmallEntries(std::pow(std::numeric_limits<double>::epsilon(), 2));
+      const auto nnz_after = A->NNZ();
+      Mpi::Print(" Dropping {} small sparse matrix entries out of {} ({:.1f}%)\n",
+                 (nnz_before - nnz_after), nnz_before,
+                 (double)(nnz_before - nnz_after) / nnz_before * 100.0);
+    }
     pc->SetOperator(*A);
```
hughcars, line 94: Does this mean that there are now two operators? I.e. within `op` there are the hypre par matrices, which are not dropped, and now there's these duplicates? Can't `op` be passed non-const ref and the member operators potentially be modified in place? in the case of `hAr` and `hAi` being matrices already.

PR #480, opened 2025-09-18
Extend infra for Serial/Parallel/GPU unit tests

Sbozzolo:
This commit extends Palace infrastructure and documentation to enable running different type of unit tests.

Now, `unit-test` will execute different tests depending on how it is run:
- tests marked as `Serial` when there is only one MPI process
- tests marked as `Parallel` when there are multiple processes
- tests marked as `GPU` when the `--device` is `cuda` or `hip`

To enable this, tests have to be explicitly marked with the relevant tag. I added an example and a tutorial to show how this is done, and I expanded upon the documentation on testing.

Explicitly marking tests as `Serial` opens to door to running multiple `Serial` tests concurrently (although, we might need to be careful if tests perform file system operations or use other share resources to avoid race conditions). `Parallel` tests are already run with as many cores as available. (Currently on `main`, `Serial` tests are being run with MPI, which means that N copies of the same identical test are being run at the same time)

The implementation takes inspiration from MFEM's unit testing approach, but with a significant difference. While MFEM defines different executables for serial, parallel (`punit`), and GPU tests (`gunit` and `pgunit`), the approach implemented keeps everything unified and allows one to reuse the same test for different purposes simply by adding the relevant execution tag.


test/unit/test-vector.cpp
```diff
@@ -0,0 +1,132 @@
...
+  // Imaginary part: each rank contributes rank * cv.Size() + sum(0 to cv.Size()-1).
+  // sum(0 to cv.Size()-1) = cv.Size() * (cv.Size() - 1) / 2.
+  double index_sum = cv.Size() * (cv.Size() - 1) / 2.0;
+  double expected_imag = cv.Size() * rank_sum + size * index_sum;
+
+  REQUIRE_THAT(sum.real(), Catch::Matchers::WithinRel(expected_real));
+  REQUIRE_THAT(sum.imag(), Catch::Matchers::WithinRel(expected_imag));
+}
+
+TEST_CASE("ComplexVector Set", "[vector][Serial][Parallel][GPU]")
```
Sbozzolo, line 109: I added this test because it uses OpenMP

docs/src/developer/testing.md
```diff
@@ -0,0 +1,211 @@
...
+These tags are inclusive, meaning that a test can be marked with multiple
+special tags, if the test is meaningful in different contexts (e.g., if a test
+supports CPU and GPU implementation at the same time). For example,
+`[Serial][Parallel][GPU]` indicates a test that should run in every case.
+
+!!! warning
+    
+    Tests without any of our special tags will never execute. Make sure to add at
+    least one.
+
```
hughcars, line 111: Is it possible to ensure that all tests have at least one such tag?

docs/src/developer/tutorial_add_new_unit_test.md
```diff
@@ -0,0 +1,165 @@
...
+}
+
+}  // namespace palace
+```
+
+This defines a new `Catch2` test case. The key components are:
+
+  - Test name (`Vector Sum - Basic`): Must be unique across the test suite
+  - Tags: `[myvector]` (arbitrary, used for filtering) and `[Serial]` (special tag, more on this later)
+  - `Approx()`: Handles floating-point comparison tolerances
```
hughcars: This seems out of date, I guess written before changing to `WithinRel`.

hughcars (changes requested):
Looks great! One question about verifying the tags, and another bit of comment debris, but basically good to go.

docs/src/developer/testing.md
Sbozzolo: I tried using the Catch2 APIs to add such check at test discovery, but I couldn't get it to work, so I added it to CI instead (style workflow).

Sbozzolo: And the CI found a case I had missed:
https://github.com/awslabs/palace/actions/runs/17991290381/job/51181679384?pr=480

hughcars (approved):

PR #481, opened 2025-09-18
Refactor BuildParSumOperator for wider usage

hughcars:
The nonlinear eigenvalue problem PR can be simplified if the `BuildParSumOperator` capability wasn't tied into `SpaceOperator`, this PR refactors to expose it. Plucks a commit from @simlapointe's branch to support complex coefficients and then fixes a typo.


simlapointe (approved):
All looks good to me!

palace/linalg/rap.cpp
```diff
@@ -734,4 +748,228 @@ ComplexVector &ComplexParOperator::GetTestLVector() const
...
+    }
+    O->SetEssentialTrueDofs(*ess_dofs, policy);
+  }
+  return O;
+}
+
+// TODO: replace with std::to_array in c++20.
+namespace detail
+{
+// Helper for conversion to std::array
```
Sbozzolo:
```suggestion
// Helper for conversion to std::array.
```

PR #483, opened 2025-09-19
Tighten CI permission

Sbozzolo:
From CodeQL:

> If a GitHub Actions job or workflow has no explicit permissions set, then the repository permissions are used. Repositories created under organizations inherit the organization permissions. The organizations or repositories created before February 2023 have the default permissions set to read-write. Often these permissions do not adhere to the principle of least privilege and can be reduced to read-only, leaving the write permission only to a specific types as issues: write or pull-requests: write.

### DIFF
--- a/.github/workflows/build-and-test-linux.yml
+++ b/.github/workflows/build-and-test-linux.yml
@@ -1,5 +1,8 @@
 name: Build and Test (Linux)
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
--- a/.github/workflows/build-and-test-macos.yml
+++ b/.github/workflows/build-and-test-macos.yml
@@ -1,5 +1,8 @@
 name: Build and Test (macOS)
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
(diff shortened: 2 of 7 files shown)


cameronrutherford (changes requested):
I think you should also touch `docs.yml` and `spack.yaml` here so that we have consistent permissions configurations across the files.

i.e. `spack.yaml` doesn't have any permissions configuration and seems to have assumed read/write as default, and then `docs.yml` has this specified, just at the job level, not the workflow level. We should be consistent with our approach (of setting things at the workflow level).

Sbozzolo:
> I think you should also touch `docs.yml` and `spack.yaml` here so that we have consistent permissions configurations across the files.
> 
> i.e. `spack.yaml` doesn't have any permissions configuration and seems to have assumed read/write as default, and then `docs.yml` has this specified, just at the job level, not the workflow level. We should be consistent with our approach (of setting things at the workflow level).

I didn't change those because they are read/write, but now I updated them to be consistent.

cameronrutherford (approved):

PR #487, opened 2025-09-23
Github action to check license

hughcars:
Some files were missing licenses, this pr adds a github action to check automatically for the presence of the correct license in c, cpp, h, hpp, jl and md files.

### DIFF
--- a/test/unit/test-vector.cpp
+++ b/test/unit/test-vector.cpp
@@ -1,3 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
 #include <catch2/catch_test_macros.hpp>
 #include <catch2/generators/catch_generators_all.hpp>
 #include <catch2/matchers/catch_matchers_floating_point.hpp>


.github/workflows/style.yml
```diff
@@ -60,3 +60,12 @@ jobs:
             echo 'Configuration file validation passed'
             exit 0
           fi
+
+  check-license:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: viperproject/check-license-header@v1
```
Sbozzolo: Version 2 is available. Did you pick `v1` for any specific reason?

docs/src/config/boundaries.md
```diff
@@ -1,8 +1,5 @@
-```@raw html
 <!--- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved. --->
```
Sbozzolo, line 2: I think the old code was correct: it embedded the copyright as a HTML comment (so that one would still get a copyright notice if they scraped the page). If you move to markdown comments, that is lost. 

Also, I don't understand why, but this change pastes the comment directly to the page:
image

Sbozzolo (changes requested):

.github/license-check/config.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  {
+    "include": [
+      "**/*.md"
+    ],
+    "exclude": [
+      ".git/**"
+    ],
+    "license": "./.github/license-check/md.txt"
+  }
```
Sbozzolo, line 43: I realized that a side effect of this is that markdowns now render this on GitHub

Image

So, I think the correct thing to do is

```suggestion
  {
    "include": [
      "**/*.md"
    ],
    "exclude": [
      ".git/**",
      "docs"
    ],
    "license": "./.github/license-check/md.txt"
  },
  {
    "include": [
      "docs/**/*.md"
    ]
    "license": "./.github/license-check/md_html.txt"
  },  
```

with two different `md.txt` and `md_html.txt`

Sbozzolo (approved):
Thank you!

PR #497, opened 2025-09-26
Allow reusing NC AMR meshes in nonadaptive context

hughcars:
A mesh output from a previous NC adaptation could not be used in a non-AMR context, as the mesh loader was assuming the NC status from the config file, whereas it should be extracted from the mesh reader.

### DIFF
--- a/docs/src/config/model.md
+++ b/docs/src/config/model.md
@@ -20,7 +20,11 @@
 
 with
 
-`"Mesh" [None]` :  Input mesh file path, an absolute path is recommended.
+`"Mesh" [None]` :  Input mesh file path, an absolute path is recommended. If the provided
+mesh is nonconformal, it is assumed that it comes from a previous *Palace* solve using AMR,
+and all mesh preprocessing checks and modifications (for example
+[`model["Refinement"]["CrackInternalBoundaryElements"]`](#model%5B%22Refinement%22%5D)), are
+skipped .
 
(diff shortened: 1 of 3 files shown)


docs/src/config/model.md
```diff
@@ -20,7 +20,11 @@
 
 with
 
-`"Mesh" [None]` :  Input mesh file path, an absolute path is recommended.
+`"Mesh" [None]` :  Input mesh file path, an absolute path is recommended. If the provided
+mesh is nonconformal, it is assumed that it comes from a previous *Palace* solve using AMR,
```
simlapointe, line 24: Is it possible for this assumption to be invalid (i.e. the mesh is nonconformal but it does not come from Palace)? If so, I'm guessing we would like to apply the preprocessing steps in that case. Is there a way to do so with the current config flags, and if not could we add a config flag to handle that case?

simlapointe (commented):
Looks good! Just one question below.

docs/src/config/model.md
hughcars: The problem we have is that the preprocessing steps can't work on a nonconformal mesh, because a lot of them rely on things like all vertices are at the end of an edge. So it's not just that we assume that, but also that we *can't* perform a lot of the steps.

simlapointe: I see, nvm my comment then.

simlapointe (approved):

PR #499, opened 2025-09-26
Update hypre

### LINKED ISSUE #498: Update hypre to 3.0.0+ to fix compatibility with CUDA 13
The version of `hypre` Palace currently uses is incompatible with CUDA 13 (see, [hypre with cuda 13+](https://github.com/hypre-space/hypre/issues/1326)). As a result, Palace does not compile with the most recent version of CUDA. This was fixed in https://github.com/hypre-space/hypre/pull/1353. This commit is part of `hypre` 3.0.0, which was released today.

Sbozzolo:
The tagged commit for `hypre` is incompatible with CUDA 13, so that Palace does not compile with the latest version of the CUDA toolkit.

`hypre` was updated in PR473 two weeks ago, so the difference with the most recent version is minimal

Closes #498

### DIFF
--- a/cmake/ExternalGitTags.cmake
+++ b/cmake/ExternalGitTags.cmake
@@ -62,7 +62,7 @@ set(EXTERN_HYPRE_GIT_BRANCH
   "Git branch for external HYPRE build"
 )
 set(EXTERN_HYPRE_GIT_TAG
-  "4dd96d0e83088890879612c58364f6c10756ed90" CACHE STRING
+  "da9f93f8d698f4caaaff35fe81655b8ad7bb91f9" CACHE STRING
   "Git tag for external HYPRE build"
 )
 


Sbozzolo:
I thought it was a trivial update because everything compiled fine locally, but I must have used some cached objects...

hughcars:
> I thought it was a trivial update because everything compiled fine locally, but I must have used some cached objects. In fact, MFEM is not compatible with Hypre 3.0.0 yet
> 
> [mfem/mfem#5042](https://github.com/mfem/mfem/issues/5042)

Yeah... I was following the hypre 3.0 developments, basically waiting for mfem to update. Once that's updated we can then pull everything forward in one. Hopefully some of our patches will also have been merged then. Right now we can treat it as a known issue that we can close with the dependency update.

PR #500, opened 2025-09-29
Replace deprecated Catch2::Approx with Catch2::Matchers

### LINKED ISSUE #486: Unit tests are using deprecated Catch2 floating point matchers
From [Catch2](https://catch2-temp.readthedocs.io/en/latest/comparing-floating-point-numbers.html#approx):
> We strongly recommend against using `Approx` when writing new code. 
> [...] `Approx `is left as-is, is considered deprecated, and should not be used in new code.

The recommended way is to use `WithinX/Y/Z` (described in the page linked above).

All our tests are using `Approx`, which is not great mostly because it sets an incorrect example. Refactoring should be straightforward (possibly easily done with AI). I added examples of using the newer matchers in my #480 PR.

laylagi:
This change removes usage of deprecated `Catch2::Approx` in favor of the more explicit `WithinAbs`/`WithinRel` matchers. 

Closes #486.

### DIFF
--- a/test/unit/test-config.cpp
+++ b/test/unit/test-config.cpp
@@ -133,6 +133,8 @@ TEST_CASE("Config Driven Solver", "[config][Serial]")
   auto jsonstream = PreprocessFile(filename.c_str());  // Apply custom palace json
   auto config = json::parse(jsonstream);
 
+  using namespace Catch::Matchers;
+
   constexpr double delta_eps = 1.0e-9;  // Precision in frequency comparisons (Hz)
   {
     auto sample_f = std::vector{0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1};
@@ -144,8 +146,7 @@ TEST_CASE("Config Driven Solver", "[config][Serial]")
 
       for (size_t i = 0; i < sample_f.size(); ++i)
       {
-        CHECK_THAT(driven_solver.sample_f[i],
-                   Catch::Matchers::WithinAbs(sample_f[i], delta_eps));
+        CHECK_THAT(driven_solver.sample_f[i], WithinAbs(sample_f[i], delta_eps));
       }
       CHECK(driven_solver.save_indices == save_indices);
       CHECK(driven_solver.prom_indices == std::vector{0, sample_f.size() - 1});
@@ -157,8 +158,7 @@ TEST_CASE("Config Driven Solver", "[config][Serial]")
 
       for (size_t i = 0; i < sample_f.size(); ++i)
       {
-        CHECK_THAT(driven_solver.sample_f[i],
-                   Catch::Matchers::WithinAbs(sample_f[i], delta_eps));
+        CHECK_THAT(driven_solver.sample_f[i], WithinAbs(sample_f[i], delta_eps));
       }
       CHECK(driven_solver.save_indices == save_indices);
       CHECK(driven_solver.prom_indices == std::vector{0, sample_f.size() - 1});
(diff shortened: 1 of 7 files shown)


test/unit/test-config.cpp
```diff
@@ -133,8 +133,6 @@ TEST_CASE("Config Driven Solver", "[config][Serial]")
   auto jsonstream = PreprocessFile(filename.c_str());  // Apply custom palace json
   auto config = json::parse(jsonstream);
 
-  using namespace Catch::Matchers;
```
Sbozzolo, line 136: Why did you remove this? Aren't the tests cleaner with `using namespace`?

Sbozzolo (commented):
Thank you! It looks good. I checked that no more `Approx`s are left and that the tests pass.

I think we can make the tests a little bit easier on the eye with the namespace. Is there a reason why you removed the `using`?

test/unit/test-config.cpp
laylagi: You are right, this can make the code more cluttered. I personally like to make dependancies more visible, but I am open to declaring the namespace if the team prefers it. I also, wanted to make the function calls consistent across all test files, including `test-vector.cpp` in which the namespace hadn't been declared. That said, I am totally fine with either approaches.

Sbozzolo: I agree with making the dependencies explicit. In this particular case, I chose to hide the namespace because this function is used all the time, so developers would have to become familiar with it quickly, but I am find with keeping it more verbose.

test/unit/test-config.cpp
```diff
@@ -144,7 +142,11 @@ TEST_CASE("Config Driven Solver", "[config][Serial]")
       config::DrivenSolverData driven_solver;
       REQUIRE_NOTHROW(driven_solver.SetUp(*config.find("driven_base_uniform_sample")));
 
-      CHECK_THAT(driven_solver.sample_f, Approx(sample_f).margin(delta_eps));
+      for (size_t i = 0; i < sample_f.size(); ++i)
+      {
+        CHECK_THAT(driven_solver.sample_f[i],
+                   Catch::Matchers::WithinAbs(sample_f[i], delta_eps));
```
hughcars: This is a bit of an annoying feature of these `WithinXXX` methods, not allowing direct vector comparisons anymore. I had a search through the docs and couldn't see anything meaningful, and you end needing something like
```
REQUIRE_THAT(sample_f, Catch::Matchers::RangeEquals(driven_solver.sample_f, [](auto x, auto y){ return std::abs(x-y) < delta_eps;}));
```
which is a mouthful, and breaks with the nice inf supporting comparators they use internally to catch2. 

I've asked a question of the catch2 devs, as I can't find it in the docs. It's plausible it's just not been thought of yet in their rush to deprecate. Additionally, `Approx` with `margin()` shouldn't have the imprecision issues of Approx, based on the page from the original issue, which are mostly to do with asymmetry of relative comparison. I don't think this is blocking, but it would be nice to upgrade to whatever solution the catch2 developers suggest once they reply.

test/unit/test-config.cpp
hughcars: I don't think `Catch::Matchers` at every call site adds much, as after the 10th repetition in quick succession, it's visual noise a developer is just filtering out at each call site. It's not a short namespace like `std`, and it often results in line breaks which increases the vertical scan distance when reading. Avoid `using namespace std;` for instance isn't so much about explicit naming, as avoiding argument dependent lookup confusions when there are overloads. 

If having it explicit is something you really want, you could use `using cm = Catch::Matchers;` but I don't think that is really any clearer than just having the `using namespace Catch::Matchers;` at the top.

laylagi: Thanks, I added `using namespace Catch::Matchers;` in [3cab1f6](https://github.com/awslabs/palace/pull/500/commits/3cab1f6a1bd9fa2c5f160dfdf0fd2650f3a763e8).

PR #493, opened 2025-09-25
Support GPUs in singularity builds

hughcars:
Addresses https://github.com/awslabs/palace/issues/489

PR #502, opened 2025-09-30
[WIP] cuda singularity build

mdmaas:
In case this helps, I got the singularity build to compile with CUDA after adding automatic search for the CUDA compiler and other suggestions by Claude code.


mdmaas:
Hmmm, the compile process seems to complete succesfully, but then I can't find the palace executable, so there is probably something wrong here.

mdmaas:
Any luck @nikosavola with the original version from #493 ?

PR #504, opened 2025-09-30
Update wave port documentation

simlapointe:
Update the wave port documentation to clarify how boundary conditions from the full 3D model are mapped to PEC and PMC in the 2D wave port eigenproblem. Also update a few places in the docs stating that frequency-dependent boundary conditions are only available for driven simulations (they are now available for both driven and eigenmode).


Sbozzolo:
The tests are not being started because of this: https://github.com/awslabs/palace/pull/435

Also, i think waveports might be another great candidate for the upcoming _Features_ section of the documentation, where we can provide a more complete and pedagogical discussion

PR #510, opened 2025-10-01
Change sign of Floquet phase factor

simlapointe:
The electric field transformation for Floquet boundary conditions was previously $E= E_p e^{+i k \cdot x}$. This changes it to $E = E_p e^{-ik \cdot x}$, which is preferable given our phasor definition $E(t) = Ee^{+i\omega t}$.


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ The format of this changelog is based on
     use in a non-amr simulation. [Issue 444](https://github.com/awslabs/palace/issues/444).
   - Fixed bug where `"CrackInternalBoundaryElements"` would result in incorrect
     results for some lumped port boundary conditions. [Issue 501](https://github.com/awslabs/palace/issues/501).
+  - Changed the sign of the Floquet phase factor from $\exp(+ik\cdot x)$ to $\exp(-ik\cdot x)$.
```
hughcars: Add the PR number and link, once the farfield gets merged the changelog will be using that style.

hughcars (approved):
LGTM. Tiny thing to fix the CHANGELOG entry. I'm pretty convinced you caught all the sign flips.

Would be nice if there was a unit test with a non-k symmetric system matrix, but I don't think that should block a correctness change like this.

PR #514, opened 2025-10-02
Use spack cache for palace in doc

hughcars:
We can allow usage of the spack cache for the palace install during doc generation, useful for cases where only markdown is changed, where the previous builds are probably fine.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -106,9 +106,7 @@ jobs:
         # Explicitly not using cache to get latest develop
       - name: Build Palace
         run: |
-          spack -e . install --only-concrete --show-log-on-error --only package --keep-stage --no-cache
-        # If you want to use the branch-local source instead (should we always do this?)
-        # spack -e . develop --path=$(pwd) local.palace@git."${{ github.head_ref || github.ref_name }}"=develop
+          spack -e . install --only-concrete --show-log-on-error --only package --keep-stage
 
       - name: Build and deploy
         env:


.github/workflows/docs.yml
```diff
@@ -106,9 +106,7 @@ jobs:
         # Explicitly not using cache to get latest develop
       - name: Build Palace
         run: |
-          spack -e . install --only-concrete --show-log-on-error --only package --keep-stage --no-cache
-        # If you want to use the branch-local source instead (should we always do this?)
-        # spack -e . develop --path=$(pwd) local.palace@git."${{ github.head_ref || github.ref_name }}"=develop
+          spack -e . install --only-concrete --show-log-on-error --only package --keep-stage
```
cameronrutherford, line 109: Above we run `spack develop`. I don't fully remember how this interaction works, but 2 runs of the pipeline (perhaps 3 with a source code modification) can verify.

My understanding of how this will work if we get a cache hit:

1. On first build, we wouldn't get a cache hit since it's a new branch with the reference `palace@git.branchname=develop`
2. On second build, we would get a cache hit, as the spack ref is identical
3. If new source code modifications are then made, we continue to get cache hits, **not grabbing new source code** when re-building palace.

This means if someone is relying on source code modifications in order to update docs, then it would be impossible for them to test bugfixes / small modifications to things and have that reflect in the docs (modifications to docs code would work...).

Alternatively, I am vaguely remembering that spack might use the SHA of the source code w/ `spack develop` in order to generate the Spack SHA of the concrete spec. If that memory is correct, then we will never get a cache hit unless a pipeline is triggered with the same source code.

So, there are three logical approaches I see:

1. Close PR, and live with re-building Palace for each doc pipeline.
2. Live with some weird behavior here where you only cache one build per branch.
3. Instead, don't run `spack develop`, and just build something like `palace@git.<LATEST_MAIN_SHA>=develop`, and get cache hits consistently for each commit in develop.

I initially wanted to move towards 3, but then this means one won't see any local branch changes that would affect docs until after a PR is merged.

If that trade-off is desired, then lets go with 3.

Otherwise, lets go with 1.

PR #517, opened 2025-10-07
CPU Profiling: Update Developer Notes

laylagi:
This PR adds documentation on Palace CPU profiling with Intel VTune.


docs/src/developer/notes.md
```diff
@@ -124,6 +124,58 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
+## Profiling *Palace* on CPUs
+
+For CPU profiling we use [Intel VTune Profiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html) as it supports profiling JIT code generated by [libxsmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_prof/), which is Palace's most performant libCEED CPU backend.
+
+### Requirements
+
+ 1. Verify your system meets the [VTune system requirements](https://www.intel.com/content/www/us/en/developer/articles/system-requirements/vtune-profiler/2025-1.html)
+ 2. [Install VTune Profiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler-download.html)
+ 3. Set VTune's root directory during Palace build
+ 4. Use `RelWithDebInfo` build type for optimal profiling results
```
cameronrutherford, line 141: Should we include a note / suggestion on a `spack.yaml` configuration that would help with this?

laylagi: Like in [.github/workflows/spack.yml](https://github.com/awslabs/palace/blob/main/.github/workflows/spack.yml)?

cameronrutherford: This would potentially be more applicable if we were using HPCToolkit, but I think it applies here too. Something like this is necessary if you want to get all the packages building with the right symbols:

```YAML
spack:
  specs:
  - palace@develop # are we missing variants here that are essential?
  packages:
    all:
      require:
        - "build_type=RelWithDebInfo"
    petsc:
      require:
        - +debug
    libxsmm:
      require:
        - +debug
    hypre:
      require:
        - +debug
```

Of course there are always many ways to do things with Spack, but I think giving some general guidance as @Sbozzolo has done in the user guide is sensible here. While we could include way more information + a full script to do something like this, I think a snippet like this would be useful as a minimum pointer for users.

Then we would also likely have to include something about using the binary `palacex86_64.bin` (or whatever it's called) instead of the palace wrapper that would be on the path... I see that this is in code below, but it might be worth making explicit.

laylagi: Ah, I see your point now. That makes sense. However, turning on `debug` for libxsmm disables optimization. 
Screenshot 2025-10-13 at 9 49 14 AM

I'm rebuilding palace with spack without the `debug` variants for libxsmm, petsc, and hypre and see if we can get a full profile.

laylagi: >Then we would also likely have to include something about using the binary `palacex86_64.bin`

Yes, that's a good point.

cameronrutherford: Hmmm well I guess it's up to you what to include as the recommended build for profiling, but I think some elementary suggestions to users would be very helpful to at least allow someone to reproduce some of the examples you give in the docs.

It might also be good to include specific potential issues (like the one in your screenshot), but that's up to your discretion.

docs/src/developer/notes.md
```diff
@@ -124,6 +124,60 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
 
+## Profiling *Palace* on CPUs
+
+For CPU profiling we use [Intel VTune Profiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html) as it supports profiling JIT code generated by [libxsmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_prof/), which is Palace's most performant libCEED CPU backend.
+
+### Requirements
+
+ 1. Verify your system meets the [VTune system requirements](https://www.intel.com/content/www/us/en/developer/articles/system-requirements/vtune-profiler/2025-1.html)
+ 2. [Install VTune Profiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler-download.html)
+ 3. Set VTune's root directory during Palace build
```
Sbozzolo:
```suggestion
 3. Set VTune's root directory during Palace build (see below)
```

docs/src/developer/notes.md
```diff
@@ -124,6 +124,60 @@ Disk IO                         // < Disk read/write time for loading the mesh f
 Total                           // < Total simulation time
 ```
 
+## Profiling *Palace* on CPUs
+
+For CPU profiling we use [Intel VTune Profiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html) as it supports profiling JIT code generated by [libxsmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_prof/), which is Palace's most performant libCEED CPU backend.
```
Sbozzolo: The way I would personally phrase this is something along these lines (to give more background for readers who are not familiar on why we need to use vtune specifically):

```
A typical *Palace* simulation spends most of its time in libCEED kernels, which, in turn, executed `libsxmm` code. Libsxmm generates code just-in-time to ensure it is the most performant on the given architecture and for the given problem. This code generation confuses most profilers. Luckily, [libsxmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_prof/) can integrate with the VTune APIs to enable profiling of jitted functions as well. 

!!! note Using a different backend?

   The following notes assume that you are using the `libsxmm` backend for libCEED. If this is not the case and you have a preference for other profilers,such as [perf](link) or [HPCToolkit](link)), you can consider using other them as well
```

docs/src/developer/notes.md
```diff
@@ -124,6 +124,60 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
+### Running a Basic Profile
+
+A flamegraph can be created using either:
+
+  - The VTune graphical user interface
+  - The command line interface
+
+Follow [Intel's official instructions](https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2024-2/launch.html) for detailed usage information.
+
+Below is a sample flamegraph resulted from profiling the rings example located at [`/examples/rings/rings.json`](https://github.com/awslabs/palace/blob/main/examples/rings/rings.json) with polynomial order P=6.
```
Sbozzolo:
```suggestion
Below is a sample flamegraph resulted from profiling the rings example located at [`/examples/rings/rings.json`](https://github.com/awslabs/palace/blob/main/examples/rings/rings.json) with polynomial order P=6 (`Solver["Order"]: 6`).
```
(or what is the correct setting)

docs/src/developer/notes.md
```diff
@@ -124,6 +124,62 @@ Disk IO                         // < Disk read/write time for loading the mesh f
 Total                           // < Total simulation time
 ```
 
+## Profiling *Palace* on CPUs
+
+A typical *Palace* simulation spends most of its time in libCEED kernels, which, in turn, executed `libsxmm` code on CPUs. Libsxmm generates code just-in-time to ensure it is the most performant on the given architecture and for the given problem. This code generation confuses most profilers. Luckily, [libsxmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_prof/) can integrate with the VTune APIs to enable profiling of jitted functions as well.
+
+**Note!** Using a different backend?
+
+> The following notes assume that you are using the `libxsmm` backend for libCEED. If this is not the case and you have a preference for other profilers, such as [perf](https://perfwiki.github.io/main/) or [HPCToolkit](https://hpctoolkit.org/), you can consider using them as well.
```
Sbozzolo: This is rendered with Documenter.jl, so you can use [admonitions](https://documenter.juliadocs.org/stable/showcase/#Note-admonition)


```suggestion
!!!note Using a different libCEED backend?

    The following notes assume that you are using the `libxsmm` backend for libCEED. If this is not the case and you have a preference for other profilers, such as [perf](https://perfwiki.github.io/main/) or [HPCToolkit](https://hpctoolkit.org/), you can consider using them as well.
```

laylagi: Ha, it didn't render that way locally.

docs/src/developer/notes.md
```diff
@@ -124,6 +124,62 @@ Disk IO                         // < Disk read/write time for loading the mesh f
 Total                           // < Total simulation time
 ```
 
+## Profiling *Palace* on CPUs
+
+A typical *Palace* simulation spends most of its time in libCEED kernels, which, in turn, executed `libsxmm` code on CPUs. Libsxmm generates code just-in-time to ensure it is the most performant on the given architecture and for the given problem. This code generation confuses most profilers. Luckily, [libsxmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_prof/) can integrate with the VTune APIs to enable profiling of jitted functions as well.
+
+!!! note Using a different libCEED backend?
```
Sbozzolo:
```suggestion
!!! note "Using a different libCEED backend?"
```

docs/src/developer/notes.md
laylagi: I am having trouble re-installing Palace with Spack without those debug symbols. The SuperLU installation fails probably due to version mismatches. I'll merge this now and include the Spack notes in a separate PR, and probably in a tutorial page, when I can confirm what configuration is optimal. Thanks!

PR #518, opened 2025-10-08
Export field solutions as MFEM grid functions

simlapointe:
Adding an option (false by default) to export fields as MFEM grid functions to visualize with GLVis in addition to or instead of ParaView. This can be enabled by setting `config["Problem"]["OutputFormats"]["GridFunction"]` to `true`. The mesh and fields are saved in the `gridfunction` folder within the output folder and can be visualized with GLVis as follows (for the spheres example with a single MPI process):

`GLVis -m postpro/gridfunction/electrostatic/mesh -g postpro/gridfunction/electrostatic/V_000001.gf -np 1`

Only 3D volume fields are currently being saved in this format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,5 +121,5 @@ Computing (CQC). Please contact the development team at
 This project is licensed under the [Apache-2.0 License](./LICENSE).
 
 This project contains a copy of [GLVis-js](https://github.com/GLVis/glvis-js),
-developed by NOTICE. See [NOTICE](docs/assets/js/glvis/NOTICE) and
-[NOTICE](docs/assets/js/glvis/LICENSE).
+developed by LLNL. See the corresponding [NOTICE](docs/assets/js/glvis/NOTICE) and
+[LICENSE](docs/assets/js/glvis/LICENSE).
--- a/docs/src/assets/js/glvis/LICENSE
+++ b/docs/src/assets/js/glvis/LICENSE
@@ -1,3 +1,5 @@
+// This file is a copy of the [GLVis-js](https://github.com/GLVis/glvis-js) license.
+
 // BSD 3-Clause License
 
 // Copyright (c) 2010-2024, Lawrence Livermore National Security, LLC
--- a/docs/src/assets/js/glvis/NOTICE
+++ b/docs/src/assets/js/glvis/NOTICE
@@ -1,3 +1,5 @@
+// This file is a copy of the [GLVis-js](https://github.com/GLVis/glvis-js) notice.
+
 // This work was produced under the auspices of the U.S. Department of Energy by
 // Lawrence Livermore National Laboratory under Contract DE-AC52-07NA27344.
 


palace/models/postoperator.hpp
```diff
@@ -93,26 +93,65 @@ class PostOperator
...
+  {
+    return enable_paraview_output && should_write_fields(step);
+  }
+  bool should_write_gridfunction_fields() const
+  {
+    return enable_gridfunction_output && anything_to_save();
+  }
+  bool should_write_gridfunction_fields(std::size_t step) const
+  {
+    return enable_gridfunction_output && should_write_fields(step);
```
Sbozzolo: Methods should be `TitleCase` and comments should end with `.`.

palace/models/postoperator.hpp
```diff
@@ -130,6 +169,7 @@ class PostOperator
   };
   std::map<int, WavePortFieldData> port_E0;
 
+  void SetupFieldCoefficients();
```
Sbozzolo, line 173: Can you add a dosctring to this function?

palace/models/postoperator.hpp
```diff
@@ -373,6 +415,9 @@ class PostOperator
     return *A;
   }
 
+  // Access to number of padding digits
```
Sbozzolo:
```suggestion
  // Access to number of padding digits.
```

palace/models/postoperator.hpp
```diff
@@ -373,6 +415,9 @@ class PostOperator
...
+  const auto GetPadDigitsDefault() const { return pad_digits_default; }
```
Sbozzolo:
```suggestion
  constexpr auto GetPadDigitsDefault() const { return pad_digits_default; }
```
Given that `pad_digits_default` is known at compile time

palace/utils/configfile.cpp
```diff
@@ -340,6 +340,15 @@ void ProblemData::SetUp(json &config)
   verbose = problem->value("Verbose", verbose);
   output = problem->value("Output", output);
 
+  // Parse output formats
```
Sbozzolo:
```suggestion
  // Parse output formats.
```

test/unit/test-postoperator.cpp
```diff
@@ -337,3 +340,192 @@ TEST_CASE("PostOperator", "[idempotent][Serial]")
...
+      for (const auto field : expected_fields)
+      {
+        std::string filename =
+            fmt::format("{}_{:0{}d}.gf.{:0{}d}", field, 0, pad_digits, i, pad_digits);
+        CHECK(fs::exists(fs::path(iodata.problem.output) / "gridfunction" /
+                         "electrostatic" / filename));
+        CHECK(!fs::is_empty(fs::path(iodata.problem.output) / "gridfunction" /
+                            "electrostatic" / filename));
+      }
+    }
```
Sbozzolo: I think this logic can be moved to a function since it seems to be always the same
```cpp
// Check that the expected files exist and that they are not empty.
auto check_files = [&](const std::string& subdir, int step, int pad_digits,
                         const std::vector<std::string>& fields) {
    for (int i = 0; i < size; i++) {
      for (const auto& field : fields) {
        auto path = fs::path(iodata.problem.output) / "gridfunction" / subdir /
                    fmt::format("{}_{:0{}d}.gf.{:0{}d}", field, step, pad_digits, i, pad_digits);
        CHECK(fs::exists(path));
        CHECK(!fs::is_empty(path));
      }
    }
  };



check_files("electrostatic", 0, post_op.GetPadDigitsDefault(), {"E", "V", "U_e"});
```

test/unit/test-postoperator.cpp
```diff
@@ -337,3 +340,192 @@ TEST_CASE("PostOperator", "[idempotent][Serial]")
...
+  iodata.domains.materials.emplace_back().attributes = {1};
+  iodata.boundaries.pec.attributes = {1};
+  iodata.problem.output_formats.gridfunction = true;
+
+  // Setup lumped port boundary data for driven and transient.
+  auto filename = fmt::format("{}/{}", PALACE_TEST_DIR, "config/boundary_configs.json");
+  auto jsonstream = PreprocessFile(filename.c_str());  // Apply custom palace json
+  auto config = json::parse(jsonstream);
+  config::BoundaryData boundary_port;
+  REQUIRE_NOTHROW(boundary_port.SetUp(*config.find("boundaries_lumped_port_X_2")));
```
Sbozzolo, line 358: Ideally, this should not be needed for your test, right? The feature does not care if we have ports or not.

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
   Mpi::Barrier(fem_op->GetComm());
 }
 
+template <ProblemType solver_t>
+void PostOperator<solver_t>::WriteMFEMGridFunctions(double time, int step)
+{
+  BlockTimer bt(Timer::IO);
+
+  // Create output directory if it doesn't exist
+  fs::create_directories(gridfunction_output_dir);
```
Sbozzolo: This is a filesystem operation, so I think it's best not execute it on each MPI rank to avoid race conditions. (If you look at how `create_directories` is currently used, you'll see that it is protected by a `if (Mpi::Root(comm))`).

See how this is done in `SaveIteration`, because one has to be a little careful.

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+
+  // Create grid function for scalar coefficients.
+  mfem::H1_FECollection h1_fec(fespace.GetMaxElementOrder(), mesh.Dimension());
+  mfem::ParFiniteElementSpace h1_fespace(&mesh, &h1_fec);
+  mfem::ParGridFunction gridfunc_scalar(&h1_fespace);
+
+  const int local_rank = mesh.GetMyRank();
+
+  // Write grid functions using MFEM's built-in Save method
+  // Use 6-digit padding to match MFEM's pad_digits_default
```
Sbozzolo:
```suggestion
  // Write grid functions using MFEM's built-in Save method.
  // Use 6-digit padding to match MFEM's pad_digits_default.
```

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+  if constexpr (HasEGridFunction<solver_t>())
+  {
+    if (E)
+    {
+      if constexpr (HasComplexGridFunction<solver_t>())
```
Sbozzolo, line 565: There's a lot of repetition in this function, can we maybe extract some of the common functions?

Something like:
```cpp
void write_field(const auto& field, const std::string& name, const std::string& gridfunctions_output_dir) {
  if (field) {
    auto path = fs::path(gridfunction_output_dir) / 
                fmt::format("{}_{:0{}d}.gf.{:0{}d}", name, step, pad_digits_default,
                           local_rank, pad_digits_default);
    std::ofstream file(path);
    field->Save(file);
  }
};
```
Possibly also handling the real/complex components.

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+
+  auto mesh_Lc0 = units.GetMeshLengthRelativeScale();
+
+  // Write the mesh partitioning and (optionally) error indicators at the final step.
+  // Write the mesh coordinates in the same units as originally input.
+  mfem::ParMesh &mesh = E ? *E->ParFESpace()->GetParMesh() : *B->ParFESpace()->GetParMesh();
+  mesh::DimensionalizeMesh(mesh, mesh_Lc0);
+
+  // Create output directory if it doesn't exist
+  fs::create_directories(gridfunction_output_dir);
```
Sbozzolo: Same comment as above regarding MPI safety

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+
+  // Create piecewise constant finite element space for rank and error indicator
+  mfem::L2_FECollection pwconst_fec(0, mesh.Dimension());
+  mfem::FiniteElementSpace pwconst_fespace(&mesh, &pwconst_fec);
+
+  const int local_rank = mesh.GetMyRank();
+
+  // Write mesh partitioning (rank information)
```
Sbozzolo:
```suggestion
  // Write mesh partitioning (rank information).
```

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+    mfem::GridFunction rank(&pwconst_fespace);
+    rank = local_rank + 1;
+
+    fs::path rank_filename = fs::path(gridfunction_output_dir) /
+                             fmt::format("rank.gf.{:0{}d}", local_rank, pad_digits_default);
+    std::ofstream rank_file(rank_filename);
+    rank.Save(rank_file);
+  }
+
+  // Write error indicator if provided
```
Sbozzolo:
```suggestion
  // Write error indicator if provided.
```

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+
+  // Write the mesh partitioning and (optionally) error indicators at the final step.
+  // Write the mesh coordinates in the same units as originally input.
+  mfem::ParMesh &mesh = E ? *E->ParFESpace()->GetParMesh() : *B->ParFESpace()->GetParMesh();
+  mesh::DimensionalizeMesh(mesh, mesh_Lc0);
+
+  // Create output directory if it doesn't exist
+  fs::create_directories(gridfunction_output_dir);
+
+  // Create piecewise constant finite element space for rank and error indicator
```
Sbozzolo:
```suggestion
  // Create piecewise constant finite element space for rank and error indicator.
```

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+    }
+  }
+
+  if constexpr (HasBGridFunction<solver_t>())
+  {
+    if (B)
+    {
+      if constexpr (HasComplexGridFunction<solver_t>())
+      {
+        // Write real and imaginary parts separately
```
Sbozzolo:
```suggestion
        // Write real and imaginary parts separately.
```

palace/models/postoperator.cpp
```diff
@@ -519,6 +516,222 @@ void PostOperator<solver_t>::WriteFieldsFinal(const ErrorIndicator *indicator)
...
+
+        std::ofstream b_real_file(b_real_filename);
+        std::ofstream b_imag_file(b_imag_filename);
+
+        B->Real().Save(b_real_file);
+        B->Imag().Save(b_imag_file);
+      }
+      else
+      {
+        // Write real part only
```
Sbozzolo:
```suggestion
        // Write real part only.
```

test/unit/test-postoperator.cpp
simlapointe: I need something like this because driven and transient simulations require at least one excited port. The `SpaceOperator` construction will error out otherwise.

palace/models/postoperator.hpp
```diff
@@ -93,26 +93,65 @@ class PostOperator
...
 
-  // ParaView Measure & Print.
-
-  // Option to write ParaView fields at all and rate / number of iterations printed.
-  std::size_t paraview_delta_post = 0;  // printing rate for ParaView (TRANSIENT)
-  std::size_t paraview_n_post = 0;      // max printing for ParaView (OTHER SOLVERS)
-  std::vector<std::size_t> paraview_save_indices = {};  // explicit saves for ParaView
-  // Whether any paraview fields will be written.
-  bool WriteParaviewFields() const
+  // Field output format control flags
```
Sbozzolo:
```suggestion
  // Field output format control flags.
```

palace/models/postoperator.hpp
```diff
@@ -93,26 +93,65 @@ class PostOperator
...
+  bool enable_paraview_output = false;
+  bool enable_gridfunction_output = false;
+
+  // How many / which fields to output
```
Sbozzolo:
```suggestion
  // How many / which fields to output.
```

palace/models/postoperator.hpp
```diff
@@ -93,26 +93,65 @@ class PostOperator
...
+  std::size_t output_delta_post = 0;                  // printing rate (TRANSIENT)
+  std::size_t output_n_post = 0;                      // max printing (OTHER SOLVERS)
+  std::vector<std::size_t> output_save_indices = {};  // explicit saves
```
Sbozzolo: Why did you use `size_t` here? These integers are not being used to index inside an array, so I would have personally used `int` (or `unsigned int`)

test/unit/test-postoperator.cpp
Sbozzolo: Could you open an issue about this?

docs/src/guide/postprocessing.md
```diff
@@ -93,12 +93,16 @@ file. These include:
...
+solutions are written to disk for 3D visualization with [ParaView](https://www.paraview.org/)
+or [GLVis](https://glvis.org/). Various other postprocessed fields are also written to the ParaView
+or grid function (GLVis) database as available, including electric and magnetic energy density,
+surface currents, and charge density. These files are found in the `paraview/` or `gridfunction/`
+directories located in the output directory specified under
+[`config["Problem"]["Output"]`](../config/problem.md#config%5B%22Problem%22%5D). The output
+formats are specified in [`config["Problem"]["OutputFormats"]`](../config/problem.md#config%5B%22Problem%22%5D).
+
+ParaView is recommended to visualize large simulations in parallel. The grid function (GLVis)
+format can be useful to embed visualizations in webpages with its Javascript version.
```
Sbozzolo:
```suggestion
format can be useful to embed visualizations in webpages with its [Javascript version](https://github.com/GLVis/glvis-js/).
```

PR #519, opened 2025-10-08
Update hypre, MFEM, PETSC, and SLEPc

### LINKED ISSUE #498: Update hypre to 3.0.0+ to fix compatibility with CUDA 13

Sbozzolo:
The tagged commit for `hypre` is incompatible with CUDA 13, so that Palace does not compile with the latest version of the CUDA toolkit.

This commit updates a few packages to their latest versions and adds a new dependency on Umpire. Umpire is strongly recommended for hypre 3 with CUDA. I also switch to using CMake instead of autotools in hypre and not use the internal hypre device manager.

Similar to magma, umpire is automatically added to the build when buidling with CUDA or HIP.

Closes #498, #576
Spack build in docs depends on https://github.com/spack/spack-packages/pull/1855 (or on pushing the package.py to spack-packages)

### DIFF
--- a/spack_repo/local/packages/palace/package.py
+++ b/spack_repo/local/packages/palace/package.py
@@ -258,7 +258,9 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
                 )
             )
 
-        palace_with_gpu_aware_mpi = any(self.spec.satisfies(f"{var}+cuda") for var in ["openmpi", "mpich", "mvapich-plus"])
+        palace_with_gpu_aware_mpi = any(self.spec.satisfies(f"{var}+cuda") or
+                                        self.spec.satisfies(f"{var}+rocm")
+                                        for var in ["openmpi", "mpich", "mvapich-plus"])
 
         args.append(self.define("PALACE_WITH_GPU_AWARE_MPI", palace_with_gpu_aware_mpi))
 


spack_repo/local/packages/palace/package.py
```diff
@@ -17,7 +17,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     homepage = "https://github.com/awslabs/palace"
     git = "https://github.com/awslabs/palace.git"
 
-    maintainers("hughcars", "simlap", "cameronrutherford")
+    maintainers("hughcars", "simlap", "cameronrutherford", "sbozzolo")
```
cameronrutherford, line 21: Big +1 on this - I think you're as qualified as the rest of us.

spack_repo/local/packages/palace/package.py
```diff
@@ -131,6 +131,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("metis~int64", when="~int64")
 
     conflicts("^hypre+int64", msg="Palace uses HYPRE's mixedint option for 64 bit integers")
+    depends_on("hypre@:2", when="@:0.14.0")  # MFEM 4.8 is incompatible with hypre v3+
+    depends_on("hypre@3:")
```
cameronrutherford, line 135:
```suggestion
    depends_on("hypre@3:", when="@0.15.0")
```

I am just basing this on what RAJA does for example - https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/raja/package.py#L233

I don't think functionally this is necessary, but it's nice to read and perhaps easier to maintain.

I don't know if `0.15.0` would be valid right at the moment in terms of syntax, so perhaps this should be simply `when="@develop"` for the moment, and then we change this in the `0.15.0` release PR.

spack_repo/local/packages/palace/package.py
```diff
@@ -186,7 +188,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("+cuda"):
         for arch in CudaPackage.cuda_arch_values:
             cuda_variant = f"+cuda cuda_arch={arch}"
-            depends_on(f"hypre{cuda_variant}", when=f"{cuda_variant}")
+            depends_on(f"umpire{cuda_variant}", when=f"{cuda_variant}")
+            depends_on(f"hypre+umpire{cuda_variant}", when=f"{cuda_variant}")
```
cameronrutherford, line 199: Commenting this without fully reading the rest of the PR...

This is to say that we only `depend_on` Umpire when building with GPUs (CUDA or ROCm)?

cmake/ExternalGitTags.cmake
```diff
@@ -284,4 +284,18 @@ set(EXTERN_SUNDIALS_GIT_BRANCH
...
+  "URL for external UMPIRE build"
+)
+set(EXTERN_UMPIRE_GIT_BRANCH
+  "develop" CACHE STRING
+  "Git branch for external UMPIRE build"
+)
+set(EXTERN_UMPIRE_GIT_TAG
+  "ab81a4f22bc2967ee9dd3e2716cafa3f1587e7b9" CACHE STRING
+  "Git tag for external UMPIRE build"
+)
```
cameronrutherford, line 301: something something no new line at the end of the file...

I was going to nitpick about CMake formatting, but then I realized we don't have a CMake formatter running in CI? 🤔

Is this something we want / have locally that we run? If not, I would be curious to hear about why/why not.

extern/patch/umpire/cuda_compatibility.diff
cameronrutherford, line 1: I think this was merged into BLT? https://github.com/LLNL/blt/pull/735

docs/src/developer/notes.md
```diff
@@ -124,6 +124,26 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
+```bash
+spack repo list
+```
+
+Verify that *Palace* is found at the right place:
+
+```
+spack location -p palace
+```
+
```
cameronrutherford: I think you might want to add something about our repo name and needing to specify `local.palace` instead of just `palace`?

Intuitively I would've thought that `spack location -p palace` would return the Spack builtin, but then `spack location -p local.palace` would return the correct one.

https://spack.readthedocs.io/en/latest/repositories.html#search-order-and-overriding-packages

In reading the above, it seems like it depends on the order of repos in `repos.yml` that determines how it searches for a package with multiple potential definitions.

So, I imagine that `spack repo add` (if you don't have an environment configured) appends to the top of the YAML your new repo, and so it gets prioritized in search order. Which works for your example, but perhaps hides some of the nuance.

cmake/ExternalHYPRE.cmake
cameronrutherford, line 76: NIT: Why don't you put the `if(UMPIRE)` block here instead of duplicating it across the `if(CUDA)` and `if(HIP)` sections?

cameronrutherford: You could also do this with `HYPRE_DEPENDENCIES` for the Umpire addition here. Consolidating would aide readability, but again a NIT.

cmake/ExternalHYPRE.cmake
```diff
@@ -143,6 +146,14 @@ if(PALACE_WITH_CUDA)
     "--enable-cusparse"
     "--enable-device-memory-pool"
   )
+  if(PALACE_WITH_UMPIRE)
+    list(APPEND HYPRE_OPTIONS
+      "--with-umpire"
+      "--with-umpire-include=${CMAKE_INSTALL_PREFIX}/include"
+      "--with-umpire-lib-dirs=${CMAKE_INSTALL_PREFIX}/lib"
```
cameronrutherford: I am struggling to convince myself of this either way, but is this correct for say having a Spack built / external Umpire, and then building Hypre from source within Palace? I guess that would just never happen, and so this is fine...

Perhaps my concern is instead building without spack, where you have a pre-built umpire, but then want to meta-build hypre. Then again, this also sounds like a narrow use-case that would never happen, so this is likely fine. If you're building Palace anyway, and want to let it build Hypre, might as well let it build Umpire too....

cmake/ExternalMFEM.cmake
cameronrutherford, line 44: Similar comment to `ExternalHypre.cmake` - why don't we reduce duplication in the file and move the Umpire snippet.

cmake/ExternalUmpire.cmake
```diff
@@ -0,0 +1,53 @@
...
+#
+# Build Umpire.
+#
+
+set(UMPIRE_OPTIONS ${PALACE_SUPERBUILD_DEFAULT_ARGS})
+list(APPEND UMPIRE_OPTIONS
+  "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
+  "-DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}"
+  "-DENABLE_C=ON"
+  "-DENABLE_FORTRAN=ON"
```
cameronrutherford, line 13: This seems like a build-time fortran dep?

spack_repo/local/packages/palace/package.py
cameronrutherford: Following up on this, perhaps we should modify how we express out build-time fortran dependency based on this new dep...

cmake/ExternalUmpire.cmake
```diff
@@ -0,0 +1,53 @@
...
+    "-DCMAKE_HIP_COMPILER=${CMAKE_HIP_COMPILER}"
+    "-DCMAKE_HIP_ARCHITECTURES=${CMAKE_HIP_ARCHITECTURES}"
+  )
+endif()
+
+# Fix CUDA compatibility in BLT.
+set(UMPIRE_PATCH_FILES
+  # https://github.com/LLNL/blt/pull/735
+  "${CMAKE_SOURCE_DIR}/extern/patch/umpire/cuda_compatibility.diff"
+)
```
cameronrutherford: Is this only necessary when building with CUDA? I imagine not-so in ROCM...

Also this was merged into BLT?

palace/linalg/rap.cpp
```diff
@@ -113,7 +113,7 @@ mfem::HypreParMatrix &ParOperator::ParallelAssemble(bool skip_zeros) const
   if (!use_R)
   {
     const mfem::HypreParMatrix *Rt = test_fespace.Get().Dof_TrueDof_Matrix();
-    RAP = std::make_unique<mfem::HypreParMatrix>(hypre_ParCSRMatrixRAPKT(*Rt, hA, *P, 1),
+    RAP = std::make_unique<mfem::HypreParMatrix>(hypre_ParCSRMatrixRAPKT(*Rt, hA, *P, 1, 1),
```
cameronrutherford, line 116: LGTM 🤷🏻

extern/patch/umpire/cuda_compatibility.diff
Sbozzolo: Yes, but umpire comes with a specific version of `blt`, so we need to wait for umpire to update their vended `blt`.

spack_repo/local/packages/palace/package.py
Sbozzolo: > This is to say that we only depend_on Umpire when building with GPUs (CUDA or ROCm)?

This is correct

cmake/ExternalGitTags.cmake
Sbozzolo: I think my commit added the missing newline? 

As for the CMake formatter, I think we should add it

cmake/ExternalHYPRE.cmake
Sbozzolo: This would not be correct if you had a spack-build umpire, but a palace-built hypre. Your concerns would be valid if we wanted to develop/patch umpire, so we would want to point Palace to a different umpire. However, I think this is already the case for pretty much every dependency in palace.

cmake/ExternalUmpire.cmake
Sbozzolo: Yes, it does not compile if I don't add the flag

cmake/ExternalHYPRE.cmake
Sbozzolo: Indeed

palace/linalg/rap.cpp
Sbozzolo: I discussed this offline with @simlapointe

cmake/ExternalGitTags.cmake
cameronrutherford: See https://github.com/awslabs/palace/issues/521

extern/patch/umpire/cuda_compatibility.diff
cameronrutherford: Understood - sounds like an easy update once we get to that point.

cmake/ExternalUmpire.cmake
cameronrutherford: So can we also reflect this in our spack package somehow? I think we already have this when all the flags are enabled, but perhaps we should say something like `when ^hypre@3, depends_on(fortran, type=build)`...

cmake/ExternalHYPRE.cmake
cameronrutherford: I am resolving this as I think that we can re-visit if someone files an issue wanting to support this.

spack_repo/local/packages/palace/package.py
Sbozzolo: I added `0.14.1`, even though I expect the next version to be 0.15 as well.

Sbozzolo:
> Test-libceed crashes when Palace is compiled with CUDA (even when it is run on CPU)

I looked a little bit into this. I found that the crash occurs at
```
mfem::forall(m + 1, [=] MFEM_HOST_DEVICE(int i) { d_I[i] = d_I_old[i]; });
```
in `fem/libceed/operator.cpp`. The pointers seem valid at that point, so I am not entirely sure what's up.

cmake/ExternalHYPRE.cmake
```diff
@@ -7,161 +7,93 @@
...
-    "--with-cuda"
-    "--with-cuda-home=${CUDAToolkit_LIBRARY_ROOT}"
-    "--enable-curand"
-    "--enable-cusparse"
-    "--enable-device-memory-pool"
+    -DHYPRE_ENABLE_CUDA=ON
+    -DHYPRE_ENABLE_CURAND=ON
+    -DHYPRE_ENABLE_CUSPARSE=ON
+    # https://github.com/hypre-space/hypre/issues/1404
+    -DCUDAToolkit_ROOT=${CUDAToolkit_LIBRARY_ROOT}
```
hughcars, line 81: Looks like this was merged. I suspect it's better to merge this PR, then do a followup with another patch update for everyone post mfem 4.9, and fix the TODO here at the same time.

CMakeLists.txt
```diff
@@ -157,6 +157,13 @@ if(PALACE_WITH_MAGMA AND NOT (PALACE_WITH_CUDA OR PALACE_WITH_HIP))
   )
 endif()
 
+# Umpire is only for GPU builds.
+if(PALACE_WITH_CUDA OR PALACE_WITH_HIP)
+  set(PALACE_WITH_UMPIRE ON)
```
hughcars, line 162: Do we want to allow umpire to be disabled by a user when using cuda or hip? I.e set here only if not already set.

cmake/ExternalHYPRE.cmake
```diff
@@ -7,161 +7,93 @@
...
-  string(REPLACE ";" " " HYPRE_MPI_LIBRARY_DIRS "${HYPRE_MPI_LIBRARY_DIRS}")
-  string(REPLACE ";" " " HYPRE_MPI_INCLUDE_DIRS "${MPI_C_INCLUDE_DIRS}")
+# Configure Umpire support
+if(PALACE_WITH_UMPIRE)
   list(APPEND HYPRE_OPTIONS
-    "--with-MPI-libs=${HYPRE_MPI_LIBRARIES}"
-    "--with-MPI-lib-dirs=${HYPRE_MPI_LIBRARY_DIRS}"
-    "--with-MPI-include=${HYPRE_MPI_INCLUDE_DIRS}"
+    -DHYPRE_ENABLE_UMPIRE=ON
+    -Dumpire_DIR=${CMAKE_INSTALL_PREFIX}/lib/cmake/umpire
```
hughcars, line 54: Looking at their `CMakeLists.txt`
```
set_hypre_option(TPL  HYPRE_ENABLE_UMPIRE_HOST          "Use Umpire Allocator for host memory" OFF)
set_hypre_option(TPL  HYPRE_ENABLE_UMPIRE_PINNED        "Use Umpire Allocator for pinned memory" OFF)
set_hypre_option(TPL  HYPRE_ENABLE_UMPIRE_DEVICE        "Use Umpire Allocator for device memory" OFF)
set_hypre_option(TPL  HYPRE_ENABLE_UMPIRE_UM            "Use Umpire Allocator for unified memory" OFF)
```
I think from your discussion of `host-std`, if we wanted to use it on host we'd need that extra flag set? Also given unified is being disabled above, the model is that we're really only using umpire for device correct? Were any of the issues you were having with the cpu/umpire related to possibly not having umpire configured for host here?

spack_repo/local/packages/palace/package.py
```diff
@@ -131,6 +131,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("metis~int64", when="~int64")
 
     conflicts("^hypre+int64", msg="Palace uses HYPRE's mixedint option for 64 bit integers")
+    depends_on("hypre@:2", when="@:0.14.0")  # MFEM 4.8 is incompatible with hypre v3+
+    depends_on("hypre@3:", when="@0.14.1:")
```
hughcars: I think these need updating to 0.15, additionally I thought `@:0.15.0` would mean up to and including 0.15, and then `@0.16:` would be from 0.16 onwards (and also includes develop). I think the libceed constraint is something similar. Updating the dependencies would mean that the next version cut would be a minor bump anyway, so needn't be 0.15.1 as a patch release wouldn't update dependencies to introduce umpire like this.

hughcars (approved):
A few small questions about the umpire configuration, but LGTM. The git sha for the dependencies are a little out of date now, but I think we should merge this one as close to as is as possible, then do a followup with a further update to remove the various TODO in the hypre/umpire build based on their having been fixed since in hypre. That way we can isolate the changes into smaller pieces (and it'll have the benefits of all the CI changes on the way).

cmake/ExternalHYPRE.cmake
Sbozzolo: Great find. Let me investigate this.

victorapm: Hi all, I've seen performance regressions with `HYPRE_ENABLE_UMPIRE_HOST=ON`. It's not clear to me yet how to fix it. For now, I'd recommend not using the umpire HOST option

CMakeLists.txt
Sbozzolo: I am inclined to be opinionated here and use what we know (or we being are told) being the best configuration instead of adding more settings.

Sbozzolo:
> I removed the patch for BLT, but now I learned that the spack build does not work for GPU anymore. This will be fixed once umpire and camp update to version 2025.12, which should be imminent (judging from the repos).

Actually, it is BLT that needs to be released.

palace/linalg/vector.hpp
```diff
@@ -36,6 +36,7 @@ class ComplexVector
   ComplexVector(const Vector &yr, const Vector &yi);
 
   // Copy constructor from an array of complex values.
+  // on_dev determines if py is already on the device.
```
hughcars: Grammar nit: specifies rather than determines. It is making a statement about the current state of py, rather than discovering it.

cmake/ExternalHYPRE.cmake
```diff
@@ -7,162 +7,93 @@
...
   list(APPEND HYPRE_OPTIONS
-    "--with-cuda"
-    "--with-cuda-home=${CUDAToolkit_LIBRARY_ROOT}"
-    "--enable-curand"
-    "--enable-cusparse"
-    "--enable-device-memory-pool"
+    -DHYPRE_ENABLE_CUDA=ON
+    -DHYPRE_ENABLE_CURAND=ON
+    -DHYPRE_ENABLE_CUSPARSE=ON
+    # https://github.com/hypre-space/hypre/issues/1404
```
hughcars, line 80: Looks like this issue was closed, but can address this in a later PR.

Sbozzolo: Added to the list of followups: https://github.com/awslabs/palace/issues/585

palace/linalg/vector.hpp
Sbozzolo: Thank you!

cmake/ExternalHYPRE.cmake
```diff
@@ -7,162 +7,93 @@
...
 endif()
 
 # Configure BLAS/LAPACK
 if(NOT "${BLAS_LAPACK_LIBRARIES}" STREQUAL "")
-  string(REPLACE "$<SEMICOLON>" " " HYPRE_BLAS_LAPACK_LIBRARIES "${BLAS_LAPACK_LIBRARIES}")
   list(APPEND HYPRE_OPTIONS
-    "--with-blas-lib=${HYPRE_BLAS_LAPACK_LIBRARIES}"
-    "--with-lapack-lib=${HYPRE_BLAS_LAPACK_LIBRARIES}"
+    -DHYPRE_ENABLE_HYPRE_BLAS=OFF
+    -DHYPRE_ENABLE_HYPRE_LAPACK=OFF
```
cameronrutherford, line 62: This is the same as the default for the Hypre package's `lapack` variant.

This raised two questions for me:

1. Should we have `depends_on("blas")` explicitly in our package? Or is that implied as a transitive dep...
2. Should we still set these variants in our spack package, even though they are the sane defaults? I think if they are crucial to Palace working, then we should, but maybe that's too restrictive and someone might want to have Hypre build it's own blas w/ Spack.

cmake/ExternalHYPRE.cmake
```diff
@@ -7,162 +7,93 @@
...
+    -DTPL_LAPACK_LIBRARIES=${BLAS_LAPACK_LIBRARIES}
   )
 endif()
 
 # Configure GPU support
 if(PALACE_WITH_CUDA OR PALACE_WITH_HIP)
-  list(APPEND HYPRE_OPTIONS
-    "--disable-unified-memory"
-  )
+  list(APPEND HYPRE_OPTIONS -DHYPRE_ENABLE_UNIFIED_MEMORY=OFF)
```
cameronrutherford, line 70: This is a Spack variant that we do not configure `unified-memory` - I think we should set this in our package.py?

cmake/ExternalHYPRE.cmake
```diff
@@ -7,162 +7,93 @@
...
   if(PALACE_WITH_GPU_AWARE_MPI)
-    list(APPEND HYPRE_OPTIONS
-      "--enable-gpu-aware-mpi"
-    )
+    list(APPEND HYPRE_OPTIONS -DHYPRE_ENABLE_GPU_AWARE_MPI=ON)
```
cameronrutherford, line 72: This should be configured + enabled in the palace package.py in the relevant section - `gpu-aware-mpi`.

Additionally there is a `gpu-profiling` variant for Hypre that we should perhaps consider making a note of in our docs for GPU profiling...

palace/CMakeLists.txt
cameronrutherford, line 20: Likely we should spin this off into another issue, but maybe we can support variable C++ standards like Hypre does...

cmake/ExternalHYPRE.cmake
```diff
@@ -7,162 +7,93 @@
...
 endif()
 if(PALACE_WITH_HIP)
   list(APPEND HYPRE_OPTIONS
-    "--with-hip"
-    "ROCM_PATH=${ROCM_DIR}"
-    "--enable-rocrand"
-    "--enable-rocsparse"
+    -DHYPRE_ENABLE_HIP=ON
+    -DHYPRE_ENABLE_ROCRAND=ON
+    -DHYPRE_ENABLE_ROCSPARSE=ON
```
cameronrutherford: Interesting that these are auto-enabled by the spack package, but required here...

spack_repo/local/packages/palace/package.py
cameronrutherford, line 150: See notes above about missing variants / some other things that perhaps we should depend_on even though they are default.

cmake/ExternalHYPRE.cmake
Sbozzolo: Good catch!

cmake/ExternalHYPRE.cmake
Sbozzolo: Indeed, I just checked and they are not required here either.

Sbozzolo: Thank you!

palace/CMakeLists.txt
Sbozzolo: I don't think we need to change this very often, so it's probably easiest to just force it to the version we care about.

cmake/ExternalHYPRE.cmake
cameronrutherford: Same with the CUDA equivalents in case you didn't get to that either.

cmake/ExternalHYPRE.cmake
Sbozzolo: You are right. I added `depends_on("blas")`, `depends_on("lapack")` and updated our package.py to set those variants.

spack_repo/local/packages/palace/package.py
```diff
@@ -132,12 +134,19 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
     depends_on("hypre@3:")
     depends_on("hypre~complex")
+    depends_on("hypre~unified-memory")
     depends_on("hypre+shared", when="+shared")
     depends_on("hypre~shared", when="~shared")
     depends_on("hypre+mixedint", when="+int64")
     depends_on("hypre~mixedint", when="~int64")
     depends_on("hypre+openmp", when="+openmp")
     depends_on("hypre~openmp", when="~openmp")
+    depends_on("hypre+gpu-aware-mpi", when="^mpi+cuda")
```
cameronrutherford: This applies to the `PALACE_WITH_GPU_AWARE_MPI` below as well...

GPU aware MPI doesn't have to be just `cuda`, as `rocm` also supports GPU aware MPI.

I think we implicitly require `+cuda` or `+rocm` in order to support GPU aware MPI(?). If that's the case, we should change how we specify this here, and also include some warning in CMake as it seems currently possible to attempt to enable GPU Aware MPI without having CUDA / ROCm enabled.

cameronrutherford: See https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/beatnik/package.py#L36 for more of an example as well. It seems like only some MPI providers actually have GPU aware MPI support, so you have to be explicit with the providers of different virtuals...

cameronrutherford: https://github.com/spack/spack-packages/blob/98ed8846d60601a7a137d18af233c007d3c1c340/repos/spack_repo/builtin/packages/seissol/package.py#L187 - maybe this is a better example?

Sbozzolo: Thank you

spack_repo/local/packages/palace/package.py
```diff
@@ -243,13 +250,18 @@ def cmake_args(self):
...
         if self.spec.satisfies("+rocm"):
             args.append(
                 self.define(
                     "CMAKE_HIP_ARCHITECTURES", ";".join(self.spec.variants["amdgpu_target"].value)
                 )
             )
 
+        palace_with_gpu_aware_mpi = any(self.spec.satisfies(f"{var}+cuda") for var in ["openmpi", "mpich", "mvapich-plus"])
+
+        args.append(self.define("PALACE_WITH_GPU_AWARE_MPI", palace_with_gpu_aware_mpi))
```
cameronrutherford: What about `rocm`?

Sbozzolo: Oops, thank you!

cameronrutherford (approved):

PR #523, opened 2025-10-10
Enable documenter/deploy workflow, fix filter CI

Sbozzolo:
Documenter.jl, which we use to render to documentation, can provide a useful feature: posting back a status with the link of the rendered documentation. This is particularly useful with PRs: now, after the documentation is generated, you can click on the `documenter/deploy` green workflow and see the documentation.

This is a feature that available by default in Documenter, but it was not working in Palace. To find what I needed to do, I searched for `documenter/deploy` in the Documenter codebase and found one match: https://github.com/JuliaDocs/Documenter.jl/blob/5d2d03af73d54b138ec478280c9ef85bcf0b3ffb/src/deployconfig.jl#L515

Here, I learned that the function was posting a GitHub status in a try-catch block. Then, I cross referenced this with the documentation and found that we needed to enable this additional write permission.

### DIFF
--- a/.github/workflows/build-and-test-linux.yml
+++ b/.github/workflows/build-and-test-linux.yml
@@ -494,6 +494,7 @@ jobs:
           julia --project=test/examples --color=yes test/examples/runtests.jl
 
       - name: Check antenna postprocessing example
+        if: needs.filter.outputs.test == 'true'
         run: |
           julia --project=examples -e 'using Pkg; Pkg.instantiate()'
           julia --project=examples examples/antenna/plot_farfield.jl examples/antenna/postpro/farfield-rE.csv


cmake/ExternalHYPRE.cmake
cameronrutherford, line 1: I don't think this change should be here, but rather in #519?

.github/workflows/docs.yml
```diff
@@ -10,6 +10,7 @@ on:
 permissions:
   packages: write
   contents: write
+  statuses: write  # Used to report documentation build statuses
```
cameronrutherford, line 13: Verified this works as-described. Nice find!

.github/workflows/build-and-test-linux.yml
```diff
@@ -494,6 +494,7 @@ jobs:
           julia --project=test/examples --color=yes test/examples/runtests.jl
 
       - name: Check antenna postprocessing example
+        if: needs.filter.outputs.test == 'true'
```
cameronrutherford, line 497: I assume this is okay....

cmake/ExternalHYPRE.cmake
Sbozzolo: Yes, I pushed to the wrong branch!

.github/workflows/build-and-test-linux.yml
Sbozzolo: We have this in every step, see also #440

PR #525, opened 2025-10-10
Adjust EM constants

phdum-a:
Adjusts the definitions of the EM constants to CODATA recommended values, which are minute. Mostly the PR switches definitions, so that we don't require a `sqrt` and the associated "checking for constexpr sqrt" functionality in cmake. Add a tiny unit test to sanity check changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@ The format of this changelog is based on
     505](https://github.com/awslabs/palace/pull/505).
   - Changed the sign of the Floquet phase factor from $\exp(+ik\cdot x)$ to $\exp(-ik\cdot x)$.
     [PR 510](https://github.com/awslabs/palace/pull/510).
+  - Update EM constants to CODATA Recommended Values of the Fundamental Physical Constants 2022
+    [PR 525](https://github.com/awslabs/palace/pull/525).
 
 ## [0.14.0] - 2025-08-20
 


test/unit/test-constants.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+#include "utils/constants.hpp"
+
+using namespace palace;
+using namespace palace::electromagnetics;
+
+TEST_CASE("EM Constant Check", "[units][Serial]")
+{
+  // This test Should be EXACT within double, so no matcher
```
Sbozzolo:
```suggestion
  // This test Should be EXACT within double, so no matcher.
```

palace/utils/constants.hpp
```diff
@@ -12,30 +12,26 @@ namespace palace::electromagnetics
...
-// Permittivity of free space [F/m].
-static constexpr double epsilon0_ = 8.8541878176e-12;
+// Speed of light in free space [m/s].
+// Exact.
+static constexpr double c0_ = 299'792'458;
 
 // Permeability of free space [H/m].
-static constexpr double mu0_ = 4.0e-7 * M_PI;
+// CODATA value with error: 1.256 637 061 27(20) * 10^(−6)
+static constexpr double mu0_ = 1.256'637'061'27e-6;
```
Sbozzolo, line 26: Agrees with https://physics.nist.gov/cgi-bin/cuu/Value?mu0

Sbozzolo: (Commenting just to say that I checked that there was no transcription/copy&paste errors)

PR #526, opened 2025-10-13
Update Floquet BC terms, docs, and regression test

simlapointe:
Make all Floquet terms consistent with the phase factor $E = E_p e^{-ik \cdot x}$ changed in https://github.com/awslabs/palace/pull/510, add the Floquet modified equations to the documentation, and add some field probes to the regression test.


docs/src/reference.md
```diff
@@ -74,6 +74,33 @@ satisfied at all exterior boundaries by the finite element formulation. Impedanc
 boundaries are modeled using a Robin boundary condition with ``\gamma = i\omega/Z_s``, in
 which ``Z_s`` the surface impedance of the boundary, with units of impedance per square.
 
+## Floquet periodic boundary conditions
```
Sbozzolo, line 77: It's not directly related to this PR, but the documentation doesn't explain anywhere what Floquet periodic boundary conditions are, or provide links to references where one can learn about them:
Image

It would be great if you wanted to take this opportunity to add discussion about this feature (especially, with users in mind).

docs/src/reference.md
```diff
@@ -74,6 +74,33 @@ satisfied at all exterior boundaries by the finite element formulation. Impedanc
...
++ i\omega\sigma\bm{E}_p
+- \omega^2\varepsilon_r\bm{E}_p &= 0 \,,\; \bm{x}\in\Omega
+\end{aligned}
+```
+
+and given the electric field solution, the time-harmonic magnetic flux density can be calculated
+as
+
+```math
+\bm{B}_p = -\frac{1}{i\omega}\nabla\times\bm{E}_p + \frac{1}{\omega} \bm{k}_p \times \bm{E}_p \,.
```
Sbozzolo, line 101: I did not check these equations.

test/examples/runtests.jl
```diff
@@ -135,6 +135,30 @@ if "cylinder/waveguide" in cases
 end
 
 if "cylinder/floquet" in cases
+
+    # The phase of the eigenmodes is not stable across architectures, so we
+    # compare the magnitudes.
+    function test_probe_magnitude(new_data, ref_data)
+        num_probes = Int((size(new_data)[2] - 1) / 6)
+        num_modes = min(size(new_data)[1], size(ref_data)[1])
```
Sbozzolo: I think this should not be needed because we already resize the two datasets in `testcase.jl` 
```julia
            data    = CSV.File(joinpath(postprodir, file); header=1) |> DataFrame
            dataref = CSV.File(joinpath(refpostprodir, file); header=1) |> DataFrame
            if !skip_rowcount
                @test nrow(data) == nrow(dataref) || logdump(data, dataref)
            end
            data = data[1:min(nrow(data), nrow(dataref)), :]
```

test/examples/ref/cylinder/floquet/probe-B.csv
```diff
@@ -0,0 +1,16 @@
...
+ 1.40e+01,        -3.435957634383e-10,        +6.252428439847e-11,        -1.983744080230e-10,        +3.608929205000e-11,        -6.854449247823e-11,        -2.050925734520e-09,        -4.204483055819e-10,        +4.231970910882e-10,        -2.427397445462e-10,        +2.443390406601e-10,        -8.739302208999e-11,        -3.798860160629e-09,        -3.763181770481e-10,        -2.059293907786e-11,        -2.172759699352e-10,        -1.188890734239e-11,        -1.287631877887e-10,        -1.664548876520e-09
+ 1.50e+01,        +1.607660822227e-09,        +9.017345132330e-10,        +9.281876080596e-10,        +5.206116485494e-10,        +7.207549335974e-10,        +6.583845075509e-09,        +1.341507456469e-09,        -3.221659309416e-10,        +7.745200954892e-10,        -1.859960748533e-10,        +7.569417159191e-10,        +7.632810142703e-09,        +1.565904271436e-09,        +6.959278264680e-10,        +9.040698220237e-10,        +4.017900485987e-10,        +8.087056522184e-10,        +6.674591097589e-09
```
Sbozzolo: This file does not end with a newline

test/examples/runtests.jl
```diff
@@ -135,6 +135,30 @@ if "cylinder/waveguide" in cases
 end
 
 if "cylinder/floquet" in cases
+
+    # The phase of the eigenmodes is not stable across architectures, so we
+    # compare the magnitudes.
+    function test_probe_magnitude(new_data, ref_data)
+        num_probes = Int((size(new_data)[2] - 1) / 6)
```
Sbozzolo:
```suggestion
        num_probes = Int((size(new_data)[2] - 1) / 6) # 6 components per probe, 1 column with the mode number`
```

docs/src/reference.md
simlapointe: Good point, added a short description to guide/boundaries in [a67067f](https://github.com/awslabs/palace/pull/526/commits/a67067fa351372bb3a3d84ae0ad32c9a84f1ed4b)

PR #530, opened 2025-10-14
Automatically set CMAKE_CUDA_ARCHITECTURES

### LINKED ISSUE #527: Autodetect `CMAKE_CUDA_ARCHITECTURES`
Currently, Palace sets `CMAKE_CUDA_ARCHITECTURES` to 70 when no value is specified. The value of 70 seems pretty arbitrary. I think the more sensible default should be to detect if a GPU is available, and if so, uses the corresponding CUDA_ARCH.

Sbozzolo:
Starting CMake 3.24, CMake now accepts `native` as argument for CMAKE_CUDA_ARCHITECTURES. For CUDA 11.5+, this is passed down to `nvcc`. For previous versions, Cmake automatically detects the native version. Unfortunately, we cannot use this method because `magma` does not support `arch=sm_native`. So, we follow what hypre does and parse the output of `nvidia-smi`. If parsing doesn't work or nvidia-smi is not available, recover the old behavior of setting to 70.

Closes #527


cameronrutherford (approved):
Maybe I am missing something obvious, but LGTM.

I don't know why I haven't seen this added to a project before, but seems like a sensible thing to add.

`70` seems like a sane default as well.

simlapointe (approved):
Looks good! I tried it and was able to build and run on GPU without specifying any cuda architecture flag (and it picked the right architecture number).

Issue #516, opened 2025-10-02
Problems with `test-libCEED` tests on GPU


Sbozzolo:
The default libCEED backend that is picked up when one choose `/gpu/cuda` is `/gpu/cuda/gen`. 

I found that this backend crashes with this error
```
MFEM abort: /data/home/gbozzola/palace/g5/extern/libCEED/backends/cuda-gen/ceed-cuda-gen-operator.c:323 in CeedOperatorApplyAddComposite_Cuda_gen(): CUDA_ERROR_ILLEGAL_ADDRESS
 ... in function: void palace::ceed::_GLOBAL__N__c4d8f0ec_12_operator_cpp_8cc45c7e::CeedAddMult(const std::vector<CeedOperator_private*>&, const std::vector<CeedVector_private*>&, const std::vector<CeedVector_private*>&, const palace::Vector&, palace::Vector&)
 ... in file: /data/home/gbozzola/palace/palace/fem/libceed/operator.cpp:173



CUDA error: (cudaFree(dptr)) failed with error:
 --> an illegal memory access was encountered [code: 700]
 ... in function: void* mfem::CuMemFree(void*)
 ... in file: /data/home/gbozzola/palace/g5/extern/mfem/general/cuda.cpp:86
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
```

In particular, this with this combination
```
  auto mesh = "star-quad.mesh";
  auto amr = true;
  auto order = 2;

  auto bdr_integ = true;
  auto coeff_type = CoeffType::Scalar;
```
in the test for `"H1 Diffusion Integrator"`

In general, I found that tests run on the GPU are much much slower than the same tests on CPU (the libCEED tests take >10 minutes on 1 A10G, the entire palace unit test suite takes less than 3 minutes on 1 CPU).

Sbozzolo:
I investigated the performance issues and found that almost all the time is spent in `PartialAssemble` (each call is about 0.5s).
It is unclear to me where most of the time is spent here. I prepared a nsight profile to look at it and found that pretty much no time is spent on the GPU:
<img width="1688" height="471" alt="Image" src="https://github.com/user-attachments/assets/d4f1e1da-22f3-45f1-ab37-ac9a33b0bc15" />


[...]

PR #531, opened 2025-10-16
Add tutorial on intro to GPU profiling

Sbozzolo:
I was investigating #516 and realized that it was very hard to understand what was happening without NVTX. As I added NVTX for my use case, I took opportunity to document this into a short tutorial to help people get started with looking at GPU profiles.


simlapointe (approved):
Looks good! It would be nice to have a profiling example that's a bit more involved but we can do that later.

PR #532, opened 2025-10-16
Output dimensional fields

simlapointe:
Previously, nondimensionalized fields were used when outputting fields in Paraview or gridfunction (GLVis) formats. This PR instead outputs the dimensionalized fields and updates the documentation to list the units of all output fields. 

Added methods to dimensionalize Vectors and GridFunctions (in-place or not). For coefficients (e.g. `U_e`, `U_m`, `S`, etc.) the scaling is applied through an additional optional function parameter since they cannot be directly scaled.

### DIFF
--- a/palace/fem/coefficient.hpp
+++ b/palace/fem/coefficient.hpp
@@ -49,7 +49,7 @@ protected:
   }
 
 public:
-  BdrGridFunctionCoefficient(const mfem::ParMesh &mesh, const double scaling = 1.0)
+  BdrGridFunctionCoefficient(const mfem::ParMesh &mesh, double scaling = 1.0)
     : mesh(mesh), scaling(scaling)
   {
   }
@@ -91,8 +91,7 @@ private:
 
 public:
   BdrSurfaceCurrentVectorCoefficient(const mfem::ParGridFunction &B,
-                                     const MaterialOperator &mat_op,
-                                     const double scaling = 1.0)
+                                     const MaterialOperator &mat_op, double scaling = 1.0)
     : mfem::VectorCoefficient(B.VectorDim()),
       BdrGridFunctionCoefficient(*B.ParFESpace()->GetParMesh(), scaling), B(B),
       mat_op(mat_op)
@@ -152,7 +151,7 @@ private:
 public:
   BdrSurfaceFluxCoefficient(const mfem::ParGridFunction *E, const mfem::ParGridFunction *B,
                             const MaterialOperator &mat_op, bool two_sided,
-                            const mfem::Vector &x0, const double scaling = 1.0)
+                            const mfem::Vector &x0, double scaling = 1.0)
     : mfem::Coefficient(),
       BdrGridFunctionCoefficient(
           E ? *E->ParFESpace()->GetParMesh() : *B->ParFESpace()->GetParMesh(), scaling),
@@ -467,7 +466,7 @@ private:
 
 public:
   EnergyDensityCoefficient(const GridFunction &U, const MaterialOperator &mat_op,
-                           const double scaling = 1.0)
+                           double scaling = 1.0)
     : mfem::Coefficient(),
       BdrGridFunctionCoefficient(*U.ParFESpace()->GetParMesh(), scaling), U(U),
       mat_op(mat_op)
@@ -565,7 +564,7 @@ private:
 
 public:
   PoyntingVectorCoefficient(const GridFunction &E, const GridFunction &B,
-                            const MaterialOperator &mat_op, const double scaling = 1.0)
+                            const MaterialOperator &mat_op, double scaling = 1.0)
     : mfem::VectorCoefficient(E.VectorDim()),
       BdrGridFunctionCoefficient(*E.ParFESpace()->GetParMesh(), scaling), E(E), B(B),
       mat_op(mat_op)
--- a/palace/models/postoperator.cpp
+++ b/palace/models/postoperator.cpp
@@ -96,6 +96,7 @@ PostOperator<solver_t>::PostOperator(const IoData &iodata, fem_op_t<solver_t> &f
   // Add wave port boundary mode postprocessing, if available.
   if constexpr (std::is_same_v<fem_op_t<solver_t>, SpaceOperator>)
   {
+    // Add scaling factor to output dimensional wave port electric fields.
     const double scaling = units.Dimensionalize<Units::ValueType::FIELD_E>(1.0);
     for (const auto &[idx, data] : fem_op->GetWavePortOp())
     {
@@ -177,8 +178,12 @@ void PostOperator<solver_t>::SetupFieldCoefficients()
 
   if constexpr (HasEGridFunction<solver_t>())
   {
-    // E will be dimensionalized when the coefficient is evaluated, so the scaling
-    // only needs to account for the remaining ε_0 = D / E.
+    // If E is dimensionalized when the coefficients are evaluated, the scaling only needs
+    // to account for the remaining ε_0 = D / E. This assumes ProjectCoefficient(),
+    // ProjectBdrCoefficient(), or paraview->Save() for U_e, Q_sr, and Q_si are always
+    // called after the E GridFunction has been dimensionalized. To output nondimensional
+    // coefficients, omit the scaling argument and make sure E is nondimensional when the
+    // coefficients are evaluated.
     const double scaling = units.Dimensionalize<Units::ValueType::FIELD_D>(1.0) /
                            units.Dimensionalize<Units::ValueType::FIELD_E>(1.0);
 
@@ -203,8 +208,12 @@ void PostOperator<solver_t>::SetupFieldCoefficients()
 
   if constexpr (HasBGridFunction<solver_t>())
   {
-    // B will be dimensionalized when the coefficient is evaluated, so the scaling
-    // only needs to account for the remaining μ⁻¹ = H / B.
+    // If B is dimensionalized when the coefficients are evaluated, the scaling only needs
+    // to account for the remaining μ⁻¹ = H / B. This assumes ProjectCoefficient(),
+    // ProjectBdrCoefficient(), or paraview->Save() for U_m, J_sr, and J_si are always
+    // called after the B GridFunction has been dimensionalized. To output nondimensional
+    // coefficients, omit the scaling argument and make sure B is nondimensional when the
+    // coefficients are evaluated.
     const double scaling = units.Dimensionalize<Units::ValueType::FIELD_H>(1.0) /
                            units.Dimensionalize<Units::ValueType::FIELD_B>(1.0);
 
@@ -232,7 +241,9 @@ void PostOperator<solver_t>::SetupFieldCoefficients()
     // Poynting Vector.
     // S = Re{E x H⋆} = Re{E x μ⁻¹B⋆}.
     // E and B will be dimensionalized when the coefficient is evaluated, so the scaling
-    // only needs to account for the remaining μ⁻¹ = H / B.
+    // only needs to account for the remaining μ⁻¹ = H / B. As mentioned above,
+    // ProjectCoefficient(*S.get()) or paraview->Save() should always be called after
+    // E and B have been dimensionalized.
     const double scaling = units.Dimensionalize<Units::ValueType::FIELD_H>(1.0) /
                            units.Dimensionalize<Units::ValueType::FIELD_B>(1.0);
     S = std::make_unique<PoyntingVectorCoefficient>(*E, *B, fem_op->GetMaterialOp(),
--- a/palace/models/waveportoperator.cpp
+++ b/palace/models/waveportoperator.cpp
@@ -332,7 +332,7 @@ public:
   BdrSubmeshEVectorCoefficient(const GridFunction &Et, const GridFunction &En,
                                const mfem::ParSubMesh &submesh,
                                const std::unordered_map<int, int> &submesh_parent_elems,
-                               const double scaling = 1.0)
+                               double scaling = 1.0)
     : mfem::VectorCoefficient(Et.Real().VectorDim()), Et(Et), En(En), submesh(submesh),
       submesh_parent_elems(submesh_parent_elems), scaling(scaling)
   {
--- a/test/unit/test-postoperator.cpp
+++ b/test/unit/test-postoperator.cpp
@@ -489,7 +489,6 @@ auto CompareGridFunctions(std::vector<mfem::ParGridFunction> &gf1,
                           std::vector<mfem::ParGridFunction> &gf2,
                           const std::vector<double> &expected_ratios, double rtol)
 {
-  // std::vector<double> ratios;
   CHECK(gf1.size() == gf2.size());
   CHECK(gf1.size() == expected_ratios.size());
   for (auto i = 0; i < gf1.size(); i++)
@@ -497,12 +496,7 @@ auto CompareGridFunctions(std::vector<mfem::ParGridFunction> &gf1,
     // Get the gridfunction data.
     Vector v1 = gf1[i].GetTrueVector();
     Vector v2 = gf2[i].GetTrueVector();
-    // Compute the ratio v2 = v2/v1.
-    v1.UseDevice(true);
-    v2.UseDevice(true);
-    auto d_v1 = v1.Read();
-    auto d_v2 = v2.ReadWrite();
-    mfem::forall(v1.Size(), [=] MFEM_HOST_DEVICE(int i) { d_v2[i] /= d_v1[i]; });
+    v2 /= v1;
     // Get the min/max of the ratio.
     double min = v2.Min();
     double max = v2.Max();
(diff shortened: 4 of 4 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ The format of this changelog is based on
   - Added option to export field solutions as MFEM grid functions to visualize with
     [GLVis](https://glvis.org). Output formats can be specified in `config["Problem"]["OutputFormats"]`.
     [PR 518](https://github.com/awslabs/palace/pull/518).
+  - Changed Paraview and GLVis output fields from nondimensional to SI units
+    [PR 532](https://github.com/awslabs/palace/pull/532).
```
Sbozzolo: This is technically a breaking/interface change.

simlapointe: Right, moved it to the interface change section.

palace/utils/units.hpp
```diff
@@ -180,6 +219,44 @@ class Units
...
     return out;
   }
+
+  template <ValueType unit>
+  auto Nondimensionalize(const Vector &value) const
+  {
+    auto out = value;
+    out *= (1.0 / GetScaleFactor<unit>());
+    return out;
+  }
```
Sbozzolo: I think that there is a way to avoid copies using a combination of [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision.html) (now guaranteed in C++17) and [move semantics](https://en.cppreference.com/w/cpp/utility/move.html)

I think that this avoid intermediate allocations, but not the final copy:
```cpp
template <ValueType unit, typename T>
auto Dimensionalize(std::vector<T> value) const
{
  std::transform(value.begin(), value.end(), value.begin(),
                 [this](T v) { return Dimensionalize<unit>(v); });
  return value;
}
```
with
```cpp
vec = Dimensionalize<unit>(vec);
```
This would avoid the final copy too:
```cpp
vec = Dimensionalize<unit>(std::move(vec));
```

However, I do find all of this more unclear/harder-to-maintain, whereas the `DimensionalizeInPlace` is very explicit.

That said, our current `Dimensionalize` functions do both: they allocate and and copy, we should probably change them to avoid the allocation (in this PR, or we can open an issue for future work).

palace/models/postoperator.cpp
```diff
@@ -409,7 +419,71 @@ void ScaleGridFunctions(double L, int dim, bool imag, T &E, T &B, T &V, T &A)
...
+void DimensionalizeGridFunctions(Units &units, bool imag, T &E, T &B, T &V, T &A)
+{
+  if (E)
+  {
+    units.DimensionalizeInPlace<Units::ValueType::FIELD_E>(*E);
+    units.DimensionalizeInPlace<Units::ValueType::FIELD_E>(E->Real().FaceNbrData());
+    if (imag)
+    {
+      units.DimensionalizeInPlace<Units::ValueType::FIELD_E>(E->Imag().FaceNbrData());
+    }
```
Sbozzolo: As discussed offline, we can fold this into the `DimensionalizeInPlace` function and add a comment to mention what `FaceNbrData` is

palace/models/postoperator.cpp
```diff
@@ -409,7 +419,71 @@ void ScaleGridFunctions(double L, int dim, bool imag, T &E, T &B, T &V, T &A)
...
+  }
+  if (B)
+  {
+    units.DimensionalizeInPlace<Units::ValueType::FIELD_B>(*B);
+    units.DimensionalizeInPlace<Units::ValueType::FIELD_B>(B->Real().FaceNbrData());
+    if (imag)
+    {
+      units.DimensionalizeInPlace<Units::ValueType::FIELD_B>(B->Imag().FaceNbrData());
+    }
+  }
```
Sbozzolo, line 454: We can chaneg this to 
```
  if (E)
  {
    units.DimensionalizeInPlace<Units::ValueType::FIELD_E>(*E);
    units.DimensionalizeInPlace<Units::ValueType::FIELD_E>(E->Real().FaceNbrData());
    if (E.HasImag())
    {
      units.DimensionalizeInPlace<Units::ValueType::FIELD_E>(E->Imag().FaceNbrData());
    }
  }
```
so that we don't have to pass `imag`.

palace/models/postoperator.cpp
```diff
@@ -1344,4 +1421,16 @@ PostOperator<ProblemType::DRIVEN>::MeasureDomainFieldEnergyOnly<ProblemType::DRI
 template auto
 PostOperator<ProblemType::DRIVEN>::InitializeParaviewDataCollection(int ex_idx) -> void;
 
+template void ScaleGridFunctions<std::unique_ptr<GridFunction>>(
+    double, int, bool, std::unique_ptr<GridFunction> &, std::unique_ptr<GridFunction> &,
+    std::unique_ptr<GridFunction> &, std::unique_ptr<GridFunction> &);
```
Sbozzolo: This function is templated, but at the end the implementation relies on the fact that the arguments are palace GridFunctions, so maybe we can remove the template?

palace/models/postoperator.cpp
Sbozzolo: I left a comment about templates below, but if we make the assumption that the arguments are palace gridfunctions (which we implicitly do in how we access Real and Imag), we can be explicit and call `HasImag()`)

palace/fem/coefficient.hpp
```diff
@@ -139,16 +144,17 @@ class BdrSurfaceFluxCoefficient : public mfem::Coefficient,
   const MaterialOperator &mat_op;
   bool two_sided;
   const mfem::Vector &x0;
+  const double scaling;
```
Sbozzolo: Could we maybe add `scaling` to the base class `BdrGridFunctionCoefficient` instead of each individual subclass?

Sbozzolo: It would also be good to add a docstring to describe what it does and comment to describe why is it needed

palace/models/postoperator.cpp
```diff
@@ -177,45 +178,57 @@ void PostOperator<solver_t>::SetupFieldCoefficients()
   if constexpr (HasEGridFunction<solver_t>())
   {
     // Electric Energy Density.
+    // U_e = 1/2 Dᴴ E = 1/2 ε_0 Eᴴ E.
+    const double scaling = units.Dimensionalize<Units::ValueType::FIELD_D>(1.0) /
+                           units.Dimensionalize<Units::ValueType::FIELD_E>(1.0);
```
Sbozzolo, line 188: Should this be D * E?

palace/models/postoperator.cpp
```diff
@@ -177,45 +178,57 @@ void PostOperator<solver_t>::SetupFieldCoefficients()
...
     U_e = std::make_unique<EnergyDensityCoefficient<EnergyDensityType::ELECTRIC>>(
-        *E, fem_op->GetMaterialOp());
+        *E, fem_op->GetMaterialOp(), scaling);
 
     // Electric Boundary Field & Surface Charge.
     E_sr = std::make_unique<BdrFieldVectorCoefficient>(E->Real());
+    // Q_s = D ⋅ n = ε_0 E ⋅ n.
     Q_sr = std::make_unique<BdrSurfaceFluxCoefficient<SurfaceFlux::ELECTRIC>>(
-        &E->Real(), nullptr, fem_op->GetMaterialOp(), true, mfem::Vector());
+        &E->Real(), nullptr, fem_op->GetMaterialOp(), true, mfem::Vector(), scaling);
```
Sbozzolo, line 199: Here we are scaling for the same value as we did for U_e, but the units are different. If it works out fine, this is not very clear. Shouldn't this be
```
scaling = units.Dimensionalize<Units::ValueType::CHARGE>(1.0) 
```
?

palace/models/postoperator.cpp
Sbozzolo: Or, `ValueType::ENERGY` / `ValueType::LENGTH^3`?

palace/models/postoperator.cpp
```diff
@@ -177,45 +178,57 @@ void PostOperator<solver_t>::SetupFieldCoefficients()
...
       Q_si = std::make_unique<BdrSurfaceFluxCoefficient<SurfaceFlux::ELECTRIC>>(
-          &E->Imag(), nullptr, fem_op->GetMaterialOp(), true, mfem::Vector());
+          &E->Imag(), nullptr, fem_op->GetMaterialOp(), true, mfem::Vector(), scaling);
     }
   }
 
   if constexpr (HasBGridFunction<solver_t>())
   {
+    const double scaling = units.Dimensionalize<Units::ValueType::FIELD_H>(1.0) /
+                           units.Dimensionalize<Units::ValueType::FIELD_B>(1.0);
```
Sbozzolo, line 218: Same comments for this section as above

palace/utils/units.hpp
```diff
@@ -157,6 +158,44 @@ class Units
...
+
+  template <ValueType unit>
+  auto Dimensionalize(const GridFunction &value) const
+  {
+    auto out = value;
+    out *= GetScaleFactor<unit>();
+    return out;
+  }
```
Sbozzolo: These look good candidates for being templated
```suggestion
template <ValueType unit, typename T>
auto Dimensionalize(const T &value) const
{
  auto out = value;
  out *= GetScaleFactor<unit>();
  return out;
}
```

Sbozzolo: We should also add unit tests for these functions

palace/utils/units.hpp
```diff
@@ -157,6 +158,44 @@ class Units
...
+
+  template <ValueType unit, typename T>
+  void DimensionalizeInPlace(T &value) const
+  {
+    value *= GetScaleFactor<unit>();
+  }
```
Sbozzolo: Can you add docstrings to these methods?

palace/utils/units.hpp
```diff
@@ -180,6 +219,44 @@ class Units
...
+    return out;
+  }
+
+  template <ValueType unit>
+  auto Nondimensionalize(const GridFunction &value) const
+  {
+    auto out = value;
+    out *= (1.0 / GetScaleFactor<unit>());
+    return out;
+  }
```
Sbozzolo, line 293: These can also be templated.

If you want to overload only specific types, you can use SFINAE
```suggestion
template <ValueType unit, typename T>
auto Nondimensionalize(const T &value) const 
  -> std::enable_if_t<
       std::is_same_v<T, Vector> || 
       std::is_same_v<T, ComplexVector> || 
       std::is_same_v<T, mfem::ParGridFunction> || 
       std::is_same_v<T, GridFunction>, T>
{
  auto out = value;
  out *= (1.0 / GetScaleFactor<unit>());
  return out;
}
```

palace/models/waveportoperator.hpp
```diff
@@ -100,8 +100,10 @@ class WavePortData
   std::unique_ptr<mfem::VectorCoefficient> GetModeExcitationCoefficientReal() const;
   std::unique_ptr<mfem::VectorCoefficient> GetModeExcitationCoefficientImag() const;
 
-  std::unique_ptr<mfem::VectorCoefficient> GetModeFieldCoefficientReal() const;
-  std::unique_ptr<mfem::VectorCoefficient> GetModeFieldCoefficientImag() const;
+  std::unique_ptr<mfem::VectorCoefficient>
+  GetModeFieldCoefficientReal(const double scaling = 1.0) const;
+  std::unique_ptr<mfem::VectorCoefficient>
+  GetModeFieldCoefficientImag(const double scaling = 1.0) const;
```
Sbozzolo: I think the `const` is unnecessary because you are passing a double by value.

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,143 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  A = std::make_unique<GridFunction>(nd_fespace);
+  V = std::make_unique<GridFunction>(h1_fespace);
+
+  // Create fields, initialize them, and set the corresponding gridfunctions.
+  ComplexVector e(nd_fespace.GetTrueVSize()), b(rt_fespace.GetTrueVSize());
+  Vector a(rt_fespace.GetTrueVSize()), v(h1_fespace.GetTrueVSize());
+  e = std::complex<double>(1.0, 1.0);
+  b = std::complex<double>(1.0, 1.0);
+  a = 1.0;
+  v = 1.0;
```
Sbozzolo: The value `1.0` is special (especially when thinking about scalings), let's maybe pick other number.

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,143 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  MaterialOperator mat_op(iodata, mesh);
+  std::unique_ptr<mfem::VectorCoefficient> S;
+  std::unique_ptr<mfem::Coefficient> U_e, U_m;
+
+  const double scaling = 1.0;
+  U_e = std::make_unique<EnergyDensityCoefficient<EnergyDensityType::ELECTRIC>>(*E, mat_op,
+                                                                                scaling);
+  U_m = std::make_unique<EnergyDensityCoefficient<EnergyDensityType::MAGNETIC>>(*B, mat_op,
+                                                                                scaling);
+  S = std::make_unique<PoyntingVectorCoefficient>(*E, *B, mat_op, scaling);
```
Sbozzolo: `scaling = 1.0` is default, which leads to cleaner code
```suggestion
  U_e = std::make_unique<EnergyDensityCoefficient<EnergyDensityType::ELECTRIC>>(*E, mat_op);
  U_m = std::make_unique<EnergyDensityCoefficient<EnergyDensityType::MAGNETIC>>(*B, mat_op);
  S = std::make_unique<PoyntingVectorCoefficient>(*E, *B, mat_op);
```

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,143 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  // E [V/m].
+  auto scaling_E = iodata.units.Dimensionalize<Units::ValueType::FIELD_E>(1.0) * mesh_Lc0;
+  // B [Wb/m²].
+  auto scaling_B =
+      iodata.units.Dimensionalize<Units::ValueType::FIELD_B>(1.0) * mesh_Lc0 * mesh_Lc0;
+  // V [V].
+  auto scaling_V = iodata.units.Dimensionalize<Units::ValueType::VOLTAGE>(1.0);
+  // A [A/m].
+  auto scaling_A = iodata.units.Dimensionalize<Units::ValueType::CURRENT>(1.0) * mesh_Lc0;
+  // Ue = 1/2 ε_0 Eᴴ E.
```
Sbozzolo:
```suggestion
  // U_e = 1/2 ε_0 Eᴴ E.
```

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,143 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  auto scaling_Ue = iodata.units.Dimensionalize<Units::ValueType::FIELD_E>(1.0) *
+                    iodata.units.Dimensionalize<Units::ValueType::FIELD_E>(1.0);
```
Sbozzolo, line 605: I am not sure I understand this. The way I think about this is that `U_e` is energy per unit volume, and I cannot map this to just E^2.

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,143 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  U_e = std::make_unique<EnergyDensityCoefficient<EnergyDensityType::ELECTRIC>>(*E, mat_op,
+                                                                                scaling);
+  U_m = std::make_unique<EnergyDensityCoefficient<EnergyDensityType::MAGNETIC>>(*B, mat_op,
+                                                                                scaling);
+  S = std::make_unique<PoyntingVectorCoefficient>(*E, *B, mat_op, scaling);
+
+  // Expected scalings.
+  auto mesh_Lc0 = iodata.units.GetMeshLengthRelativeScale();
+  // E [V/m].
+  auto scaling_E = iodata.units.Dimensionalize<Units::ValueType::FIELD_E>(1.0) * mesh_Lc0;
```
Sbozzolo, line 595: In the probe interface, we have:
```cpp
measurement_cache.probe_E_field = units.Dimensionalize<Units::ValueType::FIELD_E>(
      nondim_measurement_cache.probe_E_field);
```

Why is this scaling different?

Shouldn't `iodata.units.Dimensionalize<Units::ValueType::FIELD_E>(1.0)` already contain all the necessary factors to scale units? 

Same comment for many of the scalings below.

Sbozzolo: The way I think about this is that dimensionalizing has to components:
1. multiplying the field values for the correct scaling factor
2. multiplying the coordinates (space and time) for the correct mesh scaling factor

I think you added `mesh_LC0` because of what happens in `ScaleGridFunctions`, but I think the correct way is to first dimensionalize `mesh_LC0` and then pass it to `ScaleGridFunctions`, instead of passing a non-dimensional number with then missing a component in the scaling that needs to be accounted post-facto. 

This way, the interface is clear and vector fields can be transformed with their corresponding `unit` in units.

test/unit/test-postoperator.cpp
Sbozzolo: Actually, it would be best to have a space-varying function. Otherwise, this test would not check that the underlying coordinates are properly rescaled.

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,143 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  // Dimensionalize.
+  mesh::DimensionalizeMesh(mesh, mesh_Lc0);
+  ScaleGridFunctions(mesh_Lc0, mesh.Dimension(), true, E, B, V, A);
+  DimensionalizeGridFunctions(iodata.units, true, E, B, V, A);
+  std::vector<double> dim_values = get_max_values();
+  // Check values are as expected.
+  for (size_t i = 0; i < dim_values.size(); i++)
+  {
+    CHECK_THAT(dim_values[i], WithinRel(values_before[i] * expected_scaling[i]));
+  }
```
Sbozzolo: I think we could make this test much more stringent by:
- Making a copy of the initial gridfunctions
- Dimensionalizing into new copies
- Taking the ratio between new and old ones
- Checking that the min and the max of the ratio is `expected_scaling`

I understand that in this case it did not matter because the grid function is constant, but I think that being constant makes the test less effective

palace/models/postoperator.cpp
Sbozzolo: We can also refactor `ScaleGridFunctions` to not take `imag`

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,143 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  // Nondimensionalize.
+  mesh::NondimensionalizeMesh(mesh, mesh_Lc0);
+  ScaleGridFunctions(1.0 / mesh_Lc0, mesh.Dimension(), true, E, B, V, A);
+  NondimensionalizeGridFunctions(iodata.units, true, E, B, V, A);
+  std::vector<double> values_after = get_max_values();
+  // Check values are back to initial values.
+  for (size_t i = 0; i < dim_values.size(); i++)
+  {
+    CHECK_THAT(values_after[i], WithinRel(values_before[i]));
+  }
```
Sbozzolo: Same here, we can check the entire grid function instead of a single point.

docs/src/guide/postprocessing.md
```diff
@@ -105,17 +105,17 @@ ParaView is recommended to visualize large simulations in parallel. The grid fun
 format can be useful to embed visualizations in webpages with its
 [Javascript version](https://github.com/GLVis/glvis-js/).
 
-All fields are written out as nondimensionalized quantities. The specific quantities
-available varies by [simulation type](problem.md#Problem-Types), but the variable names for
+All fields are written out in SI units. The specific quantities available vary by
```
Sbozzolo: Is the mesh in units of meters? If yes, it might be good to clearly say that.

palace/models/postoperator.cpp
simlapointe: E will already be dimensionalized when the coefficients are evaluated, so the scaling here only needs to account for the remaining ε_0 = D / E. I added a comment to that effect.

palace/models/postoperator.cpp
simlapointe: Same as above, E/B will already be dimensionalized when the coefficients are evaluated, so the scaling only needs to account for the remaining.

palace/models/postoperator.cpp
simlapointe: Same as above, E/B will already be dimensionalized when the coefficients are evaluated, so the scaling only needs to account for the remaining.

test/unit/test-postoperator.cpp
simlapointe: The remaining units to get to energy per volume are in ε_0. Since ε_0 doesn't change when evaluating the coefficient using dimensionalized or nondimensionalized quantities I don't put this in the expected `scaling_Ue`.

test/unit/test-postoperator.cpp
simlapointe: The extra `mesh_Lc0` in the expected scalings are because of the calls to `DimensionalizeMesh` and `ScaleGridFunctions`. If we did not call these functions then the scalings would be the same as the probe scalings, without the `mesh_Lc0`. 

I prefer to leave the `mesh_Lc0`, `DimensionalizeMesh` and `ScaleGridFunctions` logic here unchanged because it mirrors what's in the `WriteParaviewFields` and `WriteMFEMGridFunctions` in postoperator.cpp

docs/src/guide/postprocessing.md
simlapointe: The post-processing mesh is the same as the input mesh provided by the user, with units of `config["Model"]["L0"]` m. I didn't think it needed to be mentioned since it's the same as the input, but I guess it doesn't to mention it.

simlapointe:
Thanks for the feedback @Sbozzolo. I think most comments are addressed in the latest.

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,164 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
...
+  {
+    // Get the gridfunction data.
+    Vector v1 = gf1[i].GetTrueVector();
+    Vector v2 = gf2[i].GetTrueVector();
+    // Compute the ratio v2 = v2/v1.
+    v1.UseDevice(true);
+    v2.UseDevice(true);
+    auto d_v1 = v1.Read();
+    auto d_v2 = v2.ReadWrite();
+    mfem::forall(v1.Size(), [=] MFEM_HOST_DEVICE(int i) { d_v2[i] /= d_v1[i]; });
```
Sbozzolo: Isn't `v1 /= v2` enough?

palace/fem/coefficient.hpp
```diff
@@ -548,13 +560,15 @@ class PoyntingVectorCoefficient : public mfem::VectorCoefficient,
       E.Imag().GetVectorValue(T, T.GetIntPoint(), W1);
       linalg::Cross3(W1, W2, V, true);
     }
+    V *= scaling;
   }
 
 public:
   PoyntingVectorCoefficient(const GridFunction &E, const GridFunction &B,
-                            const MaterialOperator &mat_op)
+                            const MaterialOperator &mat_op, const double scaling = 1.0)
```
Sbozzolo:
```suggestion
                            const MaterialOperator &mat_op, double scaling = 1.0)
```

palace/fem/coefficient.hpp
```diff
@@ -452,12 +463,13 @@ class EnergyDensityCoefficient : public mfem::Coefficient, public BdrGridFunctio
...
   const GridFunction &U;
   const MaterialOperator &mat_op;
-
   double GetLocalEnergyDensity(mfem::ElementTransformation &T) const;
 
 public:
-  EnergyDensityCoefficient(const GridFunction &U, const MaterialOperator &mat_op)
-    : mfem::Coefficient(), BdrGridFunctionCoefficient(*U.ParFESpace()->GetParMesh()), U(U),
+  EnergyDensityCoefficient(const GridFunction &U, const MaterialOperator &mat_op,
+                           const double scaling = 1.0)
```
Sbozzolo:
```suggestion
                          double scaling = 1.0)
```

palace/models/postoperator.cpp
```diff
@@ -96,11 +96,12 @@ PostOperator<solver_t>::PostOperator(const IoData &iodata, fem_op_t<solver_t> &f
   // Add wave port boundary mode postprocessing, if available.
   if constexpr (std::is_same_v<fem_op_t<solver_t>, SpaceOperator>)
   {
+    const double scaling = units.Dimensionalize<Units::ValueType::FIELD_E>(1.0);
```
Sbozzolo, line 100: As discussed, let us add notes about using this feature

palace/models/waveportoperator.cpp
```diff
@@ -326,13 +326,15 @@ class BdrSubmeshEVectorCoefficient : public mfem::VectorCoefficient
...
   const std::unordered_map<int, int> &submesh_parent_elems;
   mfem::IsoparametricTransformation T_loc;
+  const double scaling;
 
 public:
   BdrSubmeshEVectorCoefficient(const GridFunction &Et, const GridFunction &En,
                                const mfem::ParSubMesh &submesh,
-                               const std::unordered_map<int, int> &submesh_parent_elems)
+                               const std::unordered_map<int, int> &submesh_parent_elems,
+                               const double scaling = 1.0)
```
Sbozzolo:
```suggestion
                               double scaling = 1.0)
```

test/unit/test-postoperator.cpp
```diff
@@ -482,4 +483,164 @@ TEST_CASE("GridFunction export", "[gridfunction][Serial][Parallel]")
     check_files("eigenmode", 1, post_op.GetPadDigitsDefault(),
                 {"E_real", "E_imag", "B_real", "B_imag", "S", "U_e", "U_m"});
   }
+}
+
+auto CompareGridFunctions(std::vector<mfem::ParGridFunction> &gf1,
+                          std::vector<mfem::ParGridFunction> &gf2,
+                          const std::vector<double> &expected_ratios, double rtol)
+{
+  // std::vector<double> ratios;
```
Sbozzolo:
```suggestion
```

test/unit/test-postoperator.cpp
Sbozzolo: As discussed, if we can make this test a little more intuitive, that would really help future readers.

PR #538, opened 2025-10-22
Use ctest, parallelize tests, report coverage

### LINKED ISSUE #464: Error messages in unit tests make identifying failing test quite hard
Currently, unit tests require Palace to be compiled with `-DPALACE_MFEM_USE_EXCEPTIONS=ON` so that we can catch and aborts. This works, but has the side effect of leaving the error messages in the stdout. In turn, this makes identifying the failing test quite hard. Consider the screenshot below.

Image

The test passed, but there's an error message that clearly say that a verification failed. 

The real problem arises when the tests actually fail. For example, try identifying the real failures from this job: https://github.com/awslabs/palace/actions/runs/17218619226/job/48848348189

Clearing up the error messages would greatly help in understanding what is going wrong without having to reproduce the problem locally.

Sbozzolo:
This commit introduces a few improvements to the test framework:

#### Support for `ctest`

Tests can now be run with:
```
ctest -j $(nproc)
```
This will automatically take care of running all the tests, including the MPI tests (with two process) and the GPU tests (if applicable).

Note, OpenMP threads are not accounted here, so the total number of cores used is $(nproc) * NUM_OMP_THREADS

`ctest` is quite nice because it also handles the stdout well (meaning, it is output only when the test fails):
```
...
13/67 Test #20: serial-PostOperator - Idempotent ...............................   Passed    0.12 sec
      Start 29: serial-Schema Validation - Config with Comments
14/67 Test #18: serial-MaterialOperator IsIsotropic ............................   Passed    0.13 sec
      Start 30: serial-Schema Validation - Invalid Config
15/67 Test #23: serial-PostOperatorCSV_Restart_Helper_ExpectedFilling ..........   Passed    0.13 sec
      Start 31: serial-Schema Validation - Sub-schema by Key
16/67 Test #27: serial-Schema Validation - Embedded Schema Matches Source ......   Passed    0.11 sec
      Start 32: serial-Schema Validation - Error Message Format
17/67 Test #24: serial-PostOperatorCSV_Restart_OneExcitation ...................   Passed    0.15 sec
      Start 33: serial-Schema Validation - Range Expansion
...
```

#### script to compute coverage

`measure-test-coverage` is a new script that uses the ctest capabilities and lcov to distill coverage information and produce html reports.

`measure-test-coverage` can be used to:
1. produce the coverage.info objects by running the tests
2. merge different coverage.info objects
3. produce the html report from the coverage.info files

When run locally, `measure-test-coverage report` will probably cover most of the use cases. The need to handle different coverage.info objects might become relevant in the future for CI (if test coverage information is split across different workers)

#### Coverage is now computed as part of CI:

With all the previous advancements, we can now compute coverage as part of CI and produce a report. The report can be downloaded from the job page on GitHub actions and is uploaded to `gh-pages`:

https://awslabs.github.io/palace/previews/PR538/coverage/

I also added a clickable link to the summary of the GitHub job:
image

Closes #464

### DIFF
--- a/.github/workflows/docs-cleanup.yml
+++ b/.github/workflows/docs-cleanup.yml
@@ -31,6 +31,9 @@ jobs:
             git push --force origin gh-pages-new:gh-pages
           fi
 
+  # Coverage reports from non-PR builds (e.g., pushes to main, workflow_dispatch)
+  # are stored under coverage/<sha>. We check whether the SHA still exists in
+  # the repo and, if so, whether it is older than 30 days.
   cleanup-old-coverage:
     if: github.event_name == 'schedule'
     runs-on: ubuntu-latest
@@ -39,7 +42,10 @@ jobs:
         with:
           ref: gh-pages
 
-      - name: Remove coverage reports older than 30 days
+      - name: Fetch main branch for SHA lookups
+        run: git fetch origin main --depth=1000
+
+      - name: Remove old coverage reports
         run: |
           if [[ ! -d coverage ]]; then
             echo "No coverage directory found"
@@ -51,9 +57,20 @@ jobs:
 
           for dir in coverage/*/; do
             [[ -d "$dir" ]] || continue
-            # Use the git commit date of when the directory was added
-            COMMIT_DATE=$(git log -1 --format=%ct -- "$dir" 2>/dev/null || echo "0")
+            SHA=$(basename "$dir")
+
+            # If the SHA no longer exists in the repo, remove it
+            if ! git cat-file -e "$SHA" 2>/dev/null; then
+              echo "Removing $dir (SHA $SHA no longer exists)"
+              git rm -rf "$dir"
+              DELETED=$((DELETED + 1))
+              continue
+            fi
+
+            # SHA exists — check its commit date
+            COMMIT_DATE=$(git log -1 --format=%ct "$SHA" 2>/dev/null || echo "0")
             if [[ "$COMMIT_DATE" -lt "$CUTOFF" ]]; then
+              echo "Removing $dir (commit date older than 30 days)"
               git rm -rf "$dir"
               DELETED=$((DELETED + 1))
             fi


test/unit/test-libceed.cpp
```diff
@@ -1414,7 +1414,7 @@ void RunCeedBenchmarks(MPI_Comm comm, const std::string &input, int ref_levels,
 
 }  // namespace
 
-TEST_CASE("2D libCEED Operators", "[libCEED][Serial][Parallel][GPU]")
+TEST_CASE("2D libCEED Operators", "[libCEED][Serial][Parallel]")
```
Sbozzolo, line 1417: Removing the GPU tag because of #516 (ultimately, it is likely that these tests were never meant to be run on GPU as-is)

.github/workflows/build-and-test-linux.yml
```diff
@@ -1,7 +1,7 @@
 name: Build and Test (Linux)
 
 permissions:
-  contents: read
+  contents: write
```
Sbozzolo, line 4: Now we push to gh-pages

.github/workflows/spack.yml
```diff
@@ -186,5 +186,5 @@ jobs:
       - uses: ./.github/actions/run-regression-tests
         with:
           toolchain: gcc
-          variant: "+openmp+int64~shared"
+          variant: "+openmp+int64"
```
Sbozzolo, line 206: Removed because of https://github.com/awslabs/palace/issues/645

test/unit/test-memoryreporting.cpp
```diff
@@ -9,16 +9,6 @@
 using namespace palace;
 using namespace palace::memory_reporting;
 
-TEST_CASE("Local Memory Queries", "[memoryreporting][Serial][Parallel]")
```
Sbozzolo, line 12: On Linux, GetCurrentMemory and GetPeakMemory use different sources of
memory. The two sources can have a slightly lag depending on kernel
operations, leading this test to fail.

This lag does not matter in Palace simulations, so I just removed the
test.

.github/actions/palace-ci/action.yml
```diff
@@ -288,20 +288,43 @@ runs:
         start_timer
         eval $(spack -e . load --sh palace)
 
-        if [[ "${{ inputs.variant }}" == *"+cuda"* ]]; then
-          export GPU_FLAGS="--device cuda"
+        # Use nproc, but halve to account for hyperthreading.
+        LOGICAL_CORES=$(nproc)
+        THREADS_PER_CORE=$(lscpu | grep "Thread(s) per core" | awk '{print $4}')
```
phdum-a, line 305: Maybe this is my misunderstanding, but is this template ever going to be run on macos or a non-Ubuntu system? `lscpu` and `nproc` do not exist on mac by default.

Is it worth checking and aborting with error? If not, maybe update a small comment / description of the hardware/OS of palace-ci runners?

.github/actions/palace-ci/action.yml
```diff
@@ -288,20 +288,43 @@ runs:
...
         if [[ "${{ inputs.variant }}" == *"+openmp"* ]]; then
+          NUM_PROC_TEST=$(( NUM_PROC_TEST / 2 ))
           export OMP_NUM_THREADS=2
         else
           export OMP_NUM_THREADS=1
         fi
 
-        palace-unit-tests --skip-benchmarks ${GPU_FLAGS}
-        mpirun -np 2 $(which palace-unit-tests) --skip-benchmarks ${GPU_FLAGS}
+        cd $(spack -e . location -b palace)
```
phdum-a: Minor: How about putting sub-command in quotes? Although I hope spack does not make folders with spaces in them. Might be good to run this script through shellcheck for linting suggestion.

phdum-a: Also applies to `eval $(spack -e . load --sh palace)` above...

.github/actions/palace-ci/action.yml
```diff
@@ -288,20 +288,43 @@ runs:
...
+          CTEST_EXCLUDE=""
+        fi
+
+        # Disable OpenMPI 5.x default CPU binding so MPI processes don't starve
+        # OpenMP threads during parallel test execution. OpenMPI 5.x uses PRRTE,
+        # so we set both the PRTE and legacy OMPI MCA variables for robustness.
+        export PRTE_MCA_hwloc_default_binding_policy=none
+        export OMPI_MCA_hwloc_base_binding_policy=none
+
+        ctest -j $NUM_PROC_TEST --output-on-failure $CTEST_EXCLUDE
```
phdum-a, line 339: Maybe a quick comment here as well that the default runs all hardware compatible tests and skips catch2 benchmarks? Or maybe obvious from ctest setup...

docs/src/developer/testing.md
```diff
@@ -81,6 +142,8 @@ These benchmarks can be accelerated using MPI and/or OpenMP parallelism (when
 configured with `PALACE_WITH_OPENMP=ON`), but in all cases they are only testing
 the local operator assembly on each process.
 
+CTest skips all the benchmarks.
```
phdum-a: CTest is set up to skip all the benchmarks.

phdum-a: I think this section is ok, but slightly confusing. As I understand it `--skip-benchmark` only skips the `BENCHMARK` part of Catch2, but still runs the libCEED / [Benchmark] tests without those sections. So it's not the same as `~[Benchmark]` https://catch2-temp.readthedocs.io/en/latest/command-line.html#skip-all-benchmarks

docs/src/developer/testing.md
```diff
@@ -171,93 +226,79 @@ from your package manager.
...
 
 ```sh
-rm -f coverage*.profraw
-export LLVM_PROFILE_FILE="coverage-%p.profraw"
+./measure-test-coverage report -j8
 ```
 
-Execute the test suite in all configurations:
+Note that MPI tests run with 2 processes each, so `-j8` allows up to 4 MPI tests
+simultaneously.
```
phdum-a: Comment about OpenMP again, or should that be consolidated from above and just referenced here?

CMakeLists.txt
```diff
@@ -223,5 +223,14 @@ endif()
 message(STATUS "========================== Configuring Palace ==========================")
 include(ExternalPalace)
 
+# Create CTestTestfile manually (don't use enable_testing() as it gets overwritten)
+file(WRITE "${CMAKE_BINARY_DIR}/CTestTestfile.cmake" "
+# CMake generated Testfile for Palace superbuild
+set(CTEST_PROJECT_NAME \"palace\")
+
+# Include palace unit tests
+subdirs(\"palace-build/test/unit\")
```
phdum-a, line 233: Why the `\"` here?

scripts/measure-test-coverage
```diff
@@ -0,0 +1,435 @@
...
+    fi
+    (cd "$dir" && pwd)
+}
+
+# BUILD_DIR is the superbuild root (where ctest runs from).
+# PALACE_BUILD_DIR is the inner Palace build (where coverage data lives).
+# UNIT_TEST_EXEC is derived from BUILD_DIR.
+case "$COMMAND" in
+    generate)
+        BUILD_DIR="$(resolve_build_dir "${1:-$PALACE_ROOT/build}")"
```
phdum-a: Maybe document in the help message that somewhere that `$PALACE_ROOT/build` is the default of the folder. ~Also This might need a fail state. if `$PALACE_ROOT` is empty, this will look for `/build`...~ [Edit: Sorry fail state is already in `resolve_build_dir`]

phdum-a: Also, do you need to check quote splitting here?

scripts/measure-test-coverage
```diff
@@ -0,0 +1,435 @@
...
+    local dir="$1"
+    if [[ ! -d "$dir" ]]; then
+        error "Build directory not found: $dir"
+    fi
+    (cd "$dir" && pwd)
+}
+
+# BUILD_DIR is the superbuild root (where ctest runs from).
+# PALACE_BUILD_DIR is the inner Palace build (where coverage data lives).
+# UNIT_TEST_EXEC is derived from BUILD_DIR.
```
phdum-a, line 130: Do you want to predefined these in the global scope and then check to value not existence? Either is fine, it's just taste

scripts/measure-test-coverage
```diff
@@ -0,0 +1,435 @@
...
+    done < <(find . -name '#*.gc[dn][ao]' -print0)
+    cd "$PALACE_BUILD_DIR"
+
+    # Palace is a complex codebase that involves inlined and just-in-time
+    # compiled functions. This can lead to minor inconsistencies in the coverage
+    # files. These are not fatal, and ignoring them still produces good reports.
+    # For this reason, most `lcov` commands have the `--ignore-errors
+    # inconsistent` flag.
+
+    lcov --capture --directory . --ignore-errors inconsistent --output-file "$BUILD_DIR/coverage.info"
```
phdum-a: See https://github.com/linux-test-project/lcov/blob/d100e6cdd4c67cbe5322fa26b2ee8aa34ea7ebcf/man/lcov.1#L615

`--ignore-errors inconsistent` does not seem to be an option in lcov 1.15

scripts/measure-test-coverage
```diff
@@ -0,0 +1,435 @@
...
+        error "Unknown coverage type in: $coverage_line"
+    fi
+}
+
+# Check prerequisites.
+check_prerequisites() {
+    log "Checking prerequisites..."
+
+    [[ -f "$UNIT_TEST_EXEC" ]] || error "Unit tests not found at $UNIT_TEST_EXEC. Build Palace first."
+    command -v lcov >/dev/null || error "lcov not found."
```
phdum-a, line 201: Why only check these things?

.github/workflows/build-and-test-linux.yml
phdum-a: Maybe `contents: read`, `pages: write` or do we need to write contents?

scripts/measure-test-coverage
```diff
@@ -0,0 +1,435 @@
...
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m' # No Color
+
+log() { echo -e "${GREEN}[INFO]${NC} $*"; }
+warn() { echo -e "${YELLOW}[WARN]${NC} $*"; }
+error() { echo -e "${RED}[ERROR]${NC} $*"; exit 1; }
+
+# Portable nproc (works on macOS and Linux).
```
phdum-a: Should it really use all by default? And do we need to worry about openmp / hyper-threading again as on github?

.github/workflows/build-and-test-linux.yml
```diff
@@ -461,12 +483,37 @@ jobs:
...
+          keep_files: true
 
-          # Run Serial tests
-          ./palace-unit-tests --skip-benchmarks
-          # Run Parallel tests
-          mpirun -np $NUM_PROC_TEST_MAX ./palace-unit-tests --skip-benchmarks
+      - name: Add coverage link to summary
+        if: needs.filter.outputs.test == 'true' && matrix.with-coverage
+        run: |
+          PR_NUM=${{ github.event.pull_request.number }}
```
phdum-a: This workflow can also run on push to main and workflow_dispatch where the is not pull_request number. Need fallback?

.github/actions/palace-ci/action.yml
Sbozzolo: I never thought of running this on macOS, but I will take care of this

.github/workflows/build-and-test-linux.yml
Sbozzolo: `contents: write` is because we write to the repo, to the `gh-pages` branch. This is needed

.github/workflows/build-and-test-linux.yml
Sbozzolo: Good point! I now always push the coverage, and I added a step to clean up old coverages from gh-pages.

docs/src/developer/testing.md
Sbozzolo: Added comment

CMakeLists.txt
Sbozzolo: Because it's inside a `file(WRITE`

scripts/measure-test-coverage
Sbozzolo: Here, I check what is required independently of the compiler type. More checks are done when the compiler is detected.

spack_repo/local/packages/palace/package.py
```diff
@@ -341,36 +334,21 @@ def cmake_args(self):
...
+                strumpack_packages = ["ParMETIS", "METIS", "LAPACK", "BLAS", "MPI", "MPI_Fortran"]
                 if self.spec.satisfies("+openmp"):
                     strumpack_packages.append("OpenMP")
                 if self.spec.satisfies("+cuda"):
                     strumpack_packages.append("CUDAToolkit")
                 args.append(
-                    self.define(
-                        "STRUMPACK_REQUIRED_PACKAGES", ";".join(strumpack_packages)
-                    )
+                    self.define("STRUMPACK_REQUIRED_PACKAGES", ";".join(strumpack_packages))
```
Sbozzolo, line 351: There's some code movement because our package.py was not linted.

test/unit/CMakeLists.txt
```diff
@@ -134,3 +136,66 @@ install(
...
+
+# MPI tests  
+catch_discover_tests(unit-tests
+  TEST_PREFIX "mpi-"
+  TEST_SPEC "[Parallel]"
+  TEST_EXECUTOR ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} 2 ${MPIEXEC_PREFLAGS}
+  EXTRA_ARGS ${MPIEXEC_POSTFLAGS} --skip-benchmarks
+  PROPERTIES
+    PROCESSORS 2
+    SKIP_RETURN_CODE 4
```
Sbozzolo, line 168: This is needed to consider skipped tests as successes and not failures

test/unit/main.cpp
```diff
@@ -74,7 +74,7 @@ int main(int argc, char *argv[])
   // TODO: Create a palace::Device class that takes care of all of this.
   hypre::Initialize();
 #if defined(PALACE_WITH_SLEPC)
-  slepc::Initialize(argc, argv, nullptr, nullptr);
+  slepc::Initialize();
```
Sbozzolo, line 77: I changed this to fix the warnings with unused command line arguments.

spack_repo/local/packages/palace/package.py
```diff
@@ -29,12 +29,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
 
     # Note: 'cuda' and 'cuda_arch' variants are added by the CudaPackage
     # Note: 'rocm' and 'amdgpu_target' variants are added by the ROCmPackage
-    variant(
-        "cxxstd",
-        default="17",
-        values=("17", "20"),
-        description="C++ standard",
-    )
+    variant("cxxstd", default="17", values=("17", "20"), description="C++ standard", when="@0.16:")
```
phdum-a, line 32: Ok to add constraint to `@0.16`, although the C++20 test won't be meaningful until then. (But it should pass since it does not explicitly check for C++20)

.github/workflows/build-and-test-linux.yml
```diff
@@ -482,12 +509,41 @@ jobs:
             export LSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/test/unit/lsan_suppressions.txt"
           fi
 
-          cd $(pwd)/palace-build/palace-build/test/unit
+          # We don't compute coverage for the entire matrix because it slows
+          # everything down.
+          if [[ "${{ matrix.with-coverage }}" == 'true' ]]; then
+            scripts/measure-test-coverage report $(pwd)/palace-build
```
phdum-a, line 515: This is fine and I won't change it for this PR. I would say that the script folder `measure-test-coverage report` actually running things ctest if you pass it a build folder while it only does post-processing when given files is unexpected for me. But it's fine and documented by the script.

spack_repo/local/packages/palace/package.py
```diff
@@ -160,18 +162,14 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
-
     # NOTE: hypre+gpu-profiling is also useful: it adds NVTX annotations, which
     # are great for GPU profiling with Nsight.
 
     with when("@0.16:"):
         # +lapack means: use external lapack
         depends_on(
-            "mfem+mpi+metis+lapack@4.9: cxxstd=17",
-            patches=[
-                "patch_par_tet_mesh_fix_dev.diff",
```
phdum-a, line 172: Why remove the cxxstd=17 constraint? This is probably ok, it should just revert to it's own internal c++ standard...

test/unit/main.cpp
phdum-a: Ok. I just copied this from main. Question: does slepc does not need to build an mpi comm? In that case does it need argc,argv to know the number of procs etc. Or if you initialise without inputs it will inherit an already existing global MPI comm?

spack_repo/local/packages/palace/package.py
Sbozzolo: The constraint `@0.16:` is matched by `@develop`, so the test is already meaningful.

spack_repo/local/packages/palace/package.py
Sbozzolo: We had to specify cxxstd=17 because we were pinning a specific commit for MFEM. Now, we use MFEM 4.9 and that requires C++17, so no longer need that.

test/unit/main.cpp
Sbozzolo: I verified that SLEPC checks that MPI is initialized independently of argc/argv.

PR #541, opened 2025-10-24
Add documentation about units, move 2pi out of `units::FREQUENCY`, fix missing factor of 1e9 in `units::ENERGY`

Sbozzolo:
Units are confusing and it is very easy to get things wrong and miss subtleties, so I thought I would document them with my current best understanding. The target audience for this new documentation is developers.

In addition to added documentation, I:
- Changed printing `L0` and `T0` to `Lc` and `tc` in the log
- Moved 2 * pi out of the unit conversion for `FREQUENCY`
- I identified a missing factor of 1e9 in `units::ENERGY` that comes from the fact that our unit of time is nanoseconds and not seconds.

I added a test to verify that indeed the units of energy were wrong. The test compares the energy of a uniform electric field in a unit box and fails if I don't introduce the factor of 1e9. 

I also ran a COMSOL simulation for the antenna case. I set the lumped power to be normalized to 1 W, and I found that E_elec in COMSOL is 1.51e-9, while the reported E_elec in Palace is 1.07e1. The ratio Palace/COMSOL is approximately 1.41 * 1e-10. This cannot be explained just by a factor of 1e9, or by a sqrt(2), so more investigation would be needed.

### DIFF
--- a/docs/src/developer/notes.md
+++ b/docs/src/developer/notes.md
@@ -50,11 +50,13 @@ with `Z` impedance). This choice is related to the normalization of power across
 ports.
 
 An equivalent mental model about non-dimensional quantities in *Palace* is to
-think in terms of characteristic values instead of units: you can think that
-non-dimensional quantities are physical quantities normalized by some reference
-value. For example, the magnetic field strength `H` solved for is actually
-`H/Hc`, where `Hc` satisfies `Hc² × Z₀ × Lc² = 1 W`, with `Z₀` impedance of free
-space. Similarly, the non-dimensional electric field `E` is `E/Ec` with `Ec = Z₀ × Hc`.
+think in terms of characteristic values instead of units: the non-dimensional
+quantities are physical quantities normalized by reference values. These
+reference values are effectively chosen to satisfy the constraints on space,
+time and power given above. For example, the magnetic field strength `H` solved
+for is actually `H/Hc`, where `Hc` satisfies `Hc² × Z₀ × Lc² = 1 W`, with `Z₀`
+impedance of free space. Similarly, the non-dimensional electric field `E` is
+`E/Ec` with `Ec = Z₀ × Hc`.
 
 !!! note "Example: converting times and distances"
     
--- a/palace/models/postoperator.cpp
+++ b/palace/models/postoperator.cpp
@@ -1171,8 +1171,9 @@ auto PostOperator<solver_t>::MeasureAndPrintAll(int ex_idx, int step,
   measurement_cache.ex_idx = ex_idx;
   MeasureAllImpl();
 
-  omega = units.Dimensionalize<Units::ValueType::FREQUENCY>(omega) / (2 * M_PI);
-  post_op_csv.PrintAllCSVData(*this, measurement_cache, omega.real(), step, ex_idx);
+  std::complex<double> freq =
+      units.Dimensionalize<Units::ValueType::FREQUENCY>(omega) / (2 * M_PI);
+  post_op_csv.PrintAllCSVData(*this, measurement_cache, freq.real(), step, ex_idx);
   if (ShouldWriteParaviewFields(step))
   {
     Mpi::Print("\n");
@@ -1188,7 +1189,7 @@ auto PostOperator<solver_t>::MeasureAndPrintAll(int ex_idx, int step,
     auto ind = 1 + std::distance(output_save_indices.begin(),
                                  std::lower_bound(output_save_indices.begin(),
                                                   output_save_indices.end(), step));
-    WriteMFEMGridFunctions(omega.real(), ind);
+    WriteMFEMGridFunctions(freq.real(), ind);
     Mpi::Print(" Wrote fields to disk (grid function) at step {:d}\n", step + 1);
   }
   return measurement_cache.domain_E_field_energy_all +


docs/src/reference.md
```diff
@@ -594,6 +594,74 @@ the radial terms (i.e., those proportional to ``\mathbf{r}_0``, in the wave zone
...
+First, the mesh comes with its own unit of length. Users specify the conversion
+between one unit of length on the mesh to meters through the
+[`config["Model"]["L0"]`](config/problem.md#config%5B%Model%22%5D) parameter.
+
+Second, *Palace* defines a system of non-dimensional units where the actual
+solver operates.
+
+Third, postprocessing occurs in a custom variant of SI units where the unit of
+time is nanoseconds instead of seconds and frequencies are measured in GHz/rad
+instead of Hz (i.e., frequency is angular frequency).
```
phdum-a: Could you double check this? I thought the input and output frequencies are just GHz.

The internal frequency units are angular frequency (so we do scaling of the frequency units and add in a 2pi additionally). (see units.hpp#76)

docs/src/reference.md
```diff
@@ -594,6 +594,74 @@ the radial terms (i.e., those proportional to ``\mathbf{r}_0``, in the wave zone
...
+The non-dimensional unit system is constructed as follows:
+
+Length: Lengths are measured in units of the characteristic length `Lc`, which is
+typically the mesh size in mesh units. The extent of the mesh is 1 in these units
+(unless `Lc` is manually specified in
+[`config["Model"]["Lc"]`](config/problem.md#config%5B%Model%22%5D)).
+
+Time: One unit of non-dimensional time is the time it takes light to travel one
+`Lc` (equivalently, speed of light is 1 in these units).
+
```
simlapointe: Maybe add somewhere that the characteristic length and time scales `Lc` and `tc` used for non-dimensionalization are printed in the palace output/log (although in the log `Lc` is called `L0` so that can be slightly confusing)

docs/src/reference.md
simlapointe: As @phdum-a pointed out, I think the input and output frequencies are just GHz, see for example eig.csv for the eigenmode simulations. Let's double check that.

docs/src/reference.md
```diff
@@ -594,6 +594,74 @@ the radial terms (i.e., those proportional to ``\mathbf{r}_0``, in the wave zone
...
+following way. The unit of magnetic field strength `H` is defined so that the
+power flowing through a surface of area `Lc²` perpendicular to the direction of
+propagation in vacuum with a uniform magnetic field of strength 1 is exactly 1
+(`P = H^2 * Z * L^2`).
+
+An equivalent mental model about non-dimensional quantities in *Palace* is to
+think in terms of characteristic values instead of units: you can think that
+non-dimensional quantities are physical quantities normalized by some reference
+value. For example The magnetic field strength `H` solved for is actually
+`H/Hc`, where `Hc` satisfies `Hc² × Z₀ × Lc² = 1 W`. Similarly, the
```
simlapointe: Say what `Z0` is?

docs/src/reference.md
```diff
@@ -594,6 +594,74 @@ the radial terms (i.e., those proportional to ``\mathbf{r}_0``, in the wave zone
...
+non-dimensional electric field `E` is `E/Ec` with `Ec = Z₀ × Hc`.
+
+!!! note "Example: converting times and distances"
+    
+    Consider a spherical mesh with `L0` of 1 micron and radius of 3 in mesh units.
+    Then:
+    
+      - Mesh size in mesh units: 6
+      - Mesh size in non-dimensional units: 1
+      - Mesh radius in dimensional units: 6e-6
```
simlapointe: Guessing you mean mesh _size_? If radius it should 3e-6

simlapointe: I would add the actual units to every "in dimensional units" in this list, e.g. 6e-6 m, 0.3 m/ns, 2e-5 ns.

docs/src/reference.md
```diff
@@ -594,6 +594,74 @@ the radial terms (i.e., those proportional to ``\mathbf{r}_0``, in the wave zone
...
+      - Speed of light in non-dimensional units: 1
+      - Speed of light in dimensional units: 0.3
+      - Time for light to cross the sphere in non-dimensional units: 1
+      - Time for light to cross the sphere in dimensional units: 2e-5
+
+!!! warning "Non-SI time units"
+    
+    The dimensional unit system uses nanoseconds for time and GHz/rad for
```
simlapointe: As mentioned above, let's double check GHz/rad vs GHz for frequency inputs and outputs

docs/src/reference.md
```diff
@@ -594,6 +594,74 @@ the radial terms (i.e., those proportional to ``\mathbf{r}_0``, in the wave zone
...
+    frequency, not standard SI units. This affects unit combinations:
+    `units::POWER * units::TIME` is not equivalent to `units::ENERGY`.
+
+!!! warning "Finite element considerations"
+    
+    Changing mesh units requires additional rescaling for fields defined on
+    H(curl) and H(div) spaces to convert them to proper physical units. See the
+    `ScaleGridFunctions` function for details.
```
simlapointe: add file+line path where that function is?

docs/src/reference.md
```diff
@@ -594,6 +594,74 @@ the radial terms (i.e., those proportional to ``\mathbf{r}_0``, in the wave zone
...
+      - Speed of light in non-dimensional units: 1
+      - Speed of light in dimensional units: 0.3
+      - Time for light to cross the sphere in non-dimensional units: 1
+      - Time for light to cross the sphere in dimensional units: 2e-5
+
+!!! warning "Non-SI time units"
+    
+    The dimensional unit system uses nanoseconds for time and GHz/rad for
+    frequency, not standard SI units. This affects unit combinations:
+    `units::POWER * units::TIME` is not equivalent to `units::ENERGY`.
```
simlapointe: This is a bit confusing since looking at units.hpp I see this for time
`if constexpr (unit == ValueType::TIME) {return tc_ns;  // [ns]}`,
this for power
 `if constexpr (unit == ValueType::POWER) {return Hc * Hc * electromagnetics::Z0_ * Lc_m * Lc_m;  // [W]}`, 
and this for energy
`if constexpr (unit == ValueType::ENERGY) {return Hc * Hc * electromagnetics::Z0_ * Lc_m * Lc_m * tc_ns;  // [J] }`
which seems to suggest that `units::POWER * units::TIME == units::ENERGY`

simlapointe (changes requested):

docs/src/reference.md
Sbozzolo: Nice catch. I guess, the correct statement is that `unites::ENERGY` is not `Joules`, but nanojoules.

By construction, Hc^2 * Z0 * Lc^2 = 1 W, so 1 W * ns = 1e-9 J.

docs/src/reference.md
Sbozzolo: Thank you!

Sbozzolo: I added that as a separate entry. I had originally left it out on purpose because it can hide complexity (eg, it's important that it is 0.3 m/ns and not 3e8 m/s).

docs/src/reference.md
Sbozzolo: I am going to rename `L0` to `Lc` in the log.

docs/src/reference.md
Sbozzolo: This comment was mostly referring to the system of units defined in `units.cpp`, and indeed it is not phrased in the best way. 

What I meant is the following. If you have a non-dimensional time `t`, you can compute a non-dimensional frequency `f = 1/t`. Now, if you convert `t` to our dimensional units with `Dimensionalize`, you get `t_dim = XXX ns`. If you convert `f` to dimensional units, you get `2 * pi /  t_dim = 2 * pi * 1 / Dimensionalize<TIME>(t)`. 

It seems to be true that the input and output we take are non-angular frequencies, but this is because the application of `(Non)dimensionalize` changes the meaning of the variable, as clearly seen in, e.g., in `PostOperator`:
```cpp
measurement_cache.freq = omega;
(Later we call Dimensionalize)
```
or in 
```cpp
Mpi::Print(
          "\nIt {:d}/{:d}: ω/2π = {:.3e} GHz (total elapsed time = {:.2e} s{})\n",
          omega_i + 1, omega_sample.size(),
          iodata.units.Dimensionalize<Units::ValueType::FREQUENCY>(omega),
```

I am not sure how to best describe all of this.

Sbozzolo: See response to the comment above.

docs/src/reference.md
Sbozzolo: The line number will likely change in the future, so I guess said `in PostOperator`.

docs/src/reference.md
Sbozzolo: I moved this section to the Developers' notes. I removed the factor of 2pi from `units.hpp` and move it to the call site. I added documentation explaining that internally we use omega.

docs/src/reference.md
Sbozzolo: I moved the factor of 2pi out of the units, which makes everything clearer

palace/models/postoperatorcsv.cpp
```diff
@@ -22,7 +22,8 @@ Measurement Measurement::Dimensionalize(const Units &units,
 {
   Measurement measurement_cache;
   measurement_cache.freq =
-      units.Dimensionalize<Units::ValueType::FREQUENCY>(nondim_measurement_cache.freq);
+      units.Dimensionalize<Units::ValueType::FREQUENCY>(nondim_measurement_cache.freq) /
+      (2.0 * M_PI);
```
hughcars: Nit: `2.0 * M_PI` or `2 * M_PI`

test/unit/test-domainpostoperator.cpp
```diff
@@ -0,0 +1,85 @@
...
+  // TODO: This test can be expanded/improved to be a more robust test for the
+  // actual function, not just the units.
+
+  Units units(0.496, 1.453);
+  IoData iodata = IoData(units);
+
+  auto &material = iodata.domains.materials.emplace_back();
+  material.attributes = {1};
+
+  // Make mesh for a cube [0, 1] x [0, 1] x [0, 1].
```
hughcars: Given 1 can hide unit issues sometimes, make the cube have a non-unit side length. Should catch erroneous `/ L` etc. `MakeCartesian3D` has `sx`, `sy` and `sz` arguments, you could even make them non-equal if that doesn't make the manual calculation too complex.

test/unit/test-strattonchu.cpp
```diff
@@ -236,9 +235,8 @@ void runFarFieldTest(double freq_Hz, std::unique_ptr<mfem::Mesh> serial_mesh,
   SurfacePostOperator surf_post_op(iodata, mat_op, nd_fespace, nd_fespace);
 
   auto thetaphis = GenerateSphericalTestPoints();
-  // NOTE: units.Nondimensionalize<Units::ValueType::FREQUENCY> adds a factor of 2pi!
   double omega_rad_per_time =
-      units.Nondimensionalize<Units::ValueType::FREQUENCY>(freq_Hz / 1e9);
+      2 * M_PI * units.Nondimensionalize<Units::ValueType::FREQUENCY>(freq_Hz / 1e9);
```
hughcars, line 239: The `freq_Hz` here, is that `freq_GHz`, given the 1e9?

docs/src/developer/notes.md
```diff
@@ -5,6 +5,86 @@
 
 # Developer Notes
 
+## Units
+
+*Palace* use SI units for interactions with users (inputs and outputs). However,
```
hughcars: Palace mostly uses SI units, and where it doesn't it uses metric units (SI with prefix modifiers e.g. km, um, nJ).

palace/utils/iodata.cpp
```diff
@@ -537,25 +537,34 @@ void IoData::NondimensionalizeInputs(mfem::ParMesh &mesh)
...
+  solver.transient.delta_t =
+      units.Nondimensionalize<Units::ValueType::TIME>(solver.transient.delta_t);
 
   // Scale mesh vertices for correct nondimensionalization.
   mesh::NondimensionalizeMesh(mesh, units.GetMeshLengthRelativeScale());
 
   // Print some information.
   Mpi::Print(mesh.GetComm(),
-             "\nCharacteristic length and time scales:\n L₀ = {:.3e} m, t₀ = {:.3e} ns\n",
+             "\nCharacteristic length and time scales:\n Lc = {:.3e} m, tc = {:.3e} ns\n",
```
hughcars, line 580: The characteristic length and time are `L₀` and `t₀`, changing to Lc and tc here makes it overlap with `"Lc"` if a user provides that in the json. The reason to print these here, is to express what the actual, in metric units, nondimensionalization is. That number is the `L0 * Lc` from the `Units`, i.e. the second argument into the ctor, so if a user put `Lc = 1.4` or something in their json, then saw `Lc` being output here different to that, it might be confusing. Conversely, maybe having the mesh unit conversion performed for a user and output here in such a case might be helpful. I can see it either way, just worth considering. Often in writing down a nondim on paper, L0 and t0 are used, which is probably the origin of this.

docs/src/developer/notes.md
```diff
@@ -5,6 +5,86 @@
...
+
+The non-dimensional unit system is constructed as follows:
+
+Length: Lengths are measured in units of the characteristic length `Lc`, which is
+typically the mesh size in mesh units. The extent of the mesh is 1 in these units
+(unless `Lc` is manually specified in
+[`config["Model"]["Lc"]`](config/problem.md#config%5B%Model%22%5D)).
+
+Time: One unit of non-dimensional time is the time it takes light to travel one
+`Lc` (equivalently, speed of light is 1 in these units).
```
hughcars: the speed of light

docs/src/developer/notes.md
```diff
@@ -5,6 +5,86 @@
...
+`units.hpp`.
+
+### Non-dimensional unit system
+
+The non-dimensional unit system is constructed as follows:
+
+Length: Lengths are measured in units of the characteristic length `Lc`, which is
+typically the mesh size in mesh units. The extent of the mesh is 1 in these units
+(unless `Lc` is manually specified in
+[`config["Model"]["Lc"]`](config/problem.md#config%5B%Model%22%5D)).
```
hughcars: Extent here is a little vague, concretely it's the largest single dimension, so it is not rotationly invariant. Maybe something more concrete: "The largest dimension of the mesh is 1 in these units, for example for a mesh of a cuboid of size 10m x 20m x 40m, this would be 40m".

docs/src/developer/notes.md
```diff
@@ -5,6 +5,86 @@
...
+of free space is 1 and choosing the units of magnetic field strength `H` in the
+following way. The unit of magnetic field strength `H` is defined so that the
+power flowing through a surface of area `Lc²` perpendicular to the direction of
+propagation in vacuum with a uniform magnetic field of strength 1 is exactly 1
+(`P = H^2 * Z * L^2`, with `Z` impedence).
+
+An equivalent mental model about non-dimensional quantities in *Palace* is to
+think in terms of characteristic values instead of units: you can think that
+non-dimensional quantities are physical quantities normalized by some reference
+value. For example The magnetic field strength `H` solved for is actually
```
hughcars: instead of units: the non-dimensional quantities are physical quantities normalized by reference values. These reference values are effectively chosen to satisfy the constraints on space, time and power given above.

docs/src/developer/notes.md
```diff
@@ -5,6 +5,86 @@
...
+      - Speed of light in dimensional units: 0.3
+      - Speed of light in SI: 3e8 meters/seconds
+      - Time for light to cross the sphere in non-dimensional units: 1
+      - Time for light to cross the sphere in dimensional units: 2e-5
+      - Time for light to cross the sphere in SI: 2e-14 seconds
+
+!!! warning "Non-SI time units"
+    
+    The dimensional unit system uses nanoseconds for time, not seconds. This affects
+    unit combinations: `units::POWER * units::TIME` is not equivalent to `units::ENERGY`.
```
hughcars, line 80: Something to the effect of: "This choice of nanoseconds reflects a prioritization of the microwave engineering domain within Palace, as frequencies are reported in `GHz` and `1 ns` is the corresponding time scale.

docs/src/developer/notes.md
```diff
@@ -5,6 +5,86 @@
...
+!!! warning "Omega instead of Frequency"
+    
+    Internally, *Palace* uses the angular frequency `omega` instead of the regular
+    frequency, with both names `omega` and `freq` being used.
+
+!!! warning "Finite element considerations"
+    
+    Changing mesh units requires additional rescaling for fields defined on
+    H(curl) and H(div) spaces to convert them to proper physical units. See the
+    `ScaleGridFunctions` function in `PostOperator` for details.
```
hughcars: This is a function of the length scales inherent to the reference element transformations for Nedelec and Raviart Thomas spaces, used for Hcurl and Hdiv respectively.

palace/utils/iodata.cpp
```diff
@@ -537,25 +537,34 @@ void IoData::NondimensionalizeInputs(mfem::ParMesh &mesh)
...
+      2 * M_PI *
+      units.Nondimensionalize<Units::ValueType::FREQUENCY>(solver.eigenmode.target);
+  solver.eigenmode.target_upper =
+      2 * M_PI *
+      units.Nondimensionalize<Units::ValueType::FREQUENCY>(solver.eigenmode.target_upper);
 
   // For driven simulations:
   for (auto &f : solver.driven.sample_f)
-    f /= units.GetScaleFactor<Units::ValueType::FREQUENCY>();
+    f = 2 * M_PI * units.Nondimensionalize<Units::ValueType::FREQUENCY>(f);
```
hughcars, line 562: Nit: maybe append a `// rad.s^-1` or the like after this, just to emphasize what the conversion is doing.

palace/utils/units.hpp
```diff
@@ -74,7 +74,7 @@ class Units
     }
     else if constexpr (unit == ValueType::FREQUENCY)
     {
-      return 1.0 / (2.0 * M_PI * tc_ns);  // [GHz/rad]
+      return 1.0 / tc_ns;  // [GHz]
```
hughcars, line 77: This `2.0 * M_PI` results in a lot of call sites being noisier than needed, meaning a developer needs to remember between them. Is there value in a `ValueType::AngularFrequency`, for converting from a frequency to angular such as this? So dimensionalizing a frequency like this is the same, but angular frequency would incorporate the 2pi and change most call sites to converting between frequency in dimensional scale, and non-dimensional angular frequency instead?

hughcars (commented):
Generally LGTM, some comments on doc strings and an idea on the angular conversion.

test/unit/test-strattonchu.cpp
Sbozzolo: No, it's `freq_Hz`: `freq_Hz / 1e9` converts to `freq_GHz`, which is what `Nondimensionalize` expects.

palace/utils/iodata.cpp
Sbozzolo: Right, good point. The main reason for this change is to remove the source of confusion with L₀ not being the L0 specified in the JSON. 

I personally think that `Lc` here is not too confusing: we clearly say that the units are meters, and the documentation for `"Lc"` is also quite clear on being mesh units:
> "Lc" [0.0]` :  Characteristic length scale used for nondimensionalization, specified in mesh length units. [...]

palace/utils/units.hpp
Sbozzolo: This is how my brain works: I treat `freq` and `omega` as two distinct quantities related by `omega = 2 * pi * freq` in any given unit system. So, in my worldview, going from `freq` to `omega` should not be just a unit transformation. That said, I would not oppose to adding `ValueType::AngularFrequency` if you prefer that. To me, the most important thing would be clarify and standardize our variable names so that one can immediately understand if a factor of 2pi is included or not. 

We currently have code like:
```cpp
omega = units.Dimensionalize<Units::ValueType::FREQUENCY>(omega)
```
or 
```cpp
measurement_cache.freq = units.Dimensionalize<Units::ValueType::FREQUENCY>(nondim_measurement_cache.freq)
```

So, given a piece of Palace code, I never know if I am dealing with frequency or 2pi frequency.

hughcars: > This is how my brain works: I treat `freq` and `omega` as two distinct quantities related by `omega = 2 * pi * freq` in any given unit system. So, in my worldview, going from `freq` to `omega` should not be just a unit transformation. That said, I would not oppose to adding `ValueType::AngularFrequency` if you prefer that. To me, the most important thing would be clarify and standardize our variable names so that one can immediately understand if a factor of 2pi is included or not.
> 
> We currently have code like:
> 
> ```c++
> omega = units.Dimensionalize<Units::ValueType::FREQUENCY>(omega)
> ```
> 
> or
> 
> ```c++
> measurement_cache.freq = units.Dimensionalize<Units::ValueType::FREQUENCY>(nondim_measurement_cache.freq)
> ```
> 
> So, given a piece of Palace code, I never know if I am dealing with frequency or 2pi frequency.

Agreed, the variable name ought to be sufficient alone. `omega` is generally supposed to be an angular, and `f` a frequency. So nondimensionalizing a frequency or an angular frequency would still be the same coefficient applied (those two callsites remain the same), and converting between frequency and angular frequency is entirely outside of `units.hpp`.

simlapointe (approved):

palace/models/postoperator.cpp
```diff
@@ -1171,7 +1171,7 @@ auto PostOperator<solver_t>::MeasureAndPrintAll(int ex_idx, int step,
   measurement_cache.ex_idx = ex_idx;
   MeasureAllImpl();
 
-  omega = units.Dimensionalize<Units::ValueType::FREQUENCY>(omega);
+  omega = units.Dimensionalize<Units::ValueType::FREQUENCY>(omega) / (2 * M_PI);
```
hughcars: This would be a good example of the angular conversion problem, the `omega` passed in must be angular, but its assigned onto non-angular, so we can think of this as a notational bug really.

docs/src/developer/notes.md
Sbozzolo: I missed this, but it will be fixed in the new commit

palace/models/postoperator.cpp
Sbozzolo: Fixed this

PR #544, opened 2025-10-28
Scale Ls/Rs/Cs of cracked impedance boundaries

simlapointe:
When mesh cracking is enabled, internal impedance boundary conditions used to model a thin (2D) sheet are split into two thin sheets, effectively doubling/halving the user-specified inductance/resistance/capacitance of the impedance BC. This fix internally scales the user-specified Ls/Rs/Cs when mesh cracking is enabled such that the results are consistent with the non-cracked results.

### DIFF
--- a/palace/utils/geodata.cpp
+++ b/palace/utils/geodata.cpp
@@ -1941,6 +1941,7 @@ int AddInterfaceBdrElements(IoData &iodata, std::unique_ptr<mfem::Mesh> &orig_me
     auto crack_bdr_marker = mesh::AttrToMarker(
         orig_mesh->bdr_attributes.Size() ? orig_mesh->bdr_attributes.Max() : 0,
         crack_boundary_attributes, true);
+    std::unordered_set<int> external_attributes;
     for (int be = 0; be < orig_mesh->GetNBE(); be++)
     {
       if (crack_bdr_marker[orig_mesh->GetBdrAttribute(be) - 1])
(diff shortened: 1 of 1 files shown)


palace/utils/geodata.cpp
```diff
@@ -1926,6 +1951,7 @@ int AddInterfaceBdrElements(const IoData &iodata, std::unique_ptr<mfem::Mesh> &o
         if (e1 >= 0 && e2 >= 0)
         {
           crack_bdr_elem.insert(be);
+          iodata.boundaries.cracked_attributes.insert(orig_mesh->GetBdrAttribute(be));
```
hughcars, line 1955: Really not a fan of needing to allow mutation of `IoData` like this, but exfiltrating this information is a mess otherwise. This whole approach will also break for mixed internal/external boundary attributes, so should add some kind of warning for that scenario:
```diff
diff --git a/palace/utils/geodata.cpp b/palace/utils/geodata.cpp
index 5a5922d4..788833f2 100644
--- a/palace/utils/geodata.cpp
+++ b/palace/utils/geodata.cpp
@@ -1941,6 +1941,7 @@ int AddInterfaceBdrElements(IoData &iodata, std::unique_ptr<mfem::Mesh> &orig_me
     auto crack_bdr_marker = mesh::AttrToMarker(
         orig_mesh->bdr_attributes.Size() ? orig_mesh->bdr_attributes.Max() : 0,
         crack_boundary_attributes, true);
+    std::unordered_set<int> external_attributes;
     for (int be = 0; be < orig_mesh->GetNBE(); be++)
     {
       if (crack_bdr_marker[orig_mesh->GetBdrAttribute(be) - 1])
@@ -1953,8 +1954,27 @@ int AddInterfaceBdrElements(IoData &iodata, std::unique_ptr<mfem::Mesh> &orig_me
           crack_bdr_elem.insert(be);
           iodata.boundaries.cracked_attributes.insert(orig_mesh->GetBdrAttribute(be));
         }
+        else
+        {
+          external_attributes.insert(orig_mesh->GetBdrAttribute(be));
+        }
       }
     }
+    MFEM_VERIFY(crack_bdr_elem.empty() || !orig_mesh->Nonconforming(),
+                "Duplicating internal boundary elements for interior boundaries is not "
+                "supported for nonconforming meshes!");
+    std::vector<int> mixed_attributes;
+    std::set_intersection(iodata.boundaries.cracked_attributes.begin(),
+                          iodata.boundaries.cracked_attributes.end(),
+                          external_attributes.begin(), external_attributes.end(),
+                          std::back_inserter(mixed_attributes));
+    if (!mixed_attributes.empty())
+    {
+      MFEM_WARNING("Found boundary attribute with internal and external boundary elements: "
+                   << fmt::format("{}", fmt::join(mixed_attributes, " "))
+                   << ". Impedance boundary conditions for these attributes will give "
+                      "erroneous results, consider separating into different attributes!");
+    }
     MFEM_VERIFY(crack_bdr_elem.empty() || !orig_mesh->Nonconforming(),
                 "Duplicating internal boundary elements for interior boundaries is not "
                 "supported for nonconforming meshes!");
```

simlapointe: Thanks for reviewing and good catch! Added the warning for mixed internal/external in [72bafcd](https://github.com/awslabs/palace/pull/544/commits/72bafcd7dfb36746826f16d712753519991c225a)

Issue #543, opened 2025-10-28
Spack build currently broken


Sbozzolo:
As seen in https://github.com/awslabs/palace/actions/runs/18883127463. It fails with error:
```

5 errors found in build log:
     180    cd /tmp/runner/spack-stage/spack-stage-scnlib-3.0.1-j4wgylr2ljzv53x
            s2qmu5am55zm7omrt/spack-build-j4wgylr/benchmark/binarysize && /opt/
            spack/linux-x86_64_v3/compiler-wrapper-1.0-kr35wwg7rvdhdh26tpinvbyg
            zmd6fwrz/libexec/spack/gcc/g++   -O3 -DNDEBUG -Wno-unused -Wno-unus
            ed-but-set-variable -Wno-redundant-decls -Wno-missing-declarations 
            -MD -MT benchmark/binarysize/CMakeFiles/scn_benchmark_binarysize_sc
            anf.dir/out/scanf/binarysize_test_tmp_15.cpp.o -MF CMakeFiles/scn_b
            enchmark_binarysize_scanf.dir/out/scanf/binarysize_test_tmp_15.cpp.
            o.d -o CMakeFiles/scn_benchmark_binarysize_scanf.dir/out/scanf/bina
            rysize_test_tmp_15.cpp.o -c /tmp/runner/spack-stage/spack-stage-scn
            lib-3.0.1-j4wgylr2ljzv53xs2qmu5am55zm7omrt/spack-build-j4wgylr/benc
            hmark/binarysize/out/scanf/binarysize_test_tmp_15.cpp
     181    [ 71%] Building CXX object benchmark/binarysize/CMakeFiles/scn_benc
            hmark_binarysize_scanf.dir/out/scanf/binarysize_test_tmp_16.cpp.o
     182    cd /tmp/runner/spack-stage/spack-stage-scnlib-3.0.1-j4wgylr2ljzv53x
            s2qmu5am55zm7omrt/spack-build-j4wgylr/benchmark/binarysize && /opt/
            spack/linux-x86_64_v3/compiler-wrapper-1.0-kr35wwg7rvdhdh26tpinvbyg
            zmd6fwrz/libexec/spack/gcc/g++   -O3 -DNDEBUG -Wno-unused -Wno-unus
            ed-but-set-variable -Wno-redundant-decls -Wno-missing-declarations 
            -MD -MT benchmark/binarysize/CMakeFiles/scn_benchmark_binarysize_sc
[...]

PR #545, opened 2025-10-29
Cache .julia in linux GitHub run

Sbozzolo:
`julia-actions/cache` provides a shortcut to cache the relevant data in `~/.julia`. This should help speed up instantiation when packages do not change.


simlapointe (approved):
Looks good! Seems to have reduced the antenna post-processing test from ~5 mins to <1 min!

Sbozzolo:
I added a workaround for #543 (pinning `fast-float`) so that we can merge PRs. I don't when the spack PR will be merged.

Sbozzolo:
I also had to add a constraint on hypre<3.

PR #546, opened 2025-10-30
Fix typos

Sbozzolo:
I was working on something unrelated and happened to use [codespell](https://github.com/codespell-project/codespell
), so I thought I would use it for palace too.


simlapointe (approved):
LGTM

PR #549, opened 2025-11-03
Add +tests variant to spack

Sbozzolo:
Users installing Palace with spack were unable to run the unit tests. This PR adds a `+tests` variant that compiles the `palace-tests` target.

`palace-tests` now:
- installs the `unit-tests` executable to `bin` as `palace-unit-tests`
- installs the dynamic libraries for catch2
- installs the required data/mesh files to a `share` folder

This last one was required for spack because spack relocates the build, so that the hardcoded paths were pointing to an incorrect folder.

I renamed `unit-tests` to `palace-unit-tests` to avoid conflicts in the namespace if one adds `bin` to their `PATH`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -56,12 +56,8 @@ jobs:
             packages:
               petsc:
                 require: ~hdf5
-              # TODO: Remove this when
-              # https://github.com/spack/spack-packages/pull/2144/ is merged
-              fast-float:
-                require: "@6"
-              hypre:
-                require: "@2"
+            repos:
+            - spack_repo/local
             mirrors:
               spack:
                 binary: true
--- a/.github/workflows/spack.yml
+++ b/.github/workflows/spack.yml
@@ -116,12 +116,6 @@ jobs:
                 require: ~hdf5
               rocblas:
                 require: ~tensile
-              # TODO: Remove this when
-              # https://github.com/spack/spack-packages/pull/2144/ is merged
-              fast-float:
-                require: "@6"
-              hypre:
-                require: "@2"
             repos:
             - spack_repo/local
             mirrors:


cmake/ExternalPalace.cmake
```diff
@@ -99,9 +99,10 @@ ExternalProject_Add(palace
 # Add target for Palace unit tests
 ExternalProject_Add_Step(palace tests
   COMMAND           ${CMAKE_MAKE_PROGRAM} unit-tests
+  COMMAND           ${CMAKE_COMMAND} --install <BINARY_DIR>/test/unit --prefix ${CMAKE_INSTALL_PREFIX}
```
cameronrutherford, line 103: This `<BINARY_DIR>` syntax is new to me... I would've expected like `${CMAKE_BINARY_DIR}`?

docs/src/developer/testing.md
```diff
@@ -31,47 +31,48 @@ test assertions. If you want to also measure test coverage, turn
 `PALACE_BUILD_WITH_COVERAGE` on. See [Unit test coverage](#Unit-test-coverage)
 for more details on this.
 
-Once the build completes, run the tests with:
+Once the build completes, the `palace-unit-tests` executable will be installed in the
+same `bin/` directory as the main `palace` executable. Run the tests with:
```
cameronrutherford, line 35: Two thoughts:

1. We should explicitly call out the need to `make install` (or some equivalent) before running tests. I had plenty of people complain about tests not working and they were missing this step in a previous project.
2. Do we want to support running tests in the build directory w/o needing to install?

spack_repo/local/packages/palace/package.py
```diff
@@ -221,6 +222,7 @@ def cmake_args(self):
             self.define_from_variant("PALACE_WITH_SUNDIALS", "sundials"),
             self.define_from_variant("PALACE_WITH_SUPERLU", "superlu-dist"),
             self.define("PALACE_BUILD_EXTERNAL_DEPS", False),
+            self.define_from_variant("PALACE_MFEM_USE_EXCEPTIONS", "tests"),
```
cameronrutherford, line 228: There isn't a palace CMAKE variable that does along with this that enable/disables tests being built? Traditionally if we have the variant there, it would suggest either building or not building tests based on the flag...

spack_repo/local/packages/palace/package.py
```diff
@@ -273,6 +275,13 @@ def cmake_args(self):
 
         return args
 
+    def build(self, spec, prefix):
+        with working_dir(self.build_directory):
+            if spec.satisfies("+tests"):
+                make("palace-tests")
+            else:
+                make()
```
cameronrutherford, line 286: I suppose this is how we determine what gets built... Perhaps a CMake variable is a sane alternative, but this also makes sense to me as well...

test/unit/CMakeLists.txt
```diff
@@ -95,3 +98,35 @@ target_compile_definitions(
 
 # Status messages for test target
 message(STATUS "Configured unit-tests target for unit tests using Catch2")
+
+# Install Catch2 libraries
+install(
+  TARGETS Catch2
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+)
```
cameronrutherford: Should this be something that we add as a spack dependency? Or is CMake okay here since this is header only (or something like that...)

spack_repo/local/packages/palace/package.py
cameronrutherford, line 1: Perhaps we add something like https://spack.readthedocs.io/en/latest/packaging_guide_testing.html in another issue / PR.... Would need a way to down-select the number of tests since we would want more of a sanity check there than a full unit-test run...

cmake/ExternalPalace.cmake
Sbozzolo: `${CMAKE_BINARY_DIR}` refers to the binary dir for the main project, but here we add an external project. `<BINARY_DIR>` refers to the external project's build directory.

https://cmake.org/cmake/help/latest/module/ExternalProject.html

See here for an example of its use: https://cmake.org/cmake/help/latest/module/ExternalProject.html#miscellaneous-options

docs/src/developer/testing.md
Sbozzolo: Great questions. I think I have already tackled both of them in this PR. This block of code:
```
ExternalProject_Add_Step(palace tests
  COMMAND           ${CMAKE_MAKE_PROGRAM} unit-tests
  COMMAND           ${CMAKE_COMMAND} --install <BINARY_DIR>/test/unit --prefix ${CMAKE_INSTALL_PREFIX}
...
```

Calls `cmake --install`, so that `make install` is done as part of the `make palace-tests` target. This way, tests (and the required files) are always installed.

This might not be the most classical build strategy, but this is what Palace already does (`make` installs Palace)

spack_repo/local/packages/palace/package.py
Sbozzolo: No, we don't have a CMAKE variable. Instead, we have a different make target. So, the variant takes care of building or not building the tests in the `build` step.

test/unit/CMakeLists.txt
Sbozzolo: Good point! I'll add it to spack.

spack_repo/local/packages/palace/package.py
cameronrutherford: Should we make an issue for this?

spack_repo/local/packages/palace/package.py
```diff
@@ -205,6 +206,9 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
             depends_on(f"slepc{rocm_variant}", when=f"+slepc{rocm_variant}")
             depends_on(f"petsc{rocm_variant}", when=f"+slepc{rocm_variant}")
 
+    with when("+tests"):
+            depends_on(f"catch2@3")
```
cameronrutherford: NIT

```suggestion
            depends_on("catch2@3")
```

As a clarifying comment, this implies that we wouldn't be compatible with `catch2@4` if / when that is released.

test/unit/CMakeLists.txt
```diff
@@ -6,12 +6,20 @@
...
-FetchContent_Declare(Catch2
-  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-  GIT_TAG        v3.8.1
-)
-FetchContent_MakeAvailable(Catch2)
+if(PALACE_BUILD_EXTERNAL_DEPS)
+  Include(FetchContent)
+  FetchContent_Declare(Catch2
+    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+    GIT_TAG        v3.8.1
```
cameronrutherford, line 13: Funnily enough, this version isn't in spack...

Should we:

1. Add this to the global `Catch2` spack package?
2. Have a stricter version dependency at the spack level to match this version?

cameronrutherford (changes requested):
Looking good! Some small clarifications before we merge.

docs/src/developer/testing.md
cameronrutherford: Understood. This is sufficient then.

test/unit/CMakeLists.txt
Sbozzolo: Catch2 has several releases and they all mostly minor. 3.8.0 (in spack) is perfectly fine, so I'd say that we can just use that

spack_repo/local/packages/palace/package.py
Sbozzolo: Good point.

cameronrutherford: I like the `3:` as that means that we will instantly get a CI failure once we are incompatible... Bad for getting tests to always pass, but the easiest way to catch upstream issues.

spack_repo/local/packages/palace/package.py
Sbozzolo: Opened here: https://github.com/awslabs/palace/issues/551

Issue #506, opened 2025-10-01
Some of the palace unit test meshes are not immediately reproducible


Sbozzolo:
Palace comes with a few test meshes:
```
fichera-hex.mesh
fichera-mixed-p2.mesh
fichera-tet.mesh
star-mixed-p2.mesh
star-quad.mesh
star-tri.mesh
```

The documentation says:
> these meshes come from the [MFEM repository](https://github.com/mfem/mfem/tree/master/data)

However, only two meshes can be found at that link:
```
fichera-mixed-p2.mesh
star-mixed-p2.mesh
```
Presumably, the other ones used to be part of MFEM but got removed at some point. It might be good to find which MFEM commit/version was used for these meshes. 

PR #481 shows how one can directly use MFEM's meshes in the unit tests. This removes the need to maintain our copy of the meshes, but exposes us to potential problems if MFEM changes/removes meshes we care about.

Sbozzolo:
We need MFEM to build with CMake to use the MFEM data folder directly. This is blocked by https://github.com/spack/spack-packages/pull/143

PR #550, opened 2025-11-03
Compile MFEM externally

Sbozzolo:
This commit puts MFEM on the same footing as the other dependencies. Instead of being unconditionally built, it is now build when building external dependencies. This allowed me to add MFEM as a dependency managed spack.

In doing this, I found a new bug in spack:
https://github.com/spack/spack/issues/51505

So, I changed our approach to testing the spack package. Instead of adding a local repo, I overwrite palace in the `builtin` repo.

As a side effect, this makes the spack CI action even faster, because now MFEM can be cached.

Among the challenges I faced in implementing this is that hypre and mfem are currently being built with autotools (in spack). This makes certain things hard/annoying. For example, passing through the `MFEM_DATA_PATH` turned out to be a pain, so I decided to vend the two meshes instead (we are already doing that for other meshes).

It should be noted that hypre have switched to cmake with release 3.0, and that there's a PR open in spack-packages to use cmake in MFEM.


.github/workflows/spack.yml
```diff
@@ -188,6 +189,14 @@ jobs:
           palace-unit-tests --skip-benchmarks
           mpirun -np 2 $(which palace-unit-tests) --skip-benchmarks
 
+      - uses: julia-actions/cache@v2
+        if: needs.filter.outputs.test == 'true'
+
+      - uses: julia-actions/setup-julia@v2
+        if: needs.filter.outputs.test == 'true'
+        with:
+          version: '1'
```
Sbozzolo: Not related to mfem, but it speeds things up by a minute or so.

test/unit/test-geodata.cpp
```diff
@@ -327,9 +326,8 @@ TEST_CASE("Default IOData", "[iodata][Serial]")
   iodata.domains.materials.back().attributes = {0};
 
   // Pull from the mfem externals data folder.
-  auto ref_tet_path = fs::path(MFEM_DATA_PATH) / "ref-tetrahedron.mesh";
-  mfem::Mesh single_tet(ref_tet_path.string());
-  mfem::ParMesh pmesh(Mpi::World(), single_tet);
+  mfem::Mesh mesh = mfem::Mesh::MakeCartesian3D(1, 1, 1, mfem::Element::TETRAHEDRON);
```
Sbozzolo, line 367: I changed the mesh here because it doesn't really matter what mesh we use, and I preferred picking something that does not depend on a file.

test/unit/CMakeLists.txt
```diff
@@ -86,17 +86,13 @@ set_property(
...
          ${CMAKE_CURRENT_SOURCE_DIR}/test-strattonchu.cpp
+         ${CMAKE_CURRENT_SOURCE_DIR}/test-rap.cpp
   APPEND PROPERTY COMPILE_DEFINITIONS "PALACE_TEST_MESH_DIR=\"${CMAKE_INSTALL_PREFIX}/share/palace/test/mesh\""
 )
-set_property(
-  SOURCE
-  ${CMAKE_CURRENT_SOURCE_DIR}/test-geodata.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/test-rap.cpp
-  APPEND PROPERTY COMPILE_DEFINITIONS "MFEM_DATA_PATH=\"${CMAKE_INSTALL_PREFIX}/share/palace/test/mfem-data\""
-)
```
hughcars, line 100: So this is removing the ability to use mfem reference meshes directly, which seems unfortunate. I presume this is because in a spack build these data files don't persist. I wonder if we can have the reference meshes downloaded directly from the mfem repo during configuration instead, but that might be complicated. Rationale being it would be great to avoid having to check in duplicates of meshes which we can access from mfem, and which aren't changing any time soon.

It seems like if the test target is built we can have a configuration step that copies the meshes from the MFEM github to our build local data folder as part of configuration. We'd need a list of these meshes to be stored in the cmake, but then adding or removing them in future consists only of adding that string, rather than checking in another mesh. I'd basically like to maintain the flexibility of being able to use any mfem vended mesh without copying it into the repo.

.github/workflows/docs.yml
```diff
@@ -36,13 +36,19 @@ jobs:
           buildcache: true
           color: true
 
+        # https://github.com/spack/spack/issues/51505
+      - name: Overwrite builtin Palace package
+        run: |
+          cp -r spack_repo/local/packages/palace "$(spack location --repo builtin)/packages"
+          cp extern/patch/mfem/* "$(spack location --repo builtin)/packages/palace"
```
cameronrutherford: Interesting that we have patches in this location, even though spack wants to use them... I suppose once we merge this upstream for users it won't be that confusing of a end user workflow, but I wonder if we should have something in our docs about local development in this fashion, and how to iterate on mfem patches (maybe we do and I just haven't read it).

.github/workflows/spack.yml
```diff
@@ -51,8 +51,9 @@ jobs:
         compiler: [gcc] #, clang, intel]
         gpu: [none] #, cuda, rocm]
         blas: [openblas] # ,intel-oneapi-mkl]
-        spack_version: [develop] # ,v0.23.1]
         mpi: [openmpi] #, mpich, mvapich]
+        eigensolver: [arpack, slepc]
+        solver: [mumps, strumpack, superlu-dist]
```
cameronrutherford: This is perhaps something we just have to take a stance on, but there are likely always going to be build configurations that we miss out on.

For example before this change, we had:

-` @develop+strumpack+arpack+mumps+tests`

Now we have:

- `@develop+strumpack+arpack`
- `@develop+mumps+arpack`
- `@develop+superlu-dist+arpack`

And the same for `slepc`...

However, note that `+slepc` is the default for palace, and so is `+superlu-dist`, so those variants will always be enabled.

The question(s) then are:

- Do we want to perhaps change this matrix to be like `[+arpack~slepc, ~arpack+slepc, +arpack+slepc]`? This gets more complicated for the solver too...
- What (eigen)solver is used in testing when more than one is enabled?
- Do we want to test the full matrix of builds here? What's an acceptable subset?

I think there are some logical answers to these, but wanted to avoid adding my bias...

.github/workflows/spack.yml
cameronrutherford: Nice!! Should we add this to the docs workflow too?

Seems like quite a bit is shared between these workflows... I wonder if there's a nice way to share these files more...

1. We could create our own action... Seems excessive
2. We could combine the workflows into one file, and run with flags...

I don't think we should really do this, but it was a thought....

CHANGELOG.md
```diff
@@ -68,9 +68,23 @@ The format of this changelog is based on
...
+
+  - When building on a machine with a GPU, `CMAKE_CUDA_ARCHITECTURES` has now a
+    default value that is automatically determined to be the consistent with the
+    GPU. [PR 530](https://github.com/awslabs/palace/pull/530).
+  - The spack Palace build now supports a new variant `+tests` to compile and
+    install the unit tests. [PR
+    549](https://github.com/awslabs/palace/pull/549).
+  - MFEM is no longer unconditionally compiled alongside of Palace and can now
+    be built separately. This is used in the Palace spack build, which now
+    explicitely depends on MFEM. [PR
```
cameronrutherford:
```suggestion
    explicitly depends on MFEM. [PR
```

CHANGELOG.md
```diff
@@ -68,9 +68,23 @@ The format of this changelog is based on
     [PR 510](https://github.com/awslabs/palace/pull/510) and [PR 526](https://github.com/awslabs/palace/pull/526).
   - Update EM constants to CODATA Recommended Values of the Fundamental Physical Constants 2022
     [PR 525](https://github.com/awslabs/palace/pull/525).
+  - Fix compatibility with CUDA 12 and GCC 14+. [PR 535](https://github.com/awslabs/palace/pull/535).
   - Scaled Rs/Ls/Cs of impedance boundary conditions affected by mesh cracking, fixing bug where
     `"CrackInternalBoundaryElements"` would lead to incorrect results. [PR 544](https://github.com/awslabs/palace/pull/544).
 
+#### Build system
+
+  - When building on a machine with a GPU, `CMAKE_CUDA_ARCHITECTURES` has now a
```
cameronrutherford, line 92: NIT: I don't think you need to build w/ a GPU... Perhaps we mention the default value here for when no GPU (or more specifically `nvidia-smi`) isn't detected?

test/unit/mesh/ref-tetrahedron.mesh
cameronrutherford, line 1: I assume this file is fine to add and is just a canonical mesh?

palace/CMakeLists.txt
```diff
@@ -112,6 +112,15 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
...
 find_package(MPI REQUIRED)
 
+
+if(PALACE_WITH_SUNDIALS AND NOT PALACE_BUILD_EXTERNAL_DEPS)
+  # Spack's sundials requires MPI with C bindings, so we alias
+  # MPI_C to MPI_CXX (if it does not already exist).
+  if(NOT TARGET MPI::MPI_C AND TARGET MPI::MPI_CXX)
+    add_library(MPI::MPI_C ALIAS MPI::MPI_CXX)
+  endif()
+endif()
```
cameronrutherford: Odd that this isn't the `find_package` default behavior given the MPI C++ bindings were deprecated in 3.0 - https://www.mpi-forum.org/docs/mpi-2.2/mpi22-report/node32.htm

cmake/ExternalPalace.cmake
```diff
@@ -24,12 +23,16 @@ list(APPEND PALACE_OPTIONS
   "-DPALACE_WITH_OPENMP=${PALACE_WITH_OPENMP}"
   "-DPALACE_WITH_SLEPC=${PALACE_WITH_SLEPC}"
   "-DPALACE_WITH_ARPACK=${PALACE_WITH_ARPACK}"
+  "-DPALACE_WITH_SUNDIALS=${PALACE_WITH_SUNDIALS}"
+  "-DPALACE_WITH_GSLIB=${PALACE_WITH_GSLIB}"
   "-DANALYZE_SOURCES_CLANG_TIDY=${ANALYZE_SOURCES_CLANG_TIDY}"
   "-DANALYZE_SOURCES_CPPCHECK=${ANALYZE_SOURCES_CPPCHECK}"
-  "-DMFEM_DATA_PATH=${CMAKE_BINARY_DIR}/extern/mfem/data" # Path to meshes for testing
   "-DPALACE_BUILD_EXTERNAL_DEPS=${PALACE_BUILD_EXTERNAL_DEPS}" # For Catch2
```
cameronrutherford, line 33:
```suggestion
  "-DPALACE_BUILD_EXTERNAL_DEPS=${PALACE_BUILD_EXTERNAL_DEPS}" # For Catch2 and others
```

It isn't just Catch2 that would be built if we add this flag...

palace/CMakeLists.txt
```diff
@@ -141,15 +150,20 @@ if(NOT "${MFEM_DIR}" STREQUAL "")
...
+
+# MFEM in Spack is currently using autotools, and this find_package will fail
+# (hence the QUIET), but this is not a problem.
+find_package(MFEM QUIET CONFIG)
+if(MFEM_FOUND)
+  message(STATUS "Found MFEM: ${MFEM_VERSION} in ${MFEM_DIR}")
+  if(NOT MFEM_USE_MPI)
+    message(FATAL_ERROR "Build requires MFEM with MPI support")
+  endif()
+endif()
```
cameronrutherford:
```suggestion
if(MFEM_FOUND)
  message(STATUS "Found MFEM: ${MFEM_VERSION} in ${MFEM_DIR}")
  if(NOT MFEM_USE_MPI)
    message(FATAL_ERROR "Build requires MFEM with MPI support")
  endif()
elif(NOT "${MFEM_DIR}" STREQUAL "")
  message(STATUS "Found MFEM in ${MFEM_DIR}")
endif()
```

General idea is that it would be nice to still print the dir even if we don't use the superbuild CMake version... Also this will "just work" once we use the CMake mfem spack build...

1. My syntax might be wrong.
2. I think `elif(NOT MFEM_DIR)` would work...
3. Would `MFEM_VERSION` be defined if we use the `autotools` based build? I wouldn't think so, but at least we could print the dir we used
4. Maybe we can parse the autotools config file in order to get the version 😢

palace/CMakeLists.txt
```diff
@@ -141,15 +150,20 @@ if(NOT "${MFEM_DIR}" STREQUAL "")
...
+
+if(NOT "${HYPRE_DIR}" STREQUAL "")
+  set(HYPRE_ROOT ${HYPRE_DIR})
 endif()
-# if(MFEM_CXX_FLAGS)
-#   # Pull compiler flags from MFEM for OpenMP and optimizations
-#   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${MFEM_CXX_FLAGS}")
-# endif()
```
cameronrutherford, line 158: Why was this there in the first place 🤔

spack_repo/local/packages/palace/package.py
cameronrutherford: I think we need @Sbozzolo here 🤔

cameronrutherford (commented):
I might have missed something... LGTM! I think largely just clarifying questions.

.github/workflows/spack.yml
hughcars: Just adding in `arpack` or `strumpack` or `mumps` into the build is not going to change from the default `superlu-dist` `slepc` combination, as they are the defaults are all in. So if the goal is testing those, the defaults need to be explicitly `~`'d.

For the non spack tests, we use a pairwise test matrix, as we can't test every combination, as that would blow up the number ridiculously. Instead pairwise testing gives a reasonable hope at catching failures induced by poor interactions between dependencies, to then be investigated more directly.

spack_repo/local/packages/palace/package.py
```diff
@@ -144,6 +144,33 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
+        depends_on("mfem+strumpack", when="+strumpack")
+        depends_on("mfem~strumpack", when="~strumpack")
+        depends_on("mfem+mumps", when="+mumps")
+        depends_on("mfem~mumps", when="~mumps")
+        depends_on("mfem+sundials", when="+sundials")
+        depends_on("mfem~sundials", when="~sundials")
+        depends_on("mfem+gslib", when="+gslib")
+        depends_on("mfem~gslib", when="~gslib")
+
+        depends_on("mfem+exceptions", when="+tests")
```
hughcars: There's no catch2 dependency? Also add a block for `:0.14+test`, given it's not a backwards compatible flag.

Sbozzolo: The block is already under the `with when("@develop")`, so that it is ignored for previous versions. I expect us to change this to `@0.15:` when we realease it.

And for `catch2`, it is somewhere else:
https://github.com/awslabs/palace/blob/c7cd5ef6f50304498fcbc5a1f025e41636f9eb00/spack_repo/local/packages/palace/package.py#L209-L210

cmake/ExternalPalace.cmake
Sbozzolo: In this case it is only `Catch2`. This block is passing down some flags to the Palace build (not the superbuild), and `PALACE_BUILD_EXTERNAL_DEPS` is only used to control whether do download `Catch2` or not.

test/unit/CMakeLists.txt
Sbozzolo: The reason I had to do this is that MFEM is currently built with autotools (when installed with spack). For this reason, the information on `MFEM_DATA_PATH` is lost, and it is very hard recover. 

I'd be inclined to push this to the future, possibly tackling #506 at the same time.

.github/workflows/spack.yml
Sbozzolo: Yes, added to the docs as well.

palace/CMakeLists.txt
Sbozzolo: ¯\_(ツ)_/¯

CHANGELOG.md
Sbozzolo: Added

spack_repo/local/packages/palace/package.py
hughcars: What I was meaning is that I think `@:0.14+test` should be an error, rather than silent which is what I believe happens here. That catch2 dependency line means that if you did `palace@0.14+test`, it would compile catch2 in the dependencies, but then not generate any tests, which would be a surprise to a user.

Sbozzolo: Got it, thanks!

.github/workflows/spack.yml
Sbozzolo: I added them mostly to check that they compile correctly, but I will try translating our linux matrix to a spack matrix (given that it has better caching -> it's faster), and maybe in the future we can reduce the size of linux matrix.

palace/CMakeLists.txt
Sbozzolo: Interesting! I am not sure about this. We use MPI_CXX in a couple of other places

.github/workflows/docs.yml
Sbozzolo: I added developer documentation to discuss this

cameronrutherford: I guess we do the symlink for developer docs, but then do the `cp` approach for CI... This seems fine but perhaps we have some more consistency across these approaches?

.github/workflows/spack.yml
```diff
@@ -17,7 +17,7 @@ concurrency:
 
 env:
   # Note that each spack version needs its own registry
-  SPACK_VERSION: develop
```
cameronrutherford, line 20: In this case, it seems like we ended up having the bug merged into spack `develop` relatively quickly.

Longer term, it might be nice to have some `develop` pipelines that we allow to fail, although I don't really know how useful that would be, and surely we don't want to waste compute / duplicate work here...

cameronrutherford: Along the same lines, I can't imagine we want to test all spack versions... Perhaps we should advertise somewhere in the repo / docs what version of Spack we are compatible with?

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
+          math-libs: openblas
+          shared: +shared
+          int: +int64
+          openmp: ~openmp
+          eigensolver: +arpack
+          solver: +mumps
+          cuda: ~cuda
 
-    runs-on: palace_ubuntu-latest_16-core
+        # Fails configuring PETSc
```
cameronrutherford: Did we want to file / tag bug reports for each of these?

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
+        #   mpi: mpich
+        #   math-libs: amdblis
+        #   shared: ~shared
+        #   int: ~int64
+        #   openmp: +openmp
+        #   eigensolver: +slepc
+        #   solver: +superlu-dist
+        #   cuda: ~cuda
+
+        # Fails building MFEM with `FC` not found
```
cameronrutherford: Should this be a bug report?

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
+          mpi: mpich
+          math-libs: armpl-gcc
+          shared: +shared
+          int: +int64
+          openmp: +openmp
+          eigensolver: +arpack
+          solver: +strumpack
+          cuda: ~cuda
+
+        # Fails building libxsmm with ifx error None not found
```
cameronrutherford: Maybe this would be solved w/ a composite compiler and using `gfortran` as the fortran compiler...

.github/workflows/spack.yml
cameronrutherford: I also wonder if this one could be solved by adding `export FC=` at the top of the script... That's likely what they are expecting...

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
+          mpi: mpich
+          math-libs: amdblis
+          shared: +shared
+          int: +int64
+          openmp: ~openmp
+          eigensolver: +arpack
+          solver: +mumps
+          cuda: ~cuda
+
+        # Fails building superlu with pointer conversion error int vs int_t
```
cameronrutherford: Should this be a bug report?

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
+        #   mpi: openmpi
+        #   math-libs: amdblis
+        #   shared: ~shared
+        #   int: +int64
+        #   openmp: +openmp
+        #   eigensolver: +arpack
+        #   solver: +superlu-dist
+        #   cuda: ~cuda
+
+        # Fails with undefined reference to symbol 'sqrt@@GLIBC_2.17' in libmetis
```
cameronrutherford: Should this be a bug report? Sounds like a math library linkage issue...

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
     strategy:
       fail-fast: false # don't have one build stop others from finishing
       matrix:
-        compiler: [gcc] #, clang, intel]
-        gpu: [none] #, cuda, rocm]
-        blas: [openblas] # ,intel-oneapi-mkl]
-        spack_version: [develop] # ,v0.23.1]
-        mpi: [openmpi] #, mpich, mvapich]
+        include:  # Pairwise testing (from generate_test_matrix.py)
```
cameronrutherford: Can you maybe explain somewhere how `generate_test_matrix.py` is relevant here?

1. It looks like you commented out some tests here (for good reason), but that isn't from the Python file.
2. I would've figured that you were calling the `.py` file somewhere,  and then using that output to generate CI script... Maybe that's just a GitLab feature, and doesn't work here...

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
-          # Configure GPU variant
-          if [[ "${{ matrix.gpu }}" == 'cuda' ]]; then
-            GPU_VARIANT="+cuda cuda_arch=70"
-          elif [[ "${{ matrix.gpu }}" == 'rocm' ]]; then
-            GPU_VARIANT="+rocm amdgpu_target=gfx900"
+          # Change defaults for solver and eigensolvers (by not including them in the build).
+          if [[ "${{ matrix.solver }}" == "+strumpack" ]]; then
+            SOLVER_VARIANTS="+strumpack~superlu-dist~mumps"
+          elif [[ "${{ matrix.solver }}" == "+mumps" ]]; then
+            SOLVER_VARIANTS="+mumps~superlu-dist~strumpack"
```
cameronrutherford: Would it be interesting / relevant to do a build with both enabled?

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
           else
-            GPU_VARIANT=""
+            SOLVER_VARIANTS="+superlu-dist~strumpack~mumps"
           fi
+          if [[ "${{ matrix.eigensolver }}" == "+arpack" ]]; then
+            EIGEN_VARIANTS="+arpack~slepc"
+          else
+            EIGEN_VARIANTS="+slepc~arpack"
+          fi
```
cameronrutherford: Would enabling both be interesting / relevant?

.github/workflows/spack.yml
```diff
@@ -48,52 +48,160 @@ jobs:
...
+          
+          PALACE_SPEC="palace@develop${SOLVER_VARIANTS}${EIGEN_VARIANTS}${{ matrix.shared }}${{ matrix.int }}${{ matrix.openmp }}${{ matrix.cuda }}+tests"
 
-          # Develop requires ^compiler, not %compiler
-          if [[ "${{ matrix.spack_version }}" == 'v0.23.1' ]]; then
-            PALACE_SPEC="local.palace@develop+strumpack+arpack+mumps+tests %${{ matrix.compiler }} ${GPU_VARIANT} ^${{ matrix.mpi }} ^${{ matrix.blas }}"
+          # LLVM does not provide flang by default, so we use gcc
+          if [[ "${{ matrix.compiler }}" == "llvm" ]]; then
+            FORTRAN_COMPILER="gcc"
```
cameronrutherford: Per above. perhaps we need to run a hacky `export FC=gfortran` here to fix some of the build issues...

.github/workflows/spack.yml
```diff
@@ -105,19 +213,33 @@ jobs:
...
               petsc:
                 require: ~hdf5
-              rocblas:
-                require: ~tensile
-            repos:
-            - spack_repo/local
+              mpi:
+                require: ${{ matrix.mpi }}
+              blas:
+                require: ${{ matrix.math-libs }} threads=openmp
```
cameronrutherford: We can enforce the `threads=openmp` at the `package.py` level if it's critical.

.github/workflows/spack.yml
```diff
@@ -105,19 +213,33 @@ jobs:
...
+                require: [${{ matrix.compiler }}]
+              cxx:
+                require: [${{ matrix.compiler }}]
+              fortran:
+                require: [${FORTRAN_COMPILER}]
+              # Metis is verify fragile
+              # -fPIC seems to be required because metis always compiles some shared objects
+              # -lc seems to be required for intel compilers
+              metis:
+                require: cflags=-fPIC cppflags=-fPIC ldflags=-lc
```
cameronrutherford: If these are strict requirements, we should perhaps move them to the `palace` `package.py`.

.github/workflows/spack.yml
```diff
@@ -105,19 +213,33 @@ jobs:
...
+              parmetis:
+                require: ldflags=-lc
```
cameronrutherford: Same here...

.github/workflows/generate_test_matrix.py
cameronrutherford, line 1: Left a comment in `spack.yaml`, but it's unclear how this is used in relation to the GitHub workflow.

palace/CMakeLists.txt
cameronrutherford: 🤷🏻 I suppose it's an artifact that we leave. No big deal either way, but hopefully the context of my comment makes sense.

spack_repo/local/packages/libceed/package.py
cameronrutherford, line 1: Maybe would be easier to maintain a `patchfile` to apply to existing builtin packages instead of a whole copy of the `package.py`? Unsure... We have to decide on an approach that works for use in cases like this where we are waiting on a PR to be merged. We certainly can't upstream the palace package until the dependent PR is merged, so I don't know if it matters. That being said, a patchfile would make it much clearer what in particular we need to modfiy / how our local copy of the package differs from upstream, whereas this approach makes development trivial / easy enough.

spack_repo/local/packages/palace/package.py
```diff
@@ -50,17 +50,23 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         default=True,
         description="Build with GSLIB library for high-order field interpolation",
     )
-    variant("tests", default=False, description="Build and install unit tests")
+    variant("tests", default=False, description="Build and install unit tests",
+            when="@0.15:")
+
+    # TODO: Add lcov and other packages that might be needed to truly measure
+    # coverage.
```
cameronrutherford: WDYM by this? Are there missing dependencies not expressed in this file? Or is it just that we need to do more work in the future to support this and this is a note alluding to that?

spack_repo/local/packages/palace/package.py
```diff
@@ -156,14 +197,16 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("libceed@0.13:", when="@0.14:")
```
cameronrutherford: Lets have the `libceed` config all in the same block / section of code.

spack_repo/local/packages/palace/package.py
```diff
@@ -50,17 +50,23 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
     # Fix API mismatch between libxsmm@main and internal libceed build
     patch("palace-0.12.0.patch", when="@0.12")
 
-    # NOTE: We can't depend on git tagged versions here
-    #       https://github.com/spack/spack/issues/50171
-    #       Instead, version in environment / spec
     depends_on("c", type="build")
     depends_on("cxx", type="build")
-    depends_on("cmake@3.21:", type="build")
+    depends_on("cmake@3.24:", type="build")
```
cameronrutherford: Is this a palace version thing? i.e. does 0.14 require `3.21+` but develop is `3.24+`? Is there a feature that prompted this?

spack_repo/local/packages/palace/package.py
```diff
@@ -156,14 +197,16 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("libceed@0.13:", when="@0.14:")
 
     with when("+sundials @0.14:"):
-        depends_on("sundials")
+        depends_on("sundials+mpi+lapack")
+        depends_on("sundials~examples~examples-install")
```
cameronrutherford:
```suggestion
        depends_on("sundials+mpi+lapack~examples~examples-install")
```

Call me crazy...

spack_repo/local/packages/palace/package.py
```diff
@@ -156,14 +197,16 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
         depends_on("sundials+shared", when="+shared")
         depends_on("sundials~shared", when="~shared")
         depends_on("sundials+openmp", when="+openmp")
         depends_on("sundials~openmp", when="~openmp")
         depends_on("sundials~cuda", when="~cuda")
         depends_on("sundials~rocm", when="~rocm")
 
+    conflicts("+tests", when="@:0.14", msg="Building tests is only supported for Palace versions after 0.14")
```
cameronrutherford: I think @hughcars might have suggested this but I am pretty sure that `when="@0.15:"` at the variant level takes care of this conflict, although I guess this is a nice error message?


```suggestion
```

docs/src/developer/spack.md
```diff
@@ -0,0 +1,170 @@
...
+get started.
+
+## Developing Palace with Spack Environments
+
+The recommended approach for developing *Palace* with Spack is to use Spack
+environments. An environment is essentially a self-contained workspace defined
+by a `spack.yaml` file that specifies exactly what you want to build and how you
+want to build it. This keeps your development setup isolated and reproducible.
+
+To get started, create a dedicated directory for your environment. You might
```
cameronrutherford, line 32: NIT - you don't technically ever need to do directory based environments, but I like them as they keep me sane.

Otherwise they are stored in some cursed location in `$SPACK_ROOT`, and then likely somehow tied to `~/.spack`... I don't think you should comment about this rabbit hole...

See:

- https://spack-tutorial.readthedocs.io/en/latest/tutorial_environments.html#managed-versus-independent-environments

I would point to `spack env --help` as I can't find other links / docs on this:

```
    track        track an environment from a directory in Spack
    untrack      track an environment from a directory in Spack
```


Track / untrack is useful for when you want to `spack uninstall` and not nuke things that are being used by environments.... Again quite a rabbit hole...

docs/src/developer/spack.md
```diff
@@ -0,0 +1,170 @@
...
+
+```sh
+spack -e palace_spack external find --all
+```
+
+This command will modify your `spack.yaml` file to include references to system
+packages like compilers, MPI implementations, and common libraries. Spack will
+then use these instead of building them from scratch, which can save hours on
+the first build.
+
```
cameronrutherford, line 66: We could make our GHCR cache public, and then include some notes about using that cache here / making sure re-use is true, enabling splicing and all that good stuff.... perhaps that's a longer conversation, but I also think that if we just add GPG signing we could do this today.

docs/src/developer/spack.md
```diff
@@ -0,0 +1,170 @@
...
+you make changes to the *Palace* source code, you can rebuild by simply running
+`spack install` again from within the environment. Spack will detect what's
+changed and recompile only what's necessary.
+
+The beauty of this development mode is that it extends to other packages as
+well. If you need to modify MFEM alongside *Palace*, you can add another entry
+to the `develop` section of your `spack.yaml` pointing to your local MFEM
+checkout. Spack will then manage both packages in development mode, rebuilding
+each as needed when you make changes.
+
```
cameronrutherford, line 105: You could also include some note here if you want about using the CMake based build after the first spack build. i.e. you can do something like this once you have the first build done and you are just modifying Palace source code (very rough commands from memory):

```bash
cd ~/repos/palace
cd palace-<spack-hash>-build
source spack-build-env.txt
# optionally
make clean
make -j
``` 

This way you can iterate without calling out to Spack / Python interpreter which is moderately faster when doing small / frequent iterations - certainly way faster now that this doesn't trigger an MFEM rebuild.

docs/src/developer/spack.md
```diff
@@ -0,0 +1,170 @@
...
+The first step is to add *Palace*'s Spack repository to your Spack
+configuration. Assuming your *Palace* checkout is in `~/repos/palace`, you would
+run:
+
+```sh
+spack repo add ~/repos/palace/spack_repo/local
+```
+
+You should see a confirmation message like `==> Added repo with namespace 'local'.` The namespace `local` is significant because it allows you to
```
cameronrutherford, line 121: NIT - your line length is too long... Maybe we need a markdown formatter.

docs/src/developer/spack.md
```diff
@@ -0,0 +1,170 @@
...
+distinguish between your development version of the package and the official one
+in Spack's built-in repository. When you refer to `local.palace`, you're
+explicitly using the recipe from your personal repository, while
+`builtin.palace` refers to the official recipe that ships with Spack.
+
+To verify that everything is working, try making a small change to the package
+file. For example, you could edit `spack_repo/local/packages/palace/package.py`
```
cameronrutherford, line 128: or run `spack edit local.palace` in order to open the file in your `$EDITOR`.

docs/src/developer/spack.md
```diff
@@ -0,0 +1,170 @@
...
+version and the official one.
+
+One important detail is that the `local` repository takes priority over
+`builtin`. This means that if you just run `spack info palace` without a prefix,
+Spack will use your local version. While this is convenient, we'll continue
+using the explicit `local.` prefix in examples for clarity.
+
+#### Working Around a Spack Bug with Patches
+
+There's currently a known issue in Spack version 1.1.0 that affects packages
```
cameronrutherford:
```suggestion
There's currently a known issue in Spack version 1.0.1 that affects packages
```

I think this version is wrong... Maybe you also actually mean `1.0.2`?

docs/src/developer/spack.md
```diff
@@ -0,0 +1,170 @@
...
+
+If you see MFEM's package information, you're all set. This workaround ensures
+that when Spack processes patches for *Palace*, it can find MFEM in the same
+repository namespace and avoid the bug.
+
+## Iterating on Package Changes
+
+With your local repository configured, you can now iterate on *Palace*'s
+`package.py` file. After making changes, you can test them by installing
+*Palace* again. You can also use `spack spec local.palace@develop` to see how
```
cameronrutherford, line 178: I think `spack spec` and `spack concretize` would be doing the same thing here, but `spack spec` is nice because you get more / better information IIRC... Not sure if you should change anything here...

.github/workflows/spack.yml
Sbozzolo: Oops, actually this was an attempt to debug the OpenMP issues. It is not needed.

.github/workflows/spack.yml
Sbozzolo: Unfortunately, the official [getting started](https://spack.readthedocs.io/en/latest/getting_started.html) page for spack always checks out the most recent `develop`, which is not great.

I opened this: https://github.com/spack/spack/issues/51547

.github/workflows/spack.yml
Sbozzolo: I don't think we can easily produce a good bug report. We don't have a minimal reproducer and we can only point to some failed github actions jobs, which won't really help the maintainers. 

Same for the others below. I think we should do our due diligence before opening bug reports.

.github/workflows/spack.yml
Sbozzolo: I had tried but it didn't work.

.github/workflows/spack.yml
Sbozzolo: I thought about that, but also thought that part of the point of the test is to verify that we can use the intel compiler, including ifort.

.github/workflows/spack.yml
Sbozzolo: Same comment as above on good bug reports.

.github/workflows/spack.yml
Sbozzolo: Same comment as above on good bug reports.

.github/workflows/spack.yml
Sbozzolo: `generate_test_matrix.py` uses `allpairspy` to combine options to generate good coverage with reduced number of tests. We used `allpairspy` for our `linux` tests as well, but the script was not saved, making the matrix not reproducible and hiding the underlying constraints. 

I added `generate_test_matrix` so that it is clear how the test matrix was chosen and what assumptions went in. This becomes relevant if one wants to add/remove options down the line.

.github/workflows/spack.yml
Sbozzolo: No, this is Hugh's comment from above: if we don't disable certain solvers, the tests won't use the other ones (because we have certain defaults and we don't change the JSON files). We must disable certain solver so that we can test the others.

.github/workflows/spack.yml
Sbozzolo: Same as SOLVER_VARIANTS

.github/workflows/spack.yml
Sbozzolo: I had tried this and it didn't work.

.github/workflows/spack.yml
Sbozzolo: I think that these are bugs in the build system for metis/superlu. They are not strictly required, and the right thing to do is probably fixing them upstream. I didn't do it because it is substantial work.

.github/workflows/spack.yml
cameronrutherford: Right but what I am trying to get at is this explanation should be written into the code somehow, or perhaps in a README in the `.github/workflows` folder. As it stands, I would argue it's unclear how that file is necessary at all.

spack_repo/local/packages/palace/package.py
Sbozzolo: This package.py turns on the one cmake knob needed to enable coverage in Palace. However, there might be extra packages needed to truly extract coverage and make and html report. This is partially related to #538, so we will need to see.

spack_repo/local/packages/palace/package.py
Sbozzolo: This was missed in #398

.github/workflows/spack.yml
cameronrutherford: I don't think it's unreasonable to at least open a bug report with a link to the failing pipeline, even if you don't have a strong reproducer / minimum example. The more complex example that runs in a GitHub workflow is still useful as a bug report to our dependent projects IMO, and it's easy enough for them to reproduce if they spin up their own GitHub action.

I am also suggesting this as it would be good to have some tracking of the bug... If we don't open a downstream issue, then we should open an issue in Palace to track the failing builds?

That, or we should minimize and officially state our supported platforms.

.github/workflows/spack.yml
cameronrutherford: I agree in terms of a runtime test, but sometimes there are weird bugs that can happen at just the build stage, not the test stage. It would be unfortunate if someone added a bug that breaks the build with both solvers enabled, forcing you to have 2 builds to benchmark and compare solvers.

It would be an even more obscure bug to happen at the test stage as well.

docs/src/developer/spack.md
Sbozzolo: No, I meant 1.1.0, which, until recently, was the tip of `develop`

.github/workflows/spack.yml
cameronrutherford: If they aren't strictly required, can we just remove them?

If the solution is to fix upstream, the least we can do is then file a bug report, enforce this in the palace package.py, and then add a comment to our `package.py` referencing the issue.

spack_repo/local/packages/palace/package.py
cameronrutherford: So is it sufficient then to say (I would need to check if this was in `0.14`) that Palace up to 0.14 requires 3.21:, but 0.15: requires 3.24:? If so, our package should state as much.

docs/src/developer/spack.md
Sbozzolo: They are not the same: with `spack spec` you can ask for a spec for any package/variant, and not necessarily concretize the current env.

.github/workflows/spack.yml
```diff
@@ -182,17 +345,40 @@ jobs:
         run: spack -e . install --only-concrete --keep-stage --show-log-on-error --only package --no-cache
 
       - name: Run Unit Tests
-        if: needs.filter.outputs.test == 'true' && matrix.gpu == 'none' # Skip tests for GPU builds
+        if: needs.filter.outputs.test == 'true'
+        env:
+          OMP_NUM_THREADS: 2
```
hughcars: Always using threads? Looks like this gets overridden below

.github/workflows/spack.yml
Sbozzolo: Yes, I think we should track them in Palace, and possibly report them later to the relevant downstream project. For example, at this point, it is unclear where the issues are (is it the spack package, the actual build system, ...)

.github/workflows/spack.yml
Sbozzolo: We can add a case that builds everything.

.github/workflows/spack.yml
Sbozzolo: Got it work with https://spack.readthedocs.io/en/latest/env_vars_yaml.html

.github/workflows/spack.yml
Sbozzolo: No, we cannot remove them. What I mean, they are not strictly required for the package itself, but they become required in our setup. For example, `-lc` is to link to the C standard library. If I don't add this, compilation fails with some linker issues with the Intel compiler. I cannot tell you if this is (a) an upstream issue in the metis build system, (b) an issue of the intel compiler, (c) an issue in the upstream package.py, (d) an issue of our particular setup (e.g., when the intel compiler got confused with the Android NDK).

spack_repo/local/packages/libceed/package.py
Sbozzolo: I tried with a patchfile and it became very messy, so I ended up landing on copying over the files we need to our repo.

docs/src/developer/spack.md
Sbozzolo: Yes, I think this is for the future

spack_repo/local/packages/palace/package.py
Sbozzolo: Right, thank you for pointing that out

.github/workflows/spack.yml
cameronrutherford: Hmmm I think that would suggest that this is an edge case bug with the intel compilers, but to your point, hard to narrow down past that. It would be nice to verify this and perhaps only conditionally apply these things based on compiler, but that's probably overkill.

spack_repo/local/packages/palace/package.py
Sbozzolo: Yes, I think it's good to have this error message

Sbozzolo:
I found that some static builds were broken (because our build system relied on transitively linking libraries through MFEM) and fixed that.

In this, I also found that the package.py for SUNDIALS in the released `spack-packages` is broken because it refers to an non-existing commit. This was fixed [here](https://github.com/spack/spack-packages/commit/9b37b243d47faed5d54dd194f6e7d15332cc8fcb), but it is not part of the released packages. So, as of right now, I have resorted to just copying the spack recipes we need with the fixes we need. We can discuss how to best do this.

Following @cameronrutherford's suggestion, I added an `env` file and managed to get another combination to build, but one of the regression tests fails: https://github.com/awslabs/palace/actions/runs/19281378654/job/55133032500
I tried reproducing the same palace spec locally to see if the issue was with the choice of solvers, but the test passes locally.

I also think I might have found an issue with the PETSc configure error in `-lomp` not being passed depending on the config for petsc/superlu, but I could not verify this on GitHub actions.

.github/workflows/spack.yml
hughcars: One suggestion would be to open the relevant issue and then link it here, or at least the issue number.

.github/workflows/spack.yml
hughcars: Which version of the interl compiler stack was failing? There was a problem I kept running into with the intel stack and fortran a long time ago https://github.com/awslabs/palace/issues/295 I'm guessing this is tangential but just in case. Totally fine disabling the test in some fashion and linking against an issue.

.github/workflows/spack.yml
```diff
@@ -131,9 +282,61 @@ jobs:
...
+        with:
+          mpi: intelmpi
+
+      - name: Configure Intel oneAPI compiler
+        if: needs.filter.outputs.test == 'true' && matrix.compiler == 'intel-oneapi-compilers'
+        run: |
+          sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp \
+                                  intel-oneapi-compiler-fortran \
+                                  intel-oneapi-mpi \
+                                  intel-oneapi-mpi-devel
```
hughcars: Do you think there's an argument for the compiler specifications being part of the spack build themselves? I.e. can we have `^gcc@15.0` or the equivalent for the intel/clang (https://spack.readthedocs.io/en/latest/build_systems/inteloneapipackage.html covers the oneapi for instance), but use the version built by spack? I could imagine three possible benefits: 1) spack configures the compilers to work with everything downstream appropriately, so no extra environment things to do, 2) keeping the versions controlled more precisely than "whatever's coming out of apt", as in the past we've been stung by things like intel updating major versions under our feet, requiring rollbacks/pinning (for example https://github.com/awslabs/palace/issues/295) and 3) if we want to say we support a min/max of gcc versions for instance, we could have those two as matrix values in the CI. 

Can you see other possible upside/downsides to this type of approach? It does lean a lot more on spack for things, but in theory the CI is then full, from scratch, specs that are being blessed by us. As in theory each of those compilers would be building from source with the compiler present within the system to begin with. It would also lean a lot more on the spack cache. I'm not coming down hard on either side, it's just an idea to consider.

docs/src/developer/spack.md
```diff
@@ -0,0 +1,173 @@
...
+
+The first step is to add *Palace*'s Spack repository to your Spack
+configuration. Assuming your *Palace* checkout is in `~/repos/palace`, you would
+run:
+
+```sh
+spack repo add ~/repos/palace/spack_repo/local
+```
+
+You should see a confirmation message like `==> Added repo with namespace 'local'.` The namespace `local` is significant because it allows you to
```
hughcars, line 121: Nit: line length.

docs/src/developer/spack.md
```diff
@@ -0,0 +1,173 @@
...
+
+To verify that everything is working, try making a small change to the package
+file. For example, you could edit `spack_repo/local/packages/palace/package.py`
+and change the docstring to all capital letters. Then run:
+
+```sh
+spack info local.palace | head -n 4
+```
+
+You should see your modified docstring. Compare this with the output of `spack info builtin.palace | head -n 4` to see the difference between your local
```
hughcars, line 135: Nit: line length.

docs/src/developer/spack.md
hughcars: Wouldn't `touch spack_repo/local/packages/palace/package.py` work to trigger this without changing a thing?

docs/src/developer/spack.md
```diff
@@ -0,0 +1,173 @@
...
+
+There's currently a known issue in Spack version 1.1.0 that affects packages
+using patches when they come from non-builtin repositories (see the [bug
+report](https://github.com/spack/spack/issues/51505)). Since *Palace* uses
+patches, you'll run into this problem. The workaround is straightforward but a
+bit inelegant: you need to add a copy of the MFEM package to your local
+repository as well.
+
+Rather than maintaining a duplicate copy of the MFEM recipe, you can create a
+symbolic link from Spack's built-in MFEM package to your local repository:
```
hughcars, line 156: This PR is introducing a copy into the local spack_repo, so this instruction I don't think makes sense any more, given this command would clash.

test/unit/test-rap.cpp
```diff
@@ -26,8 +26,7 @@ TEST_CASE("BuildParSumOperator", "[rap][Serial][Parallel]")
   IoData iodata(units);
   iodata.domains.materials.emplace_back().attributes = {1};
 
-  // Pull from the mfem externals data folder.
-  auto ref_tet_path = fs::path(MFEM_DATA_PATH) / "ref-tetrahedron.mesh";
+  auto ref_tet_path = fs::path(PALACE_TEST_MESH_DIR) / "ref-tetrahedron.mesh";
```
hughcars: If you really wanted, we can construct a single tet mesh directly. Functions to do so would probably be fairly simple and could get a lot of reuse. That would avoid the `ref-tetrahedron.mesh` being needed. We can also temporarily punt on this until the mfem cmake spack pr goes through then can go back to pulling their data directly as you've described previously.

spack_repo/local/packages/sundials/package.py
```diff
@@ -0,0 +1,798 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+import sys
```
hughcars: I'm not a fan of effectively forking of these three spack recipes, not least with the `spack.md` suggesting a user/developer could get away with softlinking the builtin recipe if needed. Can you explain the rationale for this again and why we can't use the builtin right now? And if so, what's the path to not having these recipes to maintain? It's another thing to track pulling updates to these from `spack-packages` into here.

palace/CMakeLists.txt
```diff
@@ -137,19 +137,31 @@ find_package(Eigen3 REQUIRED CONFIG)
 message(STATUS "Found Eigen: ${Eigen3_VERSION} in ${Eigen3_DIR}")
 
 # Find MFEM (recent MFEM targets link to target Threads::Threads, for some reason)
+find_package(Threads REQUIRED)
+
+# MFEM in Spack is currently using autotools, and this find_package will fail
+# (hence the QUIET), but this is not a problem.
```
hughcars, line 184: Hopefully mfem will be merging the cmake pr shortly after 4.9, which is before EOY, so we can fix this shortly after the hypre 3 upgrade which is necessary before mfem 4.9

docs/src/developer/spack.md
```diff
@@ -0,0 +1,173 @@
...
+repository lives, then creates a symlink to the MFEM package directory. To
+verify that the symlink was created correctly, run:
+
+```sh
+spack info local.mfem | head -n 4
+```
+
+If you see MFEM's package information, you're all set. This workaround ensures
+that when Spack processes patches for *Palace*, it can find MFEM in the same
+repository namespace and avoid the bug.
```
hughcars, line 172: Didn't get added on the review, but the spirit is in the main message:

Tried this out to do an update of the mfem dependency, and one quirk is doing `spack develop mfem` starts the aim from master, which causes the patching to fail. It's easy enough going into the folder and doing `git checkout <sha>` if you know where the sha is kept `ExternalGitTags` or the spack recipe, but it doesn't work out the box. 

Given this is one of the real strengths of this external mfem development, it might be worth adding a small section here mentioning this small pitfall, and maybe also that `spack concretize --force` is especially necessary if changing the patches themselves

test/unit/mesh/ref-tetrahedron.mesh
```diff
@@ -0,0 +1,35 @@
+MFEM mesh v1.0
```
hughcars: An alternative is a ctor that uses this:
```

/**
 * @brief Create a single tet mesh, equivalent to ref-tetrahedron.mesh
 *
 * @return Mesh
 */
mfem::Mesh SingleTetMesh()
{
  auto mesh = mfem::Mesh(3, 4, 1, 4);
  mesh.AddVertex(0.0, 0.0, 0.0);
  mesh.AddVertex(1.0, 0.0, 0.0);
  mesh.AddVertex(0.0, 1.0, 0.0);
  mesh.AddVertex(0.0, 0.0, 1.0);
  mesh.AddTet(0, 1, 2, 3);
  mesh.AddBdrTriangle(1, 2, 3, 1);
  mesh.AddBdrTriangle(0, 3, 2, 2);
  mesh.AddBdrTriangle(0, 1, 3, 3);
  mesh.AddBdrTriangle(0, 2, 1, 4);
  mesh.FinalizeTopology();
  mesh.Finalize(true,false);
  return mesh;
}

TEST_CASE("ReferenceTetrahedron", "[geodata][Serial]")
{
  auto tet1 = SingleTetMesh();
  auto ref_tet_path = fs::path(MFEM_DATA_PATH) / "ref-tetrahedron.mesh";
  mfem::Mesh tet2(ref_tet_path.string());

  REQUIRE(tet1.GetNV() == tet2.GetNV());
  REQUIRE(tet1.GetNE() == tet2.GetNE());
  for (int i = 0; i < tet1.GetNV(); i++)
  {
    REQUIRE_THAT(tet1.GetVertex(i)[0], WithinAbs(tet2.GetVertex(i)[0], 1.0e-12));
    REQUIRE_THAT(tet1.GetVertex(i)[1], WithinAbs(tet2.GetVertex(i)[1], 1.0e-12));
    REQUIRE_THAT(tet1.GetVertex(i)[2], WithinAbs(tet2.GetVertex(i)[2], 1.0e-12));
  }
  for (int i = 0; i < tet1.GetNE(); i++)
  {
    for (int j = 0; j < 4; j++)
    {
      CHECK(tet1.GetElement(i)->GetVertices()[j] == tet2.GetElement(i)->GetVertices()[j]);
    }
  }
  for (int i = 0; i < tet1.GetNBE(); i++)
  {
    for (int j = 0; j < 3; j++)
    {
      CHECK(tet1.GetBdrElement(i)->GetVertices()[j] ==
              tet2.GetBdrElement(i)->GetVertices()[j]);
    }
  }
}
```
This test passes when I build it.

.github/workflows/spack.yml
Sbozzolo: This is a spack specific problem, due to misidentification of the patch versions of the intel compilers

docs/src/developer/spack.md
Sbozzolo: I think this is something that is handled by out `format-source` script. In some cases, it forces some long lines.

docs/src/developer/spack.md
Sbozzolo: I removed the local copy in our spack repo and do everything with links now.

spack_repo/local/packages/sundials/package.py
Sbozzolo: Removed all of these in favor of PRs and patches.

test/unit/mesh/ref-tetrahedron.mesh
Sbozzolo: Thank you!

Sbozzolo:
I think I addressed the comments and changed to base to be #578 (which, in turn, is against #498). This also verified that the change compiles fine on GPUs and with the new hypre.

Some Cmake work was needed to make sure everything compiled fine with static builds as well (before, MFEM would carry a bunch of dependencies transitively).

docs/src/developer/spack.md
```diff
@@ -0,0 +1,228 @@
...
+easiest way is probably to remove all the patches and re-apply the ones you need:
+
+```bash
+cd $(spack location --repo builtin)
+git fetch
+git reset --hard origin/develop
+# Apply relevant patches
+```
+
+Patches are often more less common build options, so you might not need all of
```
hughcars: Nit: often more less

Sbozzolo: I had to re-read this sentece a couple of times to understand what I was saying, so I rephrased as:
```
It is often the case that patches are required for uncommon build configurations, so you might not need all of them.
```

.github/actions/setup-runner/action.yml
```diff
@@ -101,6 +101,26 @@ runs:
...
+              echo "Please remove $patch from the Palace repo if the PR was merged"
+            else
+              echo "Applying patch: $patch"
+              git apply "$patch"
+            fi
+          fi
+        done
+
+        # Workaround for https://github.com/spack/spack/issues/51505
+        ln -s "$(spack location --repo builtin)"/packages/mfem "$PALACE_DIR"/spack_repo/local/packages/
```
cameronrutherford: Ah.... So this is where the `.diff` files are used / come in. Makes sense why there is still some template code in other PRs then.

CHANGELOG.md
```diff
@@ -87,6 +87,20 @@ The format of this changelog is based on
...
+    default value that is automatically determined to be the consistent with the
+    GPU (when no GPU is detected, `CMAKE_CUDA_ARCHITECTURES` is set to 70). [PR
+    530](https://github.com/awslabs/palace/pull/530).
+  - The spack Palace build now supports a new variant `+tests` to compile and
+    install the unit tests. [PR
+    549](https://github.com/awslabs/palace/pull/549).
```
cameronrutherford: I think this isn't true anymore as we now support the `--test root` style instead of it being an explicit variant.

test/unit/test-geodata.cpp
```diff
@@ -22,6 +24,41 @@ using json = nlohmann::json;
 namespace fs = std::filesystem;
 using namespace Catch::Matchers;
 
+// TODO: Add this test when we can access MFEM_DATA_PATH from Spack
+// This requires MFEM to move to a CMake-based build system for spack.
```
cameronrutherford, line 28:
```suggestion
// TODO: Add this test when we can access MFEM_DATA_PATH from Spack
// This requires MFEM to move to a CMake-based build system for spack.
// Waiting on https://github.com/spack/spack-packages/pull/143
```

cameronrutherford: Would be good to reference the PR we are waiting on here...

.github/actions/setup-runner/action.yml
```diff
@@ -101,6 +101,26 @@ runs:
       run: |
         echo "$(realpath spack/bin)" >> "$GITHUB_PATH"
         echo "SPACK_DISABLE_LOCAL_CONFIG=1" >> "$GITHUB_ENV"
+
+    - name: Patch Spack Packages for known problems
+      shell: bash
+      run: |
+        PALACE_DIR=$PWD
+        cd $(spack location --repo builtin)
+        for patch in "$PALACE_DIR"/spack_repo/patches/*.diff; do
```
cameronrutherford, line 111: Is this directory correct?
```suggestion
        for patch in "$PALACE_DIR"/spack_repo/local/packages/palace/*.diff; do
```

.github/workflows/docs.yml
```diff
@@ -36,18 +36,33 @@ jobs:
           buildcache: true
           color: true
 
+      - name: Patch Spack Packages for known problems
+        shell: bash
+        run: |
+          PALACE_DIR=$PWD
+          cd $(spack location --repo builtin)
+          for patch in "$PALACE_DIR"/spack_repo/patches/*.diff; do
```
cameronrutherford, line 45: I think this is also the wrong dir? Maybe we shouldn't have this duplicated in more than one spot.


```suggestion
          for patch in "$PALACE_DIR"/spack_repo/local/packages/palace/*.diff; do
```

.github/workflows/docs.yml
```diff
@@ -36,18 +36,33 @@ jobs:
           buildcache: true
           color: true
 
+      - name: Patch Spack Packages for known problems
```
cameronrutherford: Maybe this should explicitly say that this is MFEM related.

spack_repo/local/packages/palace/package.py
```diff
@@ -139,6 +137,36 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
 
+    with when("@0.16:"):
+        # +lapack means: use external lapack
+        depends_on(
+            "mfem+mpi+metis+lapack cxxstd=17 commit=bf1b25d82e232909e1d3aa1f416dbdae9931a550",
+            patches=[
+                "patch_mesh_vis_dev.diff",
+                "patch_par_tet_mesh_fix_dev.diff",
+                "patch_gmsh_parser_performance.diff",
+                "patch_race_condition_fix.diff",
```
cameronrutherford, line 157: If spack is applying the patches here, then I don't think that we need to apply them ourselves in the CI above.

Additionally I was trying to figure out if one could use something like https://spack.readthedocs.io/en/latest/package_api.html#spack.package.BaseBuilder.setup_dependent_build_environment or using the stage_dir in order to apply the patches directly based on the `extern/patches`, but this doesn't seem easy... I think your approach is completely fine for the moment...

I just can't quite wrap my head around how this would work / if this works when we end up merging into spack main, but I trust that it will?

cameronrutherford: Maybe worth considering just moving the extern patches into the spack dir, and then having the existing meta-build just refer to the patches in this directory to ensure spack works well...

.github/actions/setup-runner/action.yml
cameronrutherford: I think the symlink is necessary, but I would hope that spack is the thing responsible for applying said patches, not the action itself.

test/unit/test-geodata.cpp
cameronrutherford, line 28: Maybe we should clarify that we are waiting on https://github.com/spack/spack-packages/pull/143 to be merged.

.github/workflows/docs.yml
```diff
@@ -36,18 +36,33 @@ jobs:
...
+        run: |
+          PALACE_DIR=$PWD
+          cd $(spack location --repo builtin)
+          for patch in "$PALACE_DIR"/spack_repo/patches/*.diff; do
+            if [ -f "$patch" ]; then
+              echo "Applying patch: $patch"
+              git apply "$patch"
+            fi
+          done
+
```
cameronrutherford: I think spack should take care of this.

CHANGELOG.md
cameronrutherford: I think this isn't true anymore as we now support the --test root style instead of it being an explicit variant.

spack_repo/local/packages/palace/patch_cuda_hip_compiler_checks.diff
cameronrutherford, line 1: I think this is reasonable, but suggested in my other comment that maybe we should add these to this dir, and have the extern folder reference these instead... I just don't know how this works with spack, but it might work fine! The RPATH part is what is triggering my spidey senses.

Additionally, I couldn't find anywhere on google that shows an example of this syntax, so I am:

1. Surprised this works
2. Wondering if you could have some `/extern/patch/mfen/*.diff` syntax instead to just make this a simpler thing that doesn't need to get updated regularly.

spack_repo/local/packages/palace/package.py
```diff
@@ -249,117 +292,13 @@ def cmake_args(self):
                 )
             )
 
-        # HYPRE is always built with external BLAS/LAPACK and may need Umpire
-        hypre_packages = ["LAPACK", "BLAS"]
-        if self.spec.satisfies("^hypre+umpire"):
-            hypre_packages.append("Umpire")
-        if self.spec.satisfies("+cuda"):
-            hypre_packages.append("CUDAToolkit")
         args.extend(
```
cameronrutherford: I guess all of these deleted lines are implemented in #519

spack_repo/local/packages/palace/package.py
```diff
@@ -188,11 +216,13 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         for arch in CudaPackage.cuda_arch_values:
             cuda_variant = f"+cuda cuda_arch={arch}"
 
-            # TODO: Remove me after blt > 0.7.1 is released
-            depends_on(f"blt@develop")
+            # We need https://github.com/llnl/blt/pull/735, which is not available
+            # in <= 0.7.1
+            depends_on("blt@0.7.2:")
```
cameronrutherford: Should this be in `+rocm` as well?

palace/CMakeLists.txt
```diff
@@ -52,9 +57,36 @@ set(SLEPC_DIR "" CACHE STRING
...
+)
+set(SCALAPACK_DIR "" CACHE STRING
+  "Path to ScaLAPACK build or installation directory (not required if already on CMAKE_PREFIX_PATH)"
+)
 
 # Enable Fortran if required
-if(PALACE_WITH_ARPACK)
+if(PALACE_WITH_ARPACK OR PALACE_WITH_STRUMPACK OR PALACE_WITH_MUMPS)
   enable_language(Fortran)
 endif()
```
cameronrutherford, line 91: I think this might need to also be updated in our `package.py`? IIRC we only have `+fortran` enabled for one solver... Might be wrong 😅

palace/CMakeLists.txt
```diff
@@ -143,26 +178,45 @@ find_package(Eigen3 REQUIRED CONFIG)
 message(STATUS "Found Eigen: ${Eigen3_VERSION} in ${Eigen3_DIR}")
 
 # Find MFEM (recent MFEM targets link to target Threads::Threads, for some reason)
+find_package(Threads REQUIRED)
```
cameronrutherford, line 181: Does this need to be in `package.py`?

palace/CMakeLists.txt
```diff
@@ -330,14 +384,217 @@ if(PALACE_WITH_OPENMP)
...
+  elseif(TARGET scalapack AND NOT TARGET SCALAPACK::SCALAPACK)
+    add_library(SCALAPACK::SCALAPACK ALIAS scalapack)
+    message(STATUS "Found ScaLAPACK: scalapack target")
+  else()
+    message(STATUS "Found ScaLAPACK: SCALAPACK::SCALAPACK target")
+  endif()
+endif()
+
+# Link ZLIB
+find_package(ZLIB)
```
cameronrutherford, line 477: I think we need to have this be a dep in our `package.py` since it seems required.

cameronrutherford: Or perhaps a dependency provides it in their exported CMake...

docs/src/developer/spack.md
cameronrutherford, line 1: Only skimmed this...

cameronrutherford (changes requested):
Looking good!! Couple of requests but I think this is in great shape.

P.S. I accidentally closed the tab while writing the review and so some comments might've been duplicated / lost.

spack_repo/local/packages/palace/package.py
Sbozzolo: There are two types of patches: patches to the source code (these ones, the same as in our super build), and patches to the spack recipes. CI applies the patches to the spack recipes (which, most of the time, are just PRs that are waiting to be merged in `spack-packages`) and spack applies the patches to the source code. I found that this is often needed because it can take several days/weeks to have PRs merged in spack-packages, and CI fails without them.

spack_repo/local/packages/palace/package.py
Sbozzolo: The lines are deleted because now MFEM is built by spack, so we need less argument forwarding to cmake.

.github/workflows/docs.yml
Sbozzolo: I added notes about this, but no, the directory is correct. These are patches for spack package.pys, not for the source code.

.github/workflows/docs.yml
Sbozzolo: A changed the title to `Patch Spack package.py files for known problems` to clarify that this is to patch the recipes themselves (and this is not MFEM related).

test/unit/test-geodata.cpp
Sbozzolo: Added

CHANGELOG.md
Sbozzolo: Fixed

spack_repo/local/packages/palace/package.py
```diff
@@ -235,6 +266,18 @@ def cmake_args(self):
...
+            if self.spec.satisfies("+mumps"):
+                args.append(self.define("MUMPS_DIR", self.spec["mumps"].prefix))
+            if self.spec.satisfies("+strumpack"):
+                args.append(self.define("STRUMPACK_DIR", self.spec["strumpack"].prefix))
+            if self.spec.satisfies("+superlu-dist"):
+                args.append(self.define("SUPERLU_DIST_DIR", self.spec["superlu-dist"].prefix))
+            args.append(self.define("METIS_DIR", self.spec["metis"].prefix))
+            args.append(self.define("PARMETIS_DIR", self.spec["parmetis"].prefix))
+            args.append(self.define("HYPRE_DIR", self.spec["hypre"].prefix))
+
```
cameronrutherford, line 460: Maybe it's just a simple `else:` here to include old fixed build configuration per offline discussion...

Sbozzolo: Yes, that's exactly what I am doing.

palace/CMakeLists.txt
Sbozzolo: I already updated this

palace/CMakeLists.txt
Sbozzolo: It's not required. MFEM might use zlib (or not), but it is up to MFEM. We grab it if we find it.

palace/CMakeLists.txt
Sbozzolo: I actually don't know where Threads comes from.

There's this comment in the MFEM Cmake:
```
# Without this, CMake 3.21.1 (and 3.20.2) run into CMake Errors like the following:
# CMake Error at config/cmake/modules/MfemCmakeUtilities.cmake:60 (add_library):
#   Target "mfem" links to target "Threads::Threads" but the target was not
#   found.  Perhaps a find_package() call is missing for an IMPORTED target, or
#   an ALIAS target is missing?
# Call Stack (most recent call first):
#   CMakeLists.txt:474 (mfem_add_library)
#
# NOTE: We need to figure out which TPL library adds the dependency on
#       "Threads::Threads" and call the next line only when that TPL library is
#       enabled.  -V. Dobrev
find_package(Threads REQUIRED)
```

PR #556, opened 2025-11-07
Electrical Current Dipole

### LINKED ISSUE #555: Electric Point Dipole Source
Hi @hughcars ,

As discussed previously, I was wondering if it would be possible to implement an **electric point source dipole** in Palace to simulate point source radiators. Point source dipoles are quite useful for modelling small radiators such as two-level systems in superconducting circuits and fluorescent objects which emit in the microwave domain. Additionally, point source dipoles are also useful for creating dipole radiation to test and probe the response of microwave structures/elements such as CPW resonators, lumped-element resonators and qubits. 

We are currently using this feature in Comsol but would like to implement our simulations in Palace. I believe this type of point source is commonly implemented in many optics solvers, such as Ansys Lumerical (https://optics.ansys.com/hc/en-us/articles/360034382794-Dipole-source-Simulation-object) and tidy3D, but could also be a useful addition to Palace.

Thanks for your consideration.

David Sommers

laylagi:
This PR adds support for electric point source dipole per @sqdlab's request. 

It includes the initial implementation and an example configuration file demonstrating usage in `examples/cylinder/driven_dipole.json`.

Closes #555 

TODOs:
- [x] Add a regression test
- [x] Add to documentation
- [x] Update change log


docs/src/config/domains.md
```diff
@@ -81,6 +87,34 @@ electromagnetic fields in the material domain.
...
+]
+```
+
+with
+
+`"Index" [None]` :  Index of this current dipole source, used in postprocessing output files.
+
+`"Direction" [None]` :  Floating point array of length equal to the model spatial dimension
+specifying the direction of the current dipole moment. The direction vector will be
+automatically normalized to unit length.
```
hughcars: `"Direction"` elsewhere supports using `"+X"` etc for the directions. These map into the 3-vectors of usual. It would be good to be consistent with that, but it will make parsing `"Direction"` in configfile.cpp more complicated, and the json schema would need to use a similar trick to that for `"Direction"` in `boundaries.json`. 

Use a similar description to other instances of `"Direction"`, something like that for lumped element,
```
`"Direction" [None]` :  Direction of the Dirac current source specifying the dipole. Axis aligned directions can be specified using
keywords: `"+X"`, `"-X"`, `"+Y"`, `"-Y"`, `"+Z"`, `"-Z"`. The direction can alternatively be specified as a
normalized array of three values, for example `[0.0, 1.0, 0.0]`.
```
You'll want to drop the `"+R"` and CoordinateSystem piece, (and ensure that in validation).

docs/src/examples/antenna.md
```diff
@@ -175,20 +175,115 @@ again, we see the expected donut shape, with maximal electric field strength on
...
+
+A visualization of the simplified model for the short dipole is shown below.
+
+```@raw html
+<br/><p align="center">
+  <img src="../../assets/examples/antenna-short-dipole-model.png" width="60%" />
+</p><br/>
+```
+
+The image shows the outer spherical domain with the current dipole source located at the center, eliminating the need for the complex wire geometry and feeding structure required in the half-wave dipole model.
```
hughcars:
```suggestion
The image shows the outer spherical domain with the current dipole source located at the center, eliminating the need for the wire geometry and feeding structure required in the half-wave dipole model.
```

docs/src/examples/antenna.md
```diff
@@ -175,20 +175,115 @@ again, we see the expected donut shape, with maximal electric field strength on
...
+
+!!! note "Changing dipole orientation"
+
+    One of the key advantages of using the current dipole operator is the ease of changing the antenna orientation. By simply modifying the `"Direction"` parameter, you can align the dipole along any axis without needing to regenerate the mesh or redefine boundary conditions. For example, `"Direction": [1, 0, 0]` creates an x-oriented dipole, while `"Direction": [0, 1, 0]` creates a y-oriented dipole. This makes it straightforward to study radiation patterns for different antenna orientations.
```
hughcars:
```suggestion
    One of the key advantages of using the current dipole operator is the ease of changing the antenna orientation. By modifying the `"Direction"` parameter, you can align the dipole along any axis without needing to regenerate a mesh. For example, `"Direction": [1, 0, 0]` creates an x-oriented dipole, while `"Direction": [0, 1, 0]` creates a y-oriented dipole.
```

docs/src/examples/antenna.md
```diff
@@ -175,20 +175,115 @@ again, we see the expected donut shape, with maximal electric field strength on
...
+
+### Results
+
+The far-field radiation pattern can be visualized using the same plotting script, specifying the short dipole model type:
```
hughcars:
```suggestion
The far-field radiation pattern can be visualized using the same plotting script used for the half wave dipole, by specifying the short dipole model type:
```

docs/src/examples/antenna.md
```diff
@@ -175,20 +175,115 @@ again, we see the expected donut shape, with maximal electric field strength on
...
+
+This will generate polar plots showing the characteristic ``|\sin\theta|^2`` pattern in the E-plane and omnidirectional pattern in the H-plane, as expected for an infinitesimal dipole.
+
+```@raw html
+<br/><p align="center">
+  <img src="../../assets/examples/antenna-short-dipole-polar.png" width="100%" />
+</p><br/>
+```
+
+The polar plots clearly show the theoretical radiation pattern for a short dipole: the E-plane exhibits the ``|\sin\theta|^2`` dependence with nulls along the dipole axis (θ = 0° and 180°) and maximum radiation perpendicular to the dipole (θ = 90°), while the H-plane shows the expected omnidirectional pattern.
```
hughcars:
```suggestion
The polar plots show the theoretical radiation pattern for a short dipole: the E-plane exhibits the ``|\sin\theta|^2`` dependence decaying to zero on the dipole axis (θ = 0° and 180°) and maximum radiation perpendicular to the dipole (θ = 90°), while the H-plane shows the expected omnidirectional pattern.
```

examples/antenna/antenna_short_dipole.json
```diff
@@ -0,0 +1,68 @@
...
+                    "SaveStep": 1
+                }
+            ]
+        },
+        "Linear":
+        {
+            "Type": "Default",
+            "KSPType": "GMRES",
+            "Tol": 1.0e-10,
+            "MaxIts": 1000
```
hughcars: This seems slightly excessive given the `"Default"`, pick a value closer to that used in other examples.

palace/models/currentdipoleoperator.hpp
```diff
@@ -0,0 +1,76 @@
...
+class CurrentDipoleData
+{
+public:
+  // Current dipole properties (stored in nondimensional form).
+  mfem::Vector direction;  // Normalized direction vector
+  double moment;           // Dipole moment magnitude
+  mfem::Vector center;     // Dipole center position
+
+  // Internal MFEM coefficient for the dipole
+  std::unique_ptr<mfem::VectorDeltaCoefficient> dipole_coeff;
```
hughcars: Given DeltaCoefficients are strange enough that Eval is disabled, I think it's neater to store them here in the CurrentDipoleData. Make this a value type, then we know that once the `std::map` is finished constructing, that the addresses of any members cannot change, and there's no need for `dipole_integrator_coeffs`.

docs/src/examples/antenna.md
```diff
@@ -175,20 +176,117 @@ again, we see the expected donut shape, with maximal electric field strength on
...
     ```
-    
+
     All the subsequent times, just make sure to start Julia with `--project` from the `examples`
-    
     folder or one of its subfolders.
 
+## Alternative: Short Dipole Using the Electric Current Dipole Operator
+
+Instead of explicitly modeling the antenna geometry with a lumped port excitation, *Palace* provides an electric current dipole operator that represents an infinitesimally short current-carrying wire. This approach is particularly useful for theoretical studies or when the detailed geometry of the feeding structure is not critical to the analysis.
```
hughcars:
```suggestion
As an alternative to explicitly modeling the antenna geometry with a lumped port excitation, *Palace* provides an electric current dipole operator that represents an infinitesimally short current-carrying wire. This approach is particularly useful for theoretical studies or when the detailed geometry of the feeding structure is not critical to the analysis.
```

docs/src/examples/antenna.md
```diff
@@ -175,20 +176,117 @@ again, we see the expected donut shape, with maximal electric field strength on
...
+
+where ``\delta`` represents the Dirac delta function, indicating that the source is concentrated at a single point in space.
+
+For a z-oriented electrical current dipole (short dipole) in free space, the far-field radiation pattern in the E-plane (xz-plane) follows
+
+```math
+|\mathbf{E}_\theta|^2 \propto |\sin\theta|^2\,,
+```
+
+with omnidirectional radiation in the H-plane (xy-plane). This differs from the half-wave dipole pattern shown earlier, which exhibits the characteristic ``\left|\frac{\cos\left(\frac{\pi}{2}\cos\theta\right)}{\sin\theta}\right|^2`` dependence.
```
hughcars:
```suggestion
with omnidirectional radiation in the H-plane (xy-plane). This differs from the half-wave dipole pattern shown earlier, which exhibits a characteristic ``\left|\frac{\cos\left(\frac{\pi}{2}\cos\theta\right)}{\sin\theta}\right|^2`` dependence.
```

palace/models/currentdipoleoperator.cpp
```diff
@@ -0,0 +1,124 @@
...
+  {
+    return;
+  }
+
+  Mpi::Print("\nConfiguring electrical current dipole source terms:\n");
+  for (const auto &[idx, data] : dipoles)
+  {
+    // Convert center coordinates back to physical units for display
+    mfem::Vector physical_center = data.center;
+    iodata.units.DimensionalizeInPlace<Units::ValueType::LENGTH>(physical_center);
```
hughcars:
```
auto physical_center = iodata.units.Dimensionalize<Units::ValueType::LENGTH>(center);
```
should be possible, but there's a bug in units.hpp you can fix with
```

  // Return a copy of value scaled by given unit.
  template <ValueType unit, typename T>
  auto Dimensionalize(T value) const
      -> std::enable_if_t<!std::is_same_v<T, Vector> && !std::is_same_v<T, ComplexVector>,
                          T>
  {
    return value * GetScaleFactor<unit>();
  }
```
for the version on line 139.

palace/models/currentdipoleoperator.cpp
```diff
@@ -0,0 +1,124 @@
...
+    double physical_moment = iodata.units.Dimensionalize<Units::ValueType::CURRENT>(
+        iodata.units.Dimensionalize<Units::ValueType::LENGTH>(data.moment));
+
+    Mpi::Print(" Dipole {:d}: \n"
+               " \tMoment = {:.3e} A·m\n"
+               " \tCenter = ({:.3e}, {:.3e}, {:.3e})\n"
+               " \tDirection = ({:.3e}, {:.3e}, {:.3e})\n",
+               idx, physical_moment, physical_center[0], physical_center[1],
+               (mesh.SpaceDimension() > 2) ? physical_center[2] : 0.0, data.direction[0],
+               data.direction[1], (mesh.SpaceDimension() > 2) ? data.direction[2] : 0.0);
```
hughcars:
```suggestion
    Mpi::Print(" Dipole {:d}: \n"
               " \tMoment = {:.3e} A·m\n"
               " \tCenter = ({:.3e}) m\n"
               " \tDirection = ({:.3e}) m\n",
               idx, physical_moment, fmt::join(physical_center, ", "),
               fmt::join(data.direction, ", "));
```
fmt can take care of the variable length for you, and add units to the lengths for consistency.

palace/models/currentdipoleoperator.cpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+}
+
+void CurrentDipoleOperator::AddExcitationDomainIntegrators(int idx, mfem::LinearForm &rhs)
+{
```
hughcars: If you move the coefficient inside of the data, then you can do this
```

void CurrentDipoleOperator::AddExcitationDomainIntegrator(int idx, mfem::LinearForm &rhs)
{
  auto it = dipoles.find(idx);
  MFEM_VERIFY(it != dipoles.end(), "Invalid dipole index!");
  rhs.AddDomainIntegrator(new mfem::VectorFEDomainLFIntegrator(it->second.coef));
}

void CurrentDipoleOperator::AddExcitationDomainIntegrators(mfem::LinearForm &rhs)
{
  // Add each dipole as a separate integrator
  for (auto &[idx, dipole] : dipoles)
  {
    // Create a VectorDeltaCoefficient for this specific dipole (RHS = -iω J_e)
    // RHS will be scaled by iω later in SpaceOperator.
    rhs.AddDomainIntegrator(new mfem::VectorFEDomainLFIntegrator(dipole.coef));
  }
}
```
and avoid the search over the map for each index you've already drawn from the map. Also nit, but when adding a single index, there is only one dipole, so drop the pluralizing `s`.

palace/models/currentdipoleoperator.cpp
```diff
@@ -0,0 +1,124 @@
...
+  moment = data.moment * current_scale * length_scale;
+
+  // Set up dipole center position
+  center.SetSize(mesh.SpaceDimension());
+  MFEM_VERIFY(data.center.size() == static_cast<std::size_t>(mesh.SpaceDimension()),
+              "Current dipole center dimension must match mesh space dimension!");
+  for (int d = 0; d < mesh.SpaceDimension(); d++)
+  {
+    center[d] = data.center[d];
+  }
```
hughcars, line 44:
```

  // Assemble the coefficient.
  coef.SetDeltaCenter(center);
  coef.SetDirection(direction);
  coef.SetScale(-moment);
  ```
  then have `coef` as a value data member in `CurrentDipoleData`, as a special case of a coefficient outside of the usual structure, given `Eval` doesn't work for it.

File: palace/models/currentdipoleoperator.hpp [hunk 14]
```diff
@@ -0,0 +1,76 @@
...
+//
+class CurrentDipoleOperator
+{
+private:
+  // Mapping from dipole index to data structure containing dipole information.
+  std::map<int, CurrentDipoleData> dipoles;
+
+  // Storage for integrator coefficients to manage their lifetime
+  mutable std::vector<std::unique_ptr<mfem::VectorDeltaCoefficient>>
+      dipole_integrator_coeffs;
```
hughcars: Delete this, given all accessors to the map are `const`, once the map is filled in the ctor, the addresses of all value members should be fixed, sidestepping the issue with std::map that entries can be moved around.

palace/utils/configfile.cpp
```diff
@@ -767,12 +767,70 @@ void DomainPostData::SetUp(json &domains)
...
+        it->find("Moment") != it->end(),
+        "Missing \"CurrentDipole\" source \"Moment\" magnitude in the configuration file!");
+    auto direction = it->find("Direction");
+    auto center = it->find("Center");
+    MFEM_VERIFY(
+        direction->is_array() && center->is_array(),
+        "\"CurrentDipole\" source \"Direction\" and \"Center\" should specify arrays "
+        "in the configuration file!");
+
+    data.direction = direction->get<std::array<double, 3>>();  // Required
```
hughcars: The `"Direction"` keyword supports strings elsewhere, so would be good to do so, I have done some refactoring to allow that in `hughcars/dipoles`.

scripts/schema/config/domains.json
```diff
@@ -87,6 +87,38 @@
...
+            "maxItems": 3
+          },
+          "Direction":
+          {
+            "type": "array",
+            "additionalItems": false,
+            "items": { "type": "number" },
+            "minItems": 3,
+            "maxItems": 3
+          }
```
hughcars, line 128: See the change on `hughcars/dipoles` to support string based direction

test/examples/runtests.jl
```diff
@@ -291,12 +292,25 @@ end
 reltol = 2.0e-2
 abstol = 1.0e-10
 
-if "antenna" in cases
-    @info "Testing antenna..."
+if "antenna/antenna_halfwave_dipole" in cases
+    @info "Testing antenna/antenna_halfwave_dipole"
```
hughcars: Add the missing `...` to match the others.

docs/src/examples/antenna.md
```diff
@@ -175,20 +176,117 @@ again, we see the expected donut shape, with maximal electric field strength on
...
+    One of the key advantages of using the current dipole operator is the ease of changing the antenna orientation. By simply modifying the `"Direction"` parameter, you can align the dipole along any axis without needing to regenerate the mesh or redefine boundary conditions. For example, `"Direction": [1, 0, 0]` creates an x-oriented dipole, while `"Direction": [0, 1, 0]` creates a y-oriented dipole using the same mesh.
+
+The following plots show the spatial distribution of the real part of the electric field in the yz-plane (left) and xy-plane (right), illustrating the near-field behavior of the short dipole antenna.
+
+```@raw html
+<br/><p align="center">
+  <img src="../../assets/examples/antenna-5a.png" width="45%" />
+  <img src="../../assets/examples/antenna-5b.png" width="45%" />
+</p>
+```
```
hughcars, line 257: Would be good to include the units in the image if possible, but not strictly necessary. "E_mag" type thing.

palace/utils/configfile.cpp
```diff
@@ -767,12 +767,70 @@ void DomainPostData::SetUp(json &domains)
...
+    it->erase("Moment");
+    MFEM_VERIFY(it->empty(),
+                "Found an unsupported configuration file keyword under \"CurrentDipole\"!\n"
+                    << it->dump(2));
+
+    // Debug
+    if constexpr (JSON_DEBUG)
+    {
+      std::cout << "Index: " << ret.first->first << '\n';
+      std::cout << "Direction: " << data.moment << '\n';
```
hughcars: `data.direction`

palace/utils/configfile.hpp
```diff
@@ -346,6 +346,25 @@ struct DomainPostData
   void SetUp(json &domains);
 };
 
+struct CurrentDipoleData
+{
+public:
+  // Current dipole direction (normalized unit vector).
+  std::array<double, 3> direction{{0.0, 0.0, 0.0}};
```
hughcars, line 353: Given saying this is stored normalized, perform normalization in configfile.cpp

examples/antenna/plot_farfield.jl
```diff
@@ -239,16 +281,84 @@ function three_d_plot(data, label, filename="farfield_3d.png")
...
+                )
+                println(
+                    "Valid model types: 'antenna_short_dipole', 'antenna_halfwave_dipole'"
+                )
+                println("Examples:")
+                println("  julia plot_farfield.jl")
+                println("  julia plot_farfield.jl model=antenna_short_dipole file=data.csv")
+                return
+            end
+        end
```
Sbozzolo: This type of command line interaction is unusual. I would suggest sticking to a more common command-line argument style. Not many cli tools set their settings with this syntax. Instead, you could use positional arguments
```
plot_farfield.jl filename model_type
```
or full keywords
```
plot_farfield.jl --model model_name --file filename
```
I believe 
```
plot_farfield.jl --model=model_name --file=filename
```
could also be fine

examples/antenna/plot_farfield.jl
```diff
@@ -239,16 +281,84 @@ function three_d_plot(data, label, filename="farfield_3d.png")
...
+            if startswith(arg, "model=")
+                model_type = split(arg, "=", limit=2)[2]
+                has_model = true
+            elseif startswith(arg, "file=")
+                filename = split(arg, "=", limit=2)[2]
+                has_file = true
+            else
+                println("Error: Invalid argument '$arg'")
+                println(
+                    "Usage: julia --project plot_farfield.jl [model=<model_type> file=<filename>]"
```
Sbozzolo: `--project` is not necessary. it's up to the user.

examples/antenna/plot_farfield.jl
```diff
@@ -239,16 +281,84 @@ function three_d_plot(data, label, filename="farfield_3d.png")
...
+                )
+                println(
+                    "Valid model types: 'antenna_short_dipole', 'antenna_halfwave_dipole'"
+                )
+                println("Examples:")
+                println("  julia plot_farfield.jl")
```
Sbozzolo: I think you can use `PROGRAM_FILE` instead of hard-coding the filename

examples/antenna/plot_farfield.jl
```diff
@@ -239,16 +281,84 @@ function three_d_plot(data, label, filename="farfield_3d.png")
...
+            elseif startswith(arg, "file=")
+                filename = split(arg, "=", limit=2)[2]
+                has_file = true
+            else
+                println("Error: Invalid argument '$arg'")
+                println(
+                    "Usage: julia --project plot_farfield.jl [model=<model_type> file=<filename>]"
+                )
+                println(
+                    "Valid model types: 'antenna_short_dipole', 'antenna_halfwave_dipole'"
```
Sbozzolo: I suggest you drop the `antenna_` part of the name.

examples/antenna/plot_farfield.jl
```diff
@@ -239,16 +281,84 @@ function three_d_plot(data, label, filename="farfield_3d.png")
...
+        println(
+            "Usage: julia --project plot_farfield.jl [model=<model_type> file=<filename>]"
+        )
+        println("Examples:")
+        println("  julia plot_farfield.jl")
+        println("  julia plot_farfield.jl model=antenna_short_dipole file=data.csv")
+        return
+    end
+
+    if !(model_type in ["antenna_short_dipole", "antenna_halfwave_dipole"])
```
Sbozzolo:
```suggestion
    if !(model_type in ("antenna_short_dipole", "antenna_halfwave_dipole"))
```
Nano-optimization, but this is a statically known list, so you don't need to create an explicit list. 
With `("a", "b")` the list is created at compile time, with `["a", "b"]` at run time (simplifying a little bit). It does not matter here, but there are cases where one would pick one vs the other.

test/unit/test-config.cpp
```diff
@@ -419,3 +419,47 @@ TEST_CASE("FarField", "[config][Serial]")
...
+    CHECK_THROWS(config::ParseStringAsDirection("~x", req));
+    CHECK_THROWS(config::ParseStringAsDirection("x+", req));
+    CHECK_THROWS(config::ParseStringAsDirection("xy", req));
+    CHECK_THROWS(config::ParseStringAsDirection("xyz", req));
+    CHECK_THROWS(config::ParseStringAsDirection("abc", req));
+  }
+
+  CHECK_THROWS(config::ParseStringAsDirection("", true));
+  CHECK_NOTHROW(config::ParseStringAsDirection("", false));
+}
```
Sbozzolo: This file does not end with a newline

examples/antenna/plot_farfield.jl
```diff
@@ -239,16 +281,84 @@ function three_d_plot(data, label, filename="farfield_3d.png")
...
+            return
+        end
+    else
+        println("Error: If arguments are provided, both model= and file= must be specified")
+        println(
+            "Usage: julia --project plot_farfield.jl [model=<model_type> file=<filename>]"
+        )
+        println("Examples:")
+        println("  julia plot_farfield.jl")
+        println("  julia plot_farfield.jl model=antenna_short_dipole file=data.csv")
```
Sbozzolo: This block of code is identical to the one above. Is there a way to refactor and have it only once?

PR #559, opened 2025-11-11
Use map instead of find_if to check for duplicate column names in tablecsv

simlapointe:
Use map instead of find_if to check for duplicate column names and allow for retrieval by name in tablecsv. find_if can get quite slow when inserting a large number of columns in a table (for example, when using more than > 10,000 probes).

### DIFF
--- a/palace/utils/tablecsv.cpp
+++ b/palace/utils/tablecsv.cpp
@@ -100,12 +100,12 @@ void Table::reserve(size_t n_rows, size_t n_cols)
 // Insert columns: map like interface.
 bool Table::insert(Column &&column)
 {
-  if (col_names.find(column.name) != col_names.end())
+  if (name_to_index.find(column.name) != name_to_index.end())
   {
     return false;
   }
+  name_to_index[column.name] = cols.size();
   auto &col = cols.emplace_back(std::move(column));
-  col_names.insert(col.name);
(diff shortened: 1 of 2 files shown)


palace/utils/tablecsv.cpp
```diff
@@ -100,13 +100,12 @@ void Table::reserve(size_t n_rows, size_t n_cols)
...
 {
-  auto it = std::find_if(cols.begin(), cols.end(),
-                         [&column](auto &c) { return c.name == column.name; });
-  if (it != cols.end())
+  if (col_names.find(column.name) != col_names.end())
   {
     return false;
   }
   auto &col = cols.emplace_back(std::move(column));
+  col_names.insert(col.name);
```
hughcars: Well that explains the nonlinear time. 

Not a huge fan of this change meaning `Column` and `col_names` are both storing the name, giving the possibility of them desyncing/duplicating, but it's neater than the alternatives, where the name is removed from `Column`, (which effectively becomes wrapper of `std::vector`). Conversely, this duplication of the names for a faster lookup like this does keep the data separation a bit cleaner, and can be thought of as "table specific" organization, to help this fast lookup.

Also, whatever change you make, apply it to the other `find_if` on line 120 in `operator[]`.

PR #560, opened 2025-11-11
Do not reuse MUMPS column ordering when dropped entries change

simlapointe:
When both `DropSmallEntries` and `ReorderReuse` are set to `true`, MUMPS sometimes fails to factorize the sparse matrix with an error such as: 

> Error in MUMPS_BUILD_SORT_INDEX: matrix structure might have changed,
  analysis (JOB=1) should be performed again

This happens when the number of dropped small entries changes from that of the previous factorization. To guard against that, we keep track of the number of dropped entries and, when it changes, we set MUMPS' `ReorderReuse` to `false`.


hughcars (approved):
A pity mumps isn't as permissive as superLU in allowing a "not quite right" sparsity pattern to dictate the factorization, but this is a reasonable work around.

PR #562, opened 2025-11-13
Fix Paraview output for transient simulations

Sbozzolo:
The arguments for `WriteParaviewFields` were inconsistent with the function signature.

output


hughcars (approved):
Verified the coaxial open example paraview output.

PR #567, opened 2025-11-24
Use symmetric block matrix for complex coarse solve

zhanggy2004:
Thanks for developing and maintaining this nice project.

For "complex coarse solve", instead of solving

```math
A_1\begin{pmatrix}x_r \\ x_i \end{pmatrix}:=\begin{pmatrix} A_r & -A_i \\ A_i & A_r \end{pmatrix} \begin{pmatrix}x_r \\ x_i \end{pmatrix}=\begin{pmatrix}y_r \\ y_i \end{pmatrix},
```

we can solve

```math
A_2\begin{pmatrix}x_r \\ -x_i \end{pmatrix}:=\begin{pmatrix} A_r & A_i \\ A_i & -A_r \end{pmatrix} \begin{pmatrix}x_r \\ -x_i \end{pmatrix}=\begin{pmatrix}y_r \\ y_i \end{pmatrix}.
```

The idea is similar to [`mfem::ComplexOperator::Convention::BLOCK_SYMMETRIC`](https://docs.mfem.org/html/complex__operator_8hpp_source.html).

When $A_r$ and $A_i$ are both symmetric, $A_2$ will also be symmetric, which can be utitlized by `MUMPS` solver for more efficient solve.

I realized that $A_r$ and $A_i$ are unsymmetric when there is periodic boundary (?), but am not sure if there are other exceptions. Please let me know your suggestions for the correct/better approach. Thanks.

### DIFF
--- a/palace/linalg/mumps.hpp
+++ b/palace/linalg/mumps.hpp
@@ -29,9 +29,9 @@ public:
            iodata.problem.type == ProblemType::ELECTROSTATIC ||
            iodata.problem.type == ProblemType::MAGNETOSTATIC)
               ? mfem::MUMPSSolver::SYMMETRIC_POSITIVE_DEFINITE
-          : iodata.solver.linear.complex_coarse_solve
-              ? mfem::MUMPSSolver::UNSYMMETRIC
-              : mfem::MUMPSSolver::SYMMETRIC_INDEFINITE,
+          : iodata.boundaries.periodic.wave_vector == std::array<double, 3>{0.0, 0.0, 0.0}
+              ? mfem::MUMPSSolver::SYMMETRIC_INDEFINITE
+              : mfem::MUMPSSolver::UNSYMMETRIC,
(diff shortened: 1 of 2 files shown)


simlapointe (commented):
This looks good! I tested it on a few different cases with Floquet periodic BCs (cylinder, nail) and without (cpw, single transmon, adapter waveguide) and it works as expected in all cases. I checked that mumps, strumpack, and superlu all converge in 1-2 iterations when `ComplexCoarseSolve == true`.

I think Floquet periodicity is the only symmetry-breaking term we have at the moment, so the `wave_vector == std::array<double, 3>{0.0, 0.0, 0.0}` check is ok (though I don't love `== 0.0`). However, if we ever encounter other symmetry-breaking scenarios, it might be better to define a single variable such as `iodata.solver.linear.is_symmetric` that gets set accordingly after reading the config. @hughcars, what do you think?

PR #570, opened 2025-11-25
Fix spack recipe for +superlu/+strumpack with cuda

Sbozzolo:
The Palace spack recipe was forcing `strumpack` to be `~cuda` regardless of the palace variant, so that it was not possible to install `palace+strumpack+cuda` because of conflicting variants. Superlu was also not correctly handled.

### DIFF
--- a/spack_repo/local/packages/palace/package.py
+++ b/spack_repo/local/packages/palace/package.py
@@ -84,8 +84,6 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
 
     with when("+superlu-dist"):
         depends_on("superlu-dist+parmetis")
-        depends_on("superlu-dist~cuda", when="~cuda")
-        depends_on("superlu-dist~rocm", when="~rocm")
         depends_on("superlu-dist+shared", when="+shared")
         depends_on("superlu-dist~shared", when="~shared")
         depends_on("superlu-dist+int64", when="+int64")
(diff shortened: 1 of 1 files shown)


spack_repo/local/packages/palace/package.py
```diff
@@ -83,7 +83,9 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         depends_on("mumps~openmp", when="~openmp")
 
     with when("+superlu-dist"):
-        depends_on("superlu-dist+parmetis~cuda~rocm")
+        depends_on("superlu-dist+parmetis")
+        depends_on("superlu-dist~cuda", when="~cuda")
+        depends_on("superlu-dist~rocm", when="~rocm")
```
cameronrutherford: Arguably we don't "depend on" `superlu-dist~cuda` when `~cuda`... I am sure that a `palace~cuda` would play well enough with `superlu-dist+cuda`.

That being said, I think that this is the better choice for user experience, but it's not a hard/fast rule from a Spack perspective. For maximizing re-use the above comment is the axiom one should follow.

Either way, clearly `depends_on("superlu-dist+parmetis~cuda~rocm")` is wrong. Nice catch.

Sbozzolo: Good point. I just copied the pattern I saw in the rest of the file, but I will fix it.

PR #576, opened 2025-12-03
Update spack for 0.15.0, add sbozzolo

hughcars:
Add v0.15.0 recipe to local spack recipe, and add sbozzolo. Reflects the state in https://github.com/spack/spack-packages/pull/2642.

Will wait until spack merges, in case of any additional changes requested.

Issue #565, opened 2025-11-20
Regression test failure with high core count


phdum-a:
I was running the regression tests arm instance with 64 cores. When I run with all cores as mpi ranks I get a test failure on `cpw/wave_uniform` mostly on field measurements (E_x, etc) and a few scattering matrix angles . When I switch to NUM_PROC=8, the test passes. 

This may be just rounding errors from the solve on a different partitioned mesh, or from double reduction, etc. Or may be some actual issue with high core count. Might be worth investigation and understanding if tolerances of the tests are set with what we expect.

Note: This was on my prom branch, but the changes do not affect the adaptive wave port example. I will rerun on main when I get the chance to see that I can reproduce this.

simlapointe:
Thanks to both of you for reporting this! The differences were small-ish with 64 cores, but could be quite significant with some other core counts. I found the root cause and have a fix in #580.

PR #578, opened 2025-12-05
Add GPU & large ARM runners, and setup-palace-ci action

Sbozzolo:
This commit runs the spack workflow on larger/GPU-accelerated runners.

Updated description: https://github.com/awslabs/palace/pull/578#issuecomment-3697000643

### OLD description, new description in https://github.com/awslabs/palace/pull/578#issuecomment-3697000643

I added a full test matrix, but I could not get everything to work (despite fixing several bugs). Some of this work conflicts with #550, so I will address more of that there. I tried quick fixes, but I could not get the builds to work, so I am leaving this for future work.

To be noted:
- `libxsmm` does not support `flang`, so I am using clang+gfortran https://github.com/libxsmm/libxsmm/issues/996
- I ran into issues with waveports at high core count (see also #565)
- There's one case that compiles but fails in the ` cylinder/floquet (periodic)` test (maybe OpenMP)
- The current cache has a problem with the prefix being too short. I temporarily switched to a new cache, but we should just reset the old one (`palace-develop`)
- `CPW (lumped ports, adaptive)` takes 12 minutes on GPU. I don't know if this is another instance of #375 
- I added a new local action to set the worker up

There is this annoying bug with oneAPI and spack where the version of `ifx` is incorrectly reported, breaking everything. I asked about this on slack, and I was told that I should manually merge the two compiler entries in the package definition (as a workaround). I added a step to take care of this, but I hope we can remove it soon.

Good news that (after a week of fixing bugs), lots of variants are working, including OpenMP cases, cases with mixed compilers, GPU+LLVM, newer dependencies (e.g., eigen 5), et cetera.

The progression I see is:
(#576 at any time)
- This PR
- #550 
- #519 
- #458

### DIFF
--- a/spack_repo/local/packages/palace/package.py
+++ b/spack_repo/local/packages/palace/package.py
@@ -196,7 +196,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("+cuda"):
         # GPU-aware MPI
         for var in ["openmpi", "mpich", "mvapich-plus"]:
-            depends_on(f"hypre+gpu-aware", when=f"^[virtuals=mpi] {var}+cuda")
+            depends_on(f"hypre+gpu-aware-mpi", when=f"^[virtuals=mpi] {var}+cuda")
 
         for arch in CudaPackage.cuda_arch_values:
             cuda_variant = f"+cuda cuda_arch={arch}"
@@ -218,7 +218,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("+rocm"):
         for var in ["openmpi@5:", "mpich", "mvapich-plus"]:
             # GPU-aware MPI
-            depends_on(f"hypre+gpu-aware", when=f"^[virtuals=mpi] {var}+rocm")
+            depends_on(f"hypre+gpu-aware-mpi", when=f"^[virtuals=mpi] {var}+rocm")
 
         for arch in ROCmPackage.amdgpu_targets:
             rocm_variant = f"+rocm amdgpu_target={arch}"
@@ -268,7 +268,9 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
                 )
             )
 
-        palace_with_gpu_aware_mpi = any(self.spec.satisfies(f"{var}+cuda") for var in ["openmpi", "mpich", "mvapich-plus"])
+        palace_with_gpu_aware_mpi = any(self.spec.satisfies(f"{var}+cuda") or
+                                        self.spec.satisfies(f"{var}+rocm")
+                                        for var in ["openmpi", "mpich", "mvapich-plus"])
 
         args.append(self.define("PALACE_WITH_GPU_AWARE_MPI", palace_with_gpu_aware_mpi))
 


test/unit/test-libceed.cpp
```diff
@@ -1432,7 +1432,7 @@ TEST_CASE("3D libCEED Operators", "[libCEED][Serial][Parallel][GPU]")
                          amr, order);
 }
 
-TEST_CASE("2D libCEED Interpolators", "[libCEED][Interpolator][Serial][Parallel][GPU]")
+TEST_CASE("2D libCEED Interpolators", "[libCEED][Interpolator][Serial][Parallel]")
```
Sbozzolo, line 1435: Removed because of #516

test/examples/runtests.jl
```diff
@@ -206,11 +215,12 @@ if "antenna" in cases
...
         np=numprocs,
         rtol=reltol,
-        atol=50abstol
+        atol=50abstol,
+        force_gpu=test_gpu
     )
 end
 
-abstol = 2.0e-12
+abstol = 1.0e-11
```
Sbozzolo, line 271: I don't remember which case led me to change this, but I found it necessary

spack_repo/local/packages/palace/package.py
```diff
@@ -250,13 +253,38 @@ def cmake_args(self):
...
+            scalapack_libs = self.spec["scalapack"].libs
+            fortran_libs = ""
+            if "gfortran" in self.compiler.fc:
+                fortran_libs = "gfortran"
+            elif "ifort" in self.compiler.fc or "ifx" in self.compiler.fc:
+                fortran_libs = "ifport;ifcore"
+            # For other compilers (flang, etc.), don't add extra libs
+            strumpack_libs = str(scalapack_libs).replace(" ", ";")
+            if fortran_libs:
+                strumpack_libs += ";" + fortran_libs
```
Sbozzolo: I think this entire section might away with #550

.github/workflows/spack.yml
```diff
@@ -134,22 +403,110 @@ jobs:
...
           spack -e . external find --all \
             --exclude openssl \
             --exclude openssh \
             --exclude python \
             --exclude ncurses \
             --exclude bzip2 \
             --exclude xz \
-            --exclude curl
+            --exclude curl \
+            --exclude llvm-doe
```
hughcars: Given the reliance on the cache, do we really need any externals? After 1 success we'll always be getting a cache hit. I've personally noted that externals seem to cause some really weird errors occasionally, but maybe just having this exclusion list suffices.

.github/workflows/spack.yml
```diff
@@ -134,22 +403,110 @@ jobs:
       - name: Configure External Packages
         if: needs.filter.outputs.test == 'true'
         run: |
+          # Spack needs a little help finding the intel compiler
+          if [[ "${{ matrix.compiler }}" == "intel-oneapi-compilers" ]]; then
+            . /opt/intel/oneapi/setvars.sh
+          fi
```
hughcars: This is despite the intel compiler installation being handled by spack? (which I believe is what you're now doing given the deletion of the apt calls)

spack_repo/patches/pr2580.patch
```diff
@@ -0,0 +1,38 @@
+diff --git a/repos/spack_repo/builtin/packages/libceed/package.py b/repos/spack_repo/builtin/packages/libceed/package.py
+index 7a4471437d..e6efbe3ef9 100644
```
hughcars: If we're into allowing patches now, we might want to encourage libceed to allow the change suggested in https://github.com/CEED/libCEED/issues/1808 as this periodically causes mac to be unable to build libceed.

hughcars (commented):
One small question about usage of the cache, but generally LGTM.

.github/workflows/spack.yml
Sbozzolo: No, the intel compiler is being installed by apt by the `setup-palace-ci` action.

.github/workflows/spack.yml
Sbozzolo: This is a good point. Let me try removing all the externals to see if there's any issue.

.github/actions/setup-palace-ci/action.yml
```diff
@@ -0,0 +1,131 @@
...
+      shell: bash
+      env:
+        LLVM_VERSION: ${{ inputs.llvm-version }}
+      run: |
+        sudo apt-get install -y clang-${LLVM_VERSION} llvm-${LLVM_VERSION} lld-${LLVM_VERSION} lldb-${LLVM_VERSION} libomp5-${LLVM_VERSION} libomp-${LLVM_VERSION}-dev flang-${LLVM_VERSION}  
+
+    - name: Setup GNU
+      shell: bash
+      run: |
+        sudo apt-get install -y gfortran
```
hughcars: Why no need to install gcc explicitly?

Sbozzolo: I guess the AMIs I am using already come with gcc, but it does not hurt adding it here as well.

.github/workflows/spack.yml
Sbozzolo: I removed them. It works.

spack_repo/patches/pr2580.patch
Sbozzolo: PR opened and patch added https://github.com/spack/spack-packages/pull/2755

.github/actions/palace-ci/action.yml
```diff
@@ -0,0 +1,297 @@
...
+        fi
+        spack -e . compiler find && spack -e . compiler list
+
+    # Spack and oneAPI-2025.3.1 have some compatibility problems where the
+    # version of the fortran compiler is not correctly identified. As a
+    # workaround, we manually edit the relevant entry in the spack.yaml to put
+    # all the compilers under the same version. # This amounts to merging two
+    # entries (for two different versions) into a single one.
+    #
+    # TODO: Check if this issue is still relevant, if not, remove this step.
```
hughcars, line 178: If you think this is worth tracking, open an issue so we dont forget about it.

.github/actions/setup-runner/action.yml
```diff
@@ -0,0 +1,125 @@
...
+        echo 'APT::Periodic::Update-Package-Lists "0";' | sudo tee /etc/apt/apt.conf.d/20auto-upgrades
+        echo 'APT::Periodic::Unattended-Upgrade "0";' | sudo tee -a /etc/apt/apt.conf.d/20auto-upgrades
+
+    - name: Setup repositories
+      shell: bash
+      run: |
+        # Intel
+        echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
+        wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB \
+        | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null
```
hughcars, line 64: Can these be placed within the Setup actions straight after? Or is there value in having this broken out? I imagine they're fairly quick so the granularity is probably helpful even if it adds them on runs without these enabled.

.github/actions/setup-runner/action.yml
```diff
@@ -0,0 +1,125 @@
...
+      with:
+        repository: spack/spack
+        path: spack
+        ref: ${{ inputs.spack-version }}
+        
+    - name: Setup Spack environment
+      shell: bash
+      run: |
+        echo "$(realpath spack/bin)" >> "$GITHUB_PATH"
+        echo "SPACK_DISABLE_LOCAL_CONFIG=1" >> "$GITHUB_ENV"
```
hughcars, line 103: Is this done for the cmake builds too? Again, probably not that big a deal as I bet it's very fast.

.github/workflows/spack.yml
```diff
@@ -43,170 +43,137 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        solver: [MUMPS, SuperLU, STRUMPACK, Default]
+    runs-on: [self-hosted, x64, 4xlarge]
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/run-regression-tests
+        with:
+          toolchain: gcc
+          solver: ${{ matrix.solver }}
```
hughcars, line 151: With this matrix approach, if we have a failing test for one entry in the tensor product that we want to disable, how would we do that? I.e. only the STRUMPACK run on 4xlarge.

palace/CMakeLists.txt
```diff
@@ -120,6 +120,12 @@ endif()
 # Find LAPACK
 find_package(LAPACK REQUIRED)
 
+# Find Umpire and Camp (needed when CUDA/HIP is enabled)
+if(PALACE_WITH_CUDA OR PALACE_WITH_HIP)
+  find_package(umpire REQUIRED CONFIG)
+  find_package(camp REQUIRED CONFIG)
```
hughcars, line 126: What does camp do? I've not heard of it before.

spack_repo/local/packages/palace/package.py
```diff
@@ -182,6 +187,10 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("+cuda"):
         for arch in CudaPackage.cuda_arch_values:
             cuda_variant = f"+cuda cuda_arch={arch}"
+
+            # TODO: Remove me after blt > 0.7.1 is released
+            depends_on(f"blt@develop")
```
hughcars: Can this be modeled as a conflict for 0.7.1 and below? I.e. I think that will always pull develop if the conflict covers the most recent version, and then this will not require updating again (I think we're doing this for libceed for instance, depending on a version that doesn't exist yet).

spack_repo/local/packages/palace/package.py
```diff
@@ -148,7 +148,12 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         # https://github.com/libxsmm/libxsmm/issues/883
```
hughcars, line 157: Once the 0.15.0 spack recipe PR is in, I think you should open another one with these changes in addition. This seems like a really good upgrade of the general capability and robustness of the spack release.

docs/src/developer/testing.md
```diff
@@ -315,12 +315,17 @@ julia --project -e "using Pkg; Pkg.instantiate()"
...
-  - `TEST_CASES`: Space-separated list of test cases to run (default: all examples)
+The test runner supports command line arguments for configuration. Each argument can also be set via environment variables as fallbacks.
+
+**Key Options:**
+
+  - `--palace-test`: Path to *Palace* executable and optional arguments (default: "`palace`")
+  - `--num-proc-test`: Number of MPI processes (default: number of physical cores)
+  - `--test-cases`: Space-separated list of test cases to run (default: all examples)
+
+Run `julia --project runtests.jl --help` to see all available options with descriptions and defaults.
```
hughcars, line 328: A much nicer approach

.github/actions/palace-ci/action.yml
Sbozzolo: https://github.com/awslabs/palace/issues/592

.github/workflows/spack.yml
Sbozzolo: The action defines three separate `build-xxxxxx` jobs and three corresponding `test-xxxxxx` jobs. The `test` jobs define a matrix (all of which running on `4xlarge`). If you want to disable a specific combination, you can simply comment that out. For example, if you want to disable STRUMPACK for openmp, we would just have
```yaml
  test-x64-gcc-openmp:
    needs: [filter, build-x64-gcc-openmp]
    if: success()
    strategy:
      fail-fast: false
      matrix:
        include: 
         - solver: MUMPS
         - solver: SuperLU
        # - solver: STRUMPACK
         - solver: Default
    runs-on: [self-hosted, x64, 4xlarge]
    steps:
      - uses: actions/checkout@v6
      - uses: ./.github/actions/run-regression-tests
        with:
          toolchain: gcc
          variant: "+openmp+int64~shared"
          solver: ${{ matrix.solver }}
```

.github/actions/setup-runner/action.yml
Sbozzolo: I added this separately to have only one `apt update`. `apt update` pull from the internet and takes O(10) seconds. I moved it out for etiquette (I don't want to make too many server calls if I can avoid that) and shave off that tiny amount of time. This is a micro-optimization and I can move this to the steps below if you think that would be better.

.github/actions/setup-runner/action.yml
Sbozzolo: Currently, this file is not used for the cmake builds. We might need a couple of small tweaks for that.

spack_repo/local/packages/palace/package.py
Sbozzolo: Yes, we can just say `depends_on("blt@0.7.2:")`.

spack_repo/local/packages/palace/package.py
Sbozzolo: We change other stuff in #550, but yes, I agree we should update the spack recipe.

palace/CMakeLists.txt
Sbozzolo: To be honest, I don't quite know. It is used by umpire: https://camp.readthedocs.io/en/latest/sphinx/user_guide/using_camp.html#camp-used-in-umpire

.github/actions/palace-ci/action.yml
```diff
@@ -0,0 +1,297 @@
...
+      shell: bash
+      run: |
+        PALACE_DIR=$PWD
+        cd $(spack location --repo builtin)
+        for patch in "$PALACE_DIR"/spack_repo/patches/*.diff; do
+          if [ -f "$patch" ]; then
+            echo "Applying patch: $patch"
+            git apply "$patch"
+          fi
+        done
```
cameronrutherford, line 47: I'm confused by this. There is no `patches` folder in `spack_repo`. I think this should be removed?
```suggestion
```

.github/actions/palace-ci/action.yml
```diff
@@ -0,0 +1,297 @@
...
+        fi
+
+        # As of version 1.2.0-alpha, the spack concretizer hangs when using flang.
+        # Also https://github.com/libxsmm/libxsmm/issues/996
+        # TODO: Try again in the future to see if we can use it.
+        if [[ "${{ inputs.toolchain }}" == "llvm" ]]; then
+          FORTRAN_COMPILER="gcc"
+        else
+          FORTRAN_COMPILER="${{ inputs.toolchain }}"
+        fi
```
cameronrutherford, line 66: Is this perhaps something we should enforce in our `package.py`? I imagine not as this isn't an us problem, but rather an issue with `libxsmm` / spack itself. That being said, it might help users not run into this if we have a conflict / error message in our `package.py`.

.github/actions/palace-ci/action.yml
```diff
@@ -0,0 +1,297 @@
...
+
+        if [[ "${{ inputs.toolchain }}" == "intel-oneapi" ]]; then
+          MATH_LIBS="${{ inputs.math-libs || 'intel-oneapi-mkl' }}"
+          MPI_IMPL="intel-oneapi-mpi"
+          C_CXX_COMPILER="intel-oneapi-compilers"
+          FORTRAN_COMPILER="intel-oneapi-compilers"
+        else
+          MATH_LIBS="${{ inputs.math-libs || 'openblas@0.3.29' }}"
```
cameronrutherford, line 76: Is this a version we should pin / depend on in our `package.py`? It would be conditional somehow based on the blas provider... This would also mean an explicit `depends_on` for `blas` in our `package.py`...

.github/actions/palace-ci/action.yml
```diff
@@ -0,0 +1,297 @@
...
+          export OMP_NUM_THREADS=2
+        else
+          export OMP_NUM_THREADS=1
+        fi
+
+        palace-unit-tests --skip-benchmarks ${GPU_FLAGS}
+        mpirun -np 2 $(which palace-unit-tests) --skip-benchmarks ${GPU_FLAGS}
+
+    - name: Upload spack environment
+      uses: actions/upload-artifact@v4
```
cameronrutherford, line 283: Maybe a brief comment about where to find these artifacts would be nice? I assume that they are viewable somewhere in the pipeline's view on GitHub...

spack_repo/local/packages/palace/package.py
```diff
@@ -182,6 +187,11 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("+cuda"):
         for arch in CudaPackage.cuda_arch_values:
             cuda_variant = f"+cuda cuda_arch={arch}"
+
+            # We need https://github.com/llnl/blt/pull/735, which is not available
+            # in <= 0.7.1
+            depends_on("blt@0.7.2:")
```
cameronrutherford, line 206: What about in the `+rocm` section? Is this only a CUDA based requirement?

spack_repo/local/packages/palace/package.py
```diff
@@ -240,10 +250,15 @@ def cmake_args(self):
                 )
             )
 
-        # HYPRE is always built with external BLAS/LAPACK
+        # HYPRE is always built with external BLAS/LAPACK and may need Umpire
```
cameronrutherford: Only by nature of the default - https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/hypre/package.py#L87. Maybe we should enforce this variant to make doubly sure then since this seems like a hard requirement from Palace.

palace/CMakeLists.txt
cameronrutherford: To Hugh's point, I wonder if this is unnecessary as Umpire will already define the `camp` package / target in it's exported CMake - https://github.com/llnl/Umpire/blob/develop/umpire-config.cmake.in#L26.

docs/src/developer/testing.md
```diff
@@ -315,12 +315,17 @@ julia --project -e "using Pkg; Pkg.instantiate()"
 
 You need to do this step only the very first time.
 
-#### Environment Variables
+#### Command Line Arguments
 
-  - `PALACE_TEST`: Path to *Palace* executable and optional arguments (default: "`palace`")
-  - `NUM_PROC_TEST`: Number of MPI processes (default: number of physical cores)
-  - `OMP_NUM_THREADS`: Number of OpenMP threads (default: 1)
```
cameronrutherford, line 322: Is this still an option? Are we hoping that users run the `--help` to find this one?

.github/actions/run-regression-tests/action.yml
```diff
@@ -0,0 +1,77 @@
...
+        julia-version: 'release'
+        spack-version: 'develop'
+        llvm-version: ${{ inputs.toolchain == 'llvm' && '19' || 'dont-install' }}
+        setup-intel: ${{ inputs.toolchain == 'intel-oneapi' && 'true' || 'false' }}
+
+    - name: Download spack environment
+      if: inputs.testing-on-build-runner != 'true'
+      uses: actions/download-artifact@v4
+      with:
+        name: spack-env-${{ runner.arch }}-${{ inputs.toolchain }}-${{ inputs.variant || 'default' }}-${{ inputs.math-libs || 'default' }}
```
cameronrutherford, line 48: I see - this is where it is used...

.github/actions/setup-runner/action.yml
cameronrutherford: I say leave it as is - having a single `apt update` instead of multiple is definitely preferred.

.github/actions/setup-runner/action.yml
```diff
@@ -0,0 +1,125 @@
...
+      shell: bash
+      run: |
+        sudo apt-get install -y gcc g++ gfortran libgomp1
+        
+    - name: Install Spack
+      uses: actions/checkout@v6
+      with:
+        repository: spack/spack
+        path: spack
+        ref: ${{ inputs.spack-version }}
```
cameronrutherford, line 97: Personally I prefer / like this more as it's clearer what's going on, but wondering why you prefer this over say - https://github.com/spack/setup-spack?

For example you missed the color option within spack here, which I quite like personally! I think as well it can avoid some churn incase something changes from the spack side.

.github/workflows/spack.yml
```diff
@@ -43,170 +43,149 @@ jobs:
...
-
-      - name: Configure Intel oneAPI compiler
-        if: needs.filter.outputs.test == 'true' && matrix.compiler == 'intel'
-        run:
-          sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp=2024.2.1-1079 \
-          intel-oneapi-compiler-fortran=2024.2.1-1079
-
-      - name: Setup Spack
-        if: needs.filter.outputs.test == 'true'
-        uses: spack/setup-spack@v2
```
cameronrutherford, line 74: I think keeping this as a part of bootstrapping is sensible.

.github/workflows/spack.yml
```diff
@@ -43,170 +43,149 @@ jobs:
...
+    needs: filter
+    if: needs.filter.outputs.test == 'true'
+    runs-on: [self-hosted, gpu]
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/palace-ci
+        with:
+          toolchain: gcc
+          variant: +cuda
+          run-regression-tests: 'false'
```
cameronrutherford, line 68: Any intuition as to why some regression tests are disabled?

.github/actions/palace-ci/action.yml
cameronrutherford: Seems like this is just used later by the testing part of the actions...

docs/src/developer/testing.md
cameronrutherford: By "this" I mean `OMP_NUM_THREADS`

.github/actions/palace-ci/action.yml
Sbozzolo: My bias would be to not enforce anything in our package.py: this is really not Palace's fault/responsibility.

.github/actions/palace-ci/action.yml
Sbozzolo: This is a workaround because openblas@0.3.30 does not compile on some arm platforms (https://github.com/OpenMathLib/OpenBLAS/issues/5459).

I'll add a link to the issue and open an issue to keep track of this.

Sbozzolo: https://github.com/awslabs/palace/issues/598

.github/actions/palace-ci/action.yml
Sbozzolo: Added a comment to explain how they are used.

spack_repo/local/packages/palace/package.py
Sbozzolo: Yes, this is not really a requirement for Palace itself, but for one of its dependencies. We define it as dependency to enforce the version, and we need a modern version only because of a fix for CUDA.

spack_repo/local/packages/palace/package.py
Sbozzolo: Changed to enforce the default.

.github/actions/run-regression-tests/action.yml
Sbozzolo: Added comments to explain what's happening.

palace/CMakeLists.txt
Sbozzolo: I think I found it necessary for linking in static builds, where we don't have transitive linking.

.github/actions/palace-ci/action.yml
Sbozzolo: I found that I needed to patch packages.py frequently to get things to work (see https://github.com/spack/spack-packages/pulls?q=is%3Apr+author%3ASbozzolo+is%3Aclosed). 

When I found that things were broken, I would fix them and open a PR to spack-packages. Then, I would add the patch to this folder to make sure that CI would pass. I would then remove the patch once the PR got merged. At this point, all my PRs got merged, but I think that this is a useful infrastructure, so I left it here. 

I will add a file in the patch folder to explain the workflow.

docs/src/developer/testing.md
Sbozzolo: Yes, it's still an option, and yes, I expect users to run --help, as specified in the docs
> Run `julia --project runtests.jl --help` to see all available options with descriptions and defaults.

.github/workflows/spack.yml
Sbozzolo: You will see that there are two types of jobs: `build-xxx` and `build-test-xxx`. The `build` jobs have `run-regression-tests` disabled because we spawn many more runners in parallel in seperate `test` jobs

image

.github/actions/setup-runner/action.yml
Sbozzolo: Added color

.github/workflows/spack.yml
cameronrutherford: This clarified as well why I was confused looking at the pipeline graph before. Awesome 🥳

spack_repo/local/packages/palace/package.py
cameronrutherford: Hmmm maybe a nit then but would this be better?

```suggestion
            # We need https://github.com/llnl/blt/pull/735, which is not available
            # in <= 0.7.1
            depends_on("umpire %blt@0.7.2:")
```

I don't know how / if it really affects the concretizer, but this is perhaps more specifically showing what we "depend on"?

We should add `raja %blt` as well if that is the case.

.github/actions/palace-ci/action.yml
cameronrutherford: I think that's reasonable in this case.

palace/CMakeLists.txt
cameronrutherford: Sounds good to me. The `find_package` should find the same one that Umpire exports as well quite easily, and the call to that CMake function isn't expensive.

You could optionally do a `if(NOT TARGET camp)` or something like that to check but sounds annoying / less clear.

You could also just remove `find_package(camp)` entirely and that might still work... but that's just making the CMake more confusing than it needs to be.

cameronrutherford (approved):
:shipit:

Issue #572, opened 2025-11-30
Installation on WSL


Etawn:
Hello,

I tried to install Palace on WSL since there is no Windows version. I followed the steps in this comment: https://github.com/awslabs/palace/issues/485#issuecomment-3321832414. I also tried the steps provided by Volker (https://github.com/VolkerMuehlhaus/gds2palace_ihp_sg13g2) , which are very similar. I attempted both the Spack and Apptainer installations.

I am facing two issues:

1. Apptainer:
When I run git clone https://github.com/awslabs/palace.git, I get an error saying that the command did not find a proper EOF and it stops.
If I download the repository manually on Windows and then run 'singularity build palace.sif /home/github/palace/singularity/singularity.def' followed by 'apptainer shell palace.sif',
the container does not recognize the palace command afterward. I cannot run 'palace /home/.../myfile.json'.

2. Spack:
Sometimes I get checksum errors when downloading and it stops (I tried multiple times).
When it does download without download errors, launching Palace fails with the following error:

```
flori@RF:/mnt/c/Users/flori$ spack install palace
==> Error: failed to concretize `palace` for the following reasons:
     1. palace: Need at least one sparse direct solver
     2. palace: At least one eigenvalue solver is required
     3. palace: Need at least one sparse direct solver
        required because conflict is triggered when ~mumps~strumpack~superlu-dist
          required because palace requested explicitly
        required because conflict constraint
          required because palace requested explicitly
     4. palace: At least one eigenvalue solver is required
        required because conflict constraint
          required because palace requested explicitly
        required because conflict is triggered when ~arpack~slepc
          required because palace requested explicitly
```

Do you know what is the problem please ? I did not find solution on the issues' tickets or wiki and was not able to solve it myself
Sincerely

Etawn:
@Sbozzolo Thank you for the help. Please find below the report of your command : 

```
flori@RF:/mnt/c/Users/flori$ spack spec palace
 -   palace@0.14.0~arpack~cuda+gslib~int64~ipo+libxsmm~mumps~openmp~rocm+shared+slepc~strumpack+sundials+superlu-dist build_system=cmake build_type=Release commit=a428a3a32dbbd6a2a6013b3b577016c3e9425abc generator=make platform=linux os=ubuntu24.04 target=skylake %c,cxx=gcc@13.3.0
[...]

PR #579, opened 2025-12-08
Mention Spack Prerequisites

Sbozzolo:
I investigated Palace's compatibility with WSL as part of #572.

I found that Spack requires certain packages to be present, but does not advertise this in its installation process. One of the missing packages is `bzip2`, leading to errors like
```
==> No binary for xz-5.6.3-nkxqqpyo4qf26cpadny6oqsddq6vkdn7 found: installing from source
==> Installing xz-5.6.3-nkxqqpyo4qf26cpadny6oqsddq6vkdn7 [21/70]
==> Fetching https://mirror.spack.io/_source-cache/archive/a9/a95a49147b2dbb5487517acc0adcd77f9c2032cf00664eeae352405357d14a6c.tar.bz2
    [100%]    1.73 MB @   58.4 MB/s
tar (child): lbzip2: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
/usr/bin/tar: Child returned status 2
/usr/bin/tar: Error is not recoverable: exiting now
```
which are not immediately interpretable as a missing required dependency.

This commit updates the documentation to draw attention to the prerequisites.

Once the preerquisites are installed, Palace installs fine on Ubuntu/WSL.

### DIFF
--- a/docs/src/quick.md
+++ b/docs/src/quick.md
@@ -120,7 +120,7 @@ high-performance computing applications.
 Follow the [instructions on the official
 website](https://spack.readthedocs.io/en/latest/getting_started.html) to install
 Spack. This involves cloning a repository and sourcing a `setup-env` shell
-script. In this, ensure that your system comes with the minimal requirements as
+script. Ensure this shell environment comes with the minimal requirements as
 listed in the [official
 documentation](https://spack.readthedocs.io/en/latest/installing_prerequisites.html).
 Come back here once you are done with that.


docs/src/quick.md
```diff
@@ -120,7 +120,10 @@ high-performance computing applications.
 Follow the [instructions on the official
 website](https://spack.readthedocs.io/en/latest/getting_started.html) to install
 Spack. This involves cloning a repository and sourcing a `setup-env` shell
-script. Come back here once you are done with that.
+script. In this, ensure that your system comes with the minimal requirements as
```
hughcars:
```suggestion
script. Ensure this shell environment comes with the minimal requirements as
```
just to avoid confusion that the system might, but the shell might not.

PR #584, opened 2025-12-17
Add frequently asked questions

Sbozzolo:
Collecting some common installation known problems.

### DIFF
--- a/docs/src/faq.md
+++ b/docs/src/faq.md
@@ -71,6 +71,7 @@ Palace version and the example files you're using. Here's how to ensure
 compatibility:
 
  1. Check your Palace version by running `palace --version` in your terminal
+    (if `--version` is supported, if not, check the log of any Palace simulation)
  2. Navigate to the documentation for your specific version using the version selector
  3. Download example files from the GitHub repository at the tag or commit corresponding to your version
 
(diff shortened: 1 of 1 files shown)


docs/src/faq.md
```diff
@@ -0,0 +1,199 @@
...
+The [official documentation](https://awslabs.github.io/palace/stable/) includes
+a version selector at the bottom of the sidebar to help you choose the correct
+documentation version. We generally recommend using the most recent version of
+Palace unless you have specific reasons to use an older version.
+
+The most common cause of errors when running examples is a mismatch between your
+Palace version and the example files you're using. Here's how to ensure
+compatibility:
+
+ 1. Check your Palace version by running `palace --version` in your terminal
```
hughcars, line 73: `--version` was only introduced relatively recently (late in 0.13, so 0.14 is the first one with it). Probably not much of an issue, but can also suggest checking the terminal output for `vX.Y.Z-C-<sha>` or similar for example.

docs/src/faq.md
```diff
@@ -0,0 +1,199 @@
...
+ 3. If such a PR exists, you can track its progress and wait for it to be merged
+
+Once the PR is merged, run the update command shown above to pull the changes to
+your local copy of Spack.
+
+##### Installing Development Versions
+
+If you need the absolute latest version of Palace (including unreleased
+features), you can install the development version directly from the main
+branch:
```
hughcars:
```suggestion
If you need the absolute latest version of Palace (including
features not yet in a released version), you can install the development version directly from the main
branch:
```

PR #604, opened 2026-01-22
Update spack package.py to tagged MFEM, patches sundials

Sbozzolo:
Update to tagged MFEM 4.9. Applies this patch https://github.com/spack/spack-packages/pull/3289.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ The format of this changelog is based on
 
 #### New Features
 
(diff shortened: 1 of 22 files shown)


.github/license-check/md.txt
```diff
@@ -1,2 +1,4 @@
-<!--- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved. --->
```
hughcars, line 1: Ok with the license change, only thing is we want to be sure that format isn't going to cause thrash in future.

PR #610, opened 2026-01-26
MFEM dependency update

hughcars:
Updating the mfem dependency. Requires introducing a bugfix to mfem https://github.com/mfem/mfem/pull/5210 which fixes a bug introduced by https://github.com/mfem/mfem/pull/4028. The memory backing for the DenseMatrix was changed to Array, which has a semantic bug where move assignment would not alter the underlying data. The fix was to force a fallback to copy assignment if the data is not owned, as in such a scenario there can be no theft of the underlying resource.

Update: went for an alternative fix where all operator() assignment into the tensors is removed in favour of using `.Set` rather than rebinding. This isn't very ergonomic/is fairly surprising, but if MFEM doesn't agree with the fix PR, this is the approach we will need going forward.

### DIFF
--- a/spack_repo/local/packages/palace/package.py
+++ b/spack_repo/local/packages/palace/package.py
@@ -149,7 +149,8 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("@0.16:"):
         # +lapack means: use external lapack
         depends_on(
-            "mfem+mpi+metis+lapack cxxstd=17 commit=0e61a94b5f78f80d9514fd7d442a3c2e8f9412f9",
+            # TODO: update to v4.9 == d9d6526cc1749980a2ba1da16e2c1ca1e07d82ec once spack updated
+            "mfem+mpi+metis+lapack cxxstd=17 commit=d9d6526cc1749980a2ba1da16e2c1ca1e07d82ec",
             patches=[
                 "patch_par_tet_mesh_fix_dev.diff",
                 "patch_gmsh_parser_performance.diff",
--- a/cmake/ExternalGitTags.cmake
+++ b/cmake/ExternalGitTags.cmake
@@ -133,7 +133,7 @@ set(EXTERN_MFEM_GIT_BRANCH
   "Git branch for external MFEM build"
 )
 set(EXTERN_MFEM_GIT_TAG
-  "0e61a94b5f78f80d9514fd7d442a3c2e8f9412f9" CACHE STRING
+  "d9d6526cc1749980a2ba1da16e2c1ca1e07d82ec" CACHE STRING
   "Git tag for external MFEM build"
 )
 


spack_repo/local/packages/palace/package.py
```diff
@@ -149,12 +149,10 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("@0.16:"):
         # +lapack means: use external lapack
         depends_on(
-            "mfem+mpi+metis+lapack cxxstd=17 commit=bf1b25d82e232909e1d3aa1f416dbdae9931a550",
+            "mfem+mpi+metis+lapack cxxstd=17 commit=0e61a94b5f78f80d9514fd7d442a3c2e8f9412f9",
```
cameronrutherford: I wonder if @Sbozzolo has thoughts on this given https://github.com/spack/spack-packages/pull/3038/changes isn't quite merged yet.

I think something along the lines of the following is true:

> Every time we update the version of MFEM that works with Palace on develop, this temporarily breaks `spack install palace@develop` for external users until we add another PR into `spack-packages`.

Some potential solutions / thoughts:

1. Force each delta on MFEM version to trigger a new Palace patch version release. This would make it super clear to users that something is changing underneath them between versions.
2. Have some caveat in our repo / packag / docs that "develop may not have the right MFEM patch associated in the `package.py`. Use the local `spack_repo` in Palace in order to get the most up-to-date version to avoid this
3. Accept the slight delay associated with merging something into `spack-packages`, and communicate to users accordingly. Then simply have an MFEM commit associated with each Palace minor version

I am sure there's a sensible resolution to this, but perhaps discussion is warranted. IMO I am in favor of the order above, but also open to other suggestions.

hughcars: I actually believe this PR doesn't actually break develop, develop will just be using the prior mfem point. The changes introduced in the PR right now are backwards compatible. The changes currently on develop are such that we couldn't issue a patch release, without going back and effectively forking at the 0.15.0 release, which is a lot of work for a project where we're more focused on forward velocity, so this should really warrant an 0.16. There is already a pr that @Sbozzolo has on spack that I believe updates a few things.

As you say though, there is an issue that effectively each minor palace version is going to depend on its own mfem commit effectively. Our number of patches is dwindling, but getting it to zero will be tricky. If we moved past that, and had mfem pinned at their minor version releases, i.e. ExternalGitTags is set to them (or uses the tags themselves), then we'd be free of this. The patches would run the risk of being mangled however with each minor version bump, as they are not as intelligently merged as `git merge`. We're then into the spack dependency model, but give up on speed of updates to mfem. I would speculate that we could apply the patches on 4.9, and probably be ok.

cameronrutherford: Okay I think I understand what you are saying. I am satisfied with this direction, and it's good to know that these changes are also backwards compatible as well.

simlapointe (approved):

PR #613, opened 2026-01-28
Embed json schema validator in binary

hughcars:
*Note: this PR is part of a series broadly aimed at refactor config file to make it simpler to use, and drive towards a simpler testing experience. The next two branches in the sequence are `hughcars/config-constructor-conversions` and `hughcars/increased-schema-tests` which will be PR'd staged off of this. The rough model is this PR adds in the validator, the next one removes duplication and the final one is further hardening.* 

This PR introduces a dependency on `json-schema-validator` and uses this to allow checking a config against the schema directly within c++. It avoids the need for a separate tracking of the schema by embedding it within the executable, to avoid the need for file reading. This adds ~30kB to the binary, but makes it effectively self describing. It effectively adds a set of redundant checks, given many of these constraints are checked during json reading, but this will enable a follow up PR that removes these checks whilst simplifying `configfile.hpp` and `configfile.cpp`.

The validation can be done wholesale, where a top level json is provided, or using a keyword, to test only against a subset of the json. Additionally it was necessary to downgrade the schema version we are using, to match what the validator library was capable of, luckily the schema we have is actually already compliant with that version (with 1 tiny exception), so this required minimal changes other than the tag.

The error reporting from the schema follows the documentation `["Key1"]["Key2"]` style of printing, and should also print all errors it finds in the config at once, rather than failing at the first it finds and then erroring out. So a user should get a more complete description of the fixes they need to make, without retrying so much.

The check config github action will eventually become redundant and be removed, but for now I am leaving it in as a follow up PR hardens the schema to cover as much as possible.

- [x] Rebase on main
- [x] Update CHANGELOG with PR number

### DIFF
--- a/palace/utils/iodata.cpp
+++ b/palace/utils/iodata.cpp
@@ -206,8 +206,10 @@ IoData::IoData(const char *filename, bool print) : units(1.0, 1.0), init(false)
     std::string err = ValidateConfig(config);
     if (!err.empty())
     {
-      Mpi::Warning("Configuration file validation failed!\n{}", err);
-      Mpi::Abort(EXIT_FAILURE);
+      Mpi::Warning("{}", err);
+      std::cout << std::flush;  // Flush before abort to avoid clipping the error message
+      Mpi::Barrier();
+      MFEM_ABORT("Configuration file validation failed!");
     }
   }
 


palace/utils/iodata.hpp
```diff
@@ -4,6 +4,8 @@
 #ifndef PALACE_UTILS_IODATA_HPP
 #define PALACE_UTILS_IODATA_HPP
 
+#include <string>
+#include <nlohmann/json_fwd.hpp>
```
Sbozzolo: I think this can be removed as you are using `ValidateConfig` directly.

scripts/schema/Project.toml
Sbozzolo, line 1: Is this file used in any way?

spack_repo/local/packages/palace/package.py
```diff
@@ -65,6 +65,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("lapack")
     depends_on("zlib-api")
     depends_on("nlohmann-json")
+    depends_on("nlohmann-json-schema-validator@2.4.0:")
```
Sbozzolo, line 68: Do you know that version 2.4.0 specifically is needed?

cmake/EmbedSchema.cmake
```diff
@@ -0,0 +1,92 @@
...
+file(GLOB_RECURSE SCHEMA_FILES "${SCHEMA_DIR}/*.json")
+
+# Mark schema files as configure dependencies so CMake reconfigures when they change
+set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${SCHEMA_FILES})
+
+# Generate the header file
+set(SCHEMA_HEADER_CONTENT
+"// Auto-generated file - do not edit
+// Generated from scripts/schema/*.json
+// Provides embedded JSON schema strings for runtime config validation without file dependencies.
```
Sbozzolo:
```suggestion
" // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0
// Auto-generated file - do not edit
// Generated from scripts/schema/*.json
// Provides embedded JSON schema strings for runtime config validation without file dependencies.
```

cmake/EmbedSchema.cmake
```diff
@@ -0,0 +1,92 @@
...
+  string(APPEND SCHEMA_HEADER_CONTENT
+"inline const char* ${VAR_NAME} = R\"json(
+${FILE_CONTENT})json\";
+
+")
+endforeach()
+
+# Add a map for looking up schemas by path
+string(APPEND SCHEMA_HEADER_CONTENT
+"inline const std::unordered_map<std::string, const char*>& GetSchemaMap()
```
Sbozzolo: Do we want to maybe add an embedded docstring as well?

test/unit/test-schema.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+using json = nlohmann::json;
+using namespace palace;
+
+namespace fs = std::filesystem;
+
+TEST_CASE("Schema Validation - Embedded Schema Matches Source", "[schema][Serial]")
+{
+  // Verify embedded schemas match source files (catches stale builds).
+  std::string schema_dir = fmt::format("{}/../../scripts/schema", PALACE_TEST_DIR);
```
Sbozzolo, line 21: If this is easy, could you check that this works with Spack as well? We might need to be a little bit more careful because of the fact that Spack relocates objects.

Sbozzolo: Actually, this is already tested in CI.

spack_repo/local/packages/palace/package.py
hughcars: We're making use of a feature that wasn't in 2.3, I can't remember exactly the name for it, but it was to do with `$ref` resolution.

cmake/EmbedSchema.cmake
hughcars: I wasn't thinking that users will be opening these header files generated in their build folder, but in theory maybe they'd go looking for them through their IDE.

scripts/schema/Project.toml
hughcars: Accidental commit, sending to the bin.

spack_repo/local/packages/palace/package.py
hughcars: Addressed in 9f6741c0a0779015b381801dadc2c5e4381d129e

palace/utils/iodata.hpp
hughcars: Addressed in 9f6741c0a0779015b381801dadc2c5e4381d129e

cmake/EmbedSchema.cmake
hughcars: Addressed in 9f6741c0a0779015b381801dadc2c5e4381d129e

cmake/EmbedSchema.cmake
hughcars: Addressed in 9f6741c0a0779015b381801dadc2c5e4381d129e

PR #615, opened 2026-01-29
Make IoData quadrature error loud

hughcars:
Some of the stratton chu tests were occasionally missing by 8e-6 rather than 5e-6, when very close to 0.0, in some CI runs on unrelated changes. This PR bumps that tolerance to 1e-5 which should be tight enough. 

Update: setting the quadrature rule to `p_trial` to the value that would be used in the main code path, causes the deviation. The test was using p=1 quadrature, which comes out at second order, compared to 6th order if `p_trial=3` was set. Thus the test was previously effectively underintegrating.

### DIFF
--- a/palace/fem/integrator.cpp
+++ b/palace/fem/integrator.cpp
@@ -13,6 +13,9 @@ namespace fem
 
 int DefaultIntegrationOrder::Get(const mfem::IsoparametricTransformation &T)
 {
+  MFEM_VERIFY(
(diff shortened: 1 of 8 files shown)


Sbozzolo:
Could it be that what you are running into in your #613  is not flakiness, but an actual bug, like the one @phdum-a found regarding the MFEFM gloabls set in `CheckConfiguration`?

PR #620, opened 2026-02-03
Fix eigen warning 

hughcars:
The eigen dependency was moved past 3.4 (to 5.0.0), so we've been getting warnings in the build. The fix is moving the option into the template args, and was already commented out in the code anyway.


Sbozzolo (approved):
Thank you! I saw that too and had it in my todo list.

hughcars:
This is a tiny PR, so we could avoid merging it right away to stop https://github.com/awslabs/palace/pull/326 having to rebase quite so furiously.

PR #622, opened 2026-02-04
Add support for nightly tests

### LINKED ISSUE #221: Add AMR examples to regression tests

Sbozzolo:
This PR adds support for nightly tests in Palace.

Nightly tests are automatically run from open PRs unless:
- the PR already passed a nightly test and has not changed
- the PR has a `no-nightly` label
- the PR is a draft
- the PR is trivial (e.g., it only changes documentation)

Nightly tests can be triggered immediately by adding the `trigger-nightly` label.

Before merging we should turn the PRs we don't want to run the nightly for to `draft`.

Closes #221


.github/actions/palace-ci/action.yml
```diff
@@ -315,7 +326,7 @@ runs:
         path: |
           spack.yaml
           spack.lock
-        retention-days: 1
+        retention-days: 7
```
Sbozzolo, line 329: This is so that we can re-use the artifact for the nightly.

.github/actions/palace-ci/action.yml
```diff
@@ -23,6 +23,14 @@ inputs:
...
     required: false
     default: 'false'
+  test-cases:
+    description: 'Test cases: "default", "nightly", or space-separated list'
+    required: false
+    default: 'default'
+  additional-test-cases:
+    description: 'Additional test cases to run after default tests (space-separated list)'
+    required: false
+    default: ''
```
Sbozzolo, line 33: I added this so that I can run usual tests + coarse transmon. 

I also see ourselves adding more tests and test groups so that additional control might be useful

.github/actions/setup-runner/action.yml
```diff
@@ -110,6 +110,14 @@ runs:
 
         end_timer "Setup repositories"
 
+    - name: Setup gh
```
Sbozzolo, line 113: Used to post statuses

docs/src/examples/transmon.md
```diff
@@ -220,7 +220,7 @@ participation ratios for inductive lumped elements (the Josephson junction). The
 the sapphire substrate, enabling comprehensive analysis of loss mechanisms
 throughout the device.
 
-#### Balancing Speed and Accuracy: Solver Order Considerations
+#### Balancing Speed and Accuracy: Solver Order Considerations and
```
Sbozzolo: I realized I left an "and". I'll remove it when addressing other review comments.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,141 @@
...
+#
+# Nightly tests can be triggered immediately by adding 'trigger-nightly' label
+# to a PR.
+name: Nightly Tests
+
+on:
+  schedule:
+    - cron: '0 2 * * *'  # 2 AM daily
+  pull_request:
+    types: [synchronize]
```
hughcars: If this can only be triggered on sync, what about the "label" dispatch, don't you need `labeled` too here? (Thinking about the immediate trigger scenario).

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,141 @@
...
+          script: |
+            let prs = [];
+
+            // Handle manual trigger via 'trigger-nightly' label
+            if (context.eventName === 'pull_request') {
+              const hasLabel = context.payload.pull_request.labels
+                .some(label => label.name === 'trigger-nightly');
+              if (hasLabel) {
+                prs = [context.payload.pull_request.number];
+              }
```
hughcars: If there was a nightly failure, does a change to the PR after that reset the label to pending? Say opened on day 1, nightly fails at 2 am on day 2, then some fixes are done, does the label have to be manually applied, or is it readded to the list of nightly PRs to test?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,141 @@
...
+      - uses: actions/checkout@v6
+        with:
+          ref: refs/pull/${{ matrix.pr }}/merge
+
+      - uses: ./.github/actions/palace-ci
+        with:
+          toolchain: gcc
+          test-cases: nightly
+
+      - name: Get PR head SHA
```
hughcars, line 113: When `palace-ci` fails, this step is skipped (GitHub Actions defaults to `if: success()`), so `steps.pr-sha.outputs.sha` is never set. The "Set failure status" step then hits a malformed API URL and the failure is never reported — the PR stays in `pending` with no feedback to the author. The nightly schedule retries it every night, but for persistent failures this means silent compute burn with no developer signal.

Fix: add `if: always()` here so the SHA is captured regardless of build outcome.

test/examples/baseline
```diff
@@ -100,7 +100,8 @@ spheres
 rings
 antenna/antenna_halfwave_dipole
 antenna/antenna_short_dipole
-transmon
+transmon/coarse
```
hughcars: The new test names (`transmon/coarse`, `transmon/amr`) produce `POSTPRO=coarse` and `POSTPRO=amr` from this awk split, but the actual output paths are `postpro/transmon_coarse` and `postpro/transmon_amr`. Both `EXAMPLE_DIR` and `REF_DIR` will point to wrong locations. Also `cp *.csv` won't recurse into AMR's `iterationN/` subdirectories.

test/examples/runtests.jl
```diff
@@ -152,12 +153,46 @@ end
...
+            "Re{I[3]} (A)",
+            "Im{I[3]} (A)"
+        ],
+        skip_rowcount=true,
+        gridfunction_fields=true
+    )
+end
+
+if "transmon/amr" in cases
+    @info "Testing single transmon arm..."
```
hughcars: Nit: "arm" → "amr"

.github/actions/run-regression-tests/action.yml
```diff
@@ -25,6 +25,14 @@ inputs:
     description: 'Solver to test'
     required: false
     default: 'Default'
+  test-cases:
+    description: 'Test cases: "default", "nightly", or space-separated list'
+    required: false
+    default: 'default'
+  additional-test-cases:
+    description: 'Additional test cases to run after test-cases (space-separataed list)'
```
hughcars: Nit: "separataed" → "separated"

docs/src/examples/transmon.md
hughcars: Trailing "and" in the heading.

PR #629, opened 2026-02-09
Add memory reporting

Sbozzolo:
This PR adds support for reporting memory used by Palace (peak and current). This is a particularly useful metric as it is often the case that users run out of memory. This report can help them understand memory requirements better.

I decided to brute-force this problem and use rely on the operating system to tell us how much memory is being used, instead of doing accounting myself.

For Linux, I am using MaxRSS, Resident Set Size. I've used this in the past for other scientific applications and always found reasonably accurate. It is not quite perfect because of memory shared with other processes (e.g., in shared libraries), or kernel optimizations. In practice, this does not really matter for an application like Palace where actual memory allocations dominates everything.

I am not familiar with MacOS, and reading online brought to this page https://developer.apple.com/forums/thread/105088 where I learned that I should be looking at phys_footprint in `mach`.

For peak memory, I use the POSIX `getrusage`.

I save the results to `palace.json`, which helps tracking with overall performance metrics.

I tested this on Mac Tahoe and compared the results with `/usr/bin/time -l`.
I tested this on Linux with `/usr/bin/time -v` and with `free`.

### DIFF
--- a/palace/utils/memoryreporting.cpp
+++ b/palace/utils/memoryreporting.cpp
@@ -169,22 +169,22 @@ MemoryStats ComputeNodeStats(std::string label, long local_value, MPI_Comm comm)
 
 MemoryStats GetCurrentMemoryStats(MPI_Comm comm)
 {
-  return ComputeStats("Current per-rank", GetCurrentMemory(), comm);
+  return ComputeStats("current per-rank", GetCurrentMemory(), comm);
 }
 
 MemoryStats GetPeakMemoryStats(MPI_Comm comm)
 {
-  return ComputeStats("Peak per-rank", GetPeakMemory(), comm);
+  return ComputeStats("peak per-rank", GetPeakMemory(), comm);
 }
 
 MemoryStats GetCurrentNodeMemoryStats(MPI_Comm comm)
 {
-  return ComputeNodeStats("Current per-node", GetCurrentMemory(), comm);
+  return ComputeNodeStats("current per-node", GetCurrentMemory(), comm);
 }
 
 MemoryStats GetPeakNodeMemoryStats(MPI_Comm comm)
 {
-  return ComputeNodeStats("Peak per-node", GetPeakMemory(), comm);
+  return ComputeNodeStats("peak per-node", GetPeakMemory(), comm);
 }
 
 void PrintMemoryUsage(MPI_Comm comm, const MemoryStats &stats)


palace/utils/memoryreporting.cpp
```diff
@@ -0,0 +1,138 @@
...
+// 1. It contains memory shared with other processes,
+// 2. It doesn't fully account for kernel-internal optimizations.
+// For an application like Palace, MaxRSS is often accurate at percent level or
+// below.
+//
+// For MacOS, this is very hard, but this thread suggests using phys_footprint:
+// https://developer.apple.com/forums/thread/105088
+//
+// See: man footprint
+long GetCurrentMemory()
```
hughcars, line 37: Is it worth using unsigned to get just a bit larger? Or is long going to capture anything on a sensible computer?

palace/utils/memoryreporting.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+MemoryStats ComputeStats(std::string label, long local_value, MPI_Comm comm)
+{
+  long val_min = local_value, val_max = local_value, val_sum = local_value;
+  Mpi::GlobalMin(1, &val_min, comm);
+  Mpi::GlobalMax(1, &val_max, comm);
+  Mpi::GlobalSum(1, &val_sum, comm);
+  return {label, val_min, val_max, val_sum, static_cast<double>(val_sum) / Mpi::Size(comm)};
+}
```
hughcars, line 117: Ok, so we're really looking at per shared memory block? So say we're using 2 nodes of 64 ranks, the number we'd care about would be 64 * avg, as the total in shared memory? Is it worth reporting on "Shared memory" communicators but summed? I.e. report 1 number for each node, rather than each rank within the node? This might be a bit more involved but is more like what an HPC users needs to understand how close they are to their limit? 

Or possibly reporting just total is easier and expecting users to do that arithmetic themselves. Thoughts?

palace/utils/memoryreporting.cpp
```diff
@@ -0,0 +1,138 @@
...
+MemoryStats GetPeakMemoryStats(MPI_Comm comm)
+{
+  return ComputeStats("Peak", GetPeakMemory(), comm);
+}
+
+void PrintMemoryUsage(MPI_Comm comm, const MemoryStats &stats)
+{
+  Mpi::Print(comm, "\n{} RSS: Min. {}, Max. {}, Avg. {}, Total {}\n", stats.label,
+             FormatBytes(stats.min), FormatBytes(stats.max), FormatBytes(stats.avg),
+             FormatBytes(stats.sum));
```
hughcars: RSS seems a bit opaque, can we have `"Estimate peak memory usage"` or something to be a little more user friendly?

palace/utils/memoryreporting.cpp
Sbozzolo: I used `long` because that's what `rusage` uses:
```
     struct rusage {
             struct timeval ru_utime; /* user time used */
             struct timeval ru_stime; /* system time used */
             long ru_maxrss;          /* max resident set size */
             long ru_ixrss;           /* integral shared text memory size */
             long ru_idrss;           /* integral unshared data size */
             long ru_isrss;           /* integral unshared stack size */
             long ru_minflt;          /* page reclaims */
             long ru_majflt;          /* page faults */
             long ru_nswap;           /* swaps */
             long ru_inblock;         /* block input operations */
             long ru_oublock;         /* block output operations */
             long ru_msgsnd;          /* messages sent */
             long ru_msgrcv;          /* messages received */
             long ru_nsignals;        /* signals received */
             long ru_nvcsw;           /* voluntary context switches */
             long ru_nivcsw;          /* involuntary context switches */
     };
```

So, I guess it's fine?

palace/utils/memoryreporting.cpp
Sbozzolo: Added the per-node reporting as well

palace/utils/memoryreporting.cpp
hughcars: Sounds good to me. If we ever saw an overflow there, we'd be talking to someone running a pretty enormous job so that'd just be exciting.

Sbozzolo: So, I computed this. On 64-bit machines, the max `long` is 8 PB. So, I think we are safe :)

PR #633, opened 2026-02-11
Fix complex number multiplication bug in diagonal operator

Thounthoun:
The imaginary part calculation had a typo. It was using ti twice instead of tr and ti. Fixed in both AddMult and AddMultHermitianTranspose.


hughcars:
This PR is touching hundreds of things, but advertises itself as a bugfix PR. A bugfix PR for something small like this needs to be a single commit, and the minimum change, to make it easy for humans to read and review. Undo all your other changes.

PR #637, opened 2026-02-13
Add example of test fixture

Sbozzolo:
I am peeling this of #538 to make updating #538 easier. This is independent of test parallelization and coverage anyway.

This PR introduces `fixtures.hpp` and an example of using a Catch2 test fixture to create temporary directories for tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ The format of this changelog is based on
     when computing the average of field values on interior boundaries, causing incorrect
     results for interior trace postprocessing.
     [PR 621](https://github.com/awslabs/palace/pull/621).
+  - Fixed a bug in complex diagonal operator multiplication where the imaginary part used an
+    incorrect term in `AddMult` and `AddMultHermitianTranspose`.
+    [PR 633](https://github.com/awslabs/palace/pull/633).
 
 #### Interface Changes
 
--- a/palace/linalg/operator.cpp
+++ b/palace/linalg/operator.cpp
@@ -539,7 +539,7 @@ void BaseDiagonalOperator<ComplexOperator>::AddMult(const ComplexVector &x,
                         const auto tr = DR[i] * XR[i] - DI[i] * XI[i];
                         const auto ti = DI[i] * XR[i] + DR[i] * XI[i];
                         YR[i] += ar * tr - ai * ti;
-                        YI[i] += ai * ti + ar * ti;
+                        YI[i] += ai * tr + ar * ti;
                       });
 }
 
@@ -589,7 +589,7 @@ void DiagonalOperatorHelper<BaseDiagonalOperator<ComplexOperator>, ComplexOperat
                         const auto tr = DR[i] * XR[i] + DI[i] * XI[i];
                         const auto ti = -DI[i] * XR[i] + DR[i] * XI[i];
                         YR[i] += ar * tr - ai * ti;
-                        YI[i] += ai * ti + ar * ti;
+                        YI[i] += ai * tr + ar * ti;
                       });
 }
 
--- a/spack_repo/local/packages/palace/package.py
+++ b/spack_repo/local/packages/palace/package.py
@@ -65,7 +65,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("lapack")
     depends_on("zlib-api")
     depends_on("nlohmann-json")
-    depends_on("nlohmann-json-schema-validator@2.4.0:")
+    depends_on("nlohmann-json-schema-validator@2.4.0:", when="@0.16:")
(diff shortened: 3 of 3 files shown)


docs/src/developer/tutorial_add_new_unit_test.md
```diff
@@ -166,3 +166,45 @@ TEST_CASE("MyTest Vector Sum - Different Lengths", "[myvector][Parallel]")
...
+  // temp_dir is available and will be cleaned up automatically.
+  auto file_path = temp_dir / "vector.txt";
+  Vector v;
+  v = 2;
+  {
+     std::ofstream file(file_path);
+     vec.Print(file);
+  }
+  CHECK(std::filesystem::exists(file_path));
+}
```
hughcars: Not sure this adds anything compared to above, unless you're trying to show they are non-clashing? Are they nonclashing? i.e. if you did a CHECK it doesn't exist before creating file_path in each, does it pass?

test/unit/fixtures.hpp
```diff
@@ -0,0 +1,83 @@
...
+};
+
+// Fixture that creates a temporary directory for each test and cleans it up automatically.
+template <TempDirMode Mode = TempDirMode::PerRank>
+struct TempDirFixture
+{
+  fs::path temp_dir;
+
+  TempDirFixture()
+  {
```
hughcars: Given this is a template on a concrete set of enums, maybe make this a cpp with explicit instantiation? Not wedded to the necessity, but it places compilation at the lowest level in the file graph, compared to headers which affect any consumer.

test/unit/fixtures.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace palace::test
+{
+
+namespace fs = std::filesystem;
+
+enum class TempDirMode
+{
+  PerRank,  // Each MPI rank gets its own directory
+  Shared    // All ranks share one directory
+};
```
hughcars: Do you foresee there being more modes for this? Given these are `if constexpr` forks, is there not a value in having these be separate structs? It avoids the `<>` in the fixture naming, and removes all the `if constexpr`. They seem to be only sharing that they are structs with a member `temp_dir`.

test/unit/test-schema.cpp
```diff
@@ -80,10 +81,11 @@ TEST_CASE("Schema Validation - Example Configs", "[schema][Serial]")
   }
 }
 
-TEST_CASE("Schema Validation - Config with Comments", "[schema][Serial]")
+TEST_CASE_METHOD(palace::test::TempDirFixture<>, "Schema Validation - Config with Comments",
```
hughcars: So if they were split, `::SharedTempDir` or `::PerRankTempDir`? Also not sure `Fixture` as a suffix is necessary. There's no inheritance of it, it appears to just place the test as happening within the default constructed version of the class, i.e. exposing member variables and methods? You could pass an `IoData` if you wanted?

docs/src/developer/tutorial_add_new_unit_test.md
Sbozzolo: Yeah, this test does not make sense as it is (if was checking that two tests were not clashing when running concurrently, but there is no concurrency in this PR)

test/unit/fixtures.hpp
Sbozzolo: Changed

test/unit/fixtures.hpp
Sbozzolo: You are right. I don't expect having more modes, so I changed as you suggest in the comment below

test/unit/test-schema.cpp
Sbozzolo: Changed to `PerRankTempDir` and `SharedTempDir`

docs/src/developer/tutorial_add_new_unit_test.md
```diff
@@ -166,3 +166,32 @@ TEST_CASE("MyTest Vector Sum - Different Lengths", "[myvector][Parallel]")
...
+#include "fixtures.hpp"
+
+TEST_CASE_METHOD(palace::test::PerRankTempDir, "MyTest Print", "[myvector][Serial]") {
+  // temp_dir is available and will be cleaned up automatically.
+  auto file_path = temp_dir / "vector.txt";
+  Vector v;
+  v = 1;
+  {
+     std::ofstream file(file_path);
+     vec.Print(file);
```
hughcars: should be `v`

hughcars (changes requested):
One tiny fix, but LGTM.

docs/src/developer/tutorial_add_new_unit_test.md
Sbozzolo: Thank you!

hughcars (approved):

PR #639, opened 2026-02-16
Fixes from running unit tests with asan

phdum-a:
I was running palace unit tests with ASAN and found two problems in already existing tests

1. "Dimensional field output": The vector potential `a` should be Nedelec like. (@simlapointe)
2. "BuildParSumOperator": Buffer size mismatch in MPI reduction (@hughcars)

The fixes make ASAN pass (with np 1 and np 2 ranks), but I would appreciate a quick check to confirm this is right.


simlapointe (approved):
Looks good. My bad, thanks for catching this.

PR #641, opened 2026-02-17
Fix spack package.py

Sbozzolo:
The Spack package.py had two problems:
1. it added an unconditional `depends_on("nlohmann-json-schema-validator@2.4.0:")` even if older versions of Palace do not need this
2. when `+cuda`, it added `depends_on(f"mfem+umpire{cuda_variant}", when=f"{cuda_variant}")`, but MFEM was built as part of the superbuild in previous versions of Palace


cameronrutherford (approved):
Good catches!

PR #643, opened 2026-02-19
Add option to compile with address sanitizer

### LINKED ISSUE #640: Add LLVM sanitizer support
We should add the infrastructure to the cmake to make it easy to recompile with sanitizer support (at least the easy ones like ASAN and UBSAN. There are many more now that we should consider). Also there should be unit tests that run with sanitizer support regularly.

I found some minor bugs with ASAN in existing tests when debugging my own code. See #639

Sbozzolo:
And add one test. The test does not run regression tests to keep the cost down.

Closes #640


.github/workflows/build-and-test-linux.yml
```diff
@@ -431,12 +442,13 @@ jobs:
...
             -DPALACE_WITH_ARPACK=$WITH_ARPACK \
-            -DPALACE_MFEM_USE_EXCEPTIONS=ON
+            -DPALACE_MFEM_USE_EXCEPTIONS=ON \
+            -DPALACE_BUILD_WITH_SANITIZERS=$WITH_SANITIZERS
           make -j$NUM_PROC_BUILD palace-tests
 
       - name: Run unit tests
         if: needs.filter.outputs.test == 'true'
-        timeout-minutes: 5
+        timeout-minutes: 10
```
Sbozzolo, line 451: Increased because the sanitizer makes running test slower

palace/CMakeLists.txt
```diff
@@ -138,6 +139,20 @@ if(PALACE_BUILD_WITH_COVERAGE)
...
+     AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    message(FATAL_ERROR
+      "PALACE_BUILD_WITH_SANITIZERS is not compatible with GCC on macOS Apple Silicon. "
+      "GCC does not ship libasan on this platform. Use Clang instead "
+      "(e.g. -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++)."
+    )
+  endif()
+  add_library(sanitizer_flags INTERFACE)
+  target_compile_options(sanitizer_flags INTERFACE -fsanitize=address,undefined -fno-omit-frame-pointer)
+  target_link_options(sanitizer_flags INTERFACE -fsanitize=address,undefined)
```
hughcars, line 153: I had a hard time getting asan configured to work with fortran builds, have you encountered that? In particular strumpack and sundials from memory. Do you think it's worth having a tag of very small unit tests that do a "solve" on a tiny mesh for each driver, and are just smoke tests for each driver, but which are marked for asan. That would test for anything on the basic driver pathways, without needing a big tests case from the regression suite.

Sbozzolo: I did not see problems with fortran libraries, but I will double check this.

> Do you think it's worth having a tag of very small unit tests that do a "solve" on a tiny mesh for each driver, and are just smoke tests for each driver, but which are marked for asan. That would test for anything on the basic driver pathways, without needing a big tests case from the regression suite.

Yes, I think this would be a good idea. I'll open an issue about this after the PR is merged.

Sbozzolo: I ran with MUMPS and STRUMPACK and didn't see any issue. Do you remember what problems you had?

hughcars: it was flag propagation. If you're building and running them then its probably fine. Are you catching asan bugs in them? Also did you try testing my introducing a bug, array bounds or something simple, just to convince yourself its working.

Sbozzolo: Confirmed it works with fortran as well

PR #647, opened 2026-02-20
Fix finding ScaLAPACK for Intel MKL

Sbozzolo:
MKL provides scalapack as mkl_scalapack_lp64 instead of scalapack, which leads to this failure
```
  >> 40    CMake Error at CMakeLists.txt:489 (find_library):
     41      Could not find SCALAPACK_LIBRARY using the following names: scalapack
     42
     43
     44    -- Configuring incomplete, errors occurred!
```

I also noticed we don't pass SCALAPACK_ROOT from spack.

### DIFF
--- a/palace/CMakeLists.txt
+++ b/palace/CMakeLists.txt
@@ -467,9 +467,16 @@ endif()
 
 # Find ScaLAPACK early (needed for MUMPS/STRUMPACK CMake configs)
 if(PALACE_WITH_STRUMPACK OR PALACE_WITH_MUMPS)
-  find_package(scalapack QUIET)
+  if(NOT SCALAPACK_ROOT)
+    find_package(scalapack QUIET)
+  endif()
   if(NOT TARGET SCALAPACK::SCALAPACK AND NOT TARGET scalapack)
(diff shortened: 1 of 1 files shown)


cameronrutherford:
LGTM - is this not a test case that failed previously in CI?

phdum-a (commented):
I think this is ok, but have an open question. 

I don't remember the detail of cmake search order and return in `find_library`. Looking at https://github.com/Kitware/CMake/blob/master/Help/command/find_library.rst, I read this as cmake will search name by name and the set the full path for the first match is found. But the text is a bit brief here, so I may be wrong.

So if you list:

```
scalapack mkl_scalapack_lp64 mkl_scalapack_ilp64
```

What happens when the system has mutltiple of these installed? Are you sure our cmake will pick "right" one? E.g. if you install all of MKL, what determines that it will find the `mkl_scalapack_ilp64` when it needs to as compared to always returning with `mkl_scalapack_lp64`? (e.g. are the cmake search path variables different, etc).

hughcars (approved):
If you can get CI to pass, happy to go with this. It might be possible to simplify the cmake patches for these, to make it a little simpler, but thats not necessarily in scope for this.

PR #653, opened 2026-02-27
Fix small issues in preparation for C++20

phdum-a:
GCC 16 will [set C++20 at the default](https://gcc.gnu.org/projects/cxx-status.html#cxx20) and is expected to be [released soon](https://gcc.gnu.org/develop.html#timeline). Currently compiling main with C++20 causes compiler failures, due to small issues (changes in how fmt deals with typing, ADL ambiguity). This PR fixes these issues in a way that is compatible with both C++17 and C++20, so that people can compile with both. This PR does not actually enable C++20 or any C++20 simplifications.

### DIFF
--- a/.github/workflows/spack.yml
+++ b/.github/workflows/spack.yml
@@ -77,7 +77,19 @@ jobs:
           toolchain: gcc
           variant: "+openmp+int64~shared"
           run-regression-tests: 'false'
- 
+
+  build-x64-gcc-cxx20:
+    needs: filter
+    if: needs.filter.outputs.test == 'true'
+    runs-on: [self-hosted, x64]
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/palace-ci
+        with:
+          toolchain: gcc
+          variant: " cxxstd=20"
+          run-regression-tests: 'false'
+
   build-test-x64-llvm:
     needs: filter
     if: needs.filter.outputs.test == 'true'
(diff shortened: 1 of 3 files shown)


palace/linalg/rap.cpp
```diff
@@ -953,8 +953,8 @@ BuildParSumOperator(ScalarType (&&coeff_in)[N], const OperType *(&&ops_in)[N],
   std::transform(ops_in, ops_in + N, par_ops.begin(),
                  [](const OperType *op) { return dynamic_cast<const ParOperType *>(op); });
 
-  return BuildParSumOperator(to_array<ScalarType>(std::move(coeff_in)), std::move(par_ops),
-                             set_essential);
+  return BuildParSumOperator(palace::to_array<ScalarType>(std::move(coeff_in)),
```
hughcars, line 956: Lines 937-938 also need `palace::` — those use `to_array<std::complex<double>>` which triggers ADL into `std` (where `std::to_array` lives in C++20).

palace/utils/prettyprint.hpp
```diff
@@ -60,10 +62,10 @@ inline std::size_t PrePrint(MPI_Comm comm, std::size_t w, std::size_t wv, std::s
 // integral types.
 template <template <typename...> class Container, typename T, typename... U>
 inline void PrettyPrint(const Container<T, U...> &data, T scale,
-                        const std::string &prefix = "", MPI_Comm comm = MPI_COMM_WORLD)
+                        const std::string_view prefix = "", MPI_Comm comm = MPI_COMM_WORLD)
 {
   std::size_t w = 0, lead = prefix.length();
```
hughcars, line 67: `fmt::runtime()` is correct here — `Mpi::Print` takes `fmt::format_string<T...>` which requires compile-time checking.

PR #661, opened 2026-03-09
Fix driven solver `SurfaceCurrent` with no pre-existing excitation 

### LINKED ISSUE #616: Help with using SurfaceCurrent as excitation source in Driven simulations
I am setting up the simulation of Magnetic fields generated by flow of RF-current through a planar coil. Working with the example for the dipole antenna, I managed to get the simulation running for lumped port excitation. However, to compare between coils, I would like to use a unit current to be able to better compare different coil designs. From what I've found, surface current excitation seems the way to go for this. However, I am struggling with the simulation setup for this, as  replacing the Lumped Port BC on the strip bridging the terminals of the coil results in an error during simulation setup: Verification failed: (!port_excitation_helper.Empty()) is false:
 --> Driven problems must specify at least one excitation!
How do I need to setup the BCs for surface current excitation?
Sorry if this is a dumb question - I'm a chemist fairly inexperienced in electronics :)

phdum-a:
Closes #616.

`SurfaceCurrent` can be used in the driven solver, but don't have any "Excitation" index. Instead they are added to all excitations. There is a bug where if there is no already existing excitation, this fails.

More generally, the `Excitation` interface was only designed for lumped or wave ports even through there are other times of excitations (like SurfaceCurrent, Dipole, etc) and this should be reworked to be more general and user friendly.


palace/models/portexcitations.cpp
```diff
@@ -103,14 +103,17 @@ PortExcitations::PortExcitations(const LumpedPortOperator &lumped_port_op,
...
   for (const auto &[idx, port] : surf_j_op)
   {
     current_port_idx.push_back(idx);
   }
   if (!current_port_idx.empty())
   {
+    if (excitations.empty())
+    {
+      excitations.try_emplace(1, SingleExcitationSpec{});
+    }
```
hughcars, line 116: So if there are multiple surface current ops, then this will still have all active, as the others will be following the existing route, but the first will be the excited one?

phdum-a: Not sure I understand your comment.

- `LumpedPort`s and `WavePort`s have the notion of excitation. You can turn them the drive on and off. Since the multi-excitation PR you can group the turning off into excitation numbers.
- `SurfaceCurrent` and `CurrentDipole` don't have the notion of "Excitation" in their configuration file. They are just alway on for all excitations. If you have multiple SurfaceCurrents, they will all be on at the same time.

The reason for the difference is probably that the Lumped & Wave ports modify the dissipation matrix of the system. So even when they are turned off they are important to the system. The SurfaceCurrent and Dipole and just excitation without dissipation so I guess you just assume that if you put them in the config, they should be turned always. Also, if you turn them off, you have to be happy with the default boundary condition, since there is no surface impedance specified.

The bug is that if there is no LumpedPort / WavePort already there you need a "dummy" excitation 1 to put your SurfaceCurrents into. (Actually @laylagi had this done correctly for the dipole case as well since those are also added to all excitations).

There was an actually a "TODO" here to allow adding SurfaceCurrent to selective excitations. This is an interface choice.  If you really want to give the user maximal flexibility, you might want a "ExcitationSpec" in the config that allows users to specify excitations over LumpedPort/WavePort/SurfaceCurrent/Dipole, with coefficient pre-factors (like phases or signs). And potentially user derived functions for the transient. But that interface needs to be thought out more carefully...

hughcars: You're over reading my comment. I was just wondering what happens if more than one surface current was set at once, without any lumped or waveport. I believe this will set 1 of them to "excited", and then the others will just be always on like they are usually, so effectively you get all surface and dipoles at once, not a selector.

phdum-a: Ah, ok. Yes, All surface current and dipole operators will always be "excited" and "added to the RHS" of the system, no selection possible.

phdum-a:
So I want to close this PR as this is an actual bug fix and it is waiting on documentation improvements that will take a while.

>  I also don't see references to dissipation or even the different normalizations in the boundaries page.

It's there, if you know where to look and how to read the Palace docs....

From https://awslabs.github.io/palace/dev/guide/boundaries/#Lumped-and-wave-port-excitation

> A lumped port applies a similar boundary condition to a [surface impedance](https://awslabs.github.io/palace/dev/guide/boundaries/#Impedance-boundary) boundary, but takes on a special meaning for each simulation type.

> For each port, the excitation is normalized to have unit incident power over the port boundary surface.

From https://awslabs.github.io/palace/dev/guide/boundaries/#Surface-current-excitation:

> This option prescribes a unit source surface current excitation on the given boundary in order to excite the model. It does does not prescribe any boundary condition to the model and only affects the source term on the right hand side.

I know that this is hard to find and interpret. Making the documentation more didactic is work and I have not found a place to naturally put my comment from #616 with more rewrites. For now:

- I've added a note in #696 that we should document surfacecurrent v lumpedport differences in the reference doc. 
- In the tutorial on the driven solver #702, I have a small discussion mentioning conditioning effects.

hughcars (approved):

hughcars:
Lookahead failures are unrelated, tracked in https://github.com/awslabs/palace/issues/700

PR #673, opened 2026-03-12
Move files instead of copying them in AMR

Sbozzolo:
Palace currently copies files in the iteration folders. This can be very expensive on shared filesystems or for large Paraview files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ The format of this changelog is based on
 
 #### Bug Fixes
 
+  - Improved IO performance for simulations with Adaptive Mesh Refinement. Now,
+    files from previous iterations are moved instead of being copied. The latest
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ The format of this changelog is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/).
 
+## In progress
+
+#### Bug Fixes
+
+  - Improved IO performance for simulations with Adaptive Mesh Refinement. [PR
+    673](https://github.com/awslabs/palace/pull/673)
```
hughcars: Maybe mention that the top postpro dirs will remain the most recent results, but are now symlinks. That might be surprising for users in some cases.

PR #684, opened 2026-03-23
Use ProjectBdrCoefficientTangent for lumped port excitation

hughcars:
The boundary mass matrix CG solve was a workaround for an MFEM bug in `ProjectBdrCoefficientTangent` (face DOF orientations wrong in parallel for ND order >= 2). An MFEM patch fixes this ([mfem/mfem#5280](https://github.com/mfem/mfem/pull/5280)), so we can use direct interpolation instead.


extern/patch/mfem/mfem_pr5280.diff
Sbozzolo, line 1: I suggest we start adding the pr number to the name of the file (as I am doing with the spack patches). Otherwise, it is quite hard to find the relevant PR and check if it was merged.

hughcars: They were historically named after the full branch name (as we had some which weren't going to PRs), but numbering I think is neater.

hughcars: Addressed in https://github.com/awslabs/palace/pull/684/changes/868a322b6cbafc01ded281ecf0113b546157ee2b opted to not change the two existing, as one comes from a branch never opened as a PR (and which will be superseded soon by upgrades to the mfem gmsh reader).

PR #688, opened 2026-03-26
Use PETSc version of METIS in Spack

Sbozzolo:
I found that the PETSc fork of METIS includes a few important patches that improve mesh partitioning. In turn, this also improves performance.

METIS and ParMETIS are not well maintained and there are larger issue with the packages. See

https://github.com/spack/spack-packages/pull/4000

Matches what we do in the superbuild.


examples/adapter/mesh/mesh.jl
```diff
@@ -221,7 +221,7 @@ function generate_adapter_mesh(;
     # Generate mesh
     gmsh.option.setNumber(
         "Mesh.MeshSizeMax",
-        minimum([ellipse_a, ellipse_b, rect_a/2, rect_b/2, adapter_length])
+        0.75 * minimum([ellipse_a, ellipse_b, rect_a/2, rect_b/2, adapter_length])
```
Sbozzolo, line 224: I increased resolution to improve robustness of the test for high CPU count

(cc @simlapointe)

hughcars (approved):

PR #693, opened 2026-03-31
Add Floquet port boundary condition

### LINKED ISSUE #563: Support for Floquet/Bloch Ports in Periodic Geometries
I'd like to propose an enhancement related to the simulation of periodic structures with oblique plane-wave excitation.

Palace already supports periodic (Bloch) boundary conditions, but currently there is no "Floquet Port" implementation for modal excitation and modal analysis on periodic surfaces.

In commercial FEM/FDTD solvers (CST, HFSS, COMSOL), a Floquet Port provides two key capabilities:
1. It solves the transverse eigenvalue problem on the periodic face (Floquet eigenmodes, with know. eigenvalues).
2. It allows the user to excite a specific Floquet mode (typically the fundamental one) to model oblique plane-wave illumination and to compute meaningful S-parameters for propagating and evanescent diffraction orders.

This feature is essential for:
- periodic arrays,
- metasurfaces and metagratings,
- RCWA-like structures,
- periodic lattices under arbitrary incidence,
- first-Brillouin-zone analysis.

Without a Floquet port, users can impose Bloch phase shifts on the lateral boundaries, but cannot properly define modal excitation nor extract modal transmission/reflection coefficients.

**Proposal**
I suggest adding a "FloquetPort" or "PeriodicPort" module that would:
- compute Floquet modes (propagating + evanescent) for a periodic surface,
- associate them with user-specified Bloch vectors,
- enable excitation of a selected mode,
- return S-parameters for different diffraction orders.

This feature would significantly improve Palace's capabilities for periodic EM problems.

Thanks for your work and consideration!

### LINKED ISSUE #564: Floquet vector should be normalized with respect to frequency when using periodic boundary conditions
Description:

When using periodic boundary conditions in palace, the Bloch/Floquet vector appears to be treated as a fixed value throughout a frequency sweep. However, for physically correct periodic (Floquet) simulations, the Bloch vector depends linearly on frequency.
If the frequency changes during a sweep but the Bloch vector in the JSON configuration remains constant, the simulation becomes inconsistent and produces incorrect results.
This prevents performing valid frequency sweeps under periodic boundary conditions unless the user manually updates the Bloch vector for each frequency—something that the current input format does not support.

Suggested improvement:

It would be better if the Bloch/Floquet vector specified in the JSON:
- were normalized to the frequency, or
- alternatively, could be expressed in terms of direction only, letting palace scale it internally using the vacuum wavenumber.

This would ensure correct behaviour when performing frequency sweeps and avoid misleading results.

Thanks for your work on the project!

simlapointe:
This PR adds Floquet port boundary conditions for frequency-domain driven simulations on periodic structures, with support for oblique incidence, multi-order diffraction, TE/TM/circular polarization, and adaptive frequency sweeps, which closes #563. Also adds an option to have the Floquet wave vector vary linearly with frequency in driven simulations, which closes #564.


palace/models/portexcitations.cpp
```diff
@@ -18,34 +19,38 @@ namespace palace
...
-  auto out = fmt::appender{buf};
+  auto to = [&buf](auto f, auto &&...a)  // mini-lambda for cleaner code
+  { fmt::format_to(std::back_inserter(buf), f, std::forward<decltype(a)>(a)...); };
 
   int i = 1;
   for (const auto &[idx, ex] : excitations)
   {
-    fmt::format_to(out, "Excitation{} with index {:d} has contributions from:\n",
-                   (Size() > 1) ? fmt::format(" {:d}/{:d}", i, Size()) : "", idx);
+    to("Excitation{} with index {:d} has contributions from:\n",
```
Sbozzolo: This was changed to `fmt::format_to` in https://github.com/awslabs/palace/pull/653 for C++20

docs/src/assets/examples/dielectric_grating_reflection.png
Sbozzolo, line 1: Maybe you can run generate the plots automatically as part of producing the documentation, so that we don't have to worry to keep them updated.

palace/utils/prettyprint.hpp
```diff
@@ -62,10 +60,10 @@ inline std::size_t PrePrint(MPI_Comm comm, std::size_t w, std::size_t wv, std::s
 // integral types.
 template <template <typename...> class Container, typename T, typename... U>
 inline void PrettyPrint(const Container<T, U...> &data, T scale,
-                        const std::string_view prefix = "", MPI_Comm comm = MPI_COMM_WORLD)
+                        const std::string &prefix = "", MPI_Comm comm = MPI_COMM_WORLD)
```
Sbozzolo: This was also changed in the PR related to C++20

test/examples/ref/dielectric_grating/uniform/port-floquet-S.csv
```diff
@@ -0,0 +1,7 @@
...][1]) (deg.),     |S[P1(1;0)TM][1]| (dB),arg(S[P1(1;0)TM][1]) (deg.),     |S[P1(1;1)TE][1]| (dB),arg(S[P1(1;1)TE][1]) (deg.),     |S[P1(1;1)TM][1]| (dB),arg(S[P1(1;1)TM][1]) (deg.),   |S[P2(-1;-1)TE][1]| (dB),arg(S[P2(-1;-1)TE][1]) (deg.),   |S[P2(-1;-1)TM][1]| (dB),arg(S[P2(-1;-1)TM][1]) (deg.),    |S[P2(-1;0)TE][1]| (dB),arg(S[P2(-1;0)TE][1]) (deg.),    |S[P2(-1;0)TM][1]| (dB),arg(S[P2(-1;0)TM][1]) (deg.),    |S[P2(-1;1)TE][1]| (dB),arg(S[P2(-1;1)TE][1]) (deg.),    |S[P2(-1;1)TM][1]| (dB),arg(S[P2(-1;1)TM][1]) (deg.),    |S[P2(0;-1)TE][1]| (dB),arg(S[P2(0;-1)TE][1]) (deg.),    |S[P2(0;-1)TM][1]| (dB),arg(S[P2(0;-1)TM][1]) (deg.),     |S[P2(0;0)TE][1]| (dB),arg(S[P2(0;0)TE][1]) (deg.),     |S[P2(0;0)TM][1]| (dB),arg(S[P2(0;0)TM][1]) (deg.),     |S[P2(0;1)TE][1]| (dB),arg(S[P2(0;1)TE][1]) (deg.),     |S[P2(0;1)TM][1]| (dB),arg(S[P2(0;1)TM][1]) (deg.),    |S[P2(1;-1)TE][1]| (dB),arg(S[P2(1;-1)TE][1]) (deg.),    |S[P2(1;-1)TM][1]| (dB),arg(S[P2(1;-1)TM][1]) (deg.),     |S[P2(1;0)TE][1]| (dB),arg(S[P2(1;0)TE][1]) (deg.),     |S[P2(1;0)TM][1]| (dB),arg(S[P2(1;0)TM][1]) (deg.),     |S[P2(1;1)TE][1]| (dB),arg(S[P2(1;1)TE][1]) (deg.),     |S[P2(1;1)TM][1]| (dB),arg(S[P2(1;1)TM][1]) (deg.)
```
Sbozzolo, line 1: This file looks very weird with all these NaNs. Maybe we should explain somewhere that that this is expected/fine?

test/unit/test-floquetport.cpp
Sbozzolo, line 1: Can we have more unit tests and don't rely as much on the regression test?

Sbozzolo (commented):
I have a very brief look at the PR and left some comments

docs/src/config/boundaries.md
```diff
@@ -418,6 +431,63 @@ with
...
+  - `"TM"` :  Transverse magnetic (p-polarization). Electric field in the plane of
+    incidence.
+  - `"RHC"` :  Right-hand circular polarization. Equal superposition of TE and TM with
+    90° phase shift: ``\mathbf{E} = (\hat{e}_\text{TE} + j\hat{e}_\text{TM})/\sqrt{2}``.
+  - `"LHC"` :  Left-hand circular polarization. Equal superposition of TE and TM with
+    -90° phase shift: ``\mathbf{E} = (\hat{e}_\text{TE} - j\hat{e}_\text{TM})/\sqrt{2}``.
+
+`"MaxOrder" [-1]` :  Maximum diffraction order index to include. With `MaxOrder = M`, all
+orders ``(m, n)`` with ``|m| \leq M`` and ``|n| \leq M`` are included, each with both TE
+and TM polarizations. A value of `-1` (default) enables automatic selection. A value of `0`
```
hughcars, line 550: Is there value in possibly allowing restriction to `|m + n| < M` i.e. the L1 constraint over the Linf?

docs/src/examples/dielectric_grating.md
```diff
@@ -0,0 +1,61 @@
...
+    Adaptive fast frequency sweep from 2 to 12 GHz with initial step
+    ``\Delta f = 0.05\text{ GHz}`` and tolerance ``10^{-3}``.
+
+The S-parameter magnitudes from the uniform (markers) and adaptive (lines) sweeps are shown
+below:
+
+```@raw html
+<br/><p align="center">
+  <img src="../../assets/examples/dielectric_grating_reflection.png" width="70%" />
+  <img src="../../assets/examples/dielectric_grating_transmission.png" width="70%" />
```
hughcars, line 59: At least on my browser the font in these images is much smaller than the surrounding text making it hard to read. If you're feeling fancy you could have dashed black lines marking the critical frequencies you mention in the text below, as the axis is pretty broad

Image

docs/src/guide/boundaries.md
```diff
@@ -131,6 +131,33 @@ for implementation details.
...
+    and extracts power-normalized S-parameters for each propagating order.
+
+    Floquet ports require periodic boundary conditions to be configured under
+    [`config["Boundaries"]["Periodic"]`](../config/boundaries.md#boundaries%5B%22Periodic%22%5D).
+    The `"FloquetWaveVector"` in the periodic configuration specifies the tangential
+    component of the incident wave vector, which determines the angle of incidence. For
+    normal incidence, set the wave vector to zero. For frequency sweeps at a fixed angle
+    of incidence, set `"FloquetReferenceFrequency"` to the frequency (in GHz) at which the
+    wave vector is defined. The wave vector then scales linearly with frequency:
+    k_F(f) = FloquetWaveVector × (f / FloquetReferenceFrequency), maintaining a constant
```
hughcars: Math formatting, this is plain text rn.

docs/src/examples/dielectric_grating.md
```diff
@@ -0,0 +1,61 @@
+```@raw html
```
hughcars, line 1: Can you add an image of the geometry/fields from paraview? Would make conceptualizing simpler.

docs/src/config/boundaries.md
```diff
@@ -567,7 +638,15 @@ to the receiver attribute in mesh units. If neither `"Translation"` or `"AffineT
...
-periodic boundaries in the X/Y/Z directions in radians per mesh unit.
+periodic boundaries in the X/Y/Z directions in radians per mesh unit. When used with
+`"FloquetReferenceFrequency"`, this specifies the wave vector at the reference frequency.
+
+`"FloquetReferenceFrequency" [None]` :  Optional frequency in GHz at which the
+`"FloquetWaveVector"` is defined. When specified, the Bloch wave vector scales linearly with
+frequency during a driven simulation frequency sweep: k_F(f) = FloquetWaveVector × (f /
+FloquetReferenceFrequency). This is the physically correct behavior for oblique plane wave
+incidence at a fixed angle, where k_F = (2πf/c) sin(θ). When not specified (default), the wave
+vector is held constant across all frequencies. Only supported for driven simulations.
```
hughcars: I think we could consider the alternative angle based approach for an alternative api in future, but the issue there is needing to define a relative frame/axis for the angles which i really don't like the idea of. Is there a process for mapping between that representation and this? Maybe you could expand on that in the documentation page and then link to that from here.

examples/dielectric_grating/dielectric_grating_adaptive.json
```diff
@@ -0,0 +1,82 @@
...
+      "FreqStep": 0.05,
+      "AdaptiveTol": 1.0e-3
+    },
+    "Linear":
+    {
+      "Type": "Default",
+      "KSPType": "GMRES",
+      "Tol": 1.0e-8,
+        "MaxIts": 200,
+        "ComplexCoarseSolve": true
```
hughcars: Weird indent

palace/linalg/operator.hpp
```diff
@@ -112,6 +112,51 @@ class ComplexWrapperOperator : public ComplexOperator
...
 };
 
+// Sum of two ComplexOperators: (A + B)*x = A*x + B*x. Supports owning, non-owning, and
+// mixed modes. In non-owning mode, the caller must ensure the referenced operators outlive
+// this object.
+class SumComplexOperator : public ComplexOperator
+{
+private:
+  std::unique_ptr<ComplexOperator> owned_A, owned_B;
+  const ComplexOperator *op_A, *op_B;
```
hughcars: This is a special case of the sequence approach taken below. It would be nicest if this could be syntactically equivalent to that. An even better would be if you could achieve it with a templated version, where the SumOperator and SumComplexOperator are then just specified as 

```
template <typename OperatorT>
class BaseSumOperator : public OperatorT { .... }; 

using SumOperator = BaseSumOperator<Operator>;
using SumComplexOperator = BaseSumOperator<ComplexOperator>;
```

palace/models/floquetportoperator.hpp
```diff
@@ -0,0 +1,277 @@
...
+class MaterialPropertyCoefficient;
+
+namespace config
+{
+struct FloquetPortData;
+struct PeriodicBoundaryData;
+}  // namespace config
+
+// Flags controlling which subsystems use a given Floquet mode.
+enum class FloquetModeUse : std::uint8_t
```
hughcars: Match the `: char` from the enum above.

palace/models/floquetportoperator.hpp
```diff
@@ -0,0 +1,277 @@
...
+{
+  Output = 1,  // S-parameter CSV output (user-requested ±MaxOrder)
+  Dtn = 2,     // DtN boundary correction (BZ-centered range)
+  Both = 3     // Output | Dtn
+};
+inline bool HasFlag(FloquetModeUse u, FloquetModeUse flag)
+{
+  return (static_cast<uint8_t>(u) & static_cast<uint8_t>(flag)) != 0;
+}
```
hughcars, line 52: Is this really bit twiddling to avoid having a 3 clause switch or 3 branch if/else?

palace/models/floquetportoperator.cpp
```diff
@@ -0,0 +1,1050 @@
...
+      if (pol_diff > 1e-14)
+      {
+        mode.e_pol = new_pol;
+        need_reassemble = true;
+      }
+    }
+  }
+
+  // Re-assemble mode vectors if polarization changed.
+  if (need_reassemble && nd_fespace_ptr)
```
hughcars: **Avoidable per-frequency FEM projection re-assembly.** When `FloquetReferenceFrequency` is active, the polarization vectors change with frequency, but the Fourier kernel for a given diffraction order does not. Since

```cpp
v = ∫_Γ (n × n × N_j) · e_pol exp(-i B_mn · r) dS
```

is linear in `e_pol`, we can assemble the Cartesian component projections once per order and rebuild TE/TM vectors with cheap vector AXPYs whenever the polarization changes:

```cpp
std::array<ComplexVector, 3> v_comp;  // x/y/z component projections
std::array<ComplexVector, 2> v;       // TE/TM combinations
```

Then `Initialize(omega)` only recomputes `gamma_sq`, updates the TE/TM polarization vectors, and combines the cached components; it should not re-enter FEM `LinearForm` assembly from the frequency loop or the online PROM path. I pushed this shape in `hughcars/floquet-port-clean`.

palace/models/romoperator.cpp
```diff
@@ -98,14 +98,24 @@ inline void ProjectMatInternal(MPI_Comm comm, const std::vector<Vector> &V,
   }
   Mpi::GlobalSum((n - n0) * n, Ar.data() + n0 * n, comm);
 
-  // Fill lower block of Vᴴ A V = [ ____________  |  ]
-  //                              [ vjᴴ A V[1:n0] |  ] .
+  // Compute the lower-left block directly: rows [n0, n), columns [0, n0).
+  // No symmetry assumption — works for Hermitian, anti-symmetric, or general operators.
```
hughcars, line 114: This changes the PROM projection from using the existing complex-symmetric shortcut (`Ar(i,j) = Ar(j,i)`) to explicitly computing the lower-left block for all operators. That is needed for the Floquet/non-symmetric path, but the base non-Floquet frequency-domain operators are still complex symmetric (not Hermitian): `K + iω C - ω² M`, with K/C/M assembled from symmetric bilinear forms.

Could we preserve the old symmetric fill for operators known to be complex-symmetric, and only use the general lower-left computation when the Floquet frequency-scaling contribution can break transpose symmetry? A short comment explaining that distinction would also help. I pushed a candidate version in `hughcars/floquet-port-clean`.

palace/models/floquetportoperator.cpp
```diff
@@ -0,0 +1,1050 @@
...
+    }
+  }
+  return count;
+}
+
+std::complex<double>
+FloquetPortData::ComputeDtNCorrectionCoeff(const FloquetMode &mode) const
+{
+  // Compute g_correction = g_full - g_uniform for a single mode.
+  // g_uniform uses the TE (0,0) propagation constant as the Robin reference.
```
hughcars, line 706: `gamma0` is a per-port, per-frequency quantity, but it is looked up by scanning the modes here, again in `ComputeIncidentNormalization`, and again in `FloquetPortOperator::AddExtraSystemBdrCoefficients`. Since `Initialize(omega)` already updates every `gamma_sq`, it seems cleaner to cache the specular `(0,0)` propagation constant there and expose it with a small accessor.

That removes the repeated nested scans and makes ownership of the frequency-dependent scalar explicit. I pushed this in `hughcars/floquet-port-clean` as `FloquetPortData::gamma0` / `GetGamma0()`.

palace/models/floquetportoperator.hpp
```diff
@@ -0,0 +1,277 @@
...
+
+// Represents a single diffraction order (m, n) with a specific polarization (TE/TM).
+struct FloquetMode
```
hughcars: Right now TE and TM are represented as separate `FloquetMode` entries, but most of the data is per diffraction order rather than per polarization: `m`, `n`, `B_mn`, `gamma_sq`, and `use` are identical for the TE/TM pair. Only the polarization vector and projection vector differ.

Could this be represented as a per-order object instead, with TE/TM stored directly on the order? For example:

```cpp
struct FloquetOrder
{
  int m, n;
  FloquetModeUse use;
  mfem::Vector B_mn;
  double gamma_sq = 0.0;
  std::array<mfem::Vector, 2> e_pol;       // 0 = TE, 1 = TM
  std::array<ComplexVector, 2> v;          // 0 = TE, 1 = TM
  std::array<ComplexVector, 3> v_comp;     // x/y/z component projections
};
```

Then call sites can iterate by order and polarization without duplicating the order-level state:

```cpp
for (const auto &order : orders)
{
  for (bool is_te : {true, false})
  {
    const auto p = is_te ? 0 : 1;
    auto g = ComputeDtNCorrectionCoeff(order, is_te);
    const auto &v = order.v[p];
    ...
  }
}
```

This avoids recomputing the same `gamma_sq` twice per order, makes `(0,0)` lookups more direct, and separates genuinely per-order data from per-polarization data. I pushed this order-centered layout in `hughcars/floquet-port-clean`.

palace/models/floquetportoperator.hpp
```diff
@@ -0,0 +1,277 @@
...
+  // Add excitation vector contribution directly to the RHS.
+  // The incident field is normalized to inject 1 W (nondimensional) through the port,
+  // consistent with lumped and wave port conventions.
+  bool AddExcitationVector(double omega, ComplexVector &RHS) const;
+
+  // Returns 1 W: the port mode is normalized such that ∫ (E_inc × H_inc⋆) · n̂ dS = 1.
+  double GetExcitationPower() const { return HasExcitation() ? 1.0 : 0.0; }
+
+  // Material properties at the port (nondimensional, from adjacent volume element).
+  double GetMuEpsPort() const { return mu_eps_port; }
```
hughcars: `GetMuEpsPort()`, `GetMuRPort()`, `GetPortArea()`, and `GetExcitationPower()` have no callers anywhere (including tests — Floquet S-params bypass `PrintPortS`, which is where lumped/wave `GetExcitationPower` is used). Drop if not needed? I removed them in `hughcars/floquet-port-clean`.

palace/models/floquetportoperator.hpp
```diff
@@ -0,0 +1,277 @@
...
+                      const MaterialOperator &mat_op,
+                      mfem::ParFiniteElementSpace &nd_fespace);
+
+  const FloquetPortData &GetPort(int idx) const { return ports.at(idx); }
+  auto begin() const { return ports.begin(); }
+  auto end() const { return ports.end(); }
+  auto Size() const { return ports.size(); }
+  bool Empty() const { return ports.empty(); }
+
+  void SetSuppressOutput(bool suppress) { suppress_output = suppress; }
```
hughcars: `suppress_output` is never set and never read in `floquetportoperator.cpp` — was this copied from `WavePortOperator` where it is needed? Dead code; remove the member and setter. I removed it in `hughcars/floquet-port-clean`.

palace/models/floquetportoperator.cpp
```diff
@@ -0,0 +1,1050 @@
...
+    // units). Nondimensionalize by dividing by Lc/L0.
+    for (int i = 0; i < 3; i++)
+    {
+      a1(i) = periodic.boundary_pairs[0].affine_transform[i * 4 + 3] / mesh_scale;
+      a2(i) = periodic.boundary_pairs[1].affine_transform[i * 4 + 3] / mesh_scale;
+    }
+  }
+  // Fall back to port geometry if translations are missing (e.g., auto-detected periodicity
+  // from Gmsh setPeriodic, or no BoundaryPairs at all).
+  if (a1.Norml2() < 1e-12 || a2.Norml2() < 1e-12)
```
hughcars, line 142: For Floquet ports, it looks like we require exactly two transverse periodic directions. The config can still read `BoundaryPairs` as a variable-length array, but this constructor should validate that the length is exactly 2 when a Floquet port is configured.

Right now the guard only checks `!boundary_pairs.empty()`, while the code below only uses `boundary_pairs[0]` and `boundary_pairs[1]`. With exactly one pair, `a1`/`a2` remain unset before the `Norml2()` fallback check, so this can read uninitialized vector entries. With more than two pairs, extras appear to be silently ignored. Could we verify `boundary_pairs.size() == 2` here? I pushed that constructor-side validation in `hughcars/floquet-port-clean`.

palace/models/floquetportoperator.cpp
```diff
@@ -0,0 +1,1050 @@
...
+    {
+      Mpi::Print(" Floquet port: BZ wrapping at ref freq (shift = {:d}*b1 + {:d}*b2)\n",
+                 bz_m, bz_n);
+    }
+  }
+
+  // Compute port normal and area using geodata utilities.
+  {
+    auto &mesh = *nd_fespace.GetParMesh();
+    int bdr_attr_max = mesh.bdr_attributes.Size() ? mesh.bdr_attributes.Max() : 0;
```
hughcars: This boundary marker construction appears several times, with two subtly different ways of computing `bdr_attr_max` (`mesh.bdr_attributes.Max()` here vs. scanning `GetNBE()` below). Since the scan version is needed to include internally-added boundary elements, could we factor this into a shared mesh/geodata helper instead of keeping it Floquet-specific?

The version I pushed in `hughcars/floquet-port-clean` adds a generic helper like:

```cpp
mesh::BdrAttrToMarker(mesh, attr_list, /*include_internal_bdr=*/true)
```

and reuses it from the Floquet port code.

palace/models/postoperator.cpp
```diff
@@ -1034,19 +1039,79 @@ void PostOperator<solver_t>::MeasureWavePorts() const
...
+    {
+      return;
+    }
+    // Determine if the excitation uses circular polarization (for output basis).
+    bool circular_output = false;
+    for (const auto &[idx, port] : fem_op->GetFloquetPortOp())
+    {
+      if (port.excitation == measurement_cache.ex_idx)
+      {
+        circular_output = (std::abs(port.GetIncidentAlpha(true)) > 1e-14 &&
```
hughcars: This circular-polarization detection (`|α_TE| > 1e-14 && |α_TM| > 1e-14`) is duplicated in `PostOperatorCSV::InitializeFloquetPortS`. Could this be a method on `FloquetPortData` (for example `HasCircularExcitation()`) so the logic is consolidated? I pushed that helper in `hughcars/floquet-port-clean`.

palace/models/postoperator.cpp
```diff
@@ -1034,19 +1039,79 @@ void PostOperator<solver_t>::MeasureWavePorts() const
...
 
     // Don't measure S-Matrix unless there is only one excitation per port. Also, we current
     // don't support mixing wave and lumped ports, because we need to fix consistent
     // conventions / de-embedding.
+    bool has_lumped = fem_op->GetLumpedPortOp().Size() > 0;
+    bool has_wave = fem_op->GetWavePortOp().Size() > 0;
+    bool has_floquet = !fem_op->GetFloquetPortOp().Empty();
     if (!fem_op->GetPortExcitations().IsMultipleSimple() ||
-        !((fem_op->GetLumpedPortOp().Size() > 0) xor (fem_op->GetWavePortOp().Size() > 0)))
+        !(has_lumped xor has_wave xor has_floquet))
```
hughcars: This 3-way `xor` checks parity, not “exactly one”. So the all-three case (`has_lumped && has_wave && has_floquet`) also passes this guard.

```suggestion
        (static_cast<int>(has_lumped) + static_cast<int>(has_wave) +
         static_cast<int>(has_floquet) != 1))
```

palace/models/floquetportoperator.cpp
```diff
@@ -0,0 +1,1050 @@
...
+  norm.lambda_tm_00 = omega * omega * mu_eps_port / norm.gamma_00;
+  norm.lambda_eff = std::norm(inc_alpha_te) * norm.lambda_te_00 +
+                    std::norm(inc_alpha_tm) * norm.lambda_tm_00;
+  double p_unit = norm.lambda_eff * port_area / (2.0 * omega * mu_r_port);
+  norm.c_inc = 1.0 / std::sqrt(p_unit);
+  return norm;
+}
+
+std::map<std::tuple<int, int, bool>, std::complex<double>>
+FloquetPortData::GetAllSParameters(const GridFunction &E, bool subtract_incident) const
```
hughcars: This silently uses the cached `omega0` from whatever `Initialize()` call happened last — the postprocessing path passes no frequency and just trusts driver call ordering. A future caller measuring at a different frequency gets wrong numbers with no error. Suggest passing `omega` explicitly (or at least `MFEM_ASSERT`-ing against the cache).

palace/utils/prettyprint.hpp
hughcars: This change seems unrelated to the Floquet port work. Could we leave it out of this PR unless it is needed for one of the changes here? I removed it in `hughcars/floquet-port-clean`.

hughcars (commented):
A few suggested changes, on `hughcars/floquet-port-clean` for you to cherry-pick as you want, but generally LGTM. The reference frequency UI I think works, and i prefer it to the idea of angles as that requires defining a frame. If users decide they want angles, we can possibly add another front end to convert into this representation later, but lets wait until that's needed.

palace/models/portexcitations.cpp
simlapointe: `to` is just a lambda calling `fmt:format_to` (a few lines above) so I'd think it's ok?

docs/src/config/boundaries.md
simlapointe: Not sure there is value in doing that right now. A more desirable feature we could add in the future is to let users specify exactly which `(m, n)` pairs they want instead of looping over all combinations up to MaxOrder.

docs/src/examples/dielectric_grating.md
simlapointe: Updated the plots, take a look once the docs haven been rebuilt

docs/src/guide/boundaries.md
simlapointe: Fixed

docs/src/examples/dielectric_grating.md
simlapointe: Added an image of the geometry

docs/src/config/boundaries.md
simlapointe: Added some text in boundaries.md to help users map between angles and relative freq

palace/linalg/operator.hpp
simlapointe: Refactored as suggested

palace/models/floquetportoperator.hpp
simlapointe: Moved to if/else

test/examples/ref/dielectric_grating/uniform/port-floquet-S.csv
simlapointe: Added a mention in the docs

simlapointe:
Thanks for the review @hughcars. Appreciate the suggested changes, they were spot on. Took most/all of them and addressed the other issues. Rebased just now.

palace/utils/configfile.cpp
```diff
@@ -861,6 +877,8 @@ BoundaryData::BoundaryData(const json &boundaries)
   terminal = ParseOptionalMap<TerminalData>(boundaries, "Terminal", "\"Terminal\"");
   periodic = ParseOptional<PeriodicBoundaryData>(boundaries, "Periodic");
   waveport = ParseOptionalMap<WavePortData>(boundaries, "WavePort", "\"WavePort\"");
+  floquetport =
```
hughcars, line 880: Now that `FloquetPort` has been added to `BoundaryData`, should it also be included in `Validate(const BoundaryData&)` below? That validation currently checks duplicate `Index` values across lumped/wave/surface-current/terminal entries, but not Floquet ports, even though the docs now say Floquet indices must be unique with the other boundary condition types.

The same function's excitation-index ambiguity check also only builds `excitation_map` from lumped and wave ports. A Floquet port with a numeric `Excitation` that does not match its `Index` can therefore pass validation while the rest of the port machinery assumes these indices are unambiguous. Could we add `boundaries.floquetport` to both checks and cover it in `test-config.cpp`?

palace/models/spaceoperator.cpp
```diff
@@ -59,8 +60,11 @@ SpaceOperator::SpaceOperator(const config::SolverData &solver,
     lumped_port_op(boundaries.lumpedport, units, mat_op, *mesh.back()),
     wave_port_op(boundaries, domains, solver, problem_type, units, mat_op, GetNDSpace(),
                  GetH1Space()),
+    floquet_port_op(boundaries.floquetport, boundaries.periodic, problem_type, units,
```
hughcars, line 63: With Floquet ports now owned by `SpaceOperator`, should their boundary attributes be included in `CheckBoundaryProperties()`? That function builds markers for PEC/farfield/conductivity/impedance/lumped/wave/surface-current boundaries and verifies that no boundary attribute has multiple BCs, but `floquet_port_op.GetAttrList()` is not part of the marker set.

As a result, a config can assign the same boundary attribute to a Floquet port and e.g. PEC/impedance/wave/lumped without tripping the conflict check, which can silently assemble incompatible boundary conditions. Could we add a `floquet_port_marker` there and include it in the final conflict sum (and in `aux_bdr_marker` if Floquet boundaries should participate like wave/lumped boundaries for the auxiliary H1 space)?

palace/models/postoperatorcsv.cpp
```diff
@@ -1217,6 +1220,59 @@ auto PostOperatorCSV<solver_t>::InitializePortZ(const SpaceOperator &fem_op)
...
+  }
+
+  using fmt::format;
+  floquet_port_S = TableWithCSVFile(post_dir / "port-floquet-S.csv", reload_table);
+
+  Table t;
+  t.reserve(nr_expected_measurement_rows, 100);
+  t.insert("idx", "f (GHz)", -1, 0, PrecIndexCol(solver_t), "");
+
+  bool circular_output = false;
```
hughcars: This chooses the Floquet CSV polarization basis globally: if any excited Floquet port is circular, columns are created as `RHC/LHC` for every excitation. But `PrintFloquetPortS()` chooses `TE/TM` vs `RHC/LHC` from `measurement_cache.floquet_circular_output`, which is set per current excitation in `MeasureFloquetPorts()`.

So a model with one circular Floquet excitation and one linear Floquet excitation will initialize only circular columns, then the linear excitation will try to write `TE/TM` column names that were never inserted (`Table::operator[]` throws on missing columns). Could the column setup determine the basis per `ex_idx` by finding the port whose `port.excitation == ex_idx`, or otherwise allocate both bases?

docs/src/config/boundaries.md
```diff
@@ -121,6 +125,15 @@ wave port boundary mode analysis, overriding any other boundary condition assign
 attributes. Only relevant when wave port boundaries are specified under
 [`config["Boundaries"]["WavePort"]`](#boundaries%5B%22WavePort%22%5D).
 
+`"FloquetPort"` :  Array of objects for configuring Floquet port boundary conditions for
+periodic structures. Floquet ports provide absorbing boundary conditions for diffraction
+gratings and other periodic electromagnetic devices, enabling the computation of
+diffraction efficiencies (S-parameters) for multiple propagating orders. Floquet port
+boundaries are only available for frequency domain driven simulations and **require periodic
+boundary conditions** to be configured under
+[`config["Boundaries"]["Periodic"]`](#boundaries%5B%22Periodic%22%5D) with at least two
```
hughcars: The implementation now requires exactly two periodic boundary pairs for Floquet ports (`periodic.boundary_pairs.size() == 2`), but the docs still say "at least two" here and in the detailed note below. Could this be changed to "exactly two" to match the runtime check?

Small nearby typo while touching this section: "indicidence" should be "incidence" in the `FloquetWaveVector` paragraph.

palace/models/portexcitations.cpp
hughcars: Even though `to(...)` just calls `fmt::format_to`, the generic lambda hides the format string from `fmt`’s stricter compile-time checking in newer/C++20 builds. That’s why PR #653 moved these sites to direct `fmt::format_to` with a `fmt::appender`.

So I’d keep the explicit form:

```cpp
auto out = fmt::appender{buffer};
fmt::format_to(out, "...", args...);
```

hughcars (changes requested):
A few more small fixes, typos etc, but essentially ready imo.

palace/utils/configfile.cpp
simlapointe: Good point. Included it in `Validate` and covered in `test-config.cpp`

palace/models/spaceoperator.cpp
simlapointe: Yes they should, added it just now in [775deef](https://github.com/awslabs/palace/pull/693/commits/775deefe33274d2e2971b3331e5c2c1f319bd41c)

palace/models/postoperatorcsv.cpp
simlapointe: Now done in [775deef](https://github.com/awslabs/palace/pull/693/commits/775deefe33274d2e2971b3331e5c2c1f319bd41c)

docs/src/config/boundaries.md
simlapointe: Fixed the docs in [775deef](https://github.com/awslabs/palace/pull/693/commits/775deefe33274d2e2971b3331e5c2c1f319bd41c)

palace/models/portexcitations.cpp
simlapointe: I see, reverted back to the explicit form in [775deef](https://github.com/awslabs/palace/pull/693/commits/775deefe33274d2e2971b3331e5c2c1f319bd41c)

hughcars (approved):
LGTM

PR #696, opened 2026-04-01
[WIP] Update port description in reference.md

phdum-a:
Work in progress to improve description of Ports in `reference.md` and to standardize conventions between lumped and wave ports.


phdum-a:
To Do: We should also add a short paragraph on the use of surface current vs lumped ports in the reference section. They have different physics effects and normalization. See here (https://github.com/awslabs/palace/pull/661#issuecomment-4039996565) and (https://github.com/awslabs/palace/issues/616#issuecomment-4027227373)

simlapointe:
Closing. Cherry-picked the commits into #823 and added a short note on lumped port vs surface current.

PR #698, opened 2026-04-06
Build containers

### LINKED ISSUE #489: Support building for GPUs in Singularity definition file
The [singularity/singularity.def](https://github.com/awslabs/palace/blob/main/singularity/singularity.def) definition file does not support GPUs right now.
Apptainer itself [does support CUDA and ROCm](https://apptainer.org/docs/user/main/gpu.html#) with special flags that need to be used when running.

### Suggestion
Add something like optional arguments `PALACE_WITH_CUDA`, `PALACE_WITH_HIP`, `PALACE_WITH_GPU_AWARE_MPI` to the definition file, see [Apptainer -> %arguments](https://apptainer.org/docs/user/1.2/definition_files.html#arguments). Corresponding if checks should be added for installing the compilers.

Sbozzolo:
Spack-built containers have proven powerful and flexible. Therefore, this PR changes our recommendation on how to build with Singularity and adds a new workflow to build OCI and Singularity containers.

Remove the cylinder/floquet probes because we found them unstable (the problem has a lot of symmetry and we are probably landing on a degenerate mode) and because they don't add much testing value.

Closes #502 
Closes #489


.github/actions/palace-ci/action.yml
```diff
@@ -95,10 +95,6 @@ runs:
           concretizer:
             reuse: false
             unify: true
-            # generic cpu target so that we can increase cache hits
```
Sbozzolo, line 98: Possibly helps with re-use in containers too

.github/actions/build-container/spack_env/PalaceDockerfile
```diff
@@ -0,0 +1,89 @@
...
+          git apply "$patch"; \
+        else \
+          echo "Cannot apply $patch (may already be merged)"; \
+        fi; \
+      fi; \
+    done
+
+# Export cache credentials so spack can access OCI mirrors
+ENV GITHUB_USER=${GITHUB_USER}
+ENV GITHUB_TOKEN=${GITHUB_TOKEN}
```
cameronrutherford: Should use `ARG` not `ENV` here to avoid baking the contents into the image. I know that you reset this later, but since Dockerfiles are just a combination of different layers, you can still see this in the intermediate layers, even if it isn't in the final container. Dockerfiles and git are similar in that sense, and some runtimes store there layers as git objects - [CharlieCloud](https://dl.acm.org/doi/10.1145/3624062.3624585), and it's [Spack Package](https://packages.spack.io/package.html?name=charliecloud).


```suggestion
# Export cache credentials so spack can access OCI mirrors
ARG GITHUB_USER
ARG GITHUB_TOKEN
```

You could also use a [build-time secret](https://docs.podman.io/en/v4.6.0/markdown/options/secret.html), which is a feature of most container build tools.

.github/actions/build-container/spack_env/PalaceDockerfile
```diff
@@ -0,0 +1,89 @@
...
+    awk -F: '{print $1}' | xargs strip
+{% endif %}
+
+# Modifications to the environment that are necessary to run
+RUN cd {{ paths.environment }} && \
+    spack env activate --sh -d . > activate.sh
+
+# Clear credentials from the image
+ENV GITHUB_USER=""
+ENV GITHUB_TOKEN=""
```
cameronrutherford:
```suggestion
```

See above. Nice try ;)

.github/actions/build-container/spack_env/PalaceDockerfile
cameronrutherford: I think that [hadolint](https://hadolint.com/) might have caught this if we were running this in CI? I guess it would only work if we somehow scanned the generated Dockerfile, which might be difficult with `spack containerize`.

.github/actions/build-container/spack_env/spack.yaml
```diff
@@ -0,0 +1,77 @@
...
+    spack:
+      binary: true
+      url: https://binaries.spack.io/develop
+    local-buildcache:
+      binary: true
+      url: oci://ghcr.io/awslabs/palace-develop-testing
+      signed: false
+      access_pair:
+        id_variable: GITHUB_USER
+        secret_variable: GITHUB_TOKEN
```
cameronrutherford, line 51: These are already based on environment variables, so the `ARG`/secret approach should work.

.github/actions/build-container/spack_env/spack.yaml
```diff
@@ -0,0 +1,77 @@
...
+    # Compile gcc@15, then use it to compile everything else
+  - group: compilers
+    specs:
+      # If you change this, you might need to change PalaceDockerfile so that we
+      # remove the compiler from the container
+      - gcc@15
+    override:
+      concretizer:
+        # Do not force the target to be specific, let it generic, so that we can
+        # reuse compilers across different machines.
```
cameronrutherford, line 65: This is nice 💯

.github/actions/build-container/spack_env/spack.yaml
```diff
@@ -0,0 +1,77 @@
...
+          granularity: generic
+  - group: palace
+    needs: [compilers]
+    specs:
+    - palace@develop+libxsmm+superlu-dist+mumps+sundials+strumpack+slepc+arpack+int64 %gcc@15 ^openblas@0.3.29 ^openmpi schedulers=none fabrics=ofi
+    override:
+      packages:
+        # Enable AWS EFA
+        libfabric:
+          require: fabrics=efa,lnx,mrail,rxm,shm,sockets,tcp,udp
```
cameronrutherford, line 76:
```suggestion
    - palace@develop
    override:
      packages:
        all:
          providers:
            mpi: openmpi
            blas: openblas
          require: %gcc@15
        # Enable AWS EFA
        libfabric:
          require: fabrics=efa,lnx,mrail,rxm,shm,sockets,tcp,udp
        openmpi:
          require: "schedulers=none fabrics=ofi"
        openblas:
          require: "@0.3.29"
        # Palace and dependency preferences
        palace:
          require: +libxsmm+superlu-dist+mumps+sundials+strumpack+slepc
        
```

I think this is just personal preference, but I find this to be more readable? I admit the spec groups is more new to me.

I find it annoying that I can't reference the compiler from the previous target group though... Maybe there's a way to do that / a YAML feature that does this. I am also rusty on the [syntax](https://spack.readthedocs.io/en/latest/packages_yaml.html#package-requirements).

docs/src/install.md
```diff
@@ -69,25 +69,47 @@ for an introduction.
...
-where the repository source code has been cloned to `<SOURCE_DIR>`. For more information
-about Singularity/Apptainer, see the
-[Quick Start](https://apptainer.org/docs/user/main/quick_start.html) guide in the
-Singularity/Apptainer documentation.
+For a more complete example with binary caches, custom compilers, and a
+Dockerfile template that handles *Palace*-specific setup, see
+[`.github/actions/build-container/spack_env/spack.yaml`](https://github.com/awslabs/palace/blob/main/.github/actions/build-container/spack_env/spack.yaml).
+
+You can customize your *Palace* installation by adding variants to the `palace`
+spec (e.g., compiling with other solvers or with GPU support).
```
cameronrutherford, line 112: Will we be building GPU containers in future? What about making containers accessible to users / supporting that?

.github/workflows/singularity.yml
cameronrutherford, line 1: Should we bring this build back eventually? You can also publish `.sif` files for others to consume / run.

.github/workflows/containers.yml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  packages: write
+  contents: write
```
cameronrutherford:
```suggestion
  contents: read
```

.github/actions/palace-ci/action.yml
```diff
@@ -95,10 +95,6 @@ runs:
...
-            # (we also do not care too much about hyper-optimal performance)
-            targets:
-              granularity: generic
```
cameronrutherford, line 101: Perhaps we can still specify what we are using now - is the default `targets:granularity:microarchitectures`?

.github/actions/palace-ci/action.yml
cameronrutherford: I don't quite understand how this relates to the container build. Are you saying that using micro-arch granularity will make it more likely the container will get cache hits from the regular action?

.github/actions/build-container/spack_env/PalaceDockerfile
```diff
@@ -0,0 +1,89 @@
...
+set -o noclobber \
+{{ manifest }} > {{ paths.environment }}/spack.yaml
+
+# Check if the binary cache is reachable
+RUN cd {{ paths.environment }} && spack env activate . && \
+    if spack buildcache list -L 2>/dev/null; then \
+      echo "Binary cache is reachable"; \
+    else \
+      echo "WARNING: Binary cache is not reachable, all packages will be built from source"; \
+    fi
```
cameronrutherford, line 45: NIT: Should this just fail? I imagine this is less about building from source, as sometimes this might happen, but we do want to be able to push to the buildcache at the end as that seems like the more important step so we save time for future builds.

.github/actions/build-container/action.yml
```diff
@@ -0,0 +1,119 @@
...
+        name: ${{ steps.name.outputs.image_name }}-oci
+        path: ${{ steps.name.outputs.image_name }}.tar
+        retention-days: 1
+
+    - name: Upload SIF image
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{ steps.name.outputs.image_name }}-sif
+        path: ${{ steps.name.outputs.image_name }}.sif
+        retention-days: 1
```
cameronrutherford, line 119: Should we be uploading this and the OCI image to ECR?

.github/actions/build-container/action.yml
```diff
@@ -0,0 +1,119 @@
...
+
+    - name: Build OCI image
+      shell: bash
+      run: |
+        source /tmp/timing_functions.sh
+        start_timer
+
+        sudo podman build --format docker \
+          --build-arg GITHUB_USER=$GITHUB_USER \
+          --build-arg GITHUB_TOKEN=$GITHUB_TOKEN \
```
cameronrutherford, line 61: Could do secret or `ARG`, not stored as env in layers.

.github/actions/build-container/action.yml
```diff
@@ -0,0 +1,119 @@
...
+        SHORT_HASH=$(git rev-parse --short HEAD)
+        IMAGE_NAME="palace-${ARCH}-${SHORT_HASH}"
+        echo "image_name=${IMAGE_NAME}" >> "$GITHUB_OUTPUT"
+        echo "Image name: ${IMAGE_NAME}"
+
+    - name: Generate Dockerfile
+      shell: bash
+      run: |
+        cd .github/actions/build-container/spack_env && spack -e . containerize > $GITHUB_WORKSPACE/Dockerfile
+
```
cameronrutherford, line 47: I guess hadolint could go here? Or we upload / have these Dockerfiles referenced in git somehow and just run this and other checks in a pre-commit check.

cameronrutherford (changes requested):
Apart from the way you pass the secret into the container, nice job! This will be useful for us and hopefully will enable users as well.

Maybe GPU containers won't be too hard with this ironed out, and existing GPU tests.

.github/actions/palace-ci/action.yml
Sbozzolo: Exactly!

.github/actions/build-container/spack_env/spack.yaml
Sbozzolo: I agree and I like this too, but I found it more brittle with the `groups` feature, where I would get some unexpected behaviors. I think we should let spack iron out some of the bugs there before we switch to this

docs/src/install.md
Sbozzolo: We could. The reason I didn't do it here is because you really need to compile for the target cuda_arch to get good performance. So, I didn't want to pick a "random" cuda_arch.

.github/workflows/singularity.yml
Sbozzolo: The sif file is already produced in this action.

.github/actions/palace-ci/action.yml
Sbozzolo: Yes, correct

.github/actions/build-container/spack_env/PalaceDockerfile
Sbozzolo: I kept it like that because the spack.yaml file is also useful outside of GitHub Actions (I use it all the time to build/prototype on our internal systems). In that case, I don't want necessarily care about hitting all the buildcaches

.github/actions/build-container/action.yml
Sbozzolo: Yes. Setting this up is left for future work.

docs/src/install.md
cameronrutherford: You can specify a list of cuda architectures. You don't have to just pick one.

This does depend on the dependent packages / our package though. Some packages, like kokkos, don't allow cross compilation. I would naively like to think that our dependencies would play nice and allow this cross compilation.

Sbozzolo: I checked and many of our dependencies only support one cuda arch at the time, unfortunately.

.github/workflows/containers.yml
Sbozzolo: Thank you!

.github/actions/build-container/spack_env/PalaceDockerfile
Sbozzolo: Thank you!

.github/actions/build-container/action.yml
Sbozzolo: Added!

cameronrutherford (approved):

cameronrutherford:
```
-:19 DL3003 warning: Use WORKDIR to switch to a directory
-:19 SC2046 warning: Quote this to prevent word splitting.
-:36 SC2016 info: Expressions don't expand in single quotes, use double quotes for that.
-:96 DL3003 warning: Use WORKDIR to switch to a directory
-:104 SC2046 warning: Quote this to prevent word splitting.
-:104 DL3003 warning: Use WORKDIR to switch to a directory
-:107 DL3003 warning: Use WORKDIR to switch to a directory
-:113 DL3003 warning: Use WORKDIR to switch to a directory
-:117 SC3045 warning: In POSIX sh, read -d is undefined.
-:117 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
-:127 SC2038 warning: Use 'find .. -print0 | xargs -0 ..' or 'find .. -exec .. +' to allow non-alphanumeric filenames.
-:127 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
-:134 DL3003 warning: Use WORKDIR to switch to a directory
```

Seems like hadolint is working and just reporting warnings - nice! 🎊

Sbozzolo:
> ```
> -:19 DL3003 warning: Use WORKDIR to switch to a directory
> -:19 SC2046 warning: Quote this to prevent word splitting.
> -:36 SC2016 info: Expressions don't expand in single quotes, use double quotes for that.
> -:96 DL3003 warning: Use WORKDIR to switch to a directory
> -:104 SC2046 warning: Quote this to prevent word splitting.
> -:104 DL3003 warning: Use WORKDIR to switch to a directory
> -:107 DL3003 warning: Use WORKDIR to switch to a directory
> -:113 DL3003 warning: Use WORKDIR to switch to a directory
> -:117 SC3045 warning: In POSIX sh, read -d is undefined.
> -:117 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
> -:127 SC2038 warning: Use 'find .. -print0 | xargs -0 ..' or 'find .. -exec .. +' to allow non-alphanumeric filenames.
> -:127 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
> -:134 DL3003 warning: Use WORKDIR to switch to a directory
> ```
> 
> Seems like hadolint is working and just reporting warnings - nice! 🎊

I could not fix these ones because they are in the spack template.

examples/cylinder/floquet.json
```diff
@@ -29,22 +29,7 @@
           "Index": 1,
           "Attributes": [1]
         }
-      ],
-      "Probe":
```
hughcars, line 33: This change seems unrelated to the containerization?

.github/actions/build-container/spack_env/spack.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - $env
+      - $spack/share/spack/templates
+  packages:
+    all:
+      require: ~openmp       # We don't want any package with OpenMP
+  container:
+    format: docker
+    template: PalaceDockerfile
+    images:
+      os: "ubuntu:24.04"
```
hughcars, line 35: I'm assuming that the gcc shipping here is <15, hence not using an external compiler? Would `%gcc@15:` later work with a compiler external find?

.github/workflows/containers.yml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  # Filter to allow skipping the test if no relevant changes occurred.
+  filter:
+    runs-on: ubuntu-latest
+    outputs:
+      test: ${{ steps.filter.outputs.test }}
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dorny/paths-filter@v4
```
hughcars, line 34: All the others are using `@v3`, is this worth updating for the others?

hughcars (changes requested):
Generally LGTM, just a few nits. 

There are a couple of stale references from the singularity removal:
- README.md:53-55 links to singularity/ which no longer exists after this PR — should point to the container docs in install.md
- docs/src/developer/testing.md:393 still lists singularity.yml — should be containers.yml

The canary failure is unrelated.

examples/cylinder/floquet.json
Sbozzolo: It is unrelated. I found that the job was failing (we are compiling with slightly different settings, e.g. targeting a specific microarch and with +int64). I chatted with @simlapointe and agreed that the simplest way forward was to remove the probe, as it the test was not particularly strong anyway. The probable cause was mode degeneracy.

.github/actions/build-container/spack_env/spack.yaml
Sbozzolo: We could install gcc 15, but I preferred leaving everything up to spack. This way, I don't need to worry about the underlying system/image and the available packages, and I can more easily swap compilers (e.g., if I want to try llvm).

.github/workflows/containers.yml
Sbozzolo: I already updated all the other ones in #697

hughcars (approved):

Issue #700, opened 2026-04-07
Investigate lookahead failing


Sbozzolo:
Starting with https://github.com/awslabs/palace/actions/runs/24088054281/job/70266448712, the lookahead job has started failing:
```julia
┌ Warning: Regression test error in file 'error-indicators.csv' at row 1, column Norm: 0.002999598147249 ≉ 0.00300321331256, abs_err = 3.6151653109999134e-6 (atol=1.0e-16), rel_err = 0.0012037657451372554 (rtol=0.0001)
└ @ Main /opt/actions-runner/_work/palace/palace/test/examples/testcase.jl:236
Results: Test Failed at /opt/actions-runner/_work/palace/palace/test/examples/testcase.jl:244
  Expression: last(rowcol)

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/share/julia/stdlib/v1.12/Test/src/Test.jl:680 [inlined]
 [2] macro expansion
   @ /opt/actions-runner/_work/palace/palace/test/examples/testcase.jl:244 [inlined]
 [3] macro expansion
   @ ~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/share/julia/stdlib/v1.12/Test/src/Test.jl:1776 [inlined]
 [4] testcase(testdir::String, testconfig::String, testpostpro::String; palace::Vector{String}, np::Int64, rtol::Float64, atol::Float64, excluded_columns::Vector{String}, custom_tests::Dict{Any, Any}, paraview_fields::Bool, gridfunction_fields::Bool, skip_rowcount::Bool, generate_data::Bool, device::String, linear_solver::String, eigen_solver::String)
   @ Main /opt/actions-runner/_work/palace/palace/test/examples/testcase.jl:132
Test Summary: | Pass  Fail  Total  Time
Results       |  205     1    206  1.4s
```

The only commit that could have changed something is https://github.com/mfem/mfem/commit/fff973f19282efd964bb8d068600dacea09bfd2c

Sbozzolo:
cc: @hughcars

hughcars:
### Simplex quadrature impact on Palace

MFEM mfem/mfem#5246 replaces legacy triangle/tet integration rules (some with negative weights) with positive-weight Witherden-Vincent rules. Quadrature point locations and weights change at several polynomial orders.

3 of 18 Palace regression tests fail with the new rules:

[...]

PR #701, opened 2026-04-07
Print commit hash of @develop packages

Sbozzolo:
This helps with pin-pointing commits that break lookahead, as it happened in #700 

Prints:
```
=== @develop package commit hashes ===
hypre: 73f739a8e16820567333b5712997c7b950d3815a
mfem: aed9c8ef4ac78736814fbea2788e6bacbad66649
=== end @develop hashes ===
```

### DIFF
--- a/.github/actions/palace-ci/action.yml
+++ b/.github/actions/palace-ci/action.yml
@@ -287,6 +287,15 @@ runs:
         spack -e . install --only-concrete --no-check-signature --fail-fast --show-log-on-error --only dependencies --test root -j $(nproc)
         end_timer "Build Dependencies"
 
+    - name: Print @develop commit hashes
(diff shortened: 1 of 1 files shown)


cameronrutherford (approved):
I guess we want to stay up to date with develop on dependencies? I'm used to pinning these commits and updating with some cadence. This makes sense though to track when something breaks...

simlapointe (approved):

Sbozzolo:
> I guess we want to stay up to date with develop on dependencies? I'm used to pinning these commits and updating with some cadence. This makes sense though to track when something breaks...

We have a "canary" workflow that just uses the most updated dependencies so that we can identify when things break. This PR prints out the commit hash of such dependencies so that one does not need to go by dates.

PR #676, opened 2026-03-17
Add tutorial for circuit extraction

phdum-a:
WIP

PR #702, opened 2026-04-09
Add tutorial for driven solver (adaptive and uniform)

phdum-a:
PR #676 had two tutorials: one on the driven solver generally, and one on the circuit extraction specifically.

This PR decouples these two tutorials for simplicity. It extracts the first tutorial since this is ready for review and does not depend on other text (like the port documentation in #696).


phdum-a:
Pushed 3 tiny commits to fix typo ($\Ohm$ -> $\Omega$), notation $V -> Q$ for orthogonalised basis, and python script formatting.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+
+The following are important tuning configurations:
+
+  - [`"AdaptiveTol"`](../config/solver.md#solver%5B%22Driven%22%5D): start with `1e-3` on a coarse
+    grid for S-parameter sweeps. Tighten this tolerance as needed during the validation against the
+    uniform solver or in production runs.
+  - `"Linear"/"Tol"`: set the linear solver tolerance to be substantially smaller than
+    `"AdaptiveTol"`. The adaptive solver is based on an interpolation and can be unusually sensitive
+    to errors in the linear solver. Increase this tolerance in your validation runs and ensure the
+    output is no different. Should *Palace* log warnings like `Minimal rational interpolation encountered rank-deficient matrix` try tightening this tolerance substantially.
```
Sbozzolo: Is this something we should explicitly say in the log message in Palace?

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+sample frequency point ``f``. We will also refer to these solves as "full" or "High-Dimensional
+Solves" (HDM) since they solve the full linear problem on the finite element space. Here we are
+sampling a linear grid between ``2~\mathrm{GHz}`` and ``32~\mathrm{GHz}`` in steps of ``\Delta f = 0.1~\mathrm{GHz}``, but [users can provide more complicated sample
+specification](../config/solver.md#solver%5B%22Driven%22%5D%5B%22Samples%22%5D).
+
+The linear solver tolerance `"Tol": 1.0e-12` is chosen to be very small. Such a small tolerance is
+near the limit of what *Palace*'s solvers can reach using double precision. The error of the uniform
+linear solve ``\bm{A}(\omega) \bm{x} = \bm{b}(\omega)`` also depends on the condition number of the
+system matrix ``\bm{A}(\omega)`` at each frequency ``\omega = 2 \pi f``. Importantly, if the system
+has resonances (poles) near the real axis, the linear solve may become very poorly conditioned. This
```
Sbozzolo:
```suggestion
has resonances (poles) near the real axis, the linear solve may become poorly conditioned. This
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+There are two eigenmodes of particular interest that we discovered in the previous tutorial:
+
+  - A “transmon” mode near ``4.10~\textrm{GHz}`` with ``Q = 1.8 \cdot 10^4``,
+  - A “resonator” mode near ``5.60~\textrm{GHz}`` with ``Q = 7.9 \cdot 10^3``.
+
+The eigenmode visualization shows that even the transmon mode has appreciable weight on the readout
+resonator and in the feedline.
+
+We will now perform uniform and adaptive driven simulations on this model, by exciting the
+``50~\Ohm`` resistive ports. We will use the same mesh `examples/transmon/mesh/transmon.msh2` as the
```
Sbozzolo:
```suggestion
``50~\Omega`` resistive ports. We will use the same mesh `examples/transmon/mesh/transmon.msh2` as the
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+## Revisiting the Two Coplanar Waveguide Example (Part I)
+
+We start by re-visiting the driven response of the [two CPW example](cpw.md), using the mesh
+[`mesh/cpw_lumped_0.msh`](https://github.com/awslabs/palace/blob/main/examples/cpw/mesh/cpw_lumped_0.msh)
+that places lumped ports at the end of the CPWs.
+
+### Uniform Solver
+
+First, we set up a uniform driven simulation using the configuration file
+[`cpw_lumped_uniform_convergence.json`](https://github.com/awslabs/palace/blob/main/examples/cpw/cpw_lumped_uniform_convergence.json).
```
Sbozzolo: This file was not included in the commit

examples/transmon/transmon_tutorial_driven_plots.py
```diff
@@ -0,0 +1,565 @@
...
+)
+
+mpl.rcParams.update({
+    "figure.dpi": 150,
+    "font.family": "serif",
+    "font.serif": ["Times"],
+    "mathtext.fontset": "cm",
+    "axes.formatter.use_mathtext": True,
+    "savefig.bbox": None,
+})
```
Sbozzolo: Duplicated?

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+coarse output `"Samples"` grid and at key frequencies. This helps build intuition about the response
+and the approximation error of the adaptive solver for the specific quantities of interest. Users
+should tighten tolerances and re-validate as needed before running at fine frequency resolution.
+
+The following are important tuning configurations:
+
+  - [`"AdaptiveTol"`](../config/solver.md#solver%5B%22Driven%22%5D): start with `1e-3` on a coarse
+    grid for S-parameter sweeps. Tighten this tolerance as needed during the validation against the
+    uniform solver or in production runs.
+  - `"Linear"/"Tol"`: set the linear solver tolerance to be substantially smaller than
```
Sbozzolo: We usually use the notation `["Linear"]["Tol"]`  for nested configs. To be more precise, the other doc pages often refer to configs arguments as `config["Solver"]["Linear"]["Tol"]`.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+```
+
+*Palace* calls the uniform driven solver when
+[`"AdaptiveTol"`](../config/solver.md#solver%5B%22Driven%22%5D) in `Solver/Driven` is `0.0` or not
+specified. The uniform solver will iterate over all sample points specified in
+`Solver/Driven/Samples` and independently solve the linear equation for the electric field at the
+sample frequency point ``f``. We will also refer to these solves as "full" or "High-Dimensional
+Solves" (HDM) since they solve the full linear problem on the finite element space. Here we are
+sampling a linear grid between ``2~\mathrm{GHz}`` and ``32~\mathrm{GHz}`` in steps of ``\Delta f = 0.1~\mathrm{GHz}``, but [users can provide more complicated sample
+specification](../config/solver.md#solver%5B%22Driven%22%5D%5B%22Samples%22%5D).
```
Sbozzolo:
```suggestion
specifications](../config/solver.md#solver%5B%22Driven%22%5D%5B%22Samples%22%5D).
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+
+## Driving the Transmon Model
+
+### Set-Up
+
+Let us now apply the driven and adaptive solver to a model of a transmon qubit, which was already
+discussed in the [eigenmode tutorial](transmon.md). We assume familiarity with that tutorial. As
+discussed there, the model consists of a transmon qubit and a quarter-wave coplanar waveguide
+readout resonator coupled to a feedline. There are three lumped ports in this model: ports 1 and 2
+are the ``50~\Omega`` resistive feedline terminations and port 3 is a passive LC element (``L = 14.86\,\textrm{nH}``, ``C = 5.5\,\textrm{fF}``) representing the linearised Josephson junction.
```
Sbozzolo:
```suggestion
are the ``50~\Omega`` resistive feedline terminations and port 3 is a passive LC element (``L = 14.86\,\mathrm{nH}``, ``C = 5.5\,\mathrm{fF}``) representing the linearised Josephson junction.
```
for consistency with everything else

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+
+There are two eigenmodes of particular interest that we discovered in the previous tutorial:
+
+  - A “transmon” mode near ``4.10~\textrm{GHz}`` with ``Q = 1.8 \cdot 10^4``,
+  - A “resonator” mode near ``5.60~\textrm{GHz}`` with ``Q = 7.9 \cdot 10^3``.
```
Sbozzolo:
```suggestion
  - A “transmon” mode near ``4.10~\marhrm{GHz}`` with ``Q = 1.8 \cdot 10^4``,
  - A “resonator” mode near ``5.60~\mathrm{GHz}`` with ``Q = 7.9 \cdot 10^3``.
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+
+```@raw html
+<br/><p align="center">
+  <img src="../../assets/examples/driven_ua_transmon_sparam_adaptive_rms.svg" width="80%" />
+</p><br/>
+```
+
+In both plots above, the error is more structured than it was in the CPW example. It is worse near
+the eigenmode locations and better far away. The adaptive solver also adds HDM solves in a more
+structured manner to “shadow” the eigenmodes. This behaviour is simple to interpret — the response
```
Sbozzolo:
```suggestion
structured manner to “shadow” the eigenmodes. This behavior is simple to interpret — the response
```
:P

Sbozzolo (commented):
[Largely discussed offline.]

The content looks great. I suggest breaking it up to make it more digestable (part of it could be in Features, other could be in References).

Two concerns on maintanability:
- There are hardcoded numbers throughout the text that would need to change if other things change
- The Python scripts have a lot of overlapping material

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+
+    For the transmon example, the data can be generated with the script `examples/transmon/transmon_tutorial_driven.jl` and the result plots generated with `examples/transmon/transmon_tutorial_driven_plots.py`.
+
+## Driven Solver Quick-Start
+
+*Palace* has two modes for frequency-domain driven simulations: uniform and adaptive. The uniform
+solver runs one independent FEM solve per output frequency, which is reliable but slow. The adaptive
+solver runs a small number of FEM solves and constructs a reduced-order model (ROM) that
+interpolates the response at all other output frequencies. This is fast, but requires more careful
+set-up and validation.
```
hughcars:
```suggestion
interpolates the response at all other output frequencies. This is fast, but introduces a source of error compared to the full model evaluation, which has to be controlled.
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+and the approximation error of the adaptive solver for the specific quantities of interest. Users
+should tighten tolerances and re-validate as needed before running at fine frequency resolution.
+
+The following are important tuning configurations:
+
+  - [`"AdaptiveTol"`](../config/solver.md#solver%5B%22Driven%22%5D): start with `1e-3` on a coarse
+    grid for S-parameter sweeps. Tighten this tolerance as needed during the validation against the
+    uniform solver or in production runs.
+  - `"Linear"/"Tol"`: set the linear solver tolerance to be substantially smaller than
+    `"AdaptiveTol"`. The adaptive solver is based on an interpolation and can be unusually sensitive
```
hughcars:
```suggestion
    `"AdaptiveTol"`. The adaptive solver is based on an interpolation and can be very sensitive
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+    solves per excitation (default: 20). If the log indicates that you exceed this bound, it means
+    the adaptive solver has not reached the target tolerance and this bound should be increased. If
+    the adaptive solver never converges, it might indicate a numerical problem.
+  - [`"AdaptiveConvergenceMemory"`](../config/solver.md#solver%5B%22Driven%22%5D): number of
+    consecutive "safety" samples that have to be below the tolerance threshold before convergence.
+    Defaults to 2, which is small but fast. Increase this number to 3+ if you see early termination
+    without convergence.
+
+The rest of this tutorial will illustrate the driven solvers on two different examples (CPWs and
+transmon), explain the adaptive algorithm in detail, and give more detailed user guidance.
```
hughcars:
```suggestion
The rest of this tutorial will illustrate the driven solvers on two different examples (CPWs and
and the single transmon), explain the adaptive algorithm in detail, and give more detailed user guidance.
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+The primary downside of the uniform driven solver is its slow performance. Every output sample has
+to be individually calculated, which becomes prohibitive on a fine output grid. This is the problem
+the adaptive solver addresses.
+
+To perform a driven simulation with the adaptive driven solver, we set
+[`"AdaptiveTol"`](../config/solver.md#solver%5B%22Driven%22%5D) to `1e-1` in the configuration. We
+will define this quantity more precisely later. Simulation output, like `domain-E.csv`, has the same
+structure as the uniform solver — they are evaluated on the same frequency grid specified by
+`"Samples"` in the configuration file. However, the adaptive solver constructs a different
+"internal" frequency grid on which it does HDM solves. This internal grid is automatically chosen by
```
hughcars:
```suggestion
*internal* frequency grid on which it does HDM solves. This internal grid is automatically chosen by
```
generally the same thing for most of the `""` of terms here, they're not qualified so much as emphasized.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+```
+
+This also requires a method to efficiently recover the high-dimensional field solution ``\bm{x}``
+from the reduced ``\bm{x}_r``. If we have found such a reduced model, we can quickly solve the small
+linear system at a specific ``\omega`` and recover the full solution and all output measurements. We
+refer to the procedure of finding the model as the “offline” or “training” phase, and to using it to
+calculate many output samples as the “online” phase.
+
+There is a large literature on constructing dimensional reductions of linear systems and we refer to
+the literature for an introduction [1,2]. *Palace* finds a set of ``n`` real orthogonal vectors
```
hughcars: Link the anchors.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+``\bm{Q}`` that capture the full response of the system in the frequency range of interest
+``[f_\mathrm{min}, f_\mathrm{max}]``. The projection ``\bm{K}_r = \bm{Q}^T \bm{K} \bm{Q}``,
+``\bm{b}_r = \bm{Q}^T \bm{b}``, etc., and the prolongation ``\bm{x} = \bm{Q} \bm{x}_r`` are
+particularly simple.
+
+The set of basis vectors ``\bm{Q}`` that *Palace* uses are the orthogonalized components of the HDM
+solutions ``\bm{x}^*`` at the "internal" sampling frequencies ``f_\mathrm{sample} \in [f_\mathrm{min}, f_\mathrm{max}]``. In fact, *Palace* uses ``\mathrm{Re}\bm{x}^*``,
```
hughcars: Re and Im probably need `()` or `{}`, your call.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+The Gram-Schmidt orthogonalization algorithm used to construct ``\bm{Q}`` from the HDM solutions is
+determined by the user option
+[`"AdaptiveGSOrthogonalization"`](../config/solver.md#solver%5B%22Driven%22%5D), although a user
+should generally not need to change this. By default, it uses classical Gram-Schmidt with
+reorthogonalization which is a good high-precision option.
+
+### Choosing Internal Sample Frequencies and Convergence
+
+The above construction does not tell us how to efficiently choose the location of internal sample
+frequencies ``f_\mathrm{sample}``. For this we use the greedy sampling algorithm developed by
```
hughcars:
```suggestion
The above construction does not explain how to efficiently choose the location of internal sample
frequencies ``f_\mathrm{sample}``. For this we use the greedy sampling algorithm developed by
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+```math
+\bm{u}(\omega) = \frac{\sum_i w_i \bm{u}(\omega_i) / (\omega - \omega_i)}{\sum_i w_i / (\omega -
+\omega_i)},
+```
+
+where the ``\omega_i`` are sample frequencies, ``\bm{u}(\omega_i)`` solutions at those frequencies,
+and ``w_i`` some weight coefficients fitted. To use this interpolation for our driven
+electromagnetic solver, we linearize the quadratic ``KCM`` operator as a linear operator and
+identify ``\bm{u}^T = (\bm{x}^T, i \omega \bm{x}^T)``. This type of linearization is familiar to
+physicists as going from Newton's equations (second order in time) to Hamilton's equations (first
```
hughcars:
```suggestion
identify ``\bm{u}^T = (\bm{x}^T, i \omega \bm{x}^T)``. This type of linearization may be familiar to
physicists as going from Newton's equations (second order in time) to Hamilton's equations (first
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+
+Specifically, we declare convergence when this error is below the user-defined tolerance
+``\varepsilon < \varepsilon_\mathrm{tol}``, set by
+[`"AdaptiveTol"`](../config/solver.md#solver%5B%22Driven%22%5D), for a number of consecutive
+frequency samples, set by
+[`"AdaptiveConvergenceMemory"`](../config/solver.md#solver%5B%22Driven%22%5D). Note that the norm
+``\vert\vert \bm{x} \vert\vert`` is the L2 norm in the finite element space. So this is similar to,
+but not the same as, the physical norm computed from the overlap in space ``\int \bm{E}^*(\bm{r}) \cdot \bm{E}(\bm{r}) d^3r``, which corresponds to a mass-weighting in finite-elements
+``\bm{x}^T\bm{M}\bm{x}``. This difference between L2 and mass-weighted overlap can be significant on
+highly anisotropic meshes.
```
hughcars: This seems like a good moment to mention that the indicator we use is an estimator, not a strict bound, so the memory is a way to mitigate against falsely ending when the true error is still large. A reader who doesn't catch that detail might be a bit confused why we don't trust the first sample set below the tolerance level. 

Edit: I see you bring this up in the next section, I was just getting ahead.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+measurement is difficult, such as integrating the electric field in a region where the field is
+tiny.
+
+### Adaptive Solver Problems and User Guidance
+
+The rational interpolation approach above is extremely efficient at building a basis using only a
+very small number of HDM solves. However, there are two challenges that we encounter in practical
+use.
+
+**Early termination** due to accidental convergence. As discussed above, the convergence criterion
```
hughcars: Looking at the example rendering of this at https://awslabs.github.io/palace/previews/PR702/examples/tutorial_driven_uniform_v_adaptive/, the "**" "**" isn't coming out. It might be worth fiddling the format a little to identify the breaks here, as with the paragraph break I was momentarily confused why "To avoid this" was the second point.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+consecutive samples in a row are below tolerance. This number is defined by the
+[`"AdaptiveConvergenceMemory"`](../config/solver.md#solver%5B%22Driven%22%5D) configuration. The
+default is `2`, which is small. Increasing this number to say `3` or `4`, especially in models where
+the user does not understand the convergence properties, can be helpful. However, each HDM sample
+adds a substantial computational cost to the training phase.
+
+**Numerical problems** due to finite precision. The mathematical formulation of the rational
+interpolation for the error indicator assumes that the solution at the sample points is exact.
+However, in reality there is the linear solver tolerance
+[`"Linear"/"Tol"`](../config/solver.md#solver%5B%22Linear%22%5D) that sets the error on this and can
```
hughcars: Normally we put the links with the `[]`. Can be better fixed once your other PR is merged with the good hyperlinking capability.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+[`"AdaptiveMaxSamples"`](../config/solver.md#solver%5B%22Driven%22%5D) (default: 20), after which
+the solver just uses the capped ROM to do the computation. That cap is helpful to bound the
+oversampling cases due to numerical issues. However, under normal conditions the user should
+increase [`"AdaptiveMaxSamples"`](../config/solver.md#solver%5B%22Driven%22%5D) to guarantee
+accuracy.
+
+As discussed above, a user can also force certain frequency points to be added to the PROM before
+the adaptive sampling starts by using [`"AddToPROM": true`](../config/solver.md#solver%5B%22Driven%22%5D%5B%22Samples%22%5D). This is primarily a
+debugging tool. Routine usage is not recommended — adding sample by hand negates the major benefit
+of the adaptive solver to “choose the best” samples for a given accuracy. However, this option might
```
hughcars:
```suggestion
of the adaptive solver to efficiently choose samples based on the observed approximation error. However, this option might
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+help investigate convergence issues. If adding points, the user should be careful not to
+accidentally set [`"AddToPROM": true`](../config/solver.md#solver%5B%22Driven%22%5D%5B%22Samples%22%5D) on a dense grid, since this
+will perform a very large number of HDM solves and keep them in computer memory for the duration of
```
hughcars:
```suggestion
will perform a very large number of HDM solves and keep them in memory for the duration of
```

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+well behaved. The bad relative error in the other plots is because the magnitude of ``S_{i1}`` is
+tiny to begin with, which magnifies the relative error. For example, the magnitude ``\vert S_{11} \vert`` drops down to around ``-60~\mathrm{dB} = 10^{-3}`` near ``15~\mathrm{GHz}``. That is exactly
+the location where the relative error of ``S_{11}`` spikes in the error plot above. The relative
+error in ``S_{31}``, ``S_{41}`` is always large since their values are small.
+
+A detailed discussion of error estimates and error propagation from the electric field to derived
+quantities is beyond the scope of this tutorial. From a practical perspective, we encourage users to
+validate some of the adaptive driven simulations against uniform driven simulations in order to get
+a practical sense of the error for different quantities, to help identify an appropriate adaptive
+tolerance and to diagnose convergence errors.
```
hughcars:
```suggestion
a practical sense of the error for different quantities, to help identify an appropriate adaptive
tolerance, linear solver tolerance, and to diagnose convergence errors.
```
The interplay here with how accurate the HDM is itself is worth lightly mentioning. Not worth twiddling the adaptive if your HDM samples are all bad to begin with.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+}
+```
+
+In this case we will be using the multi-excitation feature of *Palace*. Because we have specified
+both `"Excitation": 1` on port 1 and `"Excitation": 2` on port 2, *Palace* will iterate over these
+excitations separately. This makes no difference to the uniform solver, since it solves every
+excitation and frequency sample separately. However, for the adaptive solver, the projective basis
+``\bm{Q}`` is shared between all excitations and the convergence criterion ``\varepsilon < \varepsilon_\mathrm{tol}`` is on that combined basis. This means that the output of each excitation
+can be different when we simulate them together with one ROM or separately with two ROMs. However,
+these differences are within the adaptive tolerance. See the discussion above.
```
hughcars: Do you think it's worth mentioning how this can work well with AMR, as the adaptation is driven by the error in both simulations, and the mesh kept consistent between the two. In the limit of deep AMR convergence it wouldn't matter, but for the ~1e-2 errors we often work at, it can make a big difference to ensure that the same mesh is used for all the excitations. Can possibly put in one of those *info* boxes that @Sbozzolo uses in his examples.

docs/src/examples/tutorial_driven_uniform_v_adaptive.md
```diff
@@ -0,0 +1,664 @@
...
+
+[2] A. C. Antoulas, Approximation of Large-Scale Dynamical Systems. Society for Industrial and
+Applied Mathematics, 2005. doi: 10.1137/1.9780898718713.
+
+[3] D. Pradovera, “Toward a certified greedy Loewner framework with minimal sampling,” Adv Comput
+Math, vol. 49, no. 6, p. 92, Dec. 2023, doi: 10.1007/s10444-023-10091-7.
+
+[4] D. Pradovera, “Interpolatory rational model order reduction of parametric problems lacking
+uniform inf-sup stability,” SIAM J. Numer. Anal., vol. 58, no. 4, pp. 2265–2293, Jan. 2020, doi:
+10.1137/19M1269695.
```
hughcars: Nit: if you can hyperlink to the doi.

examples/cpw/cpw_tutorial_lumped_driven.jl
```diff
@@ -0,0 +1,126 @@
...
+    cpw_dir = @__DIR__
+
+    # # Run uniform sweep
+    # println("Running uniform sweep...")
+    # run(
+    #     Cmd(
+    #         `$palace_exec -np $num_processors cpw_tutorial_lumped_uniform.json`;
+    #         dir=cpw_dir
+    #     )
+    # )
```
hughcars: Debris. If running uniform is to be an option make it a kwarg. I'd bias to just deleting this, or having it be an adaptive_tol=0.0 option below.

examples/transmon/transmon_tutorial_driven.jl
```diff
@@ -0,0 +1,144 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+#=
+# README
+
+This Julia script runs Palace driven solver simulations on the transmon example, for
+the "Driven Solver: Uniform vs Adaptive" tutorial. Use the accompanying
+`transmon_tutorial_driven_plots.py` python script to generate plots once the data is
```
hughcars: What was the reason for splitting python/julia for these files? Is generating plots from julia more complicated?

hughcars (changes requested):
A few small/nit things, but this is a huge step forward for the quality of documentation. You should mention it in the CHANGELOG, as though it isn't affecting user facing behaviour, it is a big improvement in how users will understand what is a non-standard technique for those not steeped in the numerical methods. 

Only one a little bigger is it would be nice to generate the plots a) automatically from runs (possibly a long test) so that the docs are always kept in lock step with the code state, and b) generate with julia. b is mainly just a taste thing to avoid the language split, if the plots are much easier in python then I can be convinced.

hughcars (approved):

PR #704, opened 2026-04-13
Final changes for 0.16.1

hughcars:
Patch fix for 0.16.0, no new external features, only performance improvements and fixes.


Sbozzolo:
I started https://github.com/spack/spack-packages/pull/4447. It will require adding the commit hash when this is ready.

PR #708, opened 2026-04-16
Add more detailed memory reporting

Sbozzolo:
I expanded the timer system to collect memory information.

A new table is printed at the end of each run:
```
Peak Memory Growth per Rank        Min.        Max.        Tot.
==============================================================
Initialization                    3.0M        9.0M        1.1G
  Mesh Preprocessing              0.0K       40.9M      391.9M
Operator Construction            12.0M       33.0M        3.1G
Linear Solve                      0.0K        4.5M      736.5M
  Setup                           1.5M        7.5M     1021.5M
  Preconditioner                  0.0K        1.5M       31.5M
  Coarse Solve                   86.7M      170.2M       25.2G
Eigenvalue Solve                  0.0K        3.0M      225.5M
Div.-Free Projection              6.0M       79.5M        6.0G
Estimation                        0.0K        0.0K        0.0K
  Construction                    0.0K        4.5M      312.0M
  Solve                           0.0K        0.0K        0.0K
Postprocessing                    0.0K        1.5M        4.5M
  Paraview                        0.0K        0.0K        0.0K
  Grid function                   0.0K        0.0K        0.0K
Disk IO                           3.0M       48.0M      621.0M
--------------------------------------------------------------
Total                           244.8M      303.6M       49.4G
```

Each row of this table should be interpreted as "this phase pushed the peak memory by this amount".

This information is also added to the `palace.json`.

I collect both per-rank and per-node information. The above table reports per-rank information when there is only one node, otherwise it reports per-node.

In this, I slightly refactored the `Timer` interface to split the calculations into a `Finalize` function (instead of inside `Print`).

### DIFF
--- a/docs/src/developer/notes.md
+++ b/docs/src/developer/notes.md
@@ -219,8 +219,14 @@ Total                           // < Total simulation time
 
 ## Memory Reporting
 
-Memory reporting in *Palace* tracks peak RSS (Resident Set Size) at two granularities:
-per-process snapshots and per-phase growth.
+Memory reporting in *Palace* tracks **peak** RSS (Resident Set Size) at two granularities:
+per-process snapshots and per-phase growth. Goal of this reporting is to understand the
+maximum memory required by a *Palace* simulation. We do not track how much memory *Palace*
+uses at any given time, instead, we measure how much various phases of the simulation
+increase the high-water mark. This is useful to understand and reduce the total memory
+required by *Palace*. Note, this tool is not enough to comprehensively track the memory
+lifecycle because it does not see allocations/deallocations that do not increase the peak
+RSS.
 
 ### Per-process snapshots
 
@@ -232,7 +238,7 @@ solve.
 
 ### Per-phase memory growth
 
-The `Timer`/`BlockTimer` system tracks not only elapsed time but also peak RSS growth per
+The `Timer`/`BlockTimer` system tracks not only elapsed time but also **peak** RSS growth per
 phase. Every `BlockTimer` scope automatically records how much the peak RSS increased during
 that phase, using the same stack-based interruption mechanism as timing: entering a nested
 scope attributes the memory growth so far to the outer scope, then starts tracking the inner
@@ -244,9 +250,8 @@ across all ranks). On multiple nodes, it shows per-node statistics (sum of ranks
 node) with min/max/total across nodes.
 
 The `BlockTimer` constructor accepts a `count` parameter (default `true`). When `count` is
-`false`, both timing and memory tracking are disabled for that scope. This is used in tight
-loops (e.g., preconditioner application, coarse solve within iterative solvers) to avoid the
-overhead of `getrusage()` system calls.
+`false`, both timing and memory tracking are disabled for that scope. This can be used to
+exclude certain sections of code from a `BlockTimer`.
 
 Per-phase memory data is also saved to `palace.json` under the `PeakMemoryGrowthMegabytes`
 and `PeakNodeMemoryGrowthMegabytes` keys, alongside the existing `ElapsedTime` data.
@@ -256,7 +261,8 @@ and `PeakNodeMemoryGrowthMegabytes` keys, alongside the existing `ElapsedTime` d
 Peak RSS is monotonically non-decreasing (the OS high-water mark), so per-phase deltas are
 always non-negative. A phase that allocates memory temporarily and then frees it will still
 show the growth if the allocation pushed the peak. A phase showing zero growth means it did
-not exceed the previously established peak.
+not exceed the previously established peak. Note that this does not mean that no allocations
+were performed.
 
 Per-phase deltas may not sum exactly to the total because memory growth can occur between
 timed phases (e.g., during scope transitions or in code not wrapped by a `BlockTimer`).
--- a/docs/src/developer/notes.md
+++ b/docs/src/developer/notes.md
@@ -217,6 +217,52 @@ Disk IO                         // < Disk read/write time for loading the mesh f
 Total                           // < Total simulation time
 ```
 
+## Memory Reporting
+
+Memory reporting in *Palace* tracks peak RSS (Resident Set Size) at two granularities:
+per-process snapshots and per-phase growth.
+
+### Per-process snapshots
+
+The `memory_reporting` utilities (`memoryreporting.hpp`) provide functions for querying the
+current and peak RSS of a process. These are aggregated across MPI ranks (min/max/avg) and
+across nodes (by splitting the communicator with `MPI_Comm_split_type`). Two snapshots are
+printed during each simulation: current memory after mesh loading, and peak memory after the
+solve.
+
+### Per-phase memory growth
+
+The `Timer`/`BlockTimer` system tracks not only elapsed time but also peak RSS growth per
+phase. Every `BlockTimer` scope automatically records how much the peak RSS increased during
+that phase, using the same stack-based interruption mechanism as timing: entering a nested
+scope attributes the memory growth so far to the outer scope, then starts tracking the inner
+scope.
+
+The per-phase memory table is printed alongside the timing table at the end of each
+simulation. On a single node, the table shows per-rank statistics with min/max/total (sum
+across all ranks). On multiple nodes, it shows per-node statistics (sum of ranks within each
+node) with min/max/total across nodes.
+
+The `BlockTimer` constructor accepts a `count` parameter (default `true`). When `count` is
+`false`, both timing and memory tracking are disabled for that scope. This is used in tight
+loops (e.g., preconditioner application, coarse solve within iterative solvers) to avoid the
+overhead of `getrusage()` system calls.
+
+Per-phase memory data is also saved to `palace.json` under the `PeakMemoryGrowthMegabytes`
+and `PeakNodeMemoryGrowthMegabytes` keys, alongside the existing `ElapsedTime` data.
+
+### Interpreting memory data
+
+Peak RSS is monotonically non-decreasing (the OS high-water mark), so per-phase deltas are
+always non-negative. A phase that allocates memory temporarily and then frees it will still
+show the growth if the allocation pushed the peak. A phase showing zero growth means it did
+not exceed the previously established peak.
+
+Per-phase deltas may not sum exactly to the total because memory growth can occur between
+timed phases (e.g., during scope transitions or in code not wrapped by a `BlockTimer`).
+
+You should read each row as "this phase increases the peak memory by this amount".
+
 ## Profiling *Palace* on CPUs
 
 A typical *Palace* simulation spends most of its time in libCEED kernels, which, in turn, executed `libsxmm` code on CPUs. Libsxmm generates code just-in-time to ensure it is the most performant on the given architecture and for the given problem. This code generation confuses most profilers. Luckily, [libsxmm](https://libxsmm.readthedocs.io/en/latest/libxsmm_prof/) can integrate with the VTune APIs to enable profiling of jitted functions as well.
--- a/docs/src/developer/notes.md
+++ b/docs/src/developer/notes.md
@@ -245,9 +245,11 @@ scope attributes the memory growth so far to the outer scope, then starts tracki
 scope.
 
 The per-phase memory table is printed alongside the timing table at the end of each
-simulation. On a single node, the table shows per-rank statistics with min/max/total (sum
-across all ranks). On multiple nodes, it shows per-node statistics (sum of ranks within each
-node) with min/max/total across nodes.
+simulation. It has three columns:
+
+  - Per-Node: maximum growth on any single node for this phase.
+  - Total: sum of growth across all nodes for this phase.
+  - Total HWM (High Water Mark): high-water mark growth at the end of this phase summed across all the nodes.
 
 The `BlockTimer` constructor accepts a `count` parameter (default `true`). When `count` is
 `false`, both timing and memory tracking are disabled for that scope. This can be used to
--- a/docs/src/developer/notes.md
+++ b/docs/src/developer/notes.md
@@ -220,7 +220,7 @@ Total                           // < Total simulation time
 ## Memory Reporting
 
 Memory reporting in *Palace* tracks **peak** RSS (Resident Set Size) at two granularities:
-per-process snapshots and per-phase growth. Goal of this reporting is to understand the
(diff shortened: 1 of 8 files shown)


docs/src/developer/notes.md
```diff
@@ -217,6 +217,52 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
+
+The per-phase memory table is printed alongside the timing table at the end of each
+simulation. On a single node, the table shows per-rank statistics with min/max/total (sum
+across all ranks). On multiple nodes, it shows per-node statistics (sum of ranks within each
+node) with min/max/total across nodes.
+
+The `BlockTimer` constructor accepts a `count` parameter (default `true`). When `count` is
+`false`, both timing and memory tracking are disabled for that scope. This is used in tight
+loops (e.g., preconditioner application, coarse solve within iterative solvers) to avoid the
+overhead of `getrusage()` system calls.
```
hughcars: If we're really seeing an effect of `BlockTimer` construction and destruction, that suggests the timer should be hoisted out of whatever loop is in play.

palace/utils/timer.hpp
```diff
@@ -144,93 +183,220 @@ class BlockTimer
...
+  inline static std::vector<double> reduced_time_min;
+  inline static std::vector<double> reduced_time_max;
+  inline static std::vector<double> reduced_time_avg;
+  inline static std::vector<double> reduced_mem_min;
+  inline static std::vector<double> reduced_mem_max;
+  inline static std::vector<double> reduced_mem_sum;
+  inline static std::vector<double> reduced_node_mem_min;
+  inline static std::vector<double> reduced_node_mem_max;
+  inline static std::vector<double> reduced_node_mem_sum;
+  inline static int num_nodes = 0;
```
hughcars, line 203: Using statics here will prevent using BlockTimer in a unit test? I'm not sure we do presently, do you think that would be of value?

docs/src/developer/notes.md
```diff
@@ -217,6 +217,52 @@ Disk IO                         // < Disk read/write time for loading the mesh f
 Total                           // < Total simulation time
 ```
 
+## Memory Reporting
+
+Memory reporting in *Palace* tracks peak RSS (Resident Set Size) at two granularities:
+per-process snapshots and per-phase growth.
```
hughcars: Maybe add a sentence here about how the goal of this reporting is to understand the *maximum* memory required by a job, I think reporting the high water mark is very practically useful, but looking at the tables users might get confused and think that error estimation etc doesn't use any memory maybe. Additionally as developers it doesn't tell us about all wasteful memory usages, (which might also affect speed for example), just the "fattest section". A sentence to that effect here would be informative for someone skimming. I know it's covered below, but that's the most potentially surprising feature, so pays to draw attention to it very briefly before the main section.

docs/src/developer/notes.md
```diff
@@ -217,6 +217,52 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
+
+The `memory_reporting` utilities (`memoryreporting.hpp`) provide functions for querying the
+current and peak RSS of a process. These are aggregated across MPI ranks (min/max/avg) and
+across nodes (by splitting the communicator with `MPI_Comm_split_type`). Two snapshots are
+printed during each simulation: current memory after mesh loading, and peak memory after the
+solve.
+
+### Per-phase memory growth
+
+The `Timer`/`BlockTimer` system tracks not only elapsed time but also peak RSS growth per
```
hughcars:
```suggestion
The `Timer`/`BlockTimer` system tracks not only elapsed time but also *peak* RSS growth per
```
bold or italicize the peak, it's pretty key.

docs/src/developer/notes.md
```diff
@@ -217,6 +217,52 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
+phase. Every `BlockTimer` scope automatically records how much the peak RSS increased during
+that phase, using the same stack-based interruption mechanism as timing: entering a nested
+scope attributes the memory growth so far to the outer scope, then starts tracking the inner
+scope.
```
hughcars, line 245: The timer model is that an indented timing block is a subpiece of the previous indentation, i.e. can be guaranteed to exit scope before it's containing parent scope. This has the effect of leaving the "residual time" in the parent scope. With the memory, how does that work? Say 
```
Block A + *allocate 1 GB, HW 1GB*
  Block AA + *allocate 1GB, HW 2GB*
  Block AA - *deallocate 1GB*
  Block AB + *allocate 10MB, HW 2GB*
  Block AB - *deallocate 10MB*
  Block AA + *allocate 2GB, HW 3GB*
  Block AA - *deallocate 2GB*
Block A - *deallocate 1GB*
```
In this scenario it might be tricky to figure out exactly which AA, but we can punt on this until we encounter it and view it as grounds for refactoring the timer to have greater granularity there (no reason a scoped timer can't be optionally iterable, but that could make the table more difficult to read).

I don't think this is a blocker, but it's something to consider in case a more elegant idea occurs to you, high water mark memory is the most practically useful in terms of sizing to a box. More granular might have value, but that's an extension feature really.

docs/src/developer/notes.md
```diff
@@ -217,6 +217,52 @@ Disk IO                         // < Disk read/write time for loading the mesh f
...
+
+Per-phase memory data is also saved to `palace.json` under the `PeakMemoryGrowthMegabytes`
+and `PeakNodeMemoryGrowthMegabytes` keys, alongside the existing `ElapsedTime` data.
+
+### Interpreting memory data
+
+Peak RSS is monotonically non-decreasing (the OS high-water mark), so per-phase deltas are
+always non-negative. A phase that allocates memory temporarily and then frees it will still
+show the growth if the allocation pushed the peak. A phase showing zero growth means it did
+not exceed the previously established peak.
```
hughcars: Is there a metric of number of allocation requests made? I.e. if the high water never changes, but we request millions of new chunks only to deallocate straight after, that is probably not great performance wise. As above though, I think such a thing is likely an extension not for this PR.

test/unit/test-memoryreporting.cpp
```diff
@@ -80,3 +81,50 @@ TEST_CASE("Node Memory Stats Multi Process", "[memoryreporting][Parallel]")
...
+  {
+    CHECK(BlockTimer::NodeMemoryMin()[i] >= 0.0);
+    CHECK(BlockTimer::NodeMemoryMax()[i] >= BlockTimer::NodeMemoryMin()[i]);
+    CHECK(BlockTimer::NodeMemorySum()[i] >= BlockTimer::NodeMemoryMax()[i]);
+  }
+
+  // Calling Finalize again does not crash (idempotent for AMR loop usage).
+  BlockTimer::Finalize(MPI_COMM_WORLD);
+  CHECK(BlockTimer::IsFinalized());
+}
```
hughcars: No `BlockTimer` scope is ever entered in the test body. The static `BlockTimer::timer.mem_data` is zero-initialised (from the Timer constructor at timer.hpp:97-101). All reduced_* vectors are filled with zeros. Checks:

`NodeMemoryMin()[i] >= 0.0 → 0 >= 0`
`NodeMemoryMax()[i] >= NodeMemoryMin()[i] → 0 >= 0 ✓`
`NodeMemorySum()[i] >= NodeMemoryMax()[i] → 0 >= 0 ✓`

A hypothetical Finalize that simply calls `resize(NUM_TIMINGS)` on the reduced vectors and returns also passes. A hypothetical `BlockTimer` constructor/destructor that never called `MarkMemory` at all also passes.

One possible suggestion:
```cpp
TEST_CASE("BlockTimer scopes attribute memory to the correct phase",
          "[memoryreporting][Serial]")
{
  {
    BlockTimer bt(Timer::CONSTRUCT);
    std::vector<char> buf(8 * 1024 * 1024, 1);
    asm volatile("" :: "g"(buf.data()) : "memory");
  }
  BlockTimer::Finalize(MPI_COMM_WORLD);
  CHECK(BlockTimer::RankMemoryMax()[Timer::CONSTRUCT] > 0);
  CHECK(BlockTimer::RankMemoryMax()[Timer::KSP] == 0);
}
```

palace/utils/timer.hpp
```diff
@@ -144,93 +183,220 @@ class BlockTimer
...
+    const auto &m_sum = (num_nodes == 1) ? reduced_mem_sum : reduced_node_mem_sum;
+    std::string title =
+        (num_nodes == 1) ? "Peak Memory Growth per Rank" : "Peak Memory Growth per Node";
+    PrintTable(comm, title, "Min.", "Max.", "Tot.",
+               [&](int i) -> std::tuple<std::string, std::string, std::string>
+               {
+                 return {memory_reporting::FormatBytes(m_min[i]),
+                         memory_reporting::FormatBytes(m_max[i]),
+                         memory_reporting::FormatBytes(m_sum[i])};
+               });
```
hughcars: Thinking about this, are min/max/sum the most important here? We're looking at the high water mark, so min of a max isn't that informative. The important ones are per node, and total memory. Then I think we could have something interesting if you tracked the peak and the retained, something a bit like
```
                               Δ Peak RSS    Δ Current RSS
=========================================================
Mesh Preprocessing              +40.9M        +40.9M       (kept)
Operator Construction           +33.0M        +28.0M       (mostly kept)
Coarse Solve (workspace)       +170.0M         -2.0M       (temporary)
Postprocessing                   +0.0K         +0.5M       (page-touching)
```
That would tell us if a section allocates and deallocates a big block of memory (maybe we should be reusing that allocation elsewhere?) Not sure if this is possible the polling approach you have, but if it is, that could be pretty interesting to see.

palace/utils/timer.hpp
```diff
@@ -144,93 +183,220 @@ class BlockTimer
...
-      data_min[i] = data_max[i] = data_avg[i] = timer.Data((Timer::Index)i);
-    }
-
-    Mpi::GlobalMin(Timer::NUM_TIMINGS, data_min.data(), comm);
-    Mpi::GlobalMax(Timer::NUM_TIMINGS, data_max.data(), comm);
-    Mpi::GlobalSum(Timer::NUM_TIMINGS, data_avg.data(), comm);
+  // Stored reduction results (populated by Finalize).
+  inline static std::vector<double> reduced_time_min;
+  inline static std::vector<double> reduced_time_max;
+  inline static std::vector<double> reduced_time_avg;
```
hughcars: Possible can organize these for a little less copy-paste, but the LOC gain is minimal (unless you want to add more pieces to track)
```
private:
  struct ReducedSeries { std::vector<double> min, max, sum; };
  struct ReducedAvg    : ReducedSeries { std::vector<double> avg; };

  inline static ReducedAvg    reduced_time;
  inline static ReducedSeries reduced_rank_mem;
  inline static ReducedSeries reduced_node_mem;
  inline static int           num_nodes = 0;

public:
  struct Reductions {
    const ReducedAvg&    time;
    const ReducedSeries& rank_mem;
    const ReducedSeries& node_mem;
    int                  num_nodes;
  };
  static Reductions GetReductions();  // MFEM_VERIFY(IsFinalized()) once, here
```

`SaveMetadata`'s six separate accessor calls at `basesolver.cpp:305-310` collapse to one `const auto r = BlockTimer::GetReductions();`, the Finalize-before-access contract lives at one chokepoint, and the `Timer&` cleanup above can ride in the same diff.

hughcars (commented):
Useful feature, some suggestions, mostly minor things, but one possibly a bit bigger. Given this is for collecting data for decisions, I was trying to think of some extra data that might be useful in future, so we don't have to change this new table in future (and your historical samples are richer going forward).

docs/src/developer/notes.md
Sbozzolo: I misunderstood what `count` was for. I thought it was for performance, but it is to separate accounting, as I understand from here: https://github.com/awslabs/palace/pull/147

I will update the docs.

docs/src/developer/notes.md
Sbozzolo: That would be useful, but I suspect it would require more aggressive instrumentation (either continuous polling or injecting code in mallocs/news).

palace/utils/timer.hpp
Sbozzolo: > Thinking about this, are min/max/sum the most important here?

I added them as a proxy of load-balance. Min/Max essentially give me information about the root process. Sum here is the total per each phase. That's the most important one and the one I look at the most. 

I agree that they are not as useful in most cases, so I can remove them from the table and keep them on the json only. 

I thought about using the Current RSS as well, but it becomes a little harder to use/interpret because we measure Current RSS at a specific point in time (when the Timer is destructed). So, you could allocate a big block of memory and deallocate it within the same scope but before we measure Current RSS and that would be invisible to the Current RSS Timer.

docs/src/developer/notes.md
Sbozzolo: I think that for the purpose of tracking Max RSS, the current implementation will suffice, but I agree we are possibly leaving information on the table.

palace/utils/timer.hpp
Sbozzolo: Good point. I think a simple solution could be to put all the BlockTimer tests in the same TEST_CASE so that we have better control on the static.

test/unit/test-memoryreporting.cpp
Sbozzolo: I forgot, but I've already run into problems writing unit tests for my other memory reporting feature. The test measures peak memory, but if it is run in a process where peak memory is set by some other prior test, the allocation might not push the high-water-mark. We also don't know the order when tests are run. 

I think the simplest solution is to remove the test.

palace/utils/timer.hpp
hughcars: Ah yes I see, it's almost like we'd want a timer whilst it was in existence to poll continuously until its destruction, but that's getting into a more intrusive thing.

docs/src/developer/notes.md
```diff
@@ -219,8 +219,14 @@ Total                           // < Total simulation time
 
 ## Memory Reporting
 
-Memory reporting in *Palace* tracks peak RSS (Resident Set Size) at two granularities:
-per-process snapshots and per-phase growth.
+Memory reporting in *Palace* tracks **peak** RSS (Resident Set Size) at two granularities:
+per-process snapshots and per-phase growth. Goal of this reporting is to understand the
```
hughcars: Nit: `The goal`

hughcars (approved):
I think we should have a follow up to have the inside a phase deltas, but no need to hold this up.

PR #709, opened 2026-04-16
Clean up stale buildcaches

### LINKED ISSUE #524: GHCR cleanup
We currently have three different spack GHCR caches configured:

1. https://github.com/awslabs/palace/pkgs/container/palace-develop - we use this regularly
1. https://github.com/awslabs/palace/pkgs/container/palace - unused
1. https://github.com/awslabs/palace/pkgs/container/palace-v0.23.1 - unused

The last 2 caches can be completely removed as they were configured in testing CI and in order to support multiple spack versions. This isn't something we need / do anymore.

From - https://github.com/awslabs/palace/blob/main/.github/workflows/spack.yml#L127

There isn't really a great / convenient way to deal with cleanup of these caches over time. In other registries like ECR, one can configure auto-deletion after a certain amount of time, but that isn't currently possible in GHCR.

Maybe this would be the best we can do? https://github.com/marketplace/actions/ghcr-io-cleanup-action

From offline discussion with @Sbozzolo, I think a monthly wipe of our cache would be the most reasonable approach.

Why should we clean these up?

1. In theory, we will eventually run out of space in GHCR based on some upper limit in `awslabs`. I don't think that is the case yet, but eventually it would be.
1. It seems like this is slowing down some of the spack pipelines somewhat. See the following two pipelines for comparison:

- [older pipeline](https://github.com/awslabs/palace/actions/runs/16266891339/job/45924539259#step:15:1) w/ only 877 lines of output for this step.
- [more recent pipeline](https://github.com/awslabs/palace/actions/runs/18417971960/job/52486041202#step:14:1) w/ 1977 lines of output for this step.

So, while a once monthly cleanup will slow down the first pipelines we run each month, I think that this would be reasonable.

Alternatively, we could consider moving our GitHub pipelines to use ECR as the cache in the first place, which could then open up the possibility of using the same binaries to run on other AWS services we deploy. We would just have to be a little more careful about GPG keys, but would be reasonable. This would also enable some more advanced security scanning, which I am not sure we quite need / want.

Sbozzolo:
The GHCR is owned by the organization, so I needed to use `crane` to link the build artifacts to the repo. Then, I could delete them. Now, `palace-develop-testing` is linked to the repo and can be cleaned.

In this, I also fixed skipping the container workflow (because its status is required for merge)

Closes #524


.github/workflows/containers.yml
```diff
@@ -45,12 +45,13 @@ jobs:
 
   build-containers:
     needs: filter
-    if: needs.filter.outputs.test == 'true'
```
cameronrutherford: I'm surprised this was both valid syntax, and insufficient as a guard.

.github/workflows/ghcr-cleanup.yml
```diff
@@ -0,0 +1,72 @@
...
+#
+# ## How GHCR package linking works
+#
+# GitHub Packages (GHCR) can be linked to a repository, which grants repo
+# collaborators with write access the ability to manage packages. This link
+# also allows the GITHUB_TOKEN in workflows to delete package versions.
+#
+# The link is established by adding an OCI config label:
+#
+#   org.opencontainers.image.source=https://github.com/awslabs/palace
```
cameronrutherford, line 18: Who adds this label? Is this just by default added to all packages in a GHCR?

.github/workflows/ghcr-cleanup.yml
```diff
@@ -0,0 +1,72 @@
...
+#
+# We used `crane mutate --label` to add this to a tag in the package, which
+# caused GHCR to auto-link the package to this repo. Once linked, the
+# association persists at the GitHub level regardless of individual tag labels.
```
cameronrutherford, line 22: Maybe I am just naive but who is "we" here? Our CI jobs do this?

.github/workflows/ghcr-cleanup.yml
```diff
@@ -0,0 +1,72 @@
...
+#   1. Install crane: https://github.com/google/go-containerregistry
+#   2. Authenticate: echo $TOKEN | crane auth login ghcr.io -u USER --password-stdin
+#   3. Label any tag:
+#        crane mutate ghcr.io/awslabs/palace-develop-testing:index.spack \
+#          --label org.opencontainers.image.source=https://github.com/awslabs/palace
+#   4. Verify the link:
+#        gh api orgs/awslabs/packages/container/palace-develop-testing \
+#          --jq '.repository.full_name'
+#
+# Or as a GitHub Actions step:
```
cameronrutherford, line 36: I am interpreting this as "one needs to do this one-off operation on the `index.spack` in order to delete binaries in the future"?

Which begs the question - why don't we do this then in our GH Actions?

.github/workflows/ghcr-cleanup.yml
```diff
@@ -0,0 +1,72 @@
...
+#   - name: Link GHCR package to repository
+#     run: |
+#       echo "${{ secrets.GITHUB_TOKEN }}" | crane auth login ghcr.io -u ${{ github.actor }} --password-stdin
+#       crane mutate ghcr.io/awslabs/palace-develop-testing:index.spack \
+#         --label org.opencontainers.image.source=https://github.com/${{ github.repository }}
+name: Cleanup GHCR Buildcache
+
+on:
+  schedule:
+    - cron: '0 6,18 * * *'  # Twice daily at 6:00 and 18:00 UTC
```
cameronrutherford, line 52: Maybe just once a week is fine? Curious how you decided on this cadence.

.github/workflows/ghcr-cleanup.yml
```diff
@@ -0,0 +1,72 @@
...
+permissions:
+  packages: write
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      # Each Spack spec is a tagged version in the container package. We keep
+      # the most recent 500 versions (roughly 5-10 full build sets) and delete
+      # older ones. The action deletes at most 100 per run.
```
cameronrutherford, line 64: I would love to be a bit more scientific about this if possible.

I am reading this as it deletes anything past n=500.

That seems potentially too many deletions if we end up expanding the matrix of builds that we want to support. Ideally I would hope that one can just delete based on the age of an installation / the time since an installation was last used.

In other words, it would be ideal to cleanup based on a different criteria:

> Any binary older than a week, that also hasn't been used in the last week

cameronrutherford (commented):
I think the first few comments I left are just naive. I would be curious to hear if you can modify the condition for deletion slightly.

.github/workflows/containers.yml
Sbozzolo: If you add the guard here, the entire workflow is skipped and the status is never posted. If you add it in the individual step, the individual steps are skipped, but the workflow is run, so the status is posted.

.github/workflows/ghcr-cleanup.yml
Sbozzolo: This is an annotation that you should add to your Dockerfile if you want GitHub to link the package to your repo (but that spack does not add).

.github/workflows/ghcr-cleanup.yml
Sbozzolo: Correct. This is a one-off operation, but it it not cheap (if you do it naively). I thought about adding to our actions but I found that processing all the packages we have in our binary cache would take several hours.

We could do something smarter, but I figured that we won't need to do this ever again (unless we change the name of the `palace-develop-testing` package), so I didn't bother

.github/workflows/ghcr-cleanup.yml
Sbozzolo: The action can only remove up to 100 packages every time it runs.

PR #711, opened 2026-04-17
Fix container action using local spack recipe

Sbozzolo:
The PalaceDockerfile was symlinking the local copy of the palace recipe but never adding it to spack, so the container was being built with the published version

Should fix the problem in #706


hughcars (approved):

Sbozzolo:
@hughcars I don't know why the lookahead fails, but the container one passes.

hughcars:
> @hughcars I don't know why the lookahead fails, but the container one passes.

It's the same adapter issue you ran into in the past. The instability continues.

PR #714, opened 2026-04-20
Slim containers

Sbozzolo:
I looked at the container for what else we could remove to make it smaller (headers and Python). Reduces size by 30%. 

With this, I reduced the size by 4x compared to an unmodified spack Dockerfile.


cameronrutherford:
Changes LGTM but I would be relying on CI tests to ensure functionality.

Sbozzolo:
> Changes LGTM but I would be relying on CI tests to ensure functionality.

I added the commits in #713 and found it fixes the CI failure (related to the hybrid/adapter test case) (cc: @hughcars)

So, we can merge this after #713 is merged

hughcars:
You'll want to rebase this now that Armijo is in, then can probably merge.

PR #715, opened 2026-04-21
Reduce memory usage by avoiding eager allocation of H

Sbozzolo:
With the tools I added in PR #708, I am now in the position where I can look at optimizing how memory is used in Palace.

One of the first places I found is with the allocation of H.

Currently, we allocate the full Hessenberg matrix with size max_dim * max_dim, even when only fewer iterations are required. For cases where max_dim (= MaxIts) is large, this can be very significant.

Here, I allocate H incrementally with the GMRES. In my test case, incremental allocation reduced the total memory from 600 GB to 200 GB.

### DIFF
--- a/palace/linalg/iterative.cpp
+++ b/palace/linalg/iterative.cpp
@@ -507,21 +507,36 @@ void GmresSolver<OperType>::Initialize() const
     V[j].SetSize(A->Height());
     V[j].UseDevice(true);
   }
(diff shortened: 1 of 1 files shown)


hughcars (approved):
Excellent demonstration of the memory tooling improving things.

simlapointe (approved):
Good catch!

PR #716, opened 2026-04-27
Auto-gen config docs

phdum-a:
Auto-generate configuration documentation from JSON schema information with supplemented "title" / "description" tags.

- Moves all information about config options out of manually written md files into the JSON schema.
- Julia script `docs/generate_config_docs.jl` reads all schema and generates `docs/src/config/reference.md`, although much of this is just `@raw html`.
- For details see description of schema fields used, see new comment in `developer/notes.md`.
- Includes support for CommonMark inline formatting and cross-references so JSON descriptions can `@ref` to other parts of the documentation.
- Rewrite of config overview (`config/config.md`).

Note: In the JSON schema, this had to replace some of the `enum` testing with `oneOf` + `const`, since that allowed me to add descriptions to the different enum options. This needed some changes in the C++ code to validate correctly.


docs/src/config/reference.md
```diff
@@ -0,0 +1,2033 @@
...
+  <dt><a href="#config-domains">Domains</a></dt>
+  <dd><a href="#config-domains-materials">Materials</a> · <a href="#config-domains-currentdipole">CurrentDipole</a> · <a href="#config-domains-postprocessing">Domain Postprocessing</a></dd>
+  <dt><a href="#config-boundaries">Boundaries</a></dt>
+  <dd><a href="#config-boundaries-pec">PEC Boundary</a> · <a href="#config-boundaries-pmc">PMC Boundary</a> · <a href="#config-boundaries-impedance">Impedance</a> · <a href="#config-boundaries-absorbing">Absorbing Boundary</a> · <a href="#config-boundaries-conductivity">Conductivity</a> · <a href="#config-boundaries-lumpedport">Lumped Port</a> · <a href="#config-boundaries-waveport">Wave Port</a> · <a href="#config-boundaries-waveportpec">Wave Port PEC</a> · <a href="#config-boundaries-surfacecurrent">Surface Current</a> · <a href="#config-boundaries-ground">Ground Boundary</a> · <a href="#config-boundaries-zerocharge">Zero Charge Boundary</a> · <a href="#config-boundaries-terminal">Terminal</a> · <a href="#config-boundaries-periodic">Periodic Boundary</a> · <a href="#config-boundaries-postprocessing">Boundary Postprocessing</a></dd>
```
hughcars: This is going to be pretty difficult for us to edit directly in future, I wonder if we can have some kind of conversion done from within the html/css files themselves, rather than requiring us to store the html from within the markdown like this. It being generated makes it a bit easier, but I wonder if a small change later is going to result in a huge diff on this file as a consequence. Do you think this can be generated in such a way that it doesn't risk large perturbations on each regenration? This might be impractical for html though. I suspect it is, but that complexity will not be worth it, but just incase.

It would also be preferable to avoid this transformation to a monolithic file, I wonder if it's possible to have separate files that get injected together/maintain the prior separate pages but now with the new link capability and the much nicer formatting.

docs/src/config/reference.md
```diff
@@ -0,0 +1,2033 @@
+```@raw html
```
hughcars: One problem is that the rendering doesn't respect the light/dark mode but that should be relatively easy to fix (the default value here is unreadable for instance).

Image

Similarly for some nested text

Image

docs/src/developer/notes.md
```diff
@@ -16,7 +16,7 @@ given *Palace* simulation.
 
 First, the mesh comes with its own unit of length. Users specify the conversion
 between one unit of length on the mesh to meters through the
-[`config["Model"]["L0"]`](../config/problem.md#config%5B%Model%22%5D) parameter.
+[`config["Model"]["L0"]`](../config/reference.md#config-model) parameter.
```
hughcars, line 19: Linking like this is excellent! Great quality of life improvement for doc writing and also for readers of the docs.

docs/src/config/reference.md
hughcars, line 1: Given this is effectively becoming a generated file (reading the julia below), this should be marked as non-human editable, or maybe even have some kind of autogeneration on build? There's a possibility of it basically always reflecting the current state of the local schema, but maybe that's unwieldy?

palace/utils/jsonschema.cpp
```diff
@@ -267,6 +267,26 @@ json FindEnumInSchema(const json &schema, const std::string &ptr)
...
     }
   }
+  // oneOf is used in two ways: the "usual" way to constrain types and as an emum
+  // replacement using the oneOf+const pattern, which allows us to add documentation fields
+  // to enum entries. For oneOf+const we want to collect all values into an array for
+  // consistent error formatting.
+  if (current.contains("oneOf"))
+  {
+    const auto &branches = current["oneOf"];
+    // Check if oneOf + const enum: all items in oneOf have "cosnt" and no "properties"
```
hughcars: Nit: `const`

scripts/schema/config/problem.json
```diff
@@ -9,18 +10,48 @@
     "Type":
     {
       "type": "string",
-      "enum": ["Eigenmode", "Driven", "Transient", "Electrostatic", "Magnetostatic"]
+      "description": "Controls the simulation type.",
+      "oneOf": [
+        { "const": "Eigenmode",     "description": "Perform an undamped or damped eigenfrequency analysis." },
```
hughcars: undamped or damped? That's a strange distinction to be bringing up here?

hughcars (changes requested):
This looks great! I think it's probably worth merging/rebasing after https://github.com/awslabs/palace/pull/719 as there's interplay between the two, and there's a few minor UI things (dark mode for example seems to be a bit broken), but huge QoL improvement.

docs/src/config/reference.md
phdum-a: The idea is that this `reference.md` is never manually edited — just auto-generated by `generate_config_docs.jl`. I think it is a matter of taste if you want to commit auto-generated files at all or always make them on the fly before running Documenter.jl.

The amount of custom html injection is not ideal. The system mixed Documenter needs (like that the page is md to begin with and that sub-tiltes are in markdown, so they Documenter can check them as part of their build) and html needs. I would have preferred to use more Documenter built-ins and styling. I found that doing links to every item in Documenter messy. And I could not get nice compact stye — so ended up with this custom solution.

docs/src/config/reference.md
phdum-a: (see above)

scripts/schema/config/problem.json
phdum-a: This is just a copy of the current description here (https://awslabs.github.io/palace/dev/config/problem/#config[%22Problem%22]) but agree that it is silly. Changed.

palace/utils/jsonschema.cpp
phdum-a: fixed

docs/src/developer/notes.md
phdum-a: Yeah, this is just defined by `anchor_id` with is just `config-` and then the json path joined by `-`. No more special characters `[]` inferred from the title.

docs/src/config/reference.md
phdum-a: Made some style adjustments to fix this. Now uses colors and styling following Documeter admonitions, which should look ok for both dark and light themes.

phdum-a:
(Needs rebasing and conflict resolution once #719 is merged).

scripts/schema/config/solver.json
```diff
@@ -1,257 +1,954 @@
...
+          "exclusiveMinimum": 0,
+          "default": 0
+        },
+        "InitialGuess":
+        {
+          "type": "boolean",
+          "description": "Use the previous solution as an initial guess for the iterative solver.",
+          "default": true,
+          "x-palace-advanced": true
+        },
```
Sbozzolo, line 890: The default encoded in the json does not match the resolved value in some cases:
```
| Field | Schema default | Resolves differently in C++ for | Schema | C++ |
|---|---|---|---|---|
| `Linear/InitialGuess` | `true` | false for eigenmode & adaptive-driven | `solver.json:724` | `iodata.cpp:447-458` |
| `Linear/MGAuxiliarySmoother` (`mg_smooth_aux`) | `true` | false for electro/magnetostatic | `solver.json:751` | `iodata.cpp:493-504` |
| `Linear/PCMatShifted` | `false` | true for driven+AMS, boundary-mode+MG | `solver.json:808` | `iodata.cpp:473-488` |
| `Linear/AMSSingularOperator` | `false` | true for magnetostatic | `solver.json:891` | `iodata.cpp:511-513` |
| `Linear/AMGAggressiveCoarsening` | `false` | true for electro/magnetostatic/transient | `solver.json:898` | `iodata.cpp:515-519` |
```

docs/generate_config_docs.jl
```diff
@@ -0,0 +1,829 @@
...
+                end
+            end
+        end
+
+        # Collect child sub-schemas that become sub-sections
+        sub_schemas = Pair{String, AbstractDict}[]
+        fis_array = false
+
+        t = get(field, "type", "")
+        if t == "object" && haskey(field, "properties")
```
Sbozzolo, line 370: This script silently skips `enum` json types, so that `PortDirection` and `DipoleDirection` do not have an allowed list.

docs/test/test_generate_config_docs.jl
Sbozzolo, line 1: I don't think this test is being run?

docs/Project.toml
```diff
@@ -1,4 +1,7 @@
 [deps]
+CommonMark = "a80b9123-70ca-4bc0-993e-6e3bcb318db6"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
+JSONSchema = "7d188eb4-7ad8-530c-ae41-71a32a6d4692"
```
Sbozzolo: This package is not used

scripts/schema/config/solver.json
```diff
@@ -1,257 +1,954 @@
...
+        "Save":
+        {
+          "type": "integer",
+          "description": "Number of computed field modes to save to disk for [visualization with ParaView](../guide/postprocessing.md#Visualization). Files are saved in the `paraview/` (and/or `gridfunction/`) directory under [`/Problem/Output`](@ref config-problem-output).",
+          "default": 0
+        },
+        "Type":
+        {
+          "type": "string",
```
Sbozzolo: I think we should try to avoid plain strings when we can provide a list of allowed choices.

Sbozzolo (changes requested):
Great addition!

Now that we have #719, I suggest also adding a test where we check that the resolved config is compatible with the schema to reduce the risk of drifts (until we do something like #723)

docs/src/config/reference.md
hughcars: Addressed in e5554607c. `reference.md` is generated/ignored; schemas and `docs/generate_config_docs.jl` are the editable source.

scripts/schema/config/solver.json
hughcars: Addressed in ee06742cb.

docs/generate_config_docs.jl
hughcars: Addressed in ee06742cb.

docs/test/test_generate_config_docs.jl
hughcars: Addressed in ee06742cb.

docs/Project.toml
hughcars: Addressed in ee06742cb.

scripts/schema/config/solver.json
hughcars: Addressed in ee06742cb.

Sbozzolo (approved):

hughcars (approved):

PR #717, opened 2026-04-29
Add cuDSS support via mfem::CuDSSSolver

Pennycook:
[cuDSS](https://docs.nvidia.com/cuda/cudss/index.html) is a high-performance CUDA library for Direct Sparse Solvers.

This commit enables the use of cuDSS in Palace by:
- Wrapping `mfem::CuDSSSolver` in a new `palace::CuDSSSolver`;
- Enabling the use of `CuDSSSolver` for KSP and WavePort; and
- Adding cuDSS configuration options to CMake, for both Palace and MFEM.

I've opened this as a draft for now because it depends on https://github.com/mfem/mfem/pull/5124, which has not yet been merged.

An example of building with cuDSS:
```
cmake .. -DPALACE_WITH_CUDA=ON -DPALACE_WITH_CUDSS=ON -DCUDSS_DIR="/path/to/cudss/"
```


Pennycook:
I fixed the formatting, but I'm not sure what to do about the other failures.

The license check is failing because the test assumes the header will match exactly, but I added a new copyright line to two files. One possible solution would be to rewrite the test to check for the copyright notice and SPDX license identifier separately, but I think that's outside the scope of this PR.

I think the other tests are failing because they're using the Spack route of installation, which I didn't update. It's not obvious to me that there's a way to make this work, besides waiting for the mfem Spack package to expose cuDSS support.

cmake/ExternalGitTags.cmake
```diff
@@ -125,15 +125,15 @@ set(EXTERN_METIS_GIT_TAG
...
 )
 set(EXTERN_MFEM_GIT_BRANCH
-  "master" CACHE STRING
+  "cudss-dev" CACHE STRING
   "Git branch for external MFEM build"
 )
 set(EXTERN_MFEM_GIT_TAG
-  "d9d6526cc1749980a2ba1da16e2c1ca1e07d82ec" CACHE STRING
+  "e217864f168acb1d7bdb3fd7c5aa73ef81951237" CACHE STRING
   "Git tag for external MFEM build"
```
hughcars, line 137: For a POC this will work, but we have a facility for assembling patches for MFEM into palace. If we don't want to wait for the minor version bump in mfem (no idea when that'll be), the facility would be to assemble the patch for just the mfem PR, plucked onto our "palace head" state of mfem, which is 4.9 + the patch stack. This is a bit involved, so if you want I can do this for you (it's best achieved using spack nowadays).

hughcars (commented):
We'll take this for a spin, one small comment on the patching process but generally looks good.

cmake/ExternalGitTags.cmake
Pennycook: Yes, please, that would be very helpful. I only intended that commit to be temporary.

It probably still makes sense to wait for things to be merged into mfem as they are, before you assemble the patch -- I don't expect that the PR will change before it's merged, but I can't rule it out either.

hughcars (changes requested):
Hi @Pennycook, 

I took a stab at cleaning things up as mentioned in the comments and there's a draft PR onto your fork at https://github.com/Pennycook/palace/pull/1. Most of that is noise from merging origin/main into your branch. If you merge into your fork it'll reduce down to a much smaller diff. The main thing was assembling the diffs.

Pennycook:
> I took a stab at cleaning things up as mentioned in the comments and there's a draft PR onto your fork at [Pennycook#1](https://github.com/Pennycook/palace/pull/1). Most of that is noise from merging origin/main into your branch. If you merge into your fork it'll reduce down to a much smaller diff. The main thing was assembling the diffs.

Thanks again for this, @hughcars. I rebased my PR on top of origin/main and then cherry-picked these changes.

I'm seeing some runtime failures locally that I didn't see before. Would you mind taking a look to see if the merged code looks as you expected? There were a few changes that didn't show up as merge conflicts (e.g., the JSON processing moved) that I had to fix, and so I might have missed something.

hughcars:
Took cuDSS for a spin on GPU and benchmarked it against SuperLU_DIST. Short version: it's a solid win and the advantage grows with problem size.

Setup
- p4d.24xlarge, 8× A100-40GB, -np 8, "Device": "GPU"
- Built palace@develop +cuda +cudss via spack (^cuda@13.1, cuDSS 0.7.1, OpenMPI 5.0)
- Case: examples/cpw/cpw_lumped_uniform (driven sweep, 7 frequencies/level, Order: 2) with Model.Refinement.MaxIts AMR to grow the mesh; field output disabled to isolate solve time
- Only Solver.Linear.Type changed between runs (cuDSS vs SuperLU). Both drove AMR to the same mesh at every level and produced identical S-parameters (~11 digits).
- cuDSS/SuperLU run as the geometric-multigrid coarse solver, i.e. they factorize the p1 coarse level; the fine (p2) space is the full problem size.

Results

```
┌─────────────────────────┬────────────────────────┬──────────────────┬──────────────┬────────────┬───────────────┐
│                         │ p1 coarse (factorized) │ p2 fine (global) │ Coarse Solve │ Total wall │ Peak mem/rank │
├─────────────────────────┼────────────────────────┼──────────────────┼──────────────┼────────────┼───────────────┤
│ cuDSS (MaxIts=2)        │ 58,714                 │ 330,414          │ 16.5 s       │ 162.7 s    │ —             │
├─────────────────────────┼────────────────────────┼──────────────────┼──────────────┼────────────┼───────────────┤
│ SuperLU_DIST (MaxIts=2) │ 58,714                 │ 330,414          │ 48.9 s       │ 191.9 s    │ —             │
├─────────────────────────┼────────────────────────┼──────────────────┼──────────────┼────────────┼───────────────┤
│ cuDSS (MaxIts=5)        │ 869,228                │ 5,254,370        │ 98.3 s       │ 889.7 s    │ 2.8 GB        │
├─────────────────────────┼────────────────────────┼──────────────────┼──────────────┼────────────┼───────────────┤
│ SuperLU_DIST (MaxIts=5) │ 869,228                │ 5,254,370        │ 883.7 s      │ 1648.3 s   │ 4.3 GB        │
└─────────────────────────┴────────────────────────┴──────────────────┴──────────────┴────────────┴───────────────┘
```

So ~3× faster on the coarse (direct) solve at the smaller size, growing to ~9× at the 869k-DOF coarse grid (5.25M-DOF fine problem), which is ~18% → ~46% off total wall time. cuDSS also appears to use less memory (2.8 vs 4.3 GB/rank), and both MaxIts=5 runs completed with no OOM.

palace/linalg/cudss.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "cudss.hpp"
+
+#if defined(MFEM_USE_CUDSS)
+
+#include "utils/iodata.hpp"
```
hughcars: This file only uses `MatrixSymmetry`/`SymbolicFactorization` (defined in `utils/labels.hpp`) and nothing from `iodata.hpp` — the feedback commit switches this include to `utils/labels.hpp`, the actual definer, for include-what-you-use.

palace/linalg/cudss.cpp
```diff
@@ -0,0 +1,48 @@
...
+      return mfem::CuDSSSolver::SYMMETRIC_INDEFINITE;
+    case MatrixSymmetry::UNSYMMETRIC:
+      return mfem::CuDSSSolver::NONSYMMETRIC;
+  }
+  return mfem::CuDSSSolver::NONSYMMETRIC;
+}
+
+}  // namespace
+
+CuDSSSolver::CuDSSSolver(MPI_Comm comm, MatrixSymmetry sym, SymbolicFactorization reorder,
```
hughcars: cuDSS exposes only reordering *reuse*, not a strategy selection, so the `SymbolicFactorization reorder` arg is unused — the feedback commit marks it `/*reorder*/` with a one-line note, kept only for parity with the other direct-solver wrappers.

palace/linalg/cudss.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <mfem.hpp>
+
+#if defined(MFEM_USE_CUDSS)
+
+#include "utils/labels.hpp"
+
+namespace palace
+{
+
+class IoData;
```
hughcars: Unused `IoData` forward-decl — the feedback commit removes it.

palace/models/waveportoperator.cpp
```diff
@@ -8,6 +8,7 @@
 #include "linalg/ams.hpp"
 #include "linalg/arpack.hpp"
 #include "linalg/blockprecond.hpp"
+#include "linalg/cudss.hpp"
```
hughcars: `linalg/cudss.hpp` isn't used in this file — the feedback commit removes it.

palace/linalg/ksp.cpp
```diff
@@ -187,6 +191,16 @@ ConfigurePreconditionerSolver(const config::LinearSolverData &linear,
     case LinearSolver::JACOBI:
       pc = std::make_unique<JacobiSmoother<OperType>>(comm);
       break;
+    case LinearSolver::CUDSS:
+#if defined(MFEM_USE_CUDSS)
+      pc = MakeWrapperSolver<OperType, CuDSSSolver>(linear, comm, pc_mat_sym,
+                                                    linear.sym_factorization,
+                                                    linear.reorder_reuse, print);
+#else
```
hughcars: The feedback commit drops this `#else MFEM_ABORT` so the case matches the empty-guarded MUMPS one; the single "not built with cuDSS" check moves to an `MFEM_VERIFY` in `iodata.cpp` alongside the SuperLU/STRUMPACK/MUMPS guards.

scripts/schema/config-schema.json
```diff
@@ -2034,6 +2034,10 @@
         {
           "const": "Jacobi",
           "description": "Use the Jacobi preconditioner."
+        },
+        {
+          "const": "cuDSS",
```
hughcars, line 2133: New allowed enum value ⇒ backward-compatible SchemaVer REVISION. `main` is now at `1-1-0` (RationalImpedance, #770), so the feedback commit bumps the schema `$id` `1-1-0` → `1-2-0` and tags the CHANGELOG entry `SchemaVer 1-2-0` (`check-schema-version` enforces this). You might need to bump it further if something else gets in first.

hughcars (commented):
Took it for a spin, worked great! There's an issue with building with mpich in mfem, but that will fail loudly and I have a patch on its way to mfem, so i think we can merge. I have one final commit on hughcars/pr717-cudss-followup with some small fix ups. If you cherry-pick that, rebase, and maybe bump the SchemaVer again (if another bump snuck in underneath), then we're good to go. Thank you for the submission! This'll be a meaningful improvement in nvidia gpu performance.

Pennycook:
Thanks, @hughcars. I've cherry-picked that commit and rebased on main, so things should be ready for review/merge now.

One thing to note: I created this PR with "allow edits by maintainers", so if you do need to make any small changes prior to merge then feel free to push them directly to this branch. If you'd prefer to stage changes via my fork as we have been doing that's fine too -- I only mention it because it took me a while to discover this GitHub feature, and I've found it quite useful.

CHANGELOG.md
```diff
@@ -34,6 +34,30 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
     to any passive lumped-network response. SchemaVer 1-1-0
     [PR 770](https://github.com/awslabs/palace/pull/770).
 
+## In progress
```
hughcars: Looks like the rebase duplicated the whole `## In progress` section here (including the PR #804 bug-fix entry and the PR #770 RationalImpedance entry above the cuDSS entry). Please collapse this back to a single `## In progress` section and keep the cuDSS bullet under the existing `#### New Features`.

hughcars (commented):
Minor issue in the CHANGELOG, but modulo that, approved!

hughcars (approved):

CHANGELOG.md
Pennycook: Oops, sorry about that! Thanks for fixing it.

hughcars: Np. We've got a merge pileup right now, so I'll probably have to touch the CHANGELOG again, but from your perspective it's done!

hughcars:
Override merging as remaining CI issues were related to fork PRs not being allowed to push containers, unrelated to this PR.

PR #719, opened 2026-04-30
Concretize IoData defaults and write out resolved config

hughcars:
Palace now drops `config.json` alongside `palace.json` in the output directory. Every implicit default and `"Default"` sentinel is filled in so the file alone is enough to reproduce the run.

`IoData::ConcretizeDefaults` takes the parsed config by value, fills keys the user left blank (and replaces `"Default"` sentinel strings) from the resolved `IoData`, and returns the result for `main.cpp` to stream to disk. User-provided entries pass through untouched.

A schema-driven coverage gate (`SchemaCoverageGaps` in `test/unit/test-config.cpp`) walks every section of the JSON schema and asserts that every optional property with a C++ default is emitted, modulo a small justified skip-list.

Supporting cleanup:

- `EigenSolverBackend::DEFAULT` and `TimeSteppingScheme::DEFAULT` are now type-level aliases for their build-determined / canonical concrete value, so a default-constructed field is already at its concrete value and no runtime sentinel resolution is needed.
- `Solver.Eigenmode.MaxIts` and `MaxSize` defaults are resolved inside the struct's JSON constructor: `MaxIts = 1'000'000` (iteration cap only, no buffer), `MaxSize = max(2*n, n+15)` (SLEPc's Krylov-Schur formula; previously ARPACK silently used `max(20, 2*n+1)`). Same treatment for `Solver.BoundaryMode.MaxSize` and per-port `Boundaries.WavePort[].MaxSize` — these previously fed `-1` into the SLEPc/ARPACK wrappers and segfaulted any caller (including the boundary-mode unit tests) that didn't trip a defensive fallback that no longer exists. All in-tree examples set these explicitly.
- Regression harness asserts the `config.json` sidecar is produced for every example.


palace/utils/iodata_concretize.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+  if (config.contains("Boundaries"))
+  {
+    ConcretizeBoundaries(iodata.boundaries, config["Boundaries"]);
+  }
+
+  return config;
+}
+
+void IoData::WriteResolvedConfig(const json &raw_config) const
```
Sbozzolo: This will break the user's config is if `<problem.output> = "."` and the user's config is named `config.json`.

We might want to check if a file already exists before writing.

palace/utils/configfile.cpp
```diff
@@ -626,6 +632,12 @@ WavePortData::WavePortData(const json &port)
     }
   }
   n_samples = port.value("NSamples", n_samples);
+
+  // Resolve subspace sentinel: fed directly into ModeEigenSolver → SLEPc/ARPACK.
+  if (max_size <= 0)
+  {
+    max_size = std::max(2 * mode_idx, mode_idx + 15);
```
simlapointe: Previously we were using `max(20, 2 * mode_idx + 1)`, which is the SLEPc default, correct? I know the differences should be minor, but is there a particular reason to change to `max(2 * mode_idx, mode_idx + 15)`?

palace/utils/configfile.cpp
```diff
@@ -100,12 +104,14 @@ PALACE_JSON_SERIALIZE_ENUM(FrequencySampling, {{FrequencySampling::DEFAULT, "Def
                                                {FrequencySampling::POINT, "Point"}})
 
 // Helper for converting string keys to enum for TimeSteppingScheme and Excitation.
+// DEFAULT is an alias for GEN_ALPHA; placing it last ensures to_json emits
+// "GeneralizedAlpha" rather than "Default" while still accepting "Default" as input.
```
Sbozzolo: I think this is a really great feature. 

I tested it with some of our examples and found that there are still some "Defaults" left:
```
"ColumnOrdering": "Default",
"PCSide": "Default",
```

test/examples/testcase.jl
```diff
@@ -176,7 +176,7 @@ function testcase(
         @test alldirs == expected_dirs || (@show alldirs, expected_dirs; false)
         @test sort(csvfiles) == sort(csvfilesref) || (@show csvfiles, csvfilesref; false)
 
-        expected_metafiles = ["palace.json"]
+        expected_metafiles = ["palace.json", "config.json"]
         for i = 1:max_its
             push!(expected_metafiles, "iteration$(i)/palace.json")
         end
```
Sbozzolo, line 189: this should probably also take into account the resolved config file (or we should avoid copying the resolved config file in each iteration)

palace/linalg/arpack.cpp
```diff
@@ -220,7 +220,7 @@ void ArpackEigenvalueSolver::SetNumModes(int num_eig, int num_vec)
     V.reset();
   }
   nev = num_eig;
-  ncv = (num_vec > 0) ? num_vec : std::max(20, 2 * nev + 1);  // Default from SLEPc
+  ncv = num_vec;
```
Sbozzolo, line 223: Just noting that this changes the default when `num_vec <= 4`

palace/drivers/eigensolver.cpp
```diff
@@ -84,32 +84,16 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
-  if (type == EigenSolverBackend::SLEPC)
-  {
-    Mpi::Warning("SLEPc eigensolver not available, using ARPACK!\n");
-  }
-  type = EigenSolverBackend::ARPACK;
+  const EigenSolverBackend type = iodata.solver.eigenmode.type;
+#if !defined(PALACE_WITH_ARPACK) && !defined(PALACE_WITH_SLEPC)
+#error "Eigenmode solver requires building with ARPACK or SLEPc!"
+#endif
+#if !defined(PALACE_WITH_SLEPC)
```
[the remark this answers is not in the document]
Sbozzolo: This is also a different behavior from SLP, where the type is automatically downgraded

CHANGELOG.md
```diff
@@ -23,6 +23,16 @@ The format of this changelog is based on
   - Enabled voltage and impedance postprocessing for waveport and BoundaryMode by specifying a line
     from ground to conductor, either through mesh coordinates or boundary attribute
     in `config["Boundaries"]["Postprocessing"]["Impedance"]` [PR 657](https://github.com/awslabs/palace/pull/657).
+  - Palace now writes a fully-resolved copy of the run configuration to
+    `<output_dir>/config.json` at startup, with every implicit default and `"Default"`
+    sentinel filled in. The sidecar passes schema validation and can be used to re-run
+    the same simulation deterministically [PR 719](https://github.com/awslabs/palace/pull/719).
```
Sbozzolo: We should document this in the docs as well.

palace/utils/configfile.cpp
```diff
@@ -1263,6 +1275,19 @@ EigenSolverData::EigenSolverData(const json &eigenmode)
...
+  // downstream sees -1. max_size follows SLEPc's Krylov-Schur formula (applies to both
+  // ARPACK and SLEPc wrappers); max_it is a single large cap because it only bounds
+  // iteration count, with no buffers scaling with it.
+  if (max_it <= 0)
+  {
+    max_it = 1'000'000;
+  }
+  if (max_size <= 0)
+  {
+    max_size = std::max(2 * n, n + 15);
```
Sbozzolo: Maybe we can add a function for this default, since we set it in multiple places?

test/unit/test-config.cpp
```diff
@@ -637,3 +695,677 @@ TEST_CASE("ParseStringAsDirection", "[config][Serial]")
...
+                   {"Boundaries", json::object()},
+                   {"Solver", json::object()}};
+
+    IoData iodata(config, false);
+    CHECK(iodata.solver.linear.type == LinearSolver::AMS);
+    CHECK(iodata.solver.linear.krylov_solver == KrylovSolver::CG);
+    CHECK(iodata.solver.linear.ams_singular_op == 1);
+    CHECK(iodata.solver.linear.ams_max_it == 1);
+    CHECK(iodata.solver.linear.mg_cycle_it == 1);
+    CHECK(iodata.solver.linear.pc_mat_sym == MatrixSymmetry::SPD);
```
Sbozzolo: `MatrixSymmetry` is an object implicitly defined and not a config. I don't think it should be in the `solver.linear` struct (which would otherwise represent the JSON faithfully).

palace/utils/enum_string.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <iosfwd>
+#include "labels.hpp"
+
+// Declarations for the operator<< overloads generated by PALACE_JSON_SERIALIZE_ENUM in
+// configfile.cpp. Including this header is sufficient to stream-print any Palace enum that
+// has a JSON serialization mapping — no need to forward-declare individual operators.
+
+namespace palace
+{
+
```
Sbozzolo, line 18: I think we could remove this entire file by inlining the operator in `PALACE_JSON_SERIALIZE_ENUM`

palace/utils/iodata_concretize.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+void ApplyEntries(json &j, const std::vector<Entry> &entries)
+{
+  for (const auto &[key, value] : entries)
+  {
+    Concretize(j, key, value);
+  }
+}
+
+void ConcretizeProblem(const config::ProblemData &problem, json &j_problem)
```
Sbozzolo, line 48: In the future, this entire file could (should?) probably be generated from the schema in the spirit of #723

simlapointe (changes requested):
Overall LGTM. Tested it on many cases and it seems to work as expected. 

A few small things:
- This may be intended, but I find the AMR behavior slightly weird: `config.json` gets written in the `iteration1` folder (but not subsequent iterations) and a symlink to it is created in the top-level output folder. Maybe having the actual file in the top-level output folder and not in any of the `iteration*` folders makes more sense?
- Maybe unlikely, but if the input config is `config.json` and the user sets `"Output": "."`, then the resolved config overwrites the input config. Should we guard against this?
- It would be nice if the resolved config would get printed when running `--dry-run`, but maybe that's asking for too much?
- See comment below about the new `max_size`/`num_vec` default.

Sbozzolo (changes requested):
Thank you, really useful feature! I left some comments, but nothing very big. I tested with a bunch of our examples in the repo.

I am a little concerned with the amount of information that is represented in various different ways (from JSON to C++ to concretizer...), and I think this further motivates generating this code from a single source (not in this PR).

palace/utils/iodata_concretize.cpp
hughcars: Fixed in 34666127a.

test/examples/testcase.jl
hughcars: Fixed in 34666127a — resolved config is now a single top-level `<input>_resolved.json`, no per-iteration copies, and the harness checks for it there.

CHANGELOG.md
hughcars: Documented in 34666127a (new "Resolved configuration record" section in `docs/src/guide/postprocessing.md`, plus expanded `PCSide`/`ColumnOrdering` notes in `solver.md`).

palace/utils/configfile.cpp
hughcars: Fixed in 34666127a — extracted `DefaultEigenSubspaceSize(num_modes)` and routed all three call sites through it.

test/unit/test-config.cpp
hughcars: Fixed in 032a09162 — removed from `LinearSolverData` and derived at construction via `GetPreconditionerMatrixSymmetry(iodata)` in `ksp.cpp`.

palace/utils/configfile.cpp
hughcars: These are intentionally left as `"Default"` — Palace delegates the choice to the downstream library (GMRES/FGMRES selects the PC side; the sparse direct solver selects the ordering), so there is no concrete value for Palace to emit. Documented as such in `solver.md` and `postprocessing.md` (34666127a).

palace/linalg/arpack.cpp
hughcars: Acknowledged — see the explanation on the `configfile.cpp` `MaxSize` thread. Both backends now resolve an unset subspace to SLEPc's actual default `max(2*nev, nev+15)`; the change is deliberate and documented.

palace/utils/iodata_concretize.cpp
hughcars: Agreed, and #723 is the right vehicle. This hand-written concretizer plus the schema coverage gate is the stopgap until then — I'd keep codegen out of this PR's scope. The coverage test fails loudly if a schema property ever drifts out of sync in the meantime.

palace/utils/enum_string.hpp
hughcars: I kept the file but addressed the underlying duplication. There is now one `ToString`/`FromString` mapping table per enum, and both the JSON `to_json`/`from_json` shims and the resolved-config writer delegate to it, so there is a single source of truth either way. I left a small declaration header rather than inlining into `PALACE_JSON_SERIALIZE_ENUM` for two reasons: the resolved-config writer needs `ToString` without pulling in the full nlohmann serialization machinery, and keeping the tables in a `.cpp` avoids re-expanding them at every macro site. (d22986c17)

palace/utils/configfile.cpp
hughcars: The old line was commented "Default from SLEPc," but that's a mislabel — `max(20, 2*nev+1)` isn't SLEPc's default; `max(2*nev, nev+15)` is. The two backends were silently resolving an unset `MaxSize` differently, so this unifies them on SLEPc's actual formula and makes the resolved config backend-independent. The cost is a slightly smaller default for small `nev` (e.g. `nev=4`: 20→19, `nev=1`: 20→16), though these stay well above `nev` and all in-tree examples set `MaxSize` explicitly, so nothing in the repo changes.

hughcars:
Current spack CI failures will be fixed once rebased on top of #742 once that is merged.

simlapointe (approved):

Sbozzolo (approved):

PR #723, opened 2026-05-05
[WIP] JSON schema generation

ad-cqc:
Collapses Palace's three disconnected schema sources: the C++ structs, the hand-written JSON schemas under [scripts/schema/](https://github.com/awslabs/palace/tree/main/scripts/schema), and PR #716's annotated schema into one C++ source of truth. The new `palace_emit_schema` binary generates `config-schema.json` at build time from `reflect-cpp-annotated` mirrors of the config structs; the embed pipeline picks up the generated file and bakes it into `libpalace` via the existing `embed_schema.cmake` helper. The hand-written schema files are deleted.

Descriptions in the annotated types are transcribed verbatim from [awslabs/palace#716](https://github.com/awslabs/palace/pull/716), so the doc-generator output matches PR #716 by construction. CMAKE_CXX_STANDARD bumps to C++20 (`reflect-cpp` floor).

## Test plan
- [x] C++20`libpalace` + `palace` build clean.
- [x] `palace_emit_schema` produces a 2655-line schema; embedded header regenerates via the CMake custom command.
- [x] Spot checks against PR #716 descriptions pass verbatim.
- [x] 53 advanced / 4 deprecated flags land on the correct properties.
- [x]  `test/unit/test-schema.cpp` (sample configs validate against the embedded schema).
- [ ]  ` scripts/schema/ValidateConfig.jl` against the build-generated path.
- [ ]  PR 716's `docs/generate_config_docs.jl` against the generated schema. The expected output should match PR 716's committed reference.md modulo trivial ordering.


hughcars:
I don't think we will be going with this approach, as at present after our efforts to close the gap between schemas, and with the concretized default, it has less of a target to address. This combined with the fairly large increase in complexity to handle it, means I don't think this is the approach at present. We are considering some alternatives to try and close this testing gap in future.

PR #725, opened 2026-05-06
Fix build only dependencies in doc CI

hughcars:


Sbozzolo:
I think I got something wrong with skipping the container CI tasks when they should not run
image

Sbozzolo (approved):

Sbozzolo:
Ah, I thought I fixed the problem in the same way, but we have not merged it

https://github.com/awslabs/palace/pull/709/changes/789f36c02f2eea4386cc9a19561501d322a2952e

PR #726, opened 2026-05-07
schema: hoist inline definitions into named $defs for codegen

ad-cqc:
## Motivation

The current schema is correct for *validation*, but its pervasive use of inline definitions makes it hostile to *code generation* (Python, Julia, or any target that turns a JSON Schema into a set of named types in a single module):

- **Inline object schemas inherit the surrounding key name.** Two unrelated inline objects under different parents but with the same key (e.g. `Attributes`, `Direction`, `CoordinateSystem`) collide as soon as the generated types share a namespace.
- **Inline schemas inside arrays produce synthetic `…Items` types.** `SurfaceCurrent.items` was an inline object, which generators render as `SurfaceCurrentItems`. The same was true for `Elements`, producing `SurfaceCurrentElementsItems` — a name with no relationship to anything in the source of truth.
- **`oneOf`-by-`required` discriminators are anonymous.** `LumpedPort` and `SurfaceCurrent` items used `oneOf: [{required:[Attributes]}, {required:[Elements]}]` over a *single* inline object. Generators can't see two distinct shapes, so they emit one type with every field optional and lose the discriminator.
- **Inline string enums collide.** Repeated literals like `["Cartesian", "Cylindrical"]` or `["MGS", "CGS", "CGS2"]` show up at multiple call sites; without a named definition, each call site produces a separate, name-clashing enum.

### Commit 1: schema refactor (`scripts/schema/config/*.json`)

Pure refactor — **no semantic changes**. The same documents that validated before still validate, and the same documents that failed still fail.

For each of `boundaries.json`, `domains.json`, `model.json`, `problem.json`, `solver.json`:

1. **Hoist inline object schemas into `\$defs`** with unique, source-traceable names. Examples:
   - `boundaries`: `PEC`, `PMC`, `Absorbing`, `Conductivity`, `WavePort`, `WavePortPEC`, `Ground`, `ZeroCharge`, `Terminal`, `Periodic`, `BoundaryPostprocessing`
   - `domains`: `Material`, `DomainPostprocessing`, `DomainEnergy`, `Probe`, `CurrentDipole`
   - `model`: `Box`, `Sphere`
2. **Split anonymous `oneOf` discriminators into named variants**:
   - `LumpedPort.items` → `oneOf: [LumpedPortAttributes, LumpedPortElements]`
   - `SurfaceCurrent.items` → `oneOf: [SurfaceCurrentAttributes, SurfaceCurrentElements]`

   This gives codegen two distinct, well-named types plus a tagged union, instead of one type with everything optional.
3. **Promote inline string enums to named `\$defs`**: `CoordinateSystem`, `ProblemType`, `GSOrthogonalization`, `AdaptiveGSOrthogonalization`, `AdaptiveCircuitSynthesisDomainOrthogonalization`, `MaterialAxes`, `Direction`, `DipoleDirection`.
4. **Name polymorphic value shapes** that previously appeared as anonymous `oneOf` fragments:
   - `Samples` → `SamplesPoint` | `SamplesLinear` | `SamplesLog`
   - Port `Direction` → `PortDirection` | `Vector3`
   - Port `Excitation` integer → `ExcitationIndex`

### Commit 2: validator follow-ups (`palace/utils/jsonschema.cpp`, `test/unit/test-schema.cpp`)

Caught by running the schema unit tests locally; required to keep behavior consistent under the named-`\$ref` schema.

1. **`FindAllSchemasByKey` now resolves internal `#/\$defs/` refs** instead of skipping them. Previously the function `continue`d past any `\$ref` into `\$defs`, which meant schemas only reachable through a named ref (e.g. `BoundaryPostprocessing.FarField`) were no longer findable via `ValidateConfig(data, "FarField")`. Adds a depth guard against pathological cycles.
2. **`SchemaErrorHandler` enhances type-mismatch and enum errors using substring matching.** With `LumpedPort.items` now a `oneOf`, the validator wraps simple errors as `[combination: oneOf / case#0] unexpected instance type`. The exact-string check stopped matching, so the `(got string)` / valid-enum hints disappeared. Switching to `find()` makes the enhancement fire regardless of `oneOf`/`anyOf` wrapping.
3. **Two exact-string assertions in `test-schema.cpp` relaxed to substring checks.** The validator's behavior is unchanged, but the wording for `Index` failures shifted from a flat single-message form to a `oneOf`-enumerated form. Asserting on the underlying message keeps the test resilient to validator-output formatting.

## Why this is safe

- Every schema change replaces an inline schema with a `\$ref` to a `\$defs` entry whose body is byte-equivalent (modulo whitespace) to the inlined version.
- `additionalProperties: false`, `required`, and `oneOf`/`allOf` constraints are preserved at the same nesting level.
- For `LumpedPort` / `SurfaceCurrent`, the two split variants carry the full property set; only the `required` differs, so no document that was valid before is rejected now.
- `FindAllSchemasByKey` continues to return the same results for top-level property lookups; it now *also* finds keys reachable only through internal `\$defs` refs (strictly more permissive).

## Test plan

- [x] **Local: `cmake --build build/palace-build --target unit-tests && build/palace-build/test/unit/palace-unit-tests "[schema]"`** — 10 passed, 1 skipped (the `Embedded Schema Matches Source` SECTION skips by design when its source path isn't present), 0 failed, 155/155 assertions.
  - Sub-schema lookup by key for `LumpedPort` / `WavePort` / `SurfaceCurrent` / `CurrentDipole` / `Materials` / `FarField`.
  - `LumpedPort` / `SurfaceCurrent` `oneOf: [Attributes, Elements]` discrimination.
  - `PEC`/`Ground` and `PMC`/`ZeroCharge` mutual exclusion.
  - All 24 example configs validate.
  - `Problem.Type → Solver` requirements (`Driven`/`Eigenmode`/`Transient` reject if matching solver section absent; `Electrostatic`/`Magnetostatic` accept defaults).
  - Error-message format: enum hints, type-mismatch hints, oneOf-wrapped messages, additional-property failures.
- [x] CI `style.yml` "Check JSON Schema" — `scripts/validate-config` runs over every `examples/**/*.json` against the refactored schema.
- [x] CI unit-test job (`test-schema.cpp`).

### DIFF
--- a/palace/utils/jsonschema.cpp
+++ b/palace/utils/jsonschema.cpp
@@ -51,11 +51,13 @@ public:
 };
 
 // Search for a schema property by key name, collecting all matches.
-// Helper for FindSchemaByKey - populates results vector.
+// Helper for FindSchemaByKey - populates results vector. The depth parameter
+// guards against pathological self-referential $defs cycles.
 void FindAllSchemasByKey(const json &schema, const std::string &key, const json &root_defs,
-                         std::vector<json> &results)
+                         std::vector<json> &results, int depth = 0)
 {
-  if (!schema.is_object())
+  constexpr int kMaxDepth = 32;
+  if (depth > kMaxDepth || !schema.is_object())
   {
     return;
   }
@@ -90,16 +92,22 @@ void FindAllSchemasByKey(const json &schema, const std::string &key, const json 
       if (v.contains("$ref"))
       {
         std::string ref_raw = v["$ref"].get<std::string>();
-        // Skip internal $defs references (handled elsewhere).
+        // Internal $defs reference: resolve and recurse so nested-via-$ref
+        // schemas (e.g. BoundaryPostprocessing.FarField) are still findable.
         if (ref_raw.find("#/$defs/") == 0)
         {
+          std::string def_name = ref_raw.substr(8);
+          if (!defs.is_null() && defs.contains(def_name))
+          {
+            FindAllSchemasByKey(defs[def_name], key, defs, results, depth + 1);
+          }
           continue;
         }
         std::string ref = StripPathPrefix(ref_raw);
         const auto &schema_map = schema::GetSchemaMap();
         if (auto it = schema_map.find(ref); it != schema_map.end())
         {
-          FindAllSchemasByKey(json::parse(it->second), key, defs, results);
+          FindAllSchemasByKey(json::parse(it->second), key, defs, results, depth + 1);
         }
         else
         {
@@ -108,7 +116,7 @@ void FindAllSchemasByKey(const json &schema, const std::string &key, const json 
       }
       else
       {
-        FindAllSchemasByKey(v, key, defs, results);
+        FindAllSchemasByKey(v, key, defs, results, depth + 1);
       }
     }
   }
@@ -116,7 +124,7 @@ void FindAllSchemasByKey(const json &schema, const std::string &key, const json 
   // Check items for arrays.
   if (auto items_it = schema.find("items"); items_it != schema.end())
   {
-    FindAllSchemasByKey(*items_it, key, defs, results);
+    FindAllSchemasByKey(*items_it, key, defs, results, depth + 1);
   }
 }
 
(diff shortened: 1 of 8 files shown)


palace/utils/jsonschema.cpp
```diff
@@ -51,11 +51,13 @@ class EmbeddedSchemaLoader
 };
 
 // Search for a schema property by key name, collecting all matches.
-// Helper for FindSchemaByKey - populates results vector.
+// Helper for FindSchemaByKey - populates results vector. The depth parameter
+// guards against pathological self-referential $defs cycles.
 void FindAllSchemasByKey(const json &schema, const std::string &key, const json &root_defs,
-                         std::vector<json> &results)
+                         std::vector<json> &results, int depth = 0)
```
hughcars, line 57: Can we have an error/warning if the max depth triggers, that seems like a developer error.

hughcars (commented):
Generally LGTM, one small request on error/warning handling. I think we should probably hold off merging this until #719, and #716 go in.

palace/utils/jsonschema.cpp
ad-cqc: Added a warning.

PR #657, opened 2026-03-04
Add 2D boundary mode problem type and enhance waveport capabilities

simlapointe:
This PR adds a 2D boundary mode problem type along with multiple enhacements to waveports. Specifically:

-  Adds `BoundaryMode` solver for mode analysis on 2D meshes, supporting both standalone 2D meshes
    and 2D boundary submesh extracted from a 3D mesh.
- Adds support for 2D meshes in all simulation types.
- Changes waveport formulation to enable support for impedance boundary conditions, closes #265.
-  Enables voltage and impedance postprocessing for waveport and BoundaryMode by specifying a line from ground to conductor, either through mesh coordinates or boundary attribute in `config["Boundaries"]["Postprocessing"]["Impedance"]`, closes #172.
- Fixes a bug where nonconformal AMR would lead to erroneous results when waveports are present, closes #648.
- Fixes a bug in waveport reported ω/2π frequency, closes #718.

PR #727, opened 2026-05-08
BoundaryMode follow-up: AMR on 3D-derived submesh, bug fix, architecture cleanup

hughcars:
Follow-up to #657. Bug fix, one capability unlock, and removes some now-unused
wave-port multigrid plumbing the merged PR left behind.

- Fix: `BoundaryMode` on a 3D mesh no longer aborts with `Lc = 0` at `np >= 4`.
  `RemapSubMeshAttributes` left `Mesh::attributes` stale, zeroing the 3D→2D
  tangent-plane normal.
- Serial mesh preprocessing (region refinement, crack insertion, 3D→2D submesh
  extraction) runs before partitioning, so `BoundaryMode` on a 3D-derived
  submesh supports AMR and gets `Lc` from the solve mesh rather than the
  loaded mesh.
- Splits `WavePortData`'s double duty. `BoundaryModeOperator` becomes the
  `SpaceOperator` analog; a thin shared `ModeEigenSolver` with two ctors
  (bare-FE for wave-port BC, `BoundaryModeOperator&` for `BoundaryMode`)
  replaces the `ModeEigenSolverMultigridConfig*` plumbing. Net −332 LOC in
  `palace/`. No wave-port BC behaviour change.

Commits are staged for review in order.


simlapointe (commented):
Looks good overall. I tested it on many other cases beyond the regression tests (thin or thick metal, impedance or PEC metal BC, pure 2D or submesh from 3D, single chip or flip chip, oxide layers or not, etc.) and results seem to be the same as before. One exception is the submesh case with impedance BC on the metal where results differ slightly, but I believe this current implementation is correct as I had previously made a slight error in the handling of adjacent/touching wave port BCs for the extracted submesh case.

One question is: do we need the new mode_assembly.cpp/hpp files or could/should these functions just be in boundarymodeoperator.cpp or modeeigensolver.cpp? It feels a bit unnecessary to me.

simlapointe (approved):
LGTM, assuming CI checks pass.

hughcars:
> Looks good overall. I tested it on many other cases beyond the regression tests (thin or thick metal, impedance or PEC metal BC, pure 2D or submesh from 3D, single chip or flip chip, oxide layers or not, etc.) and results seem to be the same as before. One exception is the submesh case with impedance BC on the metal where results differ slightly, but I believe this current implementation is correct as I had previously made a slight error in the handling of adjacent/touching wave port BCs for the extracted submesh case.
> 
> One question is: do we need the new mode_assembly.cpp/hpp files or could/should these functions just be in boundarymodeoperator.cpp or modeeigensolver.cpp? It feels a bit unnecessary to me.

Addressed in 43f01a8f0

Issue #710, opened 2026-04-17
Test antenna/short_dipole is very sensitive to numerics


Sbozzolo:
In https://github.com/awslabs/palace/pull/689, I bumped OpenBlas to 0.3.32 and I found that the antenna/short_dipole test started failing on neoverse-v1. 

I suspected it was a regression in OpenBlas and opened https://github.com/OpenMathLib/OpenBLAS/issues/5735.
The maintainers suggest that this is not a regression, but a change in accumulation of floating point error. 

Indeed, that test is the only case I could see where OpenBlas 0.3.32 leading to different results compared to everything else. I tried moving the short dipole away from the origin and found agreement, which suggests that what we are doing is very sensitive to details. 

It might be a good idea to investigate if this is just an issue with the test, or there is something deeper on how we handle deltas.

Sbozzolo:
cc: @laylagi

hughcars:
Another data point: #728 observed the same failure for a new one api version 2026.0.0, the baselining values are different to the others again, so it seems like a floating point rng of which element the dirac ends up in. It seems the proper solution is to actually do the smearing over all d entities attached to the dirac. This will require a (d-N) -> d entity map in the parallel mesh of mfem though to do properly, which does not exist presently (there's only d-1 -> d, face neighbors). 

<details>
Summary from debugging session:

Building Palace with the OneAPI 2026.0.0 toolchain and `intel-oneapi-mkl@2026.0.0` produces numerical output that deviates from the reference by ~27% (with sign flips on the real/imag components of far-field `E_x`) on the `antenna_short_dipole` regression case. Swapping only MKL to `@2025.3.1`, with everything else identical, makes the test match reference bit-for-bit.

Only the `CurrentDipole` path is affected — `LumpedPort`-excited antenna cases (`antenna_halfwave_dipole`, `antenna_halfwave_dipole_surfacecurrent`) match reference at both MKL versions.

### Environment

- Ubuntu 24.04, x86_64 Sapphire Rapids (c7i.16xlarge used, but any SPR/SKX box should work).
- `intel-oneapi-base-toolkit-2026.0` + `intel-oneapi-compiler-dpcpp-cpp-2026.0` + `intel-oneapi-compiler-fortran-2026.0` installed via Intel's apt repo.
- `spack develop` branch.
[...]

PR #728, opened 2026-05-08
Container CI: don't cache Palace itself

hughcars:
`palace@develop` is pinned to `branch="main"` in the Spack recipe, so every
branch's CI produces the same spec hash with a different binary. The container
build shares one OCI buildcache and force-pushes unsigned, so the last pusher
wins — poisoning every other branch's next container build. Dependencies are
content-keyed, so they're fine.

Split the Palace install into `--only dependencies` (from cache) + `--only
package --no-cache` (from local source), and restrict the cache push to
`--only dependencies`. Same pattern `palace-ci` already uses for the Spack job.

Fixes the current `testcase.jl:183` metafile-list failures on main and on
open PRs.

Drive-by: relocates PR #657's CHANGELOG bullets from the released 0.16.0
section into a new `In progress` block.

### DIFF
--- a/.github/actions/palace-ci/action.yml
+++ b/.github/actions/palace-ci/action.yml
@@ -319,7 +319,8 @@ runs:
       run: |
         source /tmp/timing_functions.sh
         start_timer
-        spack -e . buildcache push --force --with-build-dependencies --unsigned --update-index local-buildcache || true
+        # Push deps only; Palace itself is branch-specific (same reason as the container build).
+        spack -e . buildcache push --force --with-build-dependencies --only dependencies --unsigned --update-index local-buildcache || true
         end_timer "Push to GHCR cache"
 
     - name: Run Regression Tests
--- a/.github/actions/palace-ci/action.yml
+++ b/.github/actions/palace-ci/action.yml
@@ -128,6 +128,9 @@ runs:
               require: fflags=-nofor-main
             libffi:
               require: "@:3.4.8"
+            # MKL 2026+ miscomputes CurrentDipole; pin to 2025.
+            intel-oneapi-mkl:
+              require: "@:2025"
         INTELEOF
         fi
 
--- a/spack_repo/local/packages/palace/package.py
+++ b/spack_repo/local/packages/palace/package.py
@@ -94,6 +94,12 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         msg="GCC does not support AddressSanitizer on macOS (Apple Silicon). Use Clang instead.",
     )
 
+    # MKL 2026+ miscomputes CurrentDipole (~27% off on antenna_short_dipole).
+    conflicts(
+        "^intel-oneapi-mkl@2026:",
+        msg="MKL 2026+ miscomputes CurrentDipole; pin to @:2025",
+    )
+
     conflicts("^mumps+int64", msg="Palace requires MUMPS without 64 bit integers")
     with when("+mumps"):
         depends_on("fortran", type="build")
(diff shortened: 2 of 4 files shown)


Sbozzolo:
Is the intel failing test another instance of #710 ?

hughcars:
> Is the intel failing test another instance of #710 ?

Unsure, I am pretty confused about this as I thought that had been fixed.

hughcars:
> Is the intel failing test another instance of #710 ?

Trying to reproduce and it looks like its one api 2026 with a bug 😒 , I'm going to try just pinning the version back and see.

Sbozzolo:
> > Is the intel failing test another instance of #710 ?
> 
> Trying to reproduce and it looks like its one api 2026 with a bug 😒 , I'm going to try just pinning the version back and see.

How do you know it's a bug? We are also observing similar errors when we move to OpenBlas 0.3.30 and according to OpenBlas folks the only difference is a change in the accumulation of floating point error. At the time, I argued that we saw errors like this only with OpenBlas, but this seems to be another instance where the test fails with only a change in the compiler/math library. I take this as indication that either the test or the implementation of the feature are too brittle.

https://github.com/OpenMathLib/OpenBLAS/issues/5735

hughcars:
> > > Is the intel failing test another instance of #710 ?
> > 
> > 
> > Trying to reproduce and it looks like its one api 2026 with a bug 😒 , I'm going to try just pinning the version back and see.
> 
> How do you know it's a bug? We are also observing similar errors when we move to OpenBlas 0.3.30 and according to OpenBlas folks the only difference is a change in the accumulation of floating point error. At the time, I argued that we saw errors like this only with OpenBlas, but this seems to be another instance where the test fails with only a change in the compiler/math library. I take this as indication that either the test or the implementation of the feature are too brittle.
> 
> [OpenMathLib/OpenBLAS#5735](https://github.com/OpenMathLib/OpenBLAS/issues/5735)

I'd forgotten about that. I guess we just have to fix it properly. For now I'm pinning the oneapi, but we've now got two symptoms so we'll have to get it fixed properly on the actual mfem side. Frustratingly their parallel mesh data structure is incapable of the necessary (d-1)->d entity map necessary, so it's not a simple fix. We can perturb the dipole off of (0,0,0), but that's masking the issue.

spack_repo/local/packages/palace/package.py
```diff
@@ -94,6 +94,12 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         msg="GCC does not support AddressSanitizer on macOS (Apple Silicon). Use Clang instead.",
     )
 
+    # MKL 2026+ miscomputes CurrentDipole (~27% off on antenna_short_dipole).
+    conflicts(
+        "^intel-oneapi-mkl@2026:",
+        msg="MKL 2026+ miscomputes CurrentDipole; pin to @:2025",
+    )
```
Sbozzolo, line 101: I wouldn't say that Palace is incompatible with oneAPI 2026 if it is only that fragile test that fails. It's more of a bug in MFEM/Palace

.github/actions/palace-ci/action.yml
```diff
@@ -128,6 +128,9 @@ runs:
               require: fflags=-nofor-main
             libffi:
               require: "@:3.4.8"
+            # MKL 2026+ miscomputes CurrentDipole; pin to 2025.
+            intel-oneapi-mkl:
+              require: "@:2025"
```
Sbozzolo, line 133: Maybe add a link to the issue?

PR #730, opened 2026-05-13
Install gcc@15 with apt for container

hughcars:
gcc@15 sits in a separate spec group with `explicit: false` (so `spack gc -y` can reap it after the build), and palace's compiler-wrapper holds gcc's install path as a string rather than a graph dep — so neither `--only dependencies` nor `--only package` materialises it, and the build fails when the wrapper tries to exec a binary that was never installed.

Fix: `spack install gcc@15` ahead of the existing two passes.

### DIFF
--- a/.github/actions/build-container/spack_env/PalaceDockerfile
+++ b/.github/actions/build-container/spack_env/PalaceDockerfile
@@ -5,6 +5,19 @@
 ARG GITHUB_USER=""
 ARG GITHUB_TOKEN=""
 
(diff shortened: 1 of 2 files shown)


Sbozzolo:
I don't think this is the correct solution.  The fact that `gcc@15` is not in the final spec is by design: gcc is not a runtime dependency and it should not be added to the final container (it is very heavy in terms of size). If you `spack install` it, you bring it back to the container

Maybe this is a spack bug related to the fix to this issue? https://github.com/spack/spack/issues/52214

Sbozzolo (approved):
Thank you!

PR #734, opened 2026-05-18
Feature: Add absolute tolerance (AbsTol) configuration for linear solver

jpsimas:


palace/linalg/ksp.cpp
```diff
@@ -58,6 +58,7 @@ ConfigureKrylovSolver(const config::LinearSolverData &linear, int verbose, MPI_C
   }
   ksp->SetInitialGuess(linear.initial_guess);
   ksp->SetRelTol(linear.tol);
+  ksp->SetAbsTol(iodata.solver.linear.abs_tol);
```
hughcars: This needs to be `linear.abs_tol`, in the inner constructors we no longer pass the whole of IoData through.

scripts/schema/config/solver.json
```diff
@@ -21,6 +21,7 @@
       {
         "Target": { "type": "number", "exclusiveMinimum": 0.0 },
         "Tol": { "type": "number", "minimum": 0.0 },
+        "AbsTol": { "type": "number", "minimum": 0.0 },
```
hughcars: This slightly suggests changing the previous to `"RelTol"`, which would be a breaking change. We could introduce it as an alias for the time being and then mark deprecated.

hughcars (changes requested):
Hi @jpsimas, 

Thank you for the submission, two minor things. I'm also interested in why you needed this tolerance setting? What were you doing that a tighter `"Tol"`, aka RelTol, was insufficient?

jpsimas:
> Hi @jpsimas,
> 
> Thank you for the submission, two minor things. I'm also interested in why you needed this tolerance setting? What were you doing that a tighter `"Tol"`, aka RelTol, was insufficient?

Yes, when running some simulations with a particular tricky antenna geometry, the initial absolute residue sometimes gets really high, so I find myself needing to do I dry run to see what is the initial absolute residual to tune the relative tolerance so the simulation stops at a low-enough residue value. This is relatively ok, until you get to AMR runs where the initial residue can value a lot between iterations. Setting the absolute tolerance to some fixed value like 1e-3 makes it a lot easier to run this type of automated runs unattended without having to worry that the initial absolute residual jumped up and ended up generating bad results

palace/linalg/ksp.cpp
jpsimas: sorry! I was working with 0.16 and didnt see that got changed. how do I fix this, should I do another pull request?

hughcars:
> > Hi @jpsimas,
> > Thank you for the submission, two minor things. I'm also interested in why you needed this tolerance setting? What were you doing that a tighter `"Tol"`, aka RelTol, was insufficient?
> 
> Yes, when running some simulations with a particular tricky antenna geometry, the initial absolute residue sometimes gets really high, so I find myself needing to do I dry run to see what is the initial absolute residual to tune the relative tolerance so the simulation stops at a low-enough residue value. This is relatively ok, until you get to AMR runs where the initial residue can value a lot between iterations. Setting the absolute tolerance to some fixed value like 1e-3 makes it a lot easier to run this type of automated runs unattended without having to worry that the initial absolute residual jumped up and ended up generating bad results

Ok, sounds reasonable. You need to also add an entry to the documentation for the new config entry.

jpsimas:
@hughcars  – I've addressed the review comments. Summary of changes:

 • palace/linalg/ksp.cpp : Changed iodata.solver.linear.abs_tol → linear.abs_tol as requested.
 • palace/utils/configfile.cpp : Added backward‑compatible "RelTol" alias: if "RelTol" is present it takes precedence
   over the deprecated "Tol".
 • palace/utils/configfile.hpp : Added double abs_tol = 1.0e-50; member to LinearSolverData.
 • scripts/schema/config/solver.json : Added "RelTol" and "AbsTol" under "Linear"; marked "Tol" as deprecated.
 • docs/src/config/solver.md : Added documentation for "RelTol" and "AbsTol" under the solver["Linear"] section.

Please let me know if any further changes are needed. Thanks!

jpsimas:
I have resolved the merge conflicts with main. Since the configuration schemas were recently consolidated, I migrated the new AbsTol and RelTol definitions directly into the new scripts/schema/config-schema.json file. I also adapted the documentation that was previously in solver.md into the JSON descriptions so it aligns with the new auto-generation setup.

palace/utils/configfile.hpp
```diff
@@ -946,6 +946,9 @@ struct LinearSolverData
   // Iterative solver relative tolerance.
   double tol = 1.0e-6;
 
+  // Iterative solver absolute tolerance.
+  double abs_tol = 1.0e-50;
```
hughcars: The abs tol is somewhat of a risky tool, relying on a user to know an appropriate number, which is very difficult to do. So I would much rather this default to 0.0 than any finite value, and thus be turned off for almost everyone.

scripts/schema/config-schema.json
```diff
@@ -2681,10 +2681,22 @@
         },
         "Tol": {
           "type": "number",
-          "description": "Relative residual convergence tolerance for the iterative linear solver.",
+          "description": "Relative residual convergence tolerance for the iterative linear solver. **Deprecated:** Use `\"RelTol\"` instead. For backward compatibility, `\"Tol\"` is still read if `\"RelTol\"` is not specified.",
+          "minimum": 0.0,
+          "deprecated": true
```
hughcars: I don't think this is the PR to make this change in. It would require going to every `"Tol"` and changing it over through out the entire code base, and that ends up being a fairly enormous change for what is a small feature. Sorry to misdirect you originally, but better to stick with `"Tol"`, then have the `"AbsTol"` be a 0.0 default added in. If we decide to revisit to switch to `"RelTol"` everywhere, we can do that as a follow up.

scripts/schema/config-schema.json
```diff
@@ -2681,10 +2681,22 @@
...
+        },
+        "RelTol": {
+          "type": "number",
+          "description": "Relative residual convergence tolerance for the iterative linear solver. If both `\"RelTol\"` and `\"Tol\"` are specified, `\"RelTol\"` takes precedence.",
           "minimum": 0.0,
           "default": 1e-06
         },
+        "AbsTol": {
+          "type": "number",
+          "description": "Absolute residual convergence tolerance for the iterative linear solver. The solver converges when `||r|| < max(RelTol * ||b||, AbsTol)`. A value of 1.0e-50 effectively disables the absolute tolerance unless it is set larger.",
```
hughcars: Delete the third sentence about 1e-50, it can become out of date and does not add anything the previous formula does not express.

hughcars (changes requested):
Hi @jpsimas, two suggestions, one being a back track to `"Tol"`, and the other a change of default. Additionally, given this an addition to the config schema, you'll need to add a `REVISION` version bump to the schema version (see https://awslabs.github.io/palace/stable/developer/notes/#Schema-versioning for details on the new scheme we have in place). Also needs a `CHANGELOG.md` entry in the In Progress section, where you should at the end of the entry include `SchemaVer: 1-1-0` (whatever you have to bump it to), so something like

```
  - Introduced `"AbsTol"` option for linear solvers, defaulting to 0.0 to allow using an absolute tolerance when defining convergence. SchemaVer 1-1-0 [PR 734](https://github.com/awslabs/palace/pull/734).
```
(but make sure it is wrapped at the same character count as the existing entries, so that format-sources passes).

CHANGELOG.md
```diff
@@ -12,24 +12,27 @@ The format of this changelog is based on
...
+  case [PR 804](https://github.com/awslabs/palace/pull/804).
 
 #### New Features
 
-  - Added a `RationalImpedance` boundary condition: a surface (Robin) impedance boundary
-    whose per-square impedance is an arbitrary rational function of frequency,
-    `Zs(s) = N(s)/D(s)` with `s = iω`, given by numerator and denominator polynomial
-    coefficients (its zeros and poles), generalizing the parallel-RLC `Impedance` boundary
-    to any passive lumped-network response. SchemaVer 1-1-0
-    [PR 770](https://github.com/awslabs/palace/pull/770).
```
hughcars, line 32: Don't make the whitespace changes here, the SchemaVer entry is now off.

palace/utils/configfile.cpp
```diff
@@ -1314,7 +1314,9 @@ LinearSolverData::LinearSolverData(const json &linear)
 {
   type = linear.value("Type", type);
   krylov_solver = linear.value("KSPType", krylov_solver);
-  tol = linear.value("Tol", tol);
+  // Prefer "RelTol" over deprecated "Tol"
+  tol = linear.value("RelTol", linear.value("Tol", tol));
```
hughcars, line 1318: No longer needed.

hughcars (changes requested):
Two nits, then should be good.

PR #735, opened 2026-05-18
Move the regression suite from Julia into Catch2

hughcars:
## Move the regression suite from Julia into Catch2

Replaces the Julia harness (`test/examples/runtests.jl` + helpers) with regression cases written in Catch2 and built into `palace-unit-tests`. Reference CSVs and per-case tolerances are unchanged.

### Architecture

- **`palace::Run`** — library entry point lifted out of `main.cpp`'s solve block. Production `main()` and the test binary both call it.
- **`test/unit/regression_helpers.{hpp,cpp}`** — `RunRegressionCase` stages each case under `$TMPDIR/palace-regression/<case>/` (symlinks of `examples/<case>/*`), runs `palace::Run`, then on rank 0 only diffs every reference CSV column-wise via Catch2's `WithinRel || WithinAbs` matchers. Source tree stays read-only.
- **`test/unit/regression/cases.cpp`** — 27 `TEST_CASE`s tagged `[Regression]`. Two shared check factories cover the magnitude-only patterns (`CompareComplexMagnitudes` for paired `Re{X}/Im{X}` columns, `TestFarfield` for the 3-component vector magnitude on `farfield-rE.csv`).

CTest entries register as `regression-<case>` (`-np 2` by default; CMake var `PALACE_REGRESSION_NUMPROC` overrides). CI bypasses CTest and invokes `mpirun -np $NUM_PROC_TEST palace-unit-tests "[Regression]"` directly so rank count tracks the runner.

See `test/unit/README.md` for the filter model, override precedence, and how to add a case.

### Validation

`ctest -R "^regression-" -j 2` → 27/27 pass (~39m on a 2026 M-series). Tampering with a reference CSV fails the suite with Catch2's matcher diagnostic naming row, column, ref/actual, and tolerance.

### Notes for reviewers

- Stacked on `hughcars/boundarymode-2d`; the 2D cases (`cavity2d/*`, `cpw2d/*`, `cpw/wave_2dmode`) only exist on that base.
- macOS workflow drops `julia-actions/setup-julia` (only `runtests.jl` used it). Linux keeps it for `plot_farfield.jl`.
- `adapter/slp` and `transmon/*` are not ported — both were dormant in Julia's default case list (slp had no fixture; transmon was cost-gated).
- Re-baselining is now `palace -np N <config> && cp postpro/...csv test/examples/ref/...`.


test/unit/regression/cases.cpp
```diff
@@ -0,0 +1,503 @@
...
+  palace::test::RunRegressionCase("transmon", "transmon_amr.json", "transmon_amr", opts);
+}
+
+// adapter_slp is a dormant entry in runtests.jl: the case is listed in
+// the conditional block but neither examples/adapter/slp.json nor
+// test/examples/ref/adapter/slp/ exists on this branch. Omit until the
+// fixtures are added.
+
+// ===========================================================================
+// 2D cases. Live on hughcars/boundarymode-2d; not yet on origin/main.
```
simlapointe: Outdated comment

.github/actions/run-regression-tests/action.yml
```diff
@@ -86,29 +92,59 @@ runs:
...
+        # palace-unit-tests --help lists them in full.
+        LINEAR_SOLVER="${{ inputs.linear-solver }}"
+        if [[ -z "$LINEAR_SOLVER" ]]; then
+          LINEAR_SOLVER="Default"
+        fi
+        SOLVER_OVERRIDE=(--palace-linear-solver "$LINEAR_SOLVER")
+
         if [[ "${{ inputs.variant }}" == *"+cuda"* ]]; then
-          export PALACE_DEVICE=GPU
+          SOLVER_OVERRIDE+=(--palace-device GPU)
```
simlapointe: I don't think `--palace-device GPU` is sufficient to run on GPU. I think `--device cuda` is required to run on GPU because otherwise the actual MFEM device is still on CPU. It's a bit  unclear to me what `--palace-device GPU` does.

I tried on my end with a GPU build and with `--palace-device GPU` it uses the cpu:
```
user:~/palace/build$ mpirun -n 1 bin/palace-unit-tests  --palace-device GPU cpw2d_thin
Device configuration: cpu
Memory configuration: host-umpire
libCEED backend: /cpu/self/xsmm/blocked
Filters: "cpw2d_thin"
...
```
while with `--device cuda` it uses the GPU:
```
user:~/palace/build$ mpirun -n 1 bin/palace-unit-tests --device cuda cpw2d_thin
Device configuration: cuda,cpu
Memory configuration: host-umpire,cuda-umpire
Use GPU-aware MPI:    no
libCEED backend: /gpu/cuda/magma
Filters: "cpw2d_thin" [GPU]
...
```

test/unit/regression/cases.cpp
```diff
@@ -0,0 +1,503 @@
...
+    {
+      return;
+    }
+    const std::size_t n_rows = std::min(a.n_rows(), r.n_rows());
+    for (std::size_t i = 0; i < n_rows; ++i)
+    {
+      auto mag_sq = [&](palace::Table &t)
+      {
+        auto sq = [](double x, double y) { return x * x + y * y; };
+        return sq(t[3].data[i], t[4].data[i]) + sq(t[5].data[i], t[6].data[i]) +
```
simlapointe: This hard codes the farfield to be columns 3-8, which is true for driven cases but NOT for eigenmode cases so the wrong columns will be compared. See `test/examples/ref/cpw/lumped_eigen/farfield-rE.csv`, farfield columns are 5-10. The previous julia script counted the six complex E-field columns from the right.

test/unit/regression/cases.cpp
```diff
@@ -0,0 +1,503 @@
...
+                                  opts);
+}
+
+TEST_CASE("cpw_lumped_uniform", "[Serial][Parallel][GPU][Regression]")
+{
+  palace::test::RegressionOptions opts;
+  opts.rtol = 2.0e-2;
+  opts.atol = 1.0e-11;
+  opts.excluded_columns = {"Maximum", "Minimum"};
+  palace::test::RunRegressionCase("cpw", "cpw_lumped_uniform.json", "lumped_uniform", opts);
```
simlapointe, line 348: This `cpw_lumped_uniform` test case has a `farfield-rE.csv` file but we are not using `TestFarfield`... This is maybe ok if the test passes, but we should double check that this is intended. I think the old julia script used the custom farfield test.

palace/driver.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "utils/memoryreporting.hpp"
+#include "utils/outputdir.hpp"
+#include "utils/timer.hpp"
+
+namespace palace
+{
+
+void Run(IoData &iodata, MPI_Comm comm, int omp_threads, const char *git_tag)
+{
+  MakeOutputFolder(iodata, comm);
```
simlapointe: Hasn't `MakeOutputFolder` already been called at palace/main.cpp:228?

test/unit/main.cpp
```diff
@@ -111,18 +157,35 @@ int main(int argc, char *argv[])
...
-  {
-    cfg.testsOrTags.emplace_back("[Parallel]");
-  }
-  else
+  // Bare invocation: auto-select by MPI world size, and defensively
+  // exclude [Regression] (which is slow and has its own explicit
+  // entry point; long regression cases are tagged [Regression][Long]
+  // so they're caught here too). With any explicit user selector,
+  // run exactly what was asked for.
+  if (!user_selected_tests)
```
simlapointe, line 163: I believe this changes explicit tag/name selection semantics for normal unit tests. For example, on a single-rank run, `bin/palace-unit-tests "[vector]"` now includes `[Parallel]`-only tests such as `Vector Sum - Real - Asymmetric Sizes`, because the runtime `[Serial]` filter is skipped whenever the user provided any selector.

test/unit/main.cpp
```diff
@@ -98,6 +132,18 @@ int main(int argc, char *argv[])
...
+
+  // Whether the user passed any explicit test selector on the command
+  // line (name, wildcard, or tag spec). Bare invocation leaves this
+  // empty and triggers our convenience defaults below; anything
+  // explicit runs exactly what was asked for. This matters for
+  // regression cases in particular: they're tagged `[Regression]`
+  // only, so auto-adding `[Serial]` or `~[Regression]` would either
+  // fold in unrelated unit tests (separate positive filters are
+  // OR'd in Catch2) or silently exclude the very test the user
+  // selected by name.
```
simlapointe: Not sure if this comment is still accurate

.github/actions/run-regression-tests/action.yml
```diff
@@ -86,29 +92,59 @@ runs:
...
+
+        # One mpirun per case. Each case runs in a fresh process so
+        # long-lived library state (libCEED kernel cache, MFEM /
+        # HYPRE singletons, PETSc heap) doesn't accumulate across the
+        # whole suite.
+        run_cases() {
+          while IFS= read -r case; do
+            [[ -z "$case" ]] && continue
+            echo "=== Running $case ==="
+            mpirun -np $NUM_PROC_TEST palace-unit-tests \
```
simlapointe: This relies on the `palace-unit-tests` binary’s compile-time `PALACE_EXAMPLES_DIR_DEFAULT` / `PALACE_REGRESSION_REF_DIR_DEFAULT` paths. Could this be fragile for the split Spack test jobs, because the binary may have been built/installed from a different checkout or Spack stage than this runner’s `$GITHUB_WORKSPACE`?

simlapointe (changes requested):
Looks pretty good, tested it on my end and things are mostly working. Flagged a few potential issues below.

test/unit/regression_helpers.cpp
```diff
@@ -0,0 +1,490 @@
...
+// is the "either condition" form: WithinRel(ref, rtol) || WithinAbs(ref, atol).
+void CompareCSVFiles(Table &actual, Table &reference, const RegressionOptions &opts)
+{
+  if (!ValidateCSVTables(actual, reference, opts) || SkipNumericComparison(opts))
+  {
+    return;
+  }
+
+  const std::size_t n_cols = std::min(actual.n_cols(), reference.n_cols());
+  const std::size_t n_rows = std::min(actual.n_rows(), reference.n_rows());
```
Sbozzolo: When `actual.n_rows() = 0` and `skip_rowcount = true`, no check is performed, even when `reference.n_rows != 0` (which should lead to a failure)

test/unit/regression_helpers.cpp
```diff
@@ -0,0 +1,490 @@
...
+      {
+        const auto name = entry.path().filename().string();
+        // postpro/ and log/ are Palace outputs; don't stage stale copies.
+        if (name == "postpro" || name == "log")
+        {
+          continue;
+        }
+        std::filesystem::create_symlink(entry.path(), stage_ / name);
+      }
+    }
```
Sbozzolo, line 338: There's a risk of deadlock here. If rook throws an exception, all the other ranks will indefinitely wait at the `Mpi::Barrier` below.

.github/actions/build-container/action.yml
```diff
@@ -10,7 +10,7 @@ runs:
         # Spack develop because we want to have access to the most recent
         # version of spack-packages
         spack-version: 'develop'
-        julia-version: 'release'
+        julia-version: 'dont-install'
```
Sbozzolo: I think we should just remove julia from the setup for self-hosted runners.

palace/main.cpp
```diff
@@ -245,90 +234,39 @@ int main(int argc, char *argv[])
     iodata.WriteResolvedConfig(config, argv[1]);
   }
 
-  BlockTimer bt1(Timer::INIT);
-  // Initialize the MFEM device and configure libCEED backend.
-  int omp_threads = utils::ConfigureOmp(), ngpu = utils::GetDeviceCount();
-  mfem::Device device(ConfigureDevice(iodata.solver.device),
-                      utils::GetDeviceId(world_comm, ngpu));
-  ConfigureCeedBackend(iodata.solver.ceed_backend);
+  // Initialise device + numerics. The BlockTimer is scoped to this block so it
```
Sbozzolo:
```suggestion
  // Initialize device + numerics. The BlockTimer is scoped to this block so it
```

CHANGELOG.md
```diff
@@ -44,6 +44,15 @@ The format of this changelog is based on
 
 #### Interface Changes
 
+  - The regression test suite has moved from the Julia harness in
+    `test/examples/runtests.jl` to Catch2 cases registered in the
+    `palace-unit-tests` binary. Run with
+    `mpirun -np $NUM_PROC_TEST palace-unit-tests "[Regression]"` or via
+    CTest as `ctest -R "^regression-"`. Reference CSVs under
+    `test/examples/ref/` are unchanged. Re-baselining is now
```
Sbozzolo: I was using the `baseline` script. It was a nice way to quickly re-baseline one or multiple experiments at the same time. Maybe we can extract the basic functionalities in a copy&pastable bash command that we add to the docs?

Sbozzolo (commented):
- We should also remove the `.gitlab-ci.yml` (it uses the Julia harness, and we don't really use it)
- I wonder if we should just take the plunge and use this as an opportunity to decouple the notion of "example" from "regression test"

Sbozzolo:
Sorry, I clicked "submit too early", I am not done yet :)

.github/actions/run-regression-tests/action.yml
```diff
@@ -86,29 +92,59 @@ runs:
...
+              "${SOLVER_OVERRIDE[@]}" \
+              "$case"
+          done
+        }
 
-        julia --project=test/examples --color=yes test/examples/runtests.jl \
-          --palace-solver ${{ inputs.solver }} \
-          ${TEST_CASES:+--test-cases "$TEST_CASES"}
+        translate_cases "${{ inputs.test-cases }}" | run_cases
```
Sbozzolo: Can we replace all of this with using `ctest` instead of manually calling the binary?

It requires some reworking on run_regreesion_test.sh, but this code relies on string parsing and seems pretty fargile.

.github/workflows/build-and-test-macos.yml
```diff
@@ -256,6 +248,11 @@ jobs:
...
       - name: Run regression tests for examples/
         if: needs.filter.outputs.test == 'true'
+        # ctest fans the [Regression] cases out, one fresh
+        # palace-unit-tests process per case (label "regression"). Each
+        # process runs at $PALACE_REGRESSION_NUMPROC MPI ranks (read by
+        # the test/unit/run_regression_test.sh wrapper at run time);
+        # ctest -j N controls how many cases run concurrently.
         timeout-minutes: 40
         env:
           NUM_PROC_TEST_MAX: '8'
```
Sbozzolo: Maybe we can just set `PALACE_REGRESSION_NUMPROC` instead of NUM_PROC is all we do is reassigning it

.github/workflows/build-and-test-linux.yml
```diff
@@ -547,6 +551,12 @@ jobs:
 
       - name: Run regression tests for examples/
         if: needs.filter.outputs.test == 'true' && !matrix.with-sanitizers
+        # ctest fans the [Regression] cases out, one fresh
+        # palace-unit-tests process per case (label "regression"). Each
+        # process runs at $PALACE_REGRESSION_NUMPROC MPI ranks (read by
+        # the test/unit/run_regression_test.sh wrapper at run time);
+        # ctest -j N controls how many cases run concurrently. Total
+        # ranks in flight = NUM_PROC_TEST * N, kept <= NUM_PROC_TEST_MAX.
```
Sbozzolo: Important comment below.

I think we should think about this a little bit more.

This is different compared to non-regression tests. For non-regression tests, `-j M` means that up to M slots are used (typically, one slot = CPU). This is not the number of jobs (because CMake knows that there are Parallel jobs, which use 2 slots each). So, if you run a full non-regression + regression test, I think that we would end up oversubscribing CPUs.

Regression tests do not set `PROCESSORS`, so ctest assigns them one slot, even if they use multiple CPUs.

It is also counter-intuitive that -j has different meaning for different classes of tests.

palace/driver.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace palace
+{
+
+class IoData;
+
+// Run a Palace solve end-to-end on an already-parsed `iodata`.
+//
+// Preconditions:
+//   - MPI has been initialised.
```
Sbozzolo:
```suggestion
//   - MPI has been initialized.
```

palace/driver.hpp
```diff
@@ -0,0 +1,27 @@
...
+//   - `mfem::Device` has been constructed with the desired backend.
+//   - libCEED, HYPRE, and (if built in) SLEPc/PETSc are initialised.
+//   - `iodata.problem.output` is set; relative mesh paths in the config
+//     are resolvable from the current working directory (or have been
+//     rewritten to absolute paths by the caller).
```
Sbozzolo, line 23: Having a list of things to need to happen doesn't seem very robust. Instead, I wonder if we should have our Initialize class/function to set these things up. 

I've always found difficult to understand what needs to be initialized and how (and ran into lots of problems with setting up the unit test framwrok).

.github/actions/run-regression-tests/action.yml
```diff
@@ -61,7 +61,13 @@ runs:
...
         source /tmp/timing_functions.sh
         start_timer
-        spack -e . install --only-concrete --no-check-signature
+        # --test root builds and installs the palace-unit-tests binary
+        # (and its example/ref data). The target is EXCLUDE_FROM_ALL, so
+        # without --test root the regression harness binary is never
+        # built and `spack load palace` can't expose it. The lock was
+        # already concretized with --test root by the build job, so the
+        # test deps (catch2) are present in the concrete environment.
+        spack -e . install --only-concrete --no-check-signature --test root
```
Sbozzolo, line 69: I am not sure we need this is we call ctest instead of calling the executable directly. (Although, having it doesn't hurt)

.github/workflows/build-and-test-linux.yml
```diff
@@ -558,8 +568,12 @@ jobs:
             source /opt/intel/oneapi/setvars.sh  # Sets PATH, MKLROOT
           fi
           if [[ "${{ matrix.math-libs }}" == 'aocl' ]]; then
-            export AOCLROOT=/opt/AMD/aocl/aocl-linux-gcc-4.1.0/gcc
-            export LD_LIBRARY_PATH=$AOCLROOT/lib:$LD_LIBRARY_PATH
+            export AOCLROOT="$AOCLROOT"
```
Sbozzolo: What's this for? 

if `AOCLROOT` is already defined, why do need to export it?

Same with `ARMPL_DIR` below

.github/workflows/build-and-test-linux.yml
```diff
@@ -573,12 +587,25 @@ jobs:
...
 
-          # Run tests
-          julia --project=test/examples -e 'using Pkg; Pkg.instantiate()'
-          julia --project=test/examples --color=yes test/examples/runtests.jl
+          # Match the old Julia harness: run the regression cases
+          # sequentially, each on NUM_PROC_TEST ranks (consumed by
+          # run_regression_test.sh). No cross-case ctest fan-out.
+          export PALACE_REGRESSION_NUMPROC=$NUM_PROC_TEST
+          cd palace-build/palace-build/test/unit
+          ctest --output-on-failure -L "^regression$"
```
Sbozzolo: If we fix the issue with the job server slots I described above, I think we should be able to remove the split bewteen unit tests and regression tests and have ctest take care of both of them at the same time.

I think this would be nice because it unifies what tests are

.github/workflows/build-and-test-macos.yml
```diff
@@ -273,6 +270,9 @@ jobs:
...
 
-          # Run tests
-          julia --project=test/examples -e 'using Pkg; Pkg.instantiate()'
-          julia --project=test/examples --color=yes test/examples/runtests.jl
+          # Match the old Julia harness: run the regression cases
+          # sequentially, each on NUM_PROC_TEST ranks (consumed by
+          # run_regression_test.sh). No cross-case ctest fan-out.
+          export PALACE_REGRESSION_NUMPROC=$NUM_PROC_TEST
+          cd palace-build/palace-build/test/unit
+          ctest --output-on-failure -L "^regression$"
```
Sbozzolo: Same comment as above on the split between unit and regression tests.

Sbozzolo: I think this also applies to the spack builds

docs/src/developer/testing.md
```diff
@@ -173,6 +178,27 @@ special tags that control when tests execute based on the runtime environment:
...
+    and skipped from the default unit-test sweep. Opt in with
+    `ctest -L "^regression$"` or
+    `palace-unit-tests "[Regression]~[Long]"`. Each case lives in
+    `test/unit/regression/cases.cpp`.
+  - `[Long]` is a *modifier* on a regression case for solves that take
+    long enough (~10 minutes) that they shouldn't run on every PR.
+    These cases are tagged `[Regression][Long]` and registered as
+    `long-*` ctest entries (label `long`); the long-tests CI workflow
+    runs them via `ctest -L "^long$"` when the `trigger-long-tests`
+    PR label is applied.
```
Sbozzolo, line 196: Does `Long` only apply for regression?

I think it should also apply to regular unit tests, so that the two axes are truly orthogonal. This would also allow us to add long unit tests, which is something we'd get for free.

.github/actions/run-regression-tests/action.yml
Sbozzolo: I also think that this is not working at he moment.

If I look at the test suite in this PR, I only see the sphere example being run.

E.g.: https://github.com/awslabs/palace/actions/runs/27211376257/job/80384686984?pr=735

test/unit/regression/cases.cpp
```diff
@@ -0,0 +1,503 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Regression cases: full Palace solves diffed against test/examples/ref/.
+// Each case carries [Serial][Parallel][GPU][Regression] (the category
+// tag is orthogonal to execution style — every case is valid at any
+// rank count and on either CPU or GPU). [Long] modifies a regression
+// case to be skipped from the default sweep and run only under the
+// long-tests CI workflow.
```
Sbozzolo: Before, we excluded some regression tests from GPU runs because they took too long. I don't see that here and I am not sure if the intersection GPU with Regression works

test/unit/CMakeLists.txt
```diff
@@ -100,10 +105,17 @@ set_property(
...
 target_compile_definitions(
   unit-tests
   PRIVATE "PALACE_TEST_DATA_DIR=\"${CMAKE_INSTALL_PREFIX}/share/palace/test/data\""
+  # Source-tree roots used by the regression suite. Baking these into the
+  # binary keeps the common path ("run the built binary") free of env
+  # plumbing. RunRegressionCase honours a PALACE_EXAMPLES_DIR /
+  # PALACE_REGRESSION_REF_DIR environment override and a Catch2 CLI flag
+  # for relocated installs; these are the last-resort defaults.
+  PRIVATE "PALACE_EXAMPLES_DIR_DEFAULT=\"${CMAKE_SOURCE_DIR}/../examples\""
+  PRIVATE "PALACE_REGRESSION_REF_DIR_DEFAULT=\"${CMAKE_CURRENT_SOURCE_DIR}/../examples/ref\""
```
Sbozzolo: This probably won't work with space because of relocation of the source code. Everything that is needed for tests has to be in `PALACE_TEST_DATA_DIR`. I think we should move the regression data there.

test/unit/CMakeLists.txt
```diff
@@ -175,6 +191,62 @@ catch_discover_tests(unit-tests
...
+     PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                 GROUP_READ GROUP_EXECUTE
+                 WORLD_READ WORLD_EXECUTE)
+
+catch_discover_tests(unit-tests
+  TEST_PREFIX "regression-"
+  TEST_SPEC "[Regression]~[Long]"
+  TEST_EXECUTOR ${CMAKE_CURRENT_BINARY_DIR}/run_regression_test.sh
+  EXTRA_ARGS ${MPIEXEC_POSTFLAGS}
+  PROPERTIES
```
Sbozzolo, line 233: See comment on PROCESSORS and job server slots

docs/src/developer/testing.md
```diff
@@ -417,76 +443,94 @@ Two special cases bypass the long test requirement:
...
+The regression machinery reads the source-tree `examples/` and
+`test/examples/ref/` roots via a two-level override chain:
 
-```bash
-julia --project runtests.jl --num-proc-test 4
-```
+| Precedence | Mechanism                                                                                                                                                                                 |
+|:---------- |:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1 (high)   | `--examples-dir` / `--regression-ref-dir` / `--regression-run-dir` CLI flags                                                                                                              |
+| 2 (low)    | Compile-time `PALACE_EXAMPLES_DIR_DEFAULT` / `PALACE_REGRESSION_REF_DIR_DEFAULT` (wired from CMake); run-dir falls back to `std::filesystem::temp_directory_path() / "palace-regression"` |
```
Sbozzolo: I think we should just treat regression test data as any other test data (possibly moving the `test/unit/data` folder to a `test/data` folder.

Sbozzolo (changes requested):
Really great feature! 

Important:
- The spack test system doesn't seem to work and only spheres get tested (see https://github.com/awslabs/palace/actions/runs/27211376257/job/80384686984?pr=735)
- I think the why we set MPI ranks conflicts with the slot count in the ctest job server, so that we might end up getting oversubscribed count of CPUs. Fixing this, would also unify the two tests categories under ctest, which I think would be very nice.

test/unit/CMakeLists.txt
```diff
@@ -102,10 +107,17 @@ set_property(
...
 target_compile_definitions(
   unit-tests
   PRIVATE "PALACE_TEST_DATA_DIR=\"${CMAKE_INSTALL_PREFIX}/share/palace/test/data\""
+  # Source-tree roots used by the regression suite. Baking these into the
+  # binary keeps the common path ("run the built binary") free of env
+  # plumbing. RunRegressionCase honours a PALACE_EXAMPLES_DIR /
+  # PALACE_REGRESSION_REF_DIR environment override and a Catch2 CLI flag
+  # for relocated installs; these are the last-resort defaults.
+  PRIVATE "PALACE_EXAMPLES_DIR_DEFAULT=\"${CMAKE_SOURCE_DIR}/../examples\""
+  PRIVATE "PALACE_REGRESSION_REF_DIR_DEFAULT=\"${CMAKE_CURRENT_SOURCE_DIR}/../examples/ref\""
```
Sbozzolo: We should just move the regression test data in `TEST_DATA_DIR`. If we do so, we don't need to do any of the `GenerateRegerssionCTest.cmake` dance because I already took care of implmenting relocation of test data.

This also reinforces the split bteween exampes and regerssion tests

test/unit/CMakeLists.txt
```diff
@@ -177,6 +192,64 @@ catch_discover_tests(unit-tests
...
+# launches via the generated `run_regression_test.sh` wrapper, which
+# reads `$PALACE_REGRESSION_NUMPROC` (default below) and exec's mpirun.
+# CTest schedules cases using the PROCESSORS property below; OpenMP builds
+# reserve twice the MPI rank count because CI runs them with OMP_NUM_THREADS=2.
+set(PALACE_REGRESSION_NUMPROC_DEFAULT 2 CACHE STRING
+    "Default MPI rank count for the [Regression] ctest registration. Overridden at run time by the PALACE_REGRESSION_NUMPROC environment variable.")
+set(PALACE_REGRESSION_PROCESSORS ${PALACE_REGRESSION_NUMPROC_DEFAULT})
+if(PALACE_WITH_OPENMP)
+  math(EXPR PALACE_REGRESSION_PROCESSORS "${PALACE_REGRESSION_PROCESSORS} * 2")
+endif()
```
Sbozzolo: I don't think the build system should know about the details of how of CI is set up. We should just divide by 2 in CI.

docs/src/developer/testing.md
```diff
@@ -417,76 +443,106 @@ Two special cases bypass the long test requirement:
...
+rank/thread slots.
 
-```bash
-julia --project runtests.jl --test-cases "spheres rings"
-```
+#### Overrides
 
-Run with custom *Palace* executable:
+The regression machinery reads the source-tree `examples/` and
+`test/examples/ref/` roots via the following override chain:
```
Sbozzolo: This system of override is pretty complex, as I say elsewhere, I think we should treat regression data as test data and get rid of this override.

Sbozzolo (changes requested):
I think the changes do not address the problem with the job slots.

Adding `PALACE_REGRESSION_PROCESSORS` adds some notion of parallelism, but there's no connection with `MPIEXEC_NUMPROC_FLAG`, so you will still have oversubscribed runs.

I looked at it and I didn't find any simple way to make the number of MPI ranks a runtime variable.

I see two options:
- we make the number of MPI ranks for regerssion tests a build time option (just pass PALACE_REGERSSIN_PROCESSORS`). This is the simplest and would remove the need for a run_regression_test.sh
- we fold this into GenerateRegressionCTest.cmake and do generation at test time instead of build

I personally would prefer going with the much simpler case of fixing the number of ranks at build time. This puts unit and regression tests on the same footing, so that one can just run `ctest -j $(nproc)` to run everything fully utilizing all the CPUs.

The default value could be 2, but on our machines it could be nproc/4 or so. This would speed CI up because many of our tests do not benefit from running with 64 cores.

palace/driver.hpp
hughcars: Agreed; this is broader runtime-lifecycle cleanup and I'm keeping it out of this regression-harness refactor. I opened https://github.com/awslabs/palace/issues/785 to track centralizing Palace initialization across `palace/main.cpp`, `test/unit/main.cpp`, and future drivers.

.github/actions/run-regression-tests/action.yml
```diff
@@ -86,29 +91,44 @@ runs:
...
+          if [[ -z "$raw" || "$raw" == "default" ]]; then
+            run_group "[Regression]~[Long]"
+          elif [[ "$raw" == "long" ]]; then
+            run_group "[Long]"
+          else
+            for case in $raw; do
+              run_group "$case"
+            done
+          fi
+        }
```
Sbozzolo: I don't undertand why this is needed and we don't use ctest directly

.github/actions/run-regression-tests/action.yml
```diff
@@ -86,29 +91,44 @@ runs:
         fi
 
         eval $(spack -e . load --sh palace)
-        julia --project=test/examples -e 'using Pkg; Pkg.instantiate()'
-        
+
+        # The installed palace-unit-tests binary was built with
+        # PALACE_TEST_DATA_DIR pointing at the installed test data. Regression
+        # configs, meshes, and references live under that tree, so the split
+        # Spack jobs can run the installed binary directly.
```
Sbozzolo:
```suggestion
```

I don't think we need this comment

.github/workflows/build-and-test-linux.yml
```diff
@@ -558,27 +572,37 @@ jobs:
...
           fi
-          export NUM_PROC_TEST=$(nproc 2> /dev/null || sysctl -n hw.ncpu)
-          if [[ "$NUM_PROC_TEST" -gt "$NUM_PROC_TEST_MAX" ]]; then
-            NUM_PROC_TEST=$NUM_PROC_TEST_MAX
+          if [[ "${{ matrix.math-libs }}" == 'armpl' ]]; then
+            export LD_LIBRARY_PATH="$ARMPL_DIR/lib:$LD_LIBRARY_PATH"
+          fi
+          export CTEST_PARALLEL_LEVEL=$(nproc 2> /dev/null || sysctl -n hw.ncpu)
+          if [[ "$CTEST_PARALLEL_LEVEL" -gt "$NUM_PROC_TEST_MAX" ]]; then
+            CTEST_PARALLEL_LEVEL=$NUM_PROC_TEST_MAX
```
Sbozzolo: Do we need a `NUM_PROC_TEST_MAX`? Why don't we just set CTEST_PARALLEL_LEVEL=$(nproc)` ?

test/unit/CMakeLists.txt
```diff
@@ -177,6 +187,50 @@ catch_discover_tests(unit-tests
...
+# prefix (`ctest -L "^regression$"` to select, `-LE "^(regression|long)$"` to
+# skip both regression flavors).
+#
+# Rank count is a build-time knob so CTest knows exactly how many slots each
+# case consumes. CI chooses both values explicitly; for OpenMP jobs it sets
+# PROCESSORS to ranks * OMP_NUM_THREADS.
+set(PALACE_REGRESSION_NUMPROC 2 CACHE STRING
+    "MPI ranks per [Regression] ctest case.")
+set(PALACE_REGRESSION_PROCESSORS ${PALACE_REGRESSION_NUMPROC} CACHE STRING
+    "CTest processor slots reserved per [Regression] case.")
```
Sbozzolo: This hard-codes the assumption on the value of OMP_NUM_THREADS as part of the build process. If users set a different OMP_NUM_THREADS at runtime, it will break the slot counting.

We maybe should just hard-code the number of OpenMP threads in the test suite (and ingore the env variable) so that there's no ambiguity and we have full accounting of the number of CPUs to be used. This can also be used to solve the problem with slot counting in unit tests.

So maybe we can have:
```
set(PALACE_TESTS_NUMPROC 2 CACHE STRING
    "MPI ranks per [Regression] ctest case.")
set(PALACE_TESTS_OMP_THREADS 1 CACHE STRING
    "Number of OpenMP threads to use for unit and regerssion tests")
PROCESSORS = TESTS_NUMPROC * TESTS_OMP_THREADS
```

This way, we can also make the number of processes for the unit tests configurable, instead of being hardcoded to 2.

Sbozzolo (approved):
Modulo CI, this is really great improvement, thank you!

simlapointe (approved):

PR #741, opened 2026-05-28
Add IncludeInSynthesis flag to opt lumped ports out of PROM basis

hughcars:
Adds a per-port `IncludeInSynthesis` boolean (default `true`) that controls whether a `LumpedPort` contributes a port-mode basis vector to the ROM when `AdaptiveCircuitSynthesis` is enabled. The boundary condition is unaffected — only the row/column in the synthesized `L⁻¹`/`C`/`R⁻¹` matrices.

Motivation: `AddLumpedPortModesForSynthesis` injects one full-FE-space basis vector per lumped port up front. On a chip-scale model with many launchers kept as 50 Ω terminations for correct physics but not needed for measurement in the circuit. Excited ports must remain included; the parser rejects `Excitation > 0` paired with `IncludeInSynthesis: false` at config-parse time.

```jsonc
{ "Index": 2, "R": 50, "IncludeInSynthesis": false, "Direction": "+X", "Attributes": [102] }
```

Two unit tests: the positive skip path (1×1 result on a known fixture) and the parse-time rejection. Schema, docs, CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ The format of this changelog is based on
 
 #### New Features
 
+  - Added `"IncludeInSynthesis"` boolean flag to lumped port configuration (default `true`).
+    When adaptive driven circuit synthesis is enabled
+    (`config["Solver"]["Driven"]["AdaptiveCircuitSynthesis"]`), this flag controls whether
+    the port contributes a port-mode basis vector to the reduced-order model. The boundary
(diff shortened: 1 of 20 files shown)


CHANGELOG.md
simlapointe, line 1: Not sure we want these extra blank lines between each feature?

simlapointe (approved):
LGTM and should be useful. One nit about changelog style.

PR #742, opened 2026-05-28
Patch MFEM DeltaCoefficient assembly

### LINKED ISSUE #598: Remove pin to version of OpenBlas
Our CI pins OpenBlas to 0.3.29 (in spack). This is because of issues like https://github.com/OpenMathLib/OpenBLAS/issues/5459, which prevent compilation on certain arm platforms. 
We should remove the pin when the issue is fixed and a new version is released.

### LINKED ISSUE #710: Test antenna/short_dipole is very sensitive to numerics

hughcars:
Fix the short-dipole lookahead instability by patching MFEM's `DeltaCoefficient`
assembly so a source centered on a shared mesh entity is distributed over all
containing elements instead of whichever element `FindPoints` returns first.

Summary:

- Add the MFEM PR #5353 patch to both the CMake superbuild and Spack recipe.
- Use `ParLinearForm` for Palace current-dipole assembly so cross-rank containing
  element counts are reduced before applying the source weight.
- Refresh the short-dipole antenna reference outputs.
- Make Spack buildcache key installation non-interactive.
- Pin JuliaFormatter to keep style checks deterministic.

The MFEM patch also initializes the boundary "stuck" return path in
`InverseElementTransformation`, removing an uninitialized reference-point source
of nondeterminism.

Addresses awslabs/palace#710.
Tracks mfem/mfem#5353.

I added a commit at the end just regenerating the documentation images to demonstrate no changes. The images are now generated with the same frequency as the rest of those in the section. 

Closes #598 
Closes #689


Sbozzolo:
Thank you, this makes sense to me. Could you double check that this also fixes #598 ? I think all you need to do to check that is rebase/cherry-pick PR #689

Sbozzolo (approved):

PR #743, opened 2026-06-01
Improve S-parameters calculations

### LINKED ISSUE #740: S-parameters seem wrong when using different lumped port R values
S-parameters calculated for port pairs that have different R values do not align with the expected values. 

For example, when simulating a simple coaxial 50 Ω cable with a 100 Ω source port (1) at one end and 50 Ω port (2) at the other end, the reflection coefficient is $\rho_1 = (Z_{line} - R_1)/(Z_{line} + R_1) = -1/3$ and thus we expect 
$S_{11} = V_1 / V_{inc} - 1 = (1+\rho_1) - 1 = -1/3$
and 
$S_{21} = V_2 / V_{inc} = \sqrt{1 - \rho_1^2} = \sqrt{8/9} \approx 0.943$.

That is what we get in other EM solvers, but in Palace we get: 
```
 S[1][1] = -3.334e-01-2.322e-06i, |S[1][1]| = -9.541e+00, arg(S[1][1]) = -1.800e+02
 S[2][1] = -1.333e+00+2.901e-03i, |S[2][1]| = +2.498e+00, arg(S[2][1]) = +1.799e+02
```
where $|S_{21}|$ is > 1 and larger than the expected value by $\sqrt{2}$. I believe this is due to an unnecessary scaling of $\sqrt{R_{src}/R_{obs}}$ in the S-parameter calculations.

simlapointe:
This PR improves S-parameters calculations in the following ways:
- Enables mixed lumped/wave port S-parameters. Closes #691.
- Fixes a bug where S-parameters calculated from lumped port pairs with differing R values are incorrect. Closes #740.
- Uses the wave port `VoltagePath`, if provided,  to fix the phase of the wave port mode, ensuring we get consistent S-parameter signs between lumped and wave ports. Also prints the wave port impedance in both the palace console/log and output folder.
- Adds a unit test for the wave port characteristic impedance calculation and a regression test for mixed lumped/wave S-parameters.


docs/src/config/boundaries.md
```diff
@@ -423,8 +423,15 @@ for the wave port problem.
 
 `"VoltagePath" [None]` :  Array of coordinate points defining an open path for computing a
 voltage line integral on the port face. Each element is a floating point array of length
-equal to the spatial dimension. When specified, enables voltage and characteristic impedance
-postprocessing for the wave port. Specified in mesh length units.
+equal to the spatial dimension. When specified, this serves two purposes: (1) it enables
+voltage and characteristic impedance ``Z_{PV}`` postprocessing for the wave port, and (2) it
+fixes the wave port mode polarity so that the integrated voltage ``\int E_{\text{mode}} \cdot dl`` along the path is real-positive. The path direction should be chosen so that it
```
hughcars: Nit: line length. 

Does this path only require 2 points for the purposes of this sign convention? I.e. could just a single point on the high and on the low be enough?

palace/models/postoperator.cpp
```diff
@@ -1250,55 +1251,69 @@ void PostOperator<solver_t>::MeasureSParameter() const
...
+    // S-parameter computation with per-port reference impedances (Kurokawa power-wave
+    // generalized S-parameters).
+    //
+    // Each port type uses its natural reference impedance: lumped ports use the user-
+    // specified R; wave ports implicitly use the line's characteristic impedance encoded
+    // in the unit-power normalization of the boundary mode (∫|E_mode×H_mode⋆|·n dS = 1).
+    //
+    // GetSParameter() returns the b-amplitude in Kurokawa convention for both port types:
+    //   - Lumped: ∫E·(E_inc/Z_s) dS = V/V_inc, where V_inc encodes the port's R
+    //   - Wave:   ∫(E×H_mode⋆)·n dS, the modal power-overlap with unit-power normalization
```
hughcars, line 1269: Does this convention have any impact on the question of peak/rms for the fields? We were previously thinking about adjusting that, but having the plain cross here without the 1/2 suggests there's a good reason to not have the half. Is the convention the same for comsol waveports?

palace/models/waveportoperator.cpp
```diff
@@ -588,6 +590,14 @@ WavePortData::WavePortData(const config::WavePortData &data,
       voltage_path.push_back(std::move(p));
     }
     voltage_n_samples = data.n_samples;
+
+    // Set up reverse transfer map (port submesh → parent mesh) and a parent-mesh
+    // GridFunction to receive the transferred mode field. This enables computing line
+    // integrals of the port mode E-field via GSLIB on the 3D parent mesh, since GSLIB
+    // requires SpaceDim == Dim and cannot work directly on the 2D-embedded port submesh.
```
hughcars, line 598: So this makes GSLIB back into a non-optional dependency, do you think there's a way of evaluating these without that? For a conventioning issue, GSLIB is a slightly hefty tool. Could it be provided by a pair of attributes? I.e. if Attribute X in the boundary is high, and Attribute Y is low, then any two elements from each in the boundary, their centroids could form the start and end locations of the 2 point path for defining the convention.

examples/coaxial/coaxial_lumped_wave.json
hughcars, line 1: Given this new example file, and how this new capability is going to be really important for interop with lumped/wave, could you add something to the example? A small section showing they can be mixed would help advertise the capability.

test/unit/test-waveportimpedance.cpp
```diff
@@ -0,0 +1,134 @@
...
+  WavePortOperator wave_port_op(iodata, mat_op, nd_fespace_palace.Get(),
+                                h1_fespace_palace.Get());
+  wave_port_op.SetSuppressOutput(true);
+
+  // Initialize the modes at the test frequency. Internal units: ω·tc. Initialize
+  // is per-port; const_cast around the GetPort() const is safe here because we
+  // know the wave_port_op is non-const in this scope.
+  const double omega_nondim =
+      2.0 * M_PI * iodata.units.Nondimensionalize<Units::ValueType::FREQUENCY>(f_GHz);
+  auto &port = const_cast<WavePortData &>(wave_port_op.GetPort(1));
```
hughcars: Not a fan of `const_cast` generally, but the trade here of avoiding the api growing and becoming more confusing is worth it. An alternative would've been to use an `Exposed` api to make a WavePortOperator which exposes more of the internal methods. Something like
```
struct ExposedWavePortOperator : public WavePortOperator
{ using WavePortOperator::WavePortOperator; }
```
which ought then to make everything private public so you can go in and do anything (e.g. call `Initialize` directly), without mutating the restricted public api.

hughcars (commented):
Generally LGTM. Would be good to demo the lumped/wave in the example given we can now mix and match, to advertise it given its going into the regression suite.

docs/src/config/boundaries.md
simlapointe: Split the line.

simlapointe: The path only requires 2 points for the purpose of fixing the sign. Actually for quasi-TEM I think the integral should be path independent so only two points are sufficient to get V and Z. For non TEM modes it's path specific though.

palace/models/postoperator.cpp
simlapointe: Shouldn't have an impact, we just have to make sure it's consistent (i.e. numerator/denom or S use the same convention, and lumped and wave ports use the same convention).

palace/models/waveportoperator.cpp
simlapointe: Yes, if you only want to fix the sign/polarity of the mode and you don't care for Z, then you can only specify the high/low attributes and use the E field at those two points. I made changes in https://github.com/awslabs/palace/pull/743/commits/6a92dbf3c2cd9522817fdd4167db83fd00fb0e50 enabling that, take a look. However, it requires the user to have distinct attributes for terminal and ground, which is not always the case (e.g. in our cpw and coaxial examples, we use a single attribute for all metal BCs).

examples/coaxial/coaxial_lumped_wave.json
simlapointe: Added a short description of the lumped/wave driven case in examples/coaxial.md in https://github.com/awslabs/palace/pull/743/commits/6a92dbf3c2cd9522817fdd4167db83fd00fb0e50

test/unit/test-waveportimpedance.cpp
simlapointe: Replaced the `const_cast` with your suggestion in https://github.com/awslabs/palace/pull/743/commits/6a92dbf3c2cd9522817fdd4167db83fd00fb0e50

palace/models/waveportoperator.hpp
```diff
@@ -122,6 +136,12 @@ class WavePortData
 
   void Initialize(double omega);
 
+  // Compute the sign of the modal E-field projected on the (low → high) direction
+  // implied by the given pair of parent-mesh boundary attributes. Returns +1, -1,
+  // or 0 if attributes were not resolvable on this mesh partition. Used by
+  // Initialize() to flip the mode polarity without requiring GSLIB.
+  [[nodiscard]] int GetModePolaritySign(int low_attr, int high_attr) const;
```
hughcars: Arg order here doesn't match the cpp.

palace/models/waveportoperator.hpp
```diff
@@ -105,6 +106,19 @@ class WavePortData
...
+  // mesh. Only allocated if the user configured a voltage path.
+  std::unique_ptr<mfem::ParTransferMap> port_nd_transfer_reverse;
+  std::unique_ptr<GridFunction> parent_E0t;
+
+  // Optional polarity attributes (parent-mesh boundary attrs [low, high]). When set
+  // and no VoltagePath is configured, the mode is flipped so that E points from low to
+  // high. Computed without GSLIB by evaluating the mode at port-submesh boundary
+  // element centroids adjacent to each attribute.
+  std::array<int, 2> polarity_attributes = {0, 0};
+  bool has_polarity_attributes = false;
```
hughcars: This is redunant, if both above are zero, it wasn't set, and thus there is no polarity attributes (I think I saw this test being used elsewhere).

palace/models/waveportoperator.cpp
```diff
@@ -588,6 +591,23 @@ WavePortData::WavePortData(const config::WavePortData &data,
...
+        mfem::ParSubMesh::CreateTransferMap(port_E0t->Real(), parent_E0t->Real()));
+  }
+
+  // Store polarity attributes (low, high) if provided. Used only when no VoltagePath is
+  // configured — VoltagePath already pins polarity unambiguously.
+  if (!has_voltage_coords && data.polarity_attributes[0] != 0 &&
+      data.polarity_attributes[1] != 0)
+  {
+    has_polarity_attributes = true;
+    polarity_attributes = data.polarity_attributes;
```
hughcars: Why can't this be a direct copy rather than an optional one, then the checking of these being mutually exclusive be done in the config? I'm also assuming the scenario of defining a voltage path integral different from the polarity attribution is going to be unneeded.

palace/models/waveportoperator.cpp
```diff
@@ -792,6 +838,102 @@ std::complex<double> WavePortData::GetSParameter(GridFunction &E) const
   return dot;
 }
 
+int WavePortData::GetModePolaritySign(int high_attr, int low_attr) const
```
hughcars, line 836: low, high in the header

hughcars (commented):
Some more tiny nits. I am fixing the spack cache issue in another PR, so don't worry about those failures for now, should be addressed by a rebase once I am done.

palace/models/waveportoperator.cpp
simlapointe: Indeed, we can check in the config and I don't foresee people wanting both. Now done in https://github.com/awslabs/palace/pull/743/commits/5cb1ba4d5e46bac50fb810b75edbb691594fe46f

hughcars (approved):
LGTM. Can merge once the spack fixes get in underneath this.

PR #746, opened 2026-06-02
Include excitation number in farfield output

### LINKED ISSUE #744: Far-field postprocessing not computed for all port excitations in multi-port driven simulations
## Description
When running a driven simulation with multiple lumped ports each having a unique `Excitation` integer (port sweep), the far-field postprocessing (`farfield-rE.csv`) is only computed for the first excitation source. The far-field data for subsequent excitations is not included in the output.

## Steps to Reproduce
1. Create a Palace JSON config with 2+ lumped ports, each with a unique `Excitation` integer:
```json
"LumpedPort": [
  { "Index": 1, "R": 50.0, "Excitation": 1, "Direction": "+Z", "Attributes": [200] },
  { "Index": 2, "R": 50.0, "Excitation": 2, "Direction": "+Z", "Attributes": [201] }
]
```

2. Run the simulation with far-field postprocessing enabled:
```json
"Boundaries": {
  "Postprocessing": {
    "FarField": { "Attributes": [1001], "NSample": 3600 }
  }
}
```

3. Check the output `postpro/farfield-rE.csv`

## Expected Behavior
The far-field should be computed for each excitation source in the port sweep. With N excitations and M far-field angles, the output should contain N×M rows per frequency or N files.

The S-parameter output (`port-S.csv`) correctly includes all excitations with columns like `|S[1][1]|`, `|S[2][1]|`, `|S[1][2]|`, `|S[2][2]|`. The VTU field output correctly creates separate `excitation_1/` and `excitation_2/` directories. Only the far-field CSV is affected.

The Palace source code (`palace/models/postoperator.hpp` line 257) shows `MeasureFarField()` is called inside `MeasureAllImpl()`, which is invoked per excitation. However, the far-field CSV writer (`palace/models/postoperatorcsv.cpp` line 618) opens the file once as `farfield-rE.csv` and appends data without an excitation index column. This means either:
1. The far-field is only measured for the first excitation (not all), or
2. The far-field is measured for all excitations but overwritten/appended without distinction

Sbozzolo:
`farfield-rE.csv` appended all the excitations in the same file without providing a clear way to distinguish them.

This PR adds the `exc` column so that they can be distinguished.

Note, this is a breaking change, as it changes the output format.

Closes #744


palace/models/postoperatorcsv.cpp
```diff
@@ -706,6 +712,7 @@ auto PostOperatorCSV<solver_t>::PrintFarFieldE(const SurfacePostOperator &surf_p
       farfield_E->table["f_re"] << measurement_cache.freq.real();
       farfield_E->table["f_im"] << measurement_cache.freq.imag();
     }
+    farfield_E->table["exc"] << static_cast<double>(m_ex_idx);
```
Sbozzolo, line 715: `Column::data` is a vector of doubles.

Sbozzolo:
Something wrong with spack :(

simlapointe (approved):

Issue #533, opened 2025-10-18
Unexpected factor 0.5 in series resistance for 2D sheet with Rs Impedance boundary


VolkerMuehlhaus:
I recently had a question on loss model for thick conductors with metal loss specified by conductivity and thickness on the surfaces, and now understand the Palace model and results for that. That part is fine.

This issue is about **thin sheets (2D)** that represent a resistor. I implemented the test case shown below, with a **resistor sheet** (single surface with impedance boundary) embedded into a lossless line. The aspect ratio of this shape is w/l=2 and the surface impedance mapped to this sheet is 10 Ohm/square, so that we expect 20 Ohm total for the resistor.

However, I get 10 Ohms only. Repeating the test with other Rs values in config.json show that I always get **half the expected resistance**. 

I evaluate the input impedance into the line at port 1, which is resistor value plus 50 Ohm from port 2 that terminates the line. **Expected 50 + 20 = 70 Ohm, but Palace result is 60 Ohm.** Geometries are small compared to wavelength, so this behaves as lumped elements more or less, no transmission line effects to be considered.

Is this a bug or do I miss something?

The mesh and Palace config file and results are attached.
[palace_resistor_simple_data.zip](https://github.com/user-attachments/files/22983427/palace_resistor_simple_data.zip)

~~

To give some context: This is for my [gds2palace workflow](https://github.com/VolkerMuehlhaus/gds2palace_ihp__sg13g2) that generated Palace models from GDSII layout files. If required, I can provide all the files to create the mesh and Palace config file. Everything works fine for regular conductors, now I wanted to add resistive sheets.

<img width="682" height="472" alt="Image" src="https://github.com/user-attachments/assets/351bc8c6-b4eb-4e3c-b28b-c7612400b439" />

<img width="742" height="477" alt="Image" src="https://github.com/user-attachments/assets/0d7ab875-4ed9-4c4f-a84e-c6bdd14eed32" />

<img width="1092" height="429" alt="Image" src="https://github.com/user-attachments/assets/38a7cc8c-591a-471c-a774-2fc16e562428" />

<img width="554" height="360" alt="Image" src="https://github.com/user-attachments/assets/9d2e268c-d101-4012-98bd-febf63a6479a" />

<img width="400" height="418" alt="Image" src="https://github.com/user-attachments/assets/6e01e0a5-0f74-409d-be9a-a07bd7814a7e" />

simlapointe:
Hi @VolkerMuehlhaus, 

[...]

PR #748, opened 2026-06-04
Per-attribute scaling of surface impedance boundary coefficients

parrangoiz:
Makes the scaling factor applied to sheet parameters (`Ls`, `Rs`, `Cs`) of impedance boundaries adjustable per-attribute rather than globally across the whole operator. This makes it possible to build models with impedance boundaries that have some attributes cracked (internal boundaries) and some attributes uncracked (external). I contributed a unit test for this as and checked the regression suite locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ The format of this changelog is based on
   - Fixed nondeterministic current-dipole assembly when the source lies on a
     shared mesh entity by distributing the underlying MFEM delta source over all
     containing elements.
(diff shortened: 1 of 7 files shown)


simlapointe (approved):
LGTM! Tested it on our cpw example (switching to an impedance BC on the metal) as well as the case from the original issue (#533) that prompted to implement this scaling and didn't see any regressions.

hughcars (changes requested):
Needs a CHANGELOG.md entry

hughcars (approved):

PR #752, opened 2026-06-08
fix: pass freq instead of omega to WriteParaviewFields in driven solver

### LINKED ISSUE #751: Paraview frequency output wrong for driven
The "time" field when reporting a driven simulation is currently reporting omega, rather than freq as expected.

saeekailas:
Fixes #751

In the driven solver's MeasureAndPrintAll, WriteParaviewFields was being 
called with omega.real() (angular frequency) instead of freq.real() (frequency 
in GHz). This caused ParaView's time axis to display wrong values.

The fix is a one-word change: omega.real() → freq.real(). The freq variable 
was already computed on the line above, and WriteMFEMGridFunctions in the 
same block was already using it correctly.


hughcars:
Hi, @saeekailas,

Thanks for the enthusiasm, but unfortunately the fix in this PR wouldn't work, if you look at the logged output on some of the unit test outputs, you would have some failures. We have another PR that I have submitted to fix it https://github.com/awslabs/palace/pull/754

PR #764, opened 2026-06-11
Fix check-test-tags CI job for multiline TEST_CASE

### LINKED ISSUE #757: check-test-tags job fails to pick up tags on multiple lines
The check-test-tags job checks that unit tests have at least one of the requisite tags (`[Serial]`, `[Parallel]`, etc.), but the grepping that's done assumes they all live in the same line. If the `TEST_CASE` description that precedes these tags is too long, the formatter splits that into >1 line and the tag check fails. We should should relax the grepping to work across multiple lines.

Sbozzolo:
In this PR, I replace the inline grep with a new script `scripts/check-test-tags`, which accumulates each
TEST_CASE/TEST_CASE_METHOD invocation across however many lines the formatter wrapped it onto before checking for a required tag. This also fixes TEST_CASE_METHOD, which the old `TEST_CASE(` grep never matched and so left entirely unchecked.

Closes #757

Example:
```
❯ ./scripts/check-test-tags test/
All test cases have required execution tags

~/repos/palace gbozzola/fix-check-test-tags-multiline ≡ ⇡
❯ ./scripts/check-test-tags test/
Error: Found TEST_CASE without required execution tags ([Serial], [Parallel], or [GPU]):
test/unit/test-rap.cpp:24
```


rupayon123:
I reviewed this locally for the multiline TEST_CASE tag issue.

Checks run:
- bash -n scripts/check-test-tags
- ./scripts/check-test-tags test/unit
- git diff --check

The script passes on the existing unit suite and handles already-wrapped TEST_CASE_METHOD calls such as test/unit/test-schema.cpp:94 and later wrapped ROM operator cases. Moving the workflow logic into scripts/check-test-tags also makes this much easier to exercise locally than the old inline grep chain.

I did not run the full C++ build/test suite; this review was limited to the tag-check script and workflow change.

laylagi:
Thanks, nice fix! 

I think there is one edge case worth tightening: the script currently checks the entire buffered macro text for `[Serial]`, `[Parallel]`, or `[GPU]`, rather than checking the actual Catch2 tag argument. That means a test can pass this validator even if the tag string is missing the required execution tag, as long as one of those strings appears in the test name or a comment.

For example, this diff still passes `scripts/check-test-tags test/unit`:

```
palace[gbozzola/fix-check-test-tags-multiline]$ git diff
diff --git a/test/unit/test-schema.cpp b/test/unit/test-schema.cpp
index 0ecef69b0..296d50dd1 100644
--- a/test/unit/test-schema.cpp
+++ b/test/unit/test-schema.cpp
@@ -92,7 +92,7 @@ TEST_CASE("Schema Validation - Example Configs", "[schema][Serial]")
 }
 
 TEST_CASE_METHOD(palace::test::PerRankTempDir, "Schema Validation - Config with Comments",
-                 "[schema][Serial]")
+                 "[schema]") // [Serial]
 {
   // Test that preprocessing (comment stripping) works with schema validation.
   auto temp_path = temp_dir / "palace_test_comments.json";
@@ -117,7 +117,7 @@ TEST_CASE_METHOD(palace::test::PerRankTempDir, "Schema Validation - Config with
   CHECK(err.empty());
 }
 
-TEST_CASE("Schema Validation - Invalid Config", "[schema][Serial]")
+TEST_CASE("Schema Validation - [Serial] Invalid Config", "[schema]")
 {
 
   SECTION("Missing required field")
palace[gbozzola/fix-check-test-tags-multiline]$ scripts/check-test-tags test/unit
All test cases have required execution tags
```

Could we tighten the checker to validate only the tag string argument?

scripts/check-test-tags
```diff
@@ -0,0 +1,87 @@
...
+
+# Accumulate each TEST_CASE / TEST_CASE_METHOD invocation across however many lines
+# clang-format wrapped it onto, then check the assembled text for a required tag. The
+# Allman brace style enforced by the formatter guarantees the macro signature is fully
+# closed before the opening "{" on its own line, which we use as the terminator.
+REQUIRED_TAGS='\[Serial\]|\[Parallel\]|\[GPU\]'
+
+MISSING=$(find "$TEST_DIR" -name '*.cpp' -print0 | xargs -0 awk '
+    # Skip single-line comments so commented-out test cases are ignored.
+    /^[[:space:]]*\/\// { next }
```
hughcars, line 60: Optional, cheap robustness: reset awk state at each file boundary. A macro never spans files, so this can never change correct behavior, but it stops a stray match (e.g. a string literal containing `TEST_CASE(`) from leaking buffer state into the next file — removing the script's hidden reliance on the separate clang-format job having already run.

```suggestion
    # Reset per file: a macro never spans files, so never carry buffer state across.
    FNR == 1 { buffering = 0; buf = "" }

    # Skip single-line comments so commented-out test cases are ignored.
    /^[[:space:]]*\/\// { next }
```

scripts/check-test-tags
```diff
@@ -0,0 +1,87 @@
...
+
+    # Start buffering when a (non-buffered) line opens a test-case macro.
```
hughcars: Worth noting the match is intentionally an unanchored substring: it also catches `TEMPLATE_TEST_CASE` (and other `*_TEST_CASE` macros) for free. Anchoring to a word boundary would silently stop checking those.

```suggestion
    # Start buffering on a test-case macro. Match is substring (unanchored) on
    # purpose: it also covers TEMPLATE_TEST_CASE and friends for free.
```

hughcars (commented):
A nice fix, two nits, but LGTM.

hughcars (approved):

laylagi (approved):

PR #765, opened 2026-06-12
Add semantic version to configuration JSON schema and coalesce schema to single file

### LINKED ISSUE #760: Add version to config schema
Currently, the configuration schema is implicitly versioned alongside Palace. Most of the time, we strive for great level of backward compatibility and this is not captured anywhere. 

Adding semantic versioning to the json schema also allows downstream tools to handle configuration compatibility better.

Sbozzolo:
The configuration schema was implicitly versioned alongside Palace, with no explicit marker of the configuration contract or its compatibility guarantees.

In this PR, I add a top-level "version" field (semantic versioning, starting at 0.1.0) to the root schema config-schema.json.

Closes #760


Sbozzolo:
cc: @ad-cqc

scripts/schema/config-schema.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
+  "version": "0.1.0",
```
ad-cqc: It should be something like `x-schema-version`. See https://json-schema.org/blog/posts/custom-annotations-will-continue

docs/src/developer/notes.md
```diff
@@ -155,6 +155,22 @@ to parse the configuration file and check that the fields are correctly specifie
...
+downstream tooling can reason about configuration compatibility.
+
+Because the schema is versioned independently, a single schema version generally spans
+several *Palace* releases. The table below records the first *Palace* release shipping
+each schema version; a schema version applies to that release and all later ones up to
+(but not including) the next entry. Add a row whenever the schema version is bumped.
+
+| Schema version | First *Palace* release |
+|:--------------:|:----------------------:|
+| `0.1.0`        | `0.17`                 |
```
hughcars: I suspect no, but is there a mechanism we can have for enforcing this pairing outside just updating here? I.e. scrape from the CHANGELOG or the like. Will we be updating the schema version inside of Palace versions? Or does a schema update have to correspond to a palace update. I would posit a schema update will bump the palace update, but not vice versa. What is the map from schema major/minor/patch to palace major/minor/patch? I think we need some clarity on how these move, you can possibly illustrate with some of the changes from the historical changelog.

hughcars (changes requested):
I think this needs some more clarity on how versioning interacts with palace versioning, but on board with the idea because of how useful it will be for downstream tooling.

scripts/schema/config-schema.json
Sbozzolo: Renamed

docs/src/developer/notes.md
ad-cqc: We could add a dedicated record about schema version update to CHANGELOG.md, or have a separate log file for the schema.

parrangoiz: I have all the same questions as @hughcars. We have to think more carefully about what the Palace semver will mean now, as historically any changes to it have always been dictated by changes to the config schema so the Palace version effectively signaled the state of the schema and nothing else (please correct me if I'm wrong). 

I'll note that there are other surfaces besides the config, such as metadata output (`palace.json`) and the whole slew of `.csv` outputs, any changes to which can also break downstream consumers and therefore we'd ideally start signaling changes to those as well. So maybe Palace version becomes a signal for changes to the inputs _and_ outputs while the schemaver is strictly for the config schema.

ad-cqc: How about to add a schema version into `palace --version` output (doesn't exist now)? So the version map between executable and config will be explicitly maintained in the code. If you bump the schema version, it results in new palace version. For now, it'd be joint but separate changes (modify schema JSONs and source code), but eventually when the schema is going to be based on code, we won't need to modify schema files anymore.

scripts/schema/config-schema.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
+  "x-schema-version": "0.1.0",
```
ad-cqc: BTW, we could use a urn with a schema version embedded into it as a proper schema identifier, `"$id": "urn:palace:schema:0.1.0"` which I think is better than a custom annotation field.

docs/src/developer/notes.md
Sbozzolo: > We have to think more carefully about what the Palace semver will mean now, as historically any changes to it have always been dictated by changes to the config schema so the Palace version effectively signaled the state of the schema and nothing else (please correct me if I'm wrong).

This is not correct. We bumped Palace version if anything in the input/output changed, as you suggest in the second part of your comment.

A change in the Palace version does not require a change in the schema, but a change in the schema required a change in the Palace version.

> How about to add a schema version into palace --version output (doesn't exist now)? 

Good idea!

Sbozzolo: I also added a section to explain the relationship between the two versions

Sbozzolo:
I:
- added more explanation
- added the version of the schema in `palace --version`
- changed the version to SchemaVer instead of SemVer
- changed the version to be an urn instead of an attribute
- added a CI test to check that the schama was bumped
- rebased on main

scripts/schema/config-schema.json
ad-cqc: I didn't account for `$id` to be used as a root location for a subschema path resolution.  The [JSON schema standard](https://json-schema.org/draft-07/draft-handrews-json-schema-01) defines `$id` (see 8.2) as the base URI that other URI references within the schema are resolved against. This messes up the subschema file path construction with error: `Cannot add a path (./config/boundaries.json) to an URN URI (urn:palace:schema:1-0-0)`.

So, we only would be able to use `$id` with a versioned URN if the schema is a single file which would not require subschema file lookup. Or, go back to `x-schema-version` custom attribute. In https://github.com/awslabs/palace/pull/723, I emit single file with `"$id": "urn:palace:schema:0.1.0"`, but I could switch to `x-schema-version`.

Sbozzolo:
I collapsed the various schema files into a single json

palace/utils/jsonschema.cpp
```diff
@@ -22,31 +23,14 @@ constexpr const char *root_schema_file = "config-schema.json";
...
-    auto it = schema_map.find(path);
-    if (it == schema_map.end())
-    {
-      throw std::runtime_error("Schema not found: " + path);
-    }
-    schema = json::parse(it->second);
-    return schema;
+    throw std::runtime_error("Unexpected external $ref: " + uri.url());
   }
 };
```
hughcars: Given this, what's the point of this function now? Can it be deleted?

hughcars (commented):
Nothing glaring to me, I'd like @ad-cqc to look again as the schema expert.

test/unit/test-schema.cpp
```diff
@@ -712,3 +714,34 @@ TEST_CASE("Schema Validator Smoke Tests", "[schema][Serial]")
...
+  // The root schema must carry a version via the standard "$id" field using a URN
+  // (see issue #760). Format: "urn:palace:schema:MODEL-REVISION-ADDITION" (SchemaVer).
+  const auto &schema_map = schema::GetSchemaMap();
+  auto it = schema_map.find("config-schema.json");
+  REQUIRE(it != schema_map.end());
+  const json schema = json::parse(it->second);
+  REQUIRE(schema.contains("$id"));
+  REQUIRE(schema["$id"].is_string());
+  const std::string id = schema["$id"];
+  INFO("Schema $id: " << id);
```
ad-cqc: Why not use `GetSchemaVersion` here and test as well?

ad-cqc (approved):
Schema is fine.

test/unit/test-schema.cpp
Sbozzolo: Good point

hughcars (approved):

PR #768, opened 2026-06-15
Cache GSLIB point locator for wave-port voltage-path integrals

simlapointe:
`WavePortData::GetExcitationVoltage` / `GetVoltage` compute the mode voltage as a line integral along the configured `VoltagePath`, evaluated via GSLIB point interpolation on the 3D parent mesh. The underlying `fem::ComputeLineIntegral` -> `InterpolateFunction` built a fresh `mfem::FindPointsGSLIB` and called `Setup()` on every invocation. That `Setup` is the expensive step — it builds a spatial hash over all parent-mesh elements (O(num_elements)) — and it depends only on the mesh, not the field values, so rebuilding it per call is pure waste.

This PR creates the setup `FindPointsGSLIB` only once at the wave port construction and reuses it. This is a performance improvement for adaptive sweeps with wave ports, results and `VoltagePath` UI should be unchanged.

### DIFF
--- a/palace/models/waveportoperator.hpp
+++ b/palace/models/waveportoperator.hpp
@@ -112,6 +112,15 @@ private:
   std::unique_ptr<mfem::ParTransferMap> port_nd_transfer_reverse;
   std::unique_ptr<GridFunction> parent_E0t;
 
+  // Cached GSLIB point locator for the voltage-path line integrals, Setup once on the
+  // (fixed) parent mesh. The geometric Setup is the dominant cost, so reusing it across
+  // every Initialize() avoids an O(num_elements) hash rebuild per frequency — critical for
+  // the circuit-synthesis dispersion fit, which evaluates the mode at many frequencies.
+  // Only allocated if the user configured a voltage path.
+#if defined(MFEM_USE_GSLIB)
+  std::unique_ptr<mfem::FindPointsGSLIB> voltage_gslib_op;
+#endif
+
   // Optional polarity attributes (parent-mesh boundary attrs [high, low], signal
   // first, ground second). When non-zero (i.e. set by the user) the mode is flipped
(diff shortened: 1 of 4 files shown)


palace/models/waveportoperator.hpp
```diff
@@ -112,6 +112,15 @@ class WavePortData
   std::unique_ptr<mfem::ParTransferMap> port_nd_transfer_reverse;
   std::unique_ptr<GridFunction> parent_E0t;
 
+  // Cached GSLIB point locator for the voltage-path line integrals, Setup once on the
+  // (fixed) parent mesh. The geometric Setup is the dominant cost, so reusing it across
+  // every Initialize() avoids an O(num_elements) hash rebuild per frequency — critical for
+  // the circuit-synthesis dispersion fit, which evaluates the mode at many frequencies.
+  // Only allocated if the user configured a voltage path.
+#if defined(MFEM_USE_GSLIB)
+  std::unique_ptr<mfem::FindPointsGSLIB> voltage_gslib_op;
```
laylagi, line 121:
```suggestion
  mutable std::unique_ptr<mfem::FindPointsGSLIB> voltage_gslib_op;
```

Thanks, LGTM. Minor - `mutable` makes the modification through a `const` method intentional.

laylagi (approved):

palace/models/waveportoperator.hpp
simlapointe: Not sure we need `mutable` here since the `const` methods that use `voltage_gslib_op` don't modify the pointer itself (what changes is the pointee's internal GSLIB state, which `mutable` doesn't control).

laylagi: Oops, that was confusing. I had somehow dropped "intentional" from the end of the sentence. You are right, we don't need it. I'm fine either way. Nice improvement, thanks!

Issue #642, opened 2026-02-18
Option for RLC series in lumped port


krono-i2:
In the current implementation, lumped port boundary conditions use an equivalent circuit modeled as a parallel RLC network. However, many electronic components (e.g., PIN diodes) are more accurately described by a series RLC equivalent circuit.
It would be highly beneficial to provide users with the option to choose between parallel and series RLC representations when defining lumped ports. This enhancement would improve modeling flexibility and allow a more accurate representation of real-world components.

Proposed feature:
- Add a parameter to select either parallel or series RLC equivalent circuits for lumped ports.

Thanks for considering this feature request!

krono-i2:
Someone interested on that? It should be simple, but it would be better to receive a feedback from the core developers.

hughcars:
Hi @krono-i2, 

That could be an interesting feature. If you are interested in giving an implementation a shot, we would be interested in a PR. My guess is the circuit description itself oughtn't to be too complicated, but the UI choices might need some careful though. If you're feeling up to it, I'm happy to review a PR.

krono-i2:
Hi @hughcars, thanks for the feedback! I went ahead and opened a PR with a different implementation.

While working on it, I also got a better understanding of why supporting a series RLC — or more generally an arbitrary impedance — is non-trivial in this framework. The core issue is that the current FEM assembly distributes R, L, and C as independent contributions to the K, C, and M matrices respectively. This works naturally for a parallel combination, where admittances add and each branch maps cleanly to one matrix. A series combination, on the other hand, produces an impedance Z = R + iωL + 1/(iωC) whose inverse does not decompose into frequency-independent terms — so it cannot be represented as a sum of polynomial-in-ω contributions to K, C, M without frequency-dependent coefficients.

This also has implications for driven frequency sweeps: the current architecture assembles the system matrices once and sweeps ω algebraically, which only works because the frequency dependence is polynomial. Supporting a general Z(ω) would require reassembling (or at least updating) matrix entries at each frequency point, which is a more significant architectural change.

One possible direction worth exploring could be introducing a dedicated frequency-dependent boundary matrix that allows arbitrary Z(ω), assembled separately from K, C, M and evaluated at each frequency step. This would be a natural extension of the existing `GetExtraSystemMatrix` infrastructure. Happy to discuss further if this seems like a worthwhile direction!

PR #770, opened 2026-06-17
Add rational surface impedance boundary condition

### LINKED ISSUE #553: Allow frequency-dependent lumped port impedance (vector input)
### 🚀 Feature Request: Allow frequency-dependent lumped port impedance (vector input)

**Summary**

Currently, AWS Palace requires a *fixed* RLC definition for lumped ports.  
This limitation prevents accurate modeling of components whose impedance varies significantly with frequency — such as varactor diodes, tunable matching networks, or MEMS capacitors.  

I propose enabling the lumped port definition to accept **frequency-dependent impedance data**, either directly as a vector (Z(f)) or via an imported file.  
During a frequency sweep, Palace would then interpolate the provided impedance values to apply the correct boundary condition at each frequency point.

---

### 📚 Background and Motivation

In RF and microwave simulations, it is common to model devices such as **varactor diodes** or **nonlinear tuning elements** using equivalent small-signal impedances extracted from circuit models or measurements.  
These impedances are *strongly frequency-dependent* even under a fixed bias condition due to parasitics (series Ls, Cp, Rs, etc.).

When performing a full-wave sweep in Palace, we can only specify a single R, L, and C value.  
This forces the user to fit a static equivalent circuit over a finite band which introduces significant errors outside the narrow fit range.

In practical use cases (e.g., varactor-tuned filters, reconfigurable antennas, or harmonic balance verification), this becomes a serious limitation.

---

### 🔍 Desired Behavior

The lumped port definition should optionally accept a **frequency-dependent impedance**.

krono-i2:
This PR adds a `RationalImpedance` boundary condition: a surface (Robin) impedance boundary whose per-square impedance is an arbitrary rational function of frequency, `Zs(s) = N(s)/D(s)` with `s = iω`, given by numerator and denominator polynomial coefficients. It generalizes the parallel-RLC `Impedance` boundary to any passive lumped-network response. It closes https://github.com/awslabs/palace/issues/553 and https://github.com/awslabs/palace/issues/642.


palace/models/spaceoperator.cpp
```diff
@@ -195,8 +198,9 @@ void SpaceOperator::CheckBoundaryProperties()
   {
     aux_bdr_marker[i] =
         (dbc_marker[i] || farfield_marker[i] || surf_sigma_marker[i] ||
-         surf_z_Rs_marker[i] || surf_z_Ls_marker[i] || lumped_port_Rs_marker[i] ||
-         lumped_port_Ls_marker[i] || wave_port_marker[i] || floquet_port_marker[i]);
+         surf_z_Rs_marker[i] || surf_z_Rs_marker[i] || surf_z_Ls_marker[i] ||
```
laylagi:
```suggestion
         surf_z_Rs_marker[i] || surf_rz_marker[i] || surf_z_Ls_marker[i] ||
```

Was this intentional? `surf_z_Rs_marker` appears twice, while `surf_rz_marker` is defined above and used later for mutual-exclusion checking. Should one of those entries be `surf_rz_marker[i]`?

palace/models/surfacerationalimpedanceoperator.cpp
```diff
@@ -0,0 +1,224 @@
...
+    for (auto attr : mesh.bdr_attributes)
+    {
+      bdr_attr_marker[attr - 1] = 1;
+    }
+    std::set<int> bdr_warn_list;
+    for (const auto &data : impedance)
+    {
+      for (auto attr : data.attributes)
+      {
+        MFEM_VERIFY(!impedance_marker[attr - 1],
```
laylagi, line 95: Could we reorder this validation so we check that `attr` is in range and exists on the mesh before indexing `impedance_marker[attr - 1]`? For example:

```cpp
    if (attr <= 0 || attr > bdr_attr_max || !bdr_attr_marker[attr - 1])
    {
      bdr_warn_list.insert(attr);
      continue;
    }

    MFEM_VERIFY(!impedance_marker[attr - 1],
                "Multiple definitions of rational impedance boundary properties for "
                "boundary attribute " << attr << "!");
    impedance_marker[attr - 1] = 1;
```

The duplicate-definition check is useful, but it should only run for valid attributes. Otherwise invalid input like boundary attribute `0` or an attribute larger than `bdr_attr_max` can index out of bounds before the warning/ignore path runs.

I noticed a couple of existing surface-BC operators have the same ordering issue, so those probably deserve a follow-up cleanup, but I don't think we should copy that pattern into new code.

palace/models/surfacerationalimpedanceoperator.cpp
```diff
@@ -0,0 +1,224 @@
...
+      const double f_ghz = omega * freq_scale / (2.0 * M_PI);
+      Mpi::Warning("Rational impedance boundary (attribute {:d}) is not passive at "
+                   "f = {:.4f} GHz: Re(Zs) = {:.3e} < 0!\n",
+                   bdr.attr_list.Size() ? bdr.attr_list[0] : -1, f_ghz, Z.real());
+      bdr.warned_passivity = true;
+    }
+    for (auto attr : bdr.attr_list)
+    {
+      const double sc = bdr.attr_scaling.at(attr);
+      const std::complex<double> coef = s / (Z * sc);  // iω·Ys per square
```
laylagi: Could we compute the coefficient directly as `s * D / (N * sc)` instead of forming `Z = N / D` and then computing `s / Z`? The two forms are algebraically equivalent when `N(s)` and `D(s)` are nonzero, but the direct admittance form is more robust. At an impedance pole where `D(s) = 0` and `N(s) != 0`, `Zs = N / D` is infinite, while the Robin contribution `s / Zs = s * D / N` should go to zero. Forming `Z` first can introduce `inf`/`nan` intermediates unnecessarily.

palace/models/surfacerationalimpedanceoperator.cpp
```diff
@@ -0,0 +1,224 @@
...
+                   "ignore them!");
+      utils::PrettyPrint(bdr_warn_list, "Boundary attribute list:");
+      Mpi::Print("\n");
+    }
+  }
+
+  // A rational surface impedance contributes only to the frequency-dependent system matrix
+  // A2(ω); it has no time-domain (transient) or static realization here.
+  MFEM_VERIFY(impedance.empty() || problem_type == ProblemType::DRIVEN ||
+                  problem_type == ProblemType::EIGENMODE ||
```
laylagi: Could you clarify and validate eigenmode support for `RationalImpedance`?
As implemented, this BC contributes through `A2(omega)`, so eigenmode solves trigger the nonlinear eigenmode path. The nonlinear paths I checked evaluate `A2` using `abs(Im(lambda))`, rather than the full complex eigenvalue. 

I tried adapting the PR’s parallel RLC example to an eigenmode config using the same equivalent rational coefficients:

```
Numerator   = [2.2588181359893346e-06, 0.0]
Denominator = [9.542690323697736e-20, 5.995849163282243e-09, 376.730313668]
```

This case should reduce to `s / Zs(s) = 1/Ls + (1/Rs)s + Cs s^2`, but the rational version still triggered nonlinear refinement through `A2(omega)` and did not converge reliably.

Could cases where `s / Zs(s)` reduces to a quadratic polynomial be assembled into K/C/M instead of `A2(omega)`? For genuinely nonquadratic rational impedances, I think we need an eigenmode validation test or documented limitations.

palace/models/surfacerationalimpedanceoperator.cpp
```diff
@@ -0,0 +1,224 @@
...
+                  problem_type == ProblemType::BOUNDARYMODE,
```
laylagi, line 117: Could you double-check whether `ProblemType::BOUNDARYMODE` should be allowed here? I may be missing it, but I don’t see `SurfaceRationalImpedanceOperator` constructed or used `BoundaryModeOperator`.

If rational impedance is not applied in boundary-mode solves yet, maybe we should remove `ProblemType::BOUNDARYMODE` from this allowed list for now, or add the missing wiring if support is intended.

laylagi (commented):
Thanks for putting this together. This is a useful addition, and the driven RLC comparison is a nice way to validate the basic behavior. I left a few comments, mostly around edge cases and solver integration.

palace/models/surfacerationalimpedanceoperator.cpp
simlapointe: Heads up that an upcoming change affects this. Once #778 is merged, `A2` will be evaluated using the full complex eigenvalue, $\lambda$, instead of $|Im(\lambda)|$.

palace/models/spaceoperator.cpp
krono-i2: Not intentional: that was a copy-paste typo. The second `surf_z_Rs_marker[i]` should be `surf_rz_marker[i]`, so rational-impedance boundaries are included in the auxiliary-space marker like the resistive/inductive ones, consistent with the mutual-exclusion check below. Fixed.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Agreed: the range/existence check now runs before indexing `impedance_marker[attr - 1]`, so an out-of-range attribute (e.g. `0` or `> bdr_attr_max`) takes the warn/ignore path instead of indexing out of bounds. Noted on the existing operators too: I've kept the new code clean and left their ordering for a separate follow-up rather than copying the pattern.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Good point: switched to the direct admittance form. I compute `Ys = D / N` once per boundary (it's loop-invariant over the attributes) and form the coefficient as `s * Ys / sc`, so at an impedance pole (`D = 0`, `N != 0`) the contribution is exactly zero with no `inf`/`nan` intermediate. The passivity check now tests `Re{Ys}` (same sign as `Re{Zs}`), and `Zs = 1/Ys` is only formed inside the warning branch, where `Ys` is guaranteed nonzero.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Thanks both. You're right that as-is the contribution goes through `A2(ω)` and triggers the nonlinear eigensolver, and since that path evaluates `A2` at `|Im(λ)|` the result for a lossy rational impedance is only approximate, which matches the unreliable convergence you saw. For this PR I've restricted `RationalImpedance` to the frequency-domain driven solver, where the per-frequency `A2(ω)` is exact, and updated the docs and schema accordingly.

@simlapointe: evaluating `A2` at the full complex eigenvalue λ is exactly what makes a rational impedance correct in eigenmode, since `Zs(λ)` is then the physical surface impedance at the true (complex) eigenfrequency rather than a real-axis approximation. So #778 removes the main obstacle here.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: You're right: `SurfaceRationalImpedanceOperator` isn't wired into the boundary-mode (2D wave-port) solve, and unlike `Conductivity` (which maps to PEC there) a rational impedance has no meaningful reduction in that eigenproblem. I'd copied the allowed problem-type set from `SurfaceConductivityOperator` without checking the wiring. Removed `BOUNDARYMODE` (and `EIGENMODE`, per the eigenmode thread). It's now `DRIVEN` only.

krono-i2:
Rebased onto current `main` (past the `v0.17.0` tag) and force-pushed.

One change worth flagging: the per-file schema configs were consolidated upstream into `scripts/schema/config-schema.json`, so I moved the `RationalImpedance` definitions there — both the `Boundaries.properties` entry and the `$defs/RationalImpedance` block — instead of the now-removed `scripts/schema/config/boundaries.json`.

I also bumped the schema `$id` from `1-0-0` to `1-1-0`: a new optional boundary block is a backward-compatible model extension, so I classified it as a REVISION and added the matching row to the SchemaVer table in `docs/src/developer/notes.md`.

Two things I'd like a maintainer to confirm:
- The "First Palace release" column for the `1-1-0` row — I put `0.17.1` as a placeholder since I can't know the target release; please correct it.
- The versioning policy couples a REVISION to a Palace version bump — happy to handle that however you prefer, or leave it for release.

I validated `examples/rational_impedance/parallel_rlc_rational.json` against the updated schema (passes). Ready for another look.

krono-i2:
Side note, unrelated to this PR's scope: while validating configs against the consolidated schema I hit what looks like a pre-existing bug in `scripts/schema/ValidateConfig.jl`. At line 18 it `cd`s into `dirname(schema_file)` and then re-opens `schema_file` by its original relative path, so a relative `schema_file` fails with `SystemError: opening file "scripts/schema/config-schema.json": No such file or directory`; an absolute path works around it. Happy to open a separate issue (or small PR) if useful — just didn't want to widen this PR's scope.

palace/models/surfacerationalimpedanceoperator.hpp
```diff
@@ -0,0 +1,84 @@
...
+//     Zs(s) = N(s) / D(s),    s = iω,
+//
+// with N and D real polynomials given as coefficient lists (highest-degree-first). The
+// roots of N are the zeros and the roots of D are the poles of Zs, so any passive lumped
+// (RLC / network) response can be represented. Unlike the parallel-RLC impedance handled by
+// SurfaceImpedanceOperator (which maps onto the constant stiffness/damping/mass matrices),
+// a general Zs(ω) is an arbitrary function of frequency and therefore contributes to the
+// frequency-dependent "extra" system matrix A2(ω), exactly like the finite-conductivity
+// Robin BC (SurfaceConductivityOperator). It is consequently available only for
+// frequency-domain problem types (driven, eigenmode, boundary mode).
```
laylagi:
```suggestion
// frequency-domain driven problems.
```

palace/models/surfacerationalimpedanceoperator.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+void SurfaceRationalImpedanceOperator::AddExtraSystemBdrCoefficients(
+    double omega, MaterialPropertyCoefficient &fbr, MaterialPropertyCoefficient &fbi)
+{
+  // Rational surface impedance boundaries Zs(s) = N(s)/D(s) with s = iω. The Robin BC adds
+  // the term iω / Zs(iω) per square to the frequency-dependent system matrix A2(ω), exactly
+  // as the parallel admittance Ys = 1/Rs + 1/(iωLs) + iωCs contributes iω·Ys to
+  // K + iωC - ω²M. Coefficients are stored nondimensionalized and highest-degree-first.
+  if (omega == 0.0)
```
laylagi, line 192: One question about the `omega == 0.0` case: can `RationalImpedance` safely assume this contribution is always zero?

It looks like driven samples can include `MinFreq: 0.0`, so could this branch be reachable? Since the coefficient is `s * D(s) / N(s)`, are there valid rational impedances where the `s -> 0` limit is finite and nonzero, for example when `N(s)` has a first-order zero at `s = 0`?

If so, should zero-frequency samples be rejected or handled with a separate limit evaluation?

laylagi (commented):
Thanks, this looks good overall. I left one small change request and one question below.

palace/models/surfacerationalimpedanceoperator.hpp
krono-i2: Good catch — that comment was stale from before the driven-only restriction. Applied.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Good catch on reachability. I've replaced the unconditional early return with the correct DC limit of the Robin coefficient iω/Zs = s·D(s)/N(s) as s→0:
- N(0) ≠ 0 (finite Zs at DC): limit 0, no contribution.
- N(0) = 0, simple transmission zero of Zs at s=0: limit D(0)/N'(0), assembled. For the parallel-RLC example N(s)=Ls·Rs·s, this gives 1/Ls, matching the DC Robin coefficient iω·Ys = iω/(iωLs) = 1/Ls of the lumped equivalent.
- N(0) = 0, double/higher zero: limit diverges (non-physical at DC), now rejected via MFEM_VERIFY.

One clarification from testing: solving at exactly f=0 is ill-posed in the frequency-domain formulation itself, independent of the boundary condition. At ω=0 only K remains (no ω²M regularization); the curl-curl operator is singular on the gradient null space and SLEPc aborts with a nan. I confirmed this fails identically for a purely resistive termination (L=C=0), so it isn't specific to reactive or rational boundaries. This fix therefore makes the ω=0 coefficient correct and stops silently dropping a real low-frequency contribution, but it doesn't make f=0 solvable — that's a global property of the driven formulation. The example sweep is kept at MinFreq > 0 accordingly.

palace/models/surfacerationalimpedanceoperator.cpp
simlapointe: Nothing to do right now, but if #778 lands in `main` first, it might be worth revisiting how eigenmode simulations with rational impedance BCs behave, and if consider eigenmode + rational should be allowed. I feel like it should at least be allowed in the case where `Zs(λ)` is a constant, linear, or quadratic function. When `Zs(λ)` is not a quadratic function we should test how our nonlinear eigensolver performs.

palace/models/surfacerationalimpedanceoperator.cpp
simlapointe: `BOUNDARYMODE` simulations already support traditional impedance BCs and conductivity BCs. See `surf_z_op` and `surf_sigma_op` in `boundarymodeoperator.cpp` and `modeeigensolver.cpp`. The boundary mode problem is solved at a fixed frequency so the frequency dependence of the rational impedance BC should not be an issue here (unlike in `EIGENMODE` simulations). I think we could add a `surf_rz_op` in `BoundaryModeOperator` that ends up contributing to `Att` and `Ann` (also need to add rational impedance attributes to `internal_bdr_attrs` for the submesh extraction).

simlapointe (changes requested):
Nice contribution, @krono-i2. @laylagi already brought up a lot of good points, I just have a few small things:

- Documentation entries for the new boundary condition should be added to `docs/src/config/config.md` and `docs/src/guide/boundaries.md`.
- It's missing an entry in `CHANGELOG.md` in the "New Features" section.
- We should add unit and regression tests. I think we could have the following: 1) config file additions should be tested in the `test-config.cpp` unit test, 2) unit test(s) similar to `test-surfaceimpedanceoperator.cpp`, and 3) a regression test based on `examples/rational_impedance`. Note that our regression tests don't use julia anymore, see `test/data/regression` and `test/unit/regression/cases.cpp` to see how examples are wired into the regression test suite.
- The new BC is not currently compatible with the circuit synthesis functionality (`"AdaptiveCircuitSynthesis": true`). That's ok for now, but we should add a guard, probably in `iodata.cpp` lines 287-293.
- I don't think the justification to not include the rational impedance BC in `BoundaryMode` simulations is quite right. I think we can enable the new BC in boundary mode, see my comment in the previous review [here](https://github.com/awslabs/palace/pull/770#discussion_r3514847249).
- Did you verify that adaptive driven sweeps work and give the same results as uniform driven sweeps? I think it should, but would be nice to confirm. You can just run your example case with `"Solver.Driven.AdaptiveTol"` set to something like 1e-3 and check that the S-params vs freq results compare well to the uniform sweep.
- As I mentioned in a previous [comment](https://github.com/awslabs/palace/pull/770#discussion_r3514830876), we might want to revisit whether eigenmode simulations should be allowed if #778 lands first.

krono-i2:
@simlapointe,thanks for the thorough review! Status of the points:

- **Docs**: the per-parameter reference is auto-generated from the schema since #716, and the `RationalImpedance` descriptions are already in `boundaries.json`, so the reference entry is covered; I've added the hand-written section to `docs/src/guide/boundaries.md`.
- **CHANGELOG**: added under a new "In progress" section.
- **Tests**: added (1) config parsing + schema-coverage tests in `test-config.cpp`, (2) `test-surfacerationalimpedanceoperator.cpp` validating the assembled coefficient against the analytic parallel-RLC admittance (including the ω = 0 limit, impedance-pole robustness, and cracked-attribute scaling, plus rejection of invalid inputs), and (3) regression cases `rational_impedance_uniform` / `rational_impedance_adaptive` wired into the new regression suite with reference data.
- **Circuit synthesis guard**: added to the `AdaptiveCircuitSynthesis` compatibility check in `iodata.cpp`.
- **Adaptive vs uniform**: verified — with `"AdaptiveTol": 1e-3` the adaptive driven sweep reproduces the 81-point uniform sweep S-parameters to machine precision (max |Δ|S11|| ≈ 6e-8 dB across all 81 frequency points). The adaptive path with the rational impedance BC is also exercised by the new `rational_impedance_adaptive` regression case.
- **Boundary mode**: found your linked comment with the wiring recipe after posting this --- implementing it in this PR: a `surf_rz_op` in `BoundaryModeOperator` (and the wave-port 2D path) contributing to `Att`/`Ann` like `Conductivity`, plus the rational impedance attributes in `internal_bdr_attrs` for the submesh extraction. You're right that the fixed operating frequency makes the frequency dependence unproblematic there, unlike eigenmode. Will push shortly.
- **Eigenmode**: agreed — I'd revisit once #778 lands, evaluating `Zs(λ)` at the full complex eigenvalue; happy to do that as a follow-up.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Agreed on all points. Happy to do that as a follow-up once #778 lands. One small precision on the foldable case: the condition is on s·Ys(s) = s/Zs(s) being a polynomial of degree ≤ 2 (the parallel-RLC family, Ys = 1/R + 1/(sL) + sC), rather than on the degree of Zs itself (e.g. a series R+sL gives s/(R+sL)), which is not polynomial and would still go through the nonlinear path. The plan would be: detect the polynomial reduction and fold it into K/C/M (standard quadratic EVP, validated against the existing Impedance boundary in eigenmode), and benchmark the nonlinear eigensolver with the full complex-λ A2 evaluation for genuinely rational Zs.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Reopening --- implementing exactly this. Will follow up here with the commit.

krono-i2: Done in [a24b34d](https://github.com/awslabs/palace/pull/770/commits/a24b34d74c6e31dfa542ca1dd6d9ed612673ec32): added `surf_rz_op` to `BoundaryModeOperator` and threaded it through
`mode_assembly::AssembleAtt`/`AssembleAnn` — both the system matrices and the
preconditioner assembly — mirroring the `Conductivity` treatment. The wave-port 2D path
gets a `port_surf_rz_op` likewise, and the rational impedance attributes are now included
in `internal_bdr_attrs` for the submesh extraction. `BOUNDARYMODE` is allowed again in
the operator, and the docs/schema availability notes are updated. Existing unit and
regression tests still pass (the ModeEigenSolver unit tests needed the new constructor
argument threaded through as well).

One question: would you like a boundary-mode regression case with a rational impedance BC
in this PR (e.g. a variant of one of the 2D mode cases), or is that better as a follow-up?

simlapointe: Thanks for the changes. One remaining thing for the BC to fully work in BoundaryMode: `palace/drivers/boundarymodesolver.cpp` was not touched. Its 3D→2D submesh extraction appends pec/auxpec/impedance/conductivity/farfield attributes to internal_bdr_attrs (lines ~115-125) but not rational_impedance. It should append rational impedance bc there too!

palace/models/surfacerationalimpedanceoperator.cpp
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  // A rational surface impedance contributes only to the frequency-dependent system matrix
+  // A2(ω) in 3D simulations; it has no time-domain (transient) or static realization here.
+  // In boundary mode (and the wave port 2D eigenproblem) the operating frequency is fixed,
+  // so Zs(iω) is simply evaluated there.
+  MFEM_VERIFY(impedance.empty() || problem_type == ProblemType::DRIVEN ||
+                  problem_type == ProblemType::BOUNDARYMODE,
+              "Rational impedance boundaries are only available for frequency-domain "
+              "driven and boundary simulation types!");
```
simlapointe: Should say "boundary mode" instead of just "boundary"

scripts/schema/config-schema.json
simlapointe: We should add `RationalImpedance` to the list of BCs not supported in circuit synthesis

examples/rational_impedance/parallel_rlc_lumped.json
simlapointe, line 1: We now have four configs for the rational_impedance example/regression test: these two in examples and the two in test/data/regression/input/rational_impedance. I think we could forego the adaptive test now that we tested the adaptive sweep works fine. And the config in test/data/regression/input/ could be a symlink to the one in example/rational_impedance, like we do for most other regression tests based on examples.

test/data/regression/input/rational_impedance/rational_adaptive.json
simlapointe, line 1: As mentioned above, we can maybe get rid of this adaptive case

test/data/regression/input/rational_impedance/rational_uniform.json
simlapointe, line 1: This file could be a symlink to the config in examples/rational_impedance to avoid duplications

test/unit/regression/cases.cpp
```diff
@@ -338,6 +338,25 @@ TEST_CASE("coaxial_lumped_wave", "[Serial][Parallel][GPU][Regression]")
...
+{
+  palace::test::RegressionOptions opts;
+  opts.rtol = 2.0e-2;
+  opts.atol = 1.0e-11;
+  opts.excluded_columns = {"Maximum", "Minimum"};
+  palace::test::RunRegressionCase("rational_impedance", "rational_uniform.json", "uniform",
+                                  opts);
+}
+
+TEST_CASE("rational_impedance_adaptive", "[Serial][Parallel][GPU][Regression]")
```
simlapointe: Let's maybe remove this adaptive case since it works fine and we didn't touch the code in romoperator.cpp

palace/utils/iodata.cpp
simlapointe, line 313: We should probably add a warning like this for rational impedance BCs, since they're not support in electrostatic

simlapointe, line 346: Add a similar warning for rational impedance

simlapointe, line 374: Add a similar warning for rational impedance

simlapointe (changes requested):
Thanks for the changes! A few more small things below and one thing I noted in a separate comment [here](https://github.com/awslabs/palace/pull/770#discussion_r3531965108) for the boundary mode solver.

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Fixed, thanks. 0bb257d8e

scripts/schema/config-schema.json
krono-i2: Added RationalImpedance to the description. 0bb257d8e

examples/rational_impedance/parallel_rlc_lumped.json
krono-i2: Done: dropped the adaptive regression case entirely (input config, reference data, and
TEST_CASE), and the regression input is now a symlink to
examples/rational_impedance/parallel_rlc_rational.json like the other example-based
regression cases. Reference data regenerated for the full sweep. 0bb257d8e

test/data/regression/input/rational_impedance/rational_adaptive.json
krono-i2: Removed (see thread above). 0bb257d8e

test/data/regression/input/rational_impedance/rational_uniform.json
krono-i2: Replaced with a symlink to the example config (see thread above). 0bb257d8e

test/unit/regression/cases.cpp
krono-i2: Removed; the remaining case now runs the example config via the symlink. 0bb257d8e

palace/utils/iodata.cpp
krono-i2: Added, mirroring the conductivity warning. 0bb257d8e

krono-i2: Added, mirroring the conductivity warning. 0bb257d8e

krono-i2: Added, mirroring the conductivity warning. 0bb257d8e

palace/models/surfacerationalimpedanceoperator.cpp
krono-i2: Good catch, the standalone boundary-mode driver has its own 3D→2D submesh extraction that I'd missed when wiring the wave-port path; rational impedance attributes are now appended to internal_bdr_attrs there too. Fixed in 0bb257d8e.

krono-i2:
@simlapointe all points from the last round are addressed in 0bb257d8e (typo, circuit-synthesis schema description, static/transient warnings, the boundary-mode driver's submesh extraction, and the regression cleanup with adaptive case dropped, input config symlinked to the example, references regenerated). Also rebased onto current main (past #772/#794); all unit and regression tests pass locally after the rebase.

docs/src/developer/notes.md
```diff
@@ -226,9 +226,10 @@ several *Palace* releases. The table below records the first *Palace* release sh
 each schema version; a schema version applies to that release and all later ones up to
 (but not including) the next entry. Add a row whenever the schema version is bumped.
 
-| Schema version | First *Palace* release | Notes                              |
-|:--------------:|:----------------------:|:---------------------------------- |
-| `1-0-0`        | `0.17`                 | First explicitly-versioned schema. |
+| Schema version | First *Palace* release | Notes                                       |
+|:--------------:|:----------------------:|:------------------------------------------- |
+| `1-0-0`        | `0.17`                 | First explicitly-versioned schema.          |
+| `1-1-0`        | `0.17.1`               | Add `RationalImpedance` boundary condition. |
```
simlapointe: @Sbozzolo, @hughcars can you confirm/comment on whether this is the expected/correct way to update the schema version and notes.md?

examples/rational_impedance/parallel_rlc_rational.json
```diff
@@ -0,0 +1,82 @@
...
+    "Device": "CPU",
+    "Driven":
+    {
+      "Samples":
+      [
+        {
+          "Type": "Linear",
+          "MinFreq": 6.0,   // GHz
+          "MaxFreq": 14.0,  // GHz
+          "FreqStep": 0.1   // GHz  (81 points across the 10 GHz notch)
```
simlapointe: Now that this config is used for the regression test, we should increase `FreqStep` so that it runs faster. 81 points is taking too long to run. 9 points (1 GHz) step is probably good enough.

simlapointe (changes requested):
Thanks for the quick changes, @krono-i2, and sorry for the multiple iterations. It's very close! See comments below.

examples/rational_impedance/parallel_rlc_rational.json
krono-i2: Done in 89464a2a5: FreqStep increased to 1 GHz (9 points) in both example configs, regression reference data regenerated accordingly; all tests pass locally.

docs/src/developer/notes.md
Sbozzolo: The PR introduces a change so that:
- old jsons are still valid (hence, not a major change)
- new jsons might not be valid for old palaces (because there is a new option) (hence, not a patch change)
So, updating the schema as minor is correct.

Given that the old jsons are still valid, there is no need to bump the minor version of Palace (unless the CSVs changed), so 0.17.1 is also correct.

So, everything here is correct

CHANGELOG.md
```diff
@@ -12,6 +12,17 @@ The format of this changelog is based on
...
+## In progress
+
+#### New Features
+
+  - Added a `RationalImpedance` boundary condition: a surface (Robin) impedance boundary
+    whose per-square impedance is an arbitrary rational function of frequency,
+    `Zs(s) = N(s)/D(s)` with `s = iω`, given by numerator and denominator polynomial
+    coefficients (its zeros and poles), generalizing the parallel-RLC `Impedance` boundary
+    to any passive lumped-network response
+    [PR 770](https://github.com/awslabs/palace/pull/770).
```
simlapointe:
```suggestion
    to any passive lumped-network response. SchemaVer 1-1-0
    [PR 770](https://github.com/awslabs/palace/pull/770).
```
From now on we will mark the schema version changes in the changelog.

docs/src/developer/notes.md
simlapointe, line 1: No need to modify this file in this PR. Palace developers will update it when a new version is released. Only provide the new schema version number in the CHANGELOG.md, as mentioned in my other comment.

simlapointe (changes requested):
Thanks for the changes, @krono-i2, it's looking very good! A couple small comments below regarding the schema versioning (in short: don't touch docs/src/developer/notes.md, only provide the new schema version number in the changelog entry). Otherwise, please rebase on the latest main and I'll then run the CI workflows to make sure everything looks good.

docs/src/developer/notes.md
simlapointe: Upon further discussion, let's not modify this `docs/src/developer/notes.md` file in every PR. The Palace developers will update it when a new Palace version is released.

docs/src/developer/notes.md
krono-i2: Applied — schema version noted in the CHANGELOG entry.

CHANGELOG.md
krono-i2: Done — reverted the notes.md change; the schema version is now noted in the CHANGELOG
entry instead. Thanks @Sbozzolo for confirming the versioning reasoning.

krono-i2:
Rebased onto latest main; notes.md restored and the schema version moved to the CHANGELOG
entry per the new convention. All unit and regression tests pass locally — ready for CI.

simlapointe (approved):
Thanks for the quick changes! It all looks good now.

There's one failing CI test due to a github permission issue for fork PRs. I just merged a fix for it in #806. After you rebase the CI tests should all pass.

krono-i2:
@simlapointe rebased onto current main (past #804/#806); resolved the CHANGELOG conflict by keeping both the Bug Fixes and New Features sections under In progress. All unit and regression tests pass locally after the rebase.

simlapointe:
Thanks for your contribution, @krono-i2! And thanks for all the quick fixes. CI is green now, I’ll go ahead and merge.

PR #775, opened 2026-06-22
Expand version argument

nikosavola:
New PR of #477 with https://github.com/awslabs/palace/pull/477#pullrequestreview-3264709394 addressed with dee4dfc52f4c6470b3e1cf97660846828b7cf6f0


palace/CMakeLists.txt
```diff
@@ -651,6 +651,48 @@ if(PALACE_WITH_GPU_AWARE_MPI)
...
+set(dep_version_mumps        "${MUMPS_VERSION}")
+set(dep_version_parmetis     "${ParMETIS_VERSION}")
+set(dep_version_scalapack    "${scalapack_VERSION}")
+set(dep_version_scn          "${scn_VERSION}")
+set(dep_version_sundials     "${SUNDIALS_VERSION}")
+set(dep_version_superlu_dist "${superlu_dist_VERSION}")
+
+foreach(dep_folder IN LISTS dependency_folders)
+  string(REPLACE "-" "_" dep_name "${dep_folder}")
+  get_dependency_git_sha("${CMAKE_BINARY_DIR}/../extern/${dep_folder}" dep_info)
```
hughcars: This assumes the superbuild source layout and cannot recover concrete Spack provenance. Could the superbuild and Spack recipe pass their authoritative revisions instead?

palace/main.cpp
```diff
@@ -45,6 +45,49 @@ static const char *GetPalaceGitTag()
   return commit;
 }
 
+static void PrintPalaceVersionInfo(MPI_Comm comm)
+{
+  if (std::strcmp(GetPalaceGitTag(), "UNKNOWN"))
```
hughcars, line 45: In non-Git builds this prints no Palace version. Please always print `PROJECT_VERSION`, with Git metadata as optional additional detail.

hughcars (commented):
Useful addition—the `-V` / `--version` interface and dependency output are valuable. My main concern is that source-tree probing does not provide authoritative provenance across superbuild and Spack/archive builds. I tested a producer-supplied approach here as a reference, not a requested wholesale rewrite:
https://github.com/awslabs/palace/tree/hughcars/pr775-simplified-version-metadata

cmake/ExternalPalace.cmake
```diff
@@ -37,6 +37,8 @@ list(APPEND PALACE_OPTIONS
   "-DPALACE_TESTS_NUMPROC=${PALACE_TESTS_NUMPROC}"
   "-DPALACE_REGRESSION_NUMPROC=${PALACE_REGRESSION_NUMPROC}"
   "-DPALACE_TESTS_OMP_THREADS=${PALACE_TESTS_OMP_THREADS}"
+  # Where the dependency source trees live, for `palace --version` reporting
+  "-DPALACE_EXTERN_SOURCE_DIR=${CMAKE_BINARY_DIR}/extern"
```
hughcars, line 41: This gives the inner build somewhere to receive the version, but the superbuild never actually sends it. Since the superbuild already knows the exact tags, can we pass them here rather than finding them again from the checked-out source?

Something like:

```cmake
if(PALACE_BUILD_EXTERNAL_DEPS)
  list(APPEND PALACE_OPTIONS
    "-DPALACE_DEP_mfem_VERSION=${EXTERN_MFEM_GIT_TAG}"
    "-DPALACE_DEP_libCEED_VERSION=${EXTERN_LIBCEED_GIT_TAG}"
    "-DPALACE_DEP_hypre_VERSION=${EXTERN_HYPRE_GIT_TAG}"
    "-DPALACE_DEP_metis_VERSION=${EXTERN_METIS_GIT_TAG}"
    "-DPALACE_DEP_parmetis_VERSION=${EXTERN_PARMETIS_GIT_TAG}"
  )

  if(PALACE_WITH_SLEPC)
    list(APPEND PALACE_OPTIONS
      "-DPALACE_DEP_petsc_VERSION=${EXTERN_PETSC_GIT_TAG}"
      "-DPALACE_DEP_slepc_VERSION=${EXTERN_SLEPC_GIT_TAG}"
    )
  endif()
  if(PALACE_WITH_ARPACK)
    list(APPEND PALACE_OPTIONS
      "-DPALACE_DEP_arpack_ng_VERSION=${EXTERN_ARPACK_GIT_TAG}"
    )
  endif()
  if(PALACE_WITH_SUNDIALS)
    list(APPEND PALACE_OPTIONS
      "-DPALACE_DEP_sundials_VERSION=${EXTERN_SUNDIALS_GIT_TAG}"
    )
  endif()
  if(PALACE_WITH_SUPERLU)
    list(APPEND PALACE_OPTIONS
      "-DPALACE_DEP_superlu_dist_VERSION=${EXTERN_SUPERLU_GIT_TAG}"
    )
  endif()
endif()
```

The same should be done for the other enabled optional dependencies. Then the source-tree lookup is actually a fallback rather than the normal route.

palace/CMakeLists.txt
```diff
@@ -677,6 +684,59 @@ if(PALACE_WITH_GPU_AWARE_MPI)
   )
 endif()
 
+# Report build dependency versions through `palace --version`. For each
+# dependency the version is resolved in order of decreasing authority:
+#   1. PALACE_DEP_<NAME>_VERSION passed in by the producer that knows the
+#      concrete provenance (e.g. a Spack recipe passing its resolved spec).
+#   2. The version number discovered by find_package.
```
hughcars, line 692: We have the same gap in the Spack recipe. It knows the exact concrete dependencies, but doesn’t pass any of them through, so `--version` can still leave out MFEM, libCEED, PETSc, and SLEPc.

Can we add something like this in `cmake_args`?

```python
dependency_versions = {
    "STRUMPACK": "strumpack",
    "arpack_ng": "arpack-ng",
    "eigen": "eigen",
    "fmt": "fmt",
    "gslib": "gslib",
    "hypre": "hypre",
    "json": "nlohmann-json",
    "libCEED": "libceed",
    "libxsmm": "libxsmm",
    "magma": "magma",
    "metis": "metis",
    "mfem": "mfem",
    "mumps": "mumps",
    "parmetis": "parmetis",
    "petsc": "petsc",
    "scalapack": "scalapack",
    "scn": "scnlib",
    "slepc": "slepc",
    "sundials": "sundials",
    "superlu_dist": "superlu-dist",
}

for cmake_name, spack_name in dependency_versions.items():
    if spack_name not in self.spec:
        continue
    dep = self.spec[spack_name]
    provenance = f"{dep.version} /{dep.dag_hash(7)}"
    args.append(
        self.define(
            f"PALACE_DEP_{cmake_name}_VERSION",
            provenance,
        )
    )
```

I included the short DAG hash because `develop`, or the same nominal dependency version, can still mean a different concrete build.

hughcars (changes requested):
Thanks for taking another pass at this. The Palace version handling looks good, and passing the external source directory fixes the layout assumption. I think the remaining part is to actually pass these values from the superbuild and Spack. Right now the new interface is there, but neither caller uses it, so we still have the original problem. I’ve left examples for both.

palace/CMakeLists.txt
nikosavola: Addressed in 222499ad

cmake/ExternalPalace.cmake
nikosavola: Addressed in 222499ad

hughcars:
Hi @nikosavola, I think my earlier feedback pushed this into a more complicated provenance interface than the feature needs. Sorry for the churn.

I put a tested simplification directly on top of your branch in [`20691d69`](https://github.com/awslabs/palace/commit/20691d69799c84f8c90cdd0a325eaf1b3db745f9):

```bash
git cherry-pick 20691d69799c84f8c90cdd0a325eaf1b3db745f9
```

It retains the complete dependency list, full Palace SHA, Spack root DAG hash, and launcher-free `-V`/`--version`, but uses one manifest and native Spack dependency traversal. The commit removes 108 net lines from the current branch and has been tested across Spack, bundled CMake, SLEPc, ARPACK, and STRUMPACK configurations.

Could you cherry-pick this commit onto your branch?

nikosavola:
Cherry-picked at HEAD -- the PR branch is now at 20691d69799c84f8c90cdd0a325eaf1b3db745f9, so no churn to worry about.

Applied as a fast-forward rather than a rebase: that commit was already a direct child of 222499ade (the previous head), so the branch just advanced onto it and your commit kept its original authorship.

Quick sanity checks on the resulting tree, all clean: no remaining references to `GetGitDescription`/`git_describe` or the `PALACE_DEP_*`/`PALACE_EXTERN_SOURCE_DIR` machinery, the Spack recipe still parses, and the generated `BuildInfo.hpp` lands in `${CMAKE_BINARY_DIR}/generated`, which is already on the target include path from the existing schema header wiring.

CI is running on the new head. Thanks for the simplification.

nikosavola:
> Hi @nikosavola, I think my earlier feedback pushed this into a more complicated provenance interface than the feature needs. Sorry for the churn.
> 
> I put a tested simplification directly on top of your branch in [`20691d69`](https://github.com/awslabs/palace/commit/20691d69799c84f8c90cdd0a325eaf1b3db745f9):
> 
> ```shell
> git cherry-pick 20691d69799c84f8c90cdd0a325eaf1b3db745f9
> ```
> 
> It retains the complete dependency list, full Palace SHA, Spack root DAG hash, and launcher-free `-V`/`--version`, but uses one manifest and native Spack dependency traversal. The commit removes 108 net lines from the current branch and has been tested across Spack, bundled CMake, SLEPc, ARPACK, and STRUMPACK configurations.
> 
> Could you cherry-pick this commit onto your branch?

This is now done

hughcars (approved):
Addressed in `20691d69`. The branch now uses one dependency manifest, native Spack dependency traversal, the full Palace SHA, and a root Spack DAG hash. I tested the Spack and bundled CMake paths, including SLEPc, ARPACK, and STRUMPACK configurations.

Issue #774, opened 2026-06-22
Clean up GHCR ci job failing


hughcars:
The CI job to clean the cache is [failing](https://github.com/awslabs/palace/actions/runs/27937563084/job/82662849447) this ticket is to triage and fix.

PR #776, opened 2026-06-23
Fix GHCR cache cleanup

hughcars:
Attempted fix for #774.

Keep the official `actions/delete-package-versions` cleanup action, pinned to the v5 commit SHA, but retry it once when the first attempt hits a transient GHCR/GitHub Packages API failure. The workflow also logs the package version count before and after cleanup so scheduled/manual runs show whether the backlog is shrinking.

Validation: `git diff --check`; YAML parsed with Ruby.


Sbozzolo (approved):
Thank you!

PR #777, opened 2026-06-23
Fix bug in mesh cracking refinement on periodic meshes

### LINKED ISSUE #773: Cracking an internal boundary on a periodic face aborts in STable3D
### Description

A driven, doubly-periodic simulation aborts during mesh setup when an internal
boundary has nodes lying on a periodic boundary:

```
MFEM abort: (r,c,f) = (...)
 ... in function: int mfem::STable3D::operator()(int, int, int) const
 ... in file: general/stable3d.cpp:112
```

### To reproduce

Minimal files attached: `mwe_periodic_crack.json` and `mwe_periodic_crack.msh`.
The model is an air unit cell, periodic in x and y (two boundary pairs), with one
internal PEC strip at mid-height that spans the full y extent, so its edges lie on
the `y_master`/`y_slave` faces. Floquet port on top, PEC ground on bottom.

```
palace mwe_periodic_crack.json
```

Setting `CrackInternalBoundaryElements` to `false` in the config makes the abort
go away and the run proceeds.

### Notes

The input mesh appears valid: the periodic faces match node-for-node, the strip is
internal (each strip face is shared by two tetrahedra), and there are no coincident
or non-manifold faces. The abort shows up only after the internal boundary is
cracked, and only when it touches the periodic seam, which suggests the seam nodes
duplicated by cracking are not reconciled by the periodic vertex mapping.

For reference, this combination is not exercised by the existing examples:
`dielectric_grating` is doubly periodic but has no internal cracked boundaries, and
the `cylinder` Floquet example is periodic in a single direction.

https://sprint-cloud.dieti.unina.it/s/ypCYYq6a6yYWE8t
https://sprint-cloud.dieti.unina.it/s/QCDbjkiAcT28xnq

simlapointe:
Periodic meshes do not support conformal mesh refinement, but the cracking refinement (`RefineCrackElements`) uses MFEM's conformal refinement (`GeneralRefinement(refinements, 0)`), leading to errors such as:
```
MFEM abort: (r,c,f) = (...)
 ... in function: int mfem::STable3D::operator()(int, int, int) const
 ... in file: general/stable3d.cpp:112
```
when apply mesh cracking to some periodic meshes, as reported in #773. 

This PR implements a new `LocalEdgeSplit` function that bypasses MFEM's conformal refinement to perform a similar edge refinement/splitting strategy for periodic meshes with mesh cracking.

### DIFF
--- a/palace/utils/geodata.cpp
+++ b/palace/utils/geodata.cpp
@@ -2290,6 +2290,317 @@ void SplitMeshElements(std::unique_ptr<mfem::Mesh> &orig_mesh, bool make_simplex
   orig_mesh->FinalizeTopology();
 }
 
+// Conformally bisect a set of mesh edges by inserting a midpoint on each and splitting
+// every tetrahedron and boundary element incident on a split edge into two. Unlike
+// mfem::Mesh::GeneralRefinement, this performs a purely local edge-fan split with no
+// marked-edge/longest-edge selection and no global conformity-closure propagation. This is
+// essential for periodic meshes: MFEM's conforming bisection keys the new midpoint on the
+// collapsed vertex pair and cannot keep the two periodic copies of a seam edge consistent,
+// which corrupts the seam (aborting later in STable3D). A local edge-fan split touches only
+// the ring of elements around each edge, leaving the rest of the mesh (and the periodic
+// identification) untouched. The inserted midpoints are later duplicated by the regular
+// interior-boundary cracking pass, which decouples the two sides of the crack.
+//
+// To split many edges in a single mesh rebuild (rather than one expensive rebuild per
+// edge), a maximal independent subset of the candidate edges is selected such that no
+// tetrahedron contains more than one selected edge; this keeps the per-element split simple
+// (each affected tet/triangle bisects into exactly two children) while processing many edges
+// at once. The caller re-invokes (via the AddInterfaceBdrElements retry loop) to handle any
+// candidates not selected in this pass. Returns the number of edges split (0 if none).
+template <typename EdgeContainer>
+int LocalEdgeSplit(std::unique_ptr<mfem::Mesh> &orig_mesh, const EdgeContainer &edges)
+{
+  // Only pure tetrahedral meshes are supported (matching the previous refinement path,
+  // which marked and bisected only TETRAHEDRON elements). The caller invokes
+  // SplitMeshElements(make_simplex=true) immediately before reaching here, so any
+  // hexahedral/prism mesh has already been converted to tetrahedra; this guard makes that
+  // contract explicit rather than relying on the upstream conversion.
+  MFEM_VERIFY(orig_mesh->Dimension() == 3, "LocalEdgeSplit only supports 3D meshes!");
+  {
+    const auto element_types = mesh::CheckElements(*orig_mesh);
+    MFEM_VERIFY(element_types.has_simplices && !element_types.has_hexahedra &&
+                    !element_types.has_prisms && !element_types.has_pyramids,
+                "LocalEdgeSplit only supports pure tetrahedral meshes!");
+  }
+  if (edges.empty())
+  {
+    return 0;
+  }
+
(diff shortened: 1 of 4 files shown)


palace/utils/geodata.cpp
```diff
@@ -2290,6 +2294,315 @@ void SplitMeshElements(std::unique_ptr<mfem::Mesh> &orig_mesh, bool make_simplex
...
+                    !element_types.has_prisms && !element_types.has_pyramids,
+                "LocalEdgeSplit only supports pure tetrahedral meshes!");
+  }
+  if (edges.empty())
+  {
+    return 0;
+  }
+
+  std::unique_ptr<mfem::Table> vert_to_elem(orig_mesh->GetVertexToElementTable());
+  auto edge_ring = [&](int v0, int v1)
```
hughcars, line 2329: Total aside, but in meshing this actually has a name, this is the "edge cavity".

palace/utils/geodata.cpp
```diff
@@ -2290,6 +2294,315 @@ void SplitMeshElements(std::unique_ptr<mfem::Mesh> &orig_mesh, bool make_simplex
...
+  const int new_nv = orig_mesh->GetNV() + num_split;
+  const int new_ne = orig_mesh->GetNE() + n_split_elem;
+  const int new_nbe = orig_mesh->GetNBE() + n_split_bdr;
+  auto new_mesh = std::make_unique<mfem::Mesh>(orig_mesh->Dimension(), new_nv, new_ne,
+                                               new_nbe, orig_mesh->SpaceDimension());
+  for (int v = 0; v < orig_mesh->GetNV(); v++)
+  {
+    new_mesh->AddVertex(orig_mesh->GetVertex(v));
+  }
+  // Midpoint coordinates (geometric average; curved geometry is fixed up below for
```
hughcars: Nit: I think this is an arithmetic average, geo would be `sqrt(x_1 x_2)`.

palace/utils/geodata.cpp
```diff
@@ -2746,50 +3059,71 @@ int AddInterfaceBdrElements(IoData &iodata, std::unique_ptr<mfem::Mesh> &orig_me
           face_to_be.clear();
           return 0;  // Mesh was converted to simplices, start over
         }
-        std::unordered_map<int, int> elem_to_refine;
-        for (const auto &[edge, adjacent_be] : coarse_crack_edge_to_be)
+        const bool periodic = !iodata.boundaries.periodic.boundary_pairs.empty();
+        if (!periodic)
```
hughcars, line 3063: Given your bisection algorithm above only supports tets, and to do mesh cracking the mesh is converted to tets already, is there any good reason not to just use your algorithm always and get rid of the move shuffle?

hughcars (commented):
One question about whether to remove the prior algorithm entirely, given you've effectively introduced an edge-cavity split. Otherwise, if your tests are passing for that issue, I think it looks good.

palace/utils/geodata.cpp
simlapointe: Fixed in [8fc245f](https://github.com/awslabs/palace/pull/777/commits/8fc245fc59069f7f564db0d02a89f075329dda3b)

palace/utils/geodata.cpp
simlapointe: I thought of doing that. Actually tried it on a handful of representative meshes and, while it gives the exact same numbers of additional mesh elements and vertices, the edge splits are not necessarily the same and thus the mesh is not identical and the results change slightly. Specifically, it seems that `LocalEdgeSplit` can give slightly worse mesh quality in those split elements. 

So to avoid introducing changes affecting all cracked meshes while fixing this bug promptly, I thought to roll this out for periodic meshes only for now. I do think it might be possible to improve `LocalEdgeSplit` in the future and use that path for all meshes, but I'd want to test it more thoroughly to make sure there are no regressions/edge cases.

hughcars (approved):
Considering opening a follow up issue to see if we can reconcile the two split styles.

PR #778, opened 2026-06-23
Use complex frequency in nonlinear BC in eigenmode simulations

simlapointe:
In eigenmode simulations, the (complex) eigenmode of the 3D system is an input to the nonlinear frequency-dependent boundary conditions (wave port, 2nd order absorbing BC, surface conductivity, rational impedance). Previously, these boundary conditions took only a real frequency as input. This PR changes that to allow the freq-dependent BCs to take a complex frequency as input so that they are consistent with the eigenmode simulation. Results of eigenmode simulations with wave ports (or other nonlinear BCs) change by roughly 1/Q.


palace/drivers/eigensolver.cpp
```diff
@@ -42,14 +42,17 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
   auto M = space_op.GetMassMatrix<ComplexOperator>(Operator::DIAG_ZERO);
 
   // Check if there are nonlinear terms and, if so, setup interpolation operator.
-  auto funcA2 = [&space_op](double omega) -> std::unique_ptr<ComplexOperator>
-  { return space_op.GetExtraSystemMatrix<ComplexOperator>(omega, Operator::DIAG_ZERO); };
+  auto funcA2 = [&space_op](std::complex<double> lambda) -> std::unique_ptr<ComplexOperator>
+  {
+    const std::complex<double> omega = lambda / std::complex<double>(0.0, 1.0);  // ω = -iλ
```
hughcars: comment and the math not being identical (though equivalent), on purpose?

palace/drivers/eigensolver.cpp
```diff
@@ -267,14 +270,7 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
   {
     // Search for eigenvalues closest to λ = iσ.
     eigen->SetShiftInvert(1i * target);
-    if (type == EigenSolverBackend::ARPACK)
-    {
-      // ARPACK searches based on eigenvalues of the transformed problem. The eigenvalue
-      // 1 / (λ - σ) will be a large-magnitude negative imaginary number for an eigenvalue
-      // λ with frequency close to but not below the target σ.
-      eigen->SetWhichEigenpairs(EigenvalueSolver::WhichType::SMALLEST_IMAGINARY);
-    }
```
hughcars, line 276: This ARPACK treatment is no longer necessary?

palace/linalg/arpack.cpp
```diff
@@ -378,13 +386,70 @@ int ArpackEigenvalueSolver::SolveInternal(int n, std::complex<double> *r,
...
+  {
+    const bool l_on_target_side = OnTargetImagSide(eig[l]);
+    const bool r_on_target_side = OnTargetImagSide(eig[r]);
+    if (l_on_target_side != r_on_target_side)
+    {
+      return l_on_target_side;
+    }
+    const double target = sigma.imag();
+    const double l_dimag = std::abs(eig[l].imag() - target);
+    const double r_dimag = std::abs(eig[r].imag() - target);
```
hughcars: `dimag` rather than `imag`? What's the `d` for?

palace/linalg/arpack.cpp
```diff
@@ -378,13 +386,70 @@ int ArpackEigenvalueSolver::SolveInternal(int n, std::complex<double> *r,
...
+  if (which_type == WhichType::TARGET_IMAGINARY)
+  {
+    std::sort(perm, perm + nev, CompareTargetImag);
+    const int num_extracted = std::min(num_conv, nev);
+    int num_target_side = 0;
+    for (int i = 0; i < num_extracted; i++)
+    {
+      if (OnTargetImagSide(eig[perm[i]]))
+      {
+        perm[num_target_side++] = perm[i];
```
hughcars, line 442: Safe self-assign as `num_target_side <= i` at all times, so the rhs is always still valid before the assignment.

palace/models/postoperator.cpp
```diff
@@ -1349,7 +1349,7 @@ void PostOperator<solver_t>::MeasureSParameter() const
       return;
     }
 
-    // Per-side wave-port de-embedding factor: exp(i·kₙ·d_offset) when the port is a
+    // Per-side wave-port de-embedding factor: exp(i·k_n·d_offset) when the port is a
```
hughcars: Nit: keep the unicode

palace/models/modeeigensolver.cpp
```diff
@@ -153,19 +190,27 @@ AssembleAnn(const FiniteElementSpace &h1_fespace, const MaterialOperator &mat_op
...
-  MaterialPropertyCoefficient poseps_h1_func(mat_op.GetAttributeToMaterial(),
-                                             normal ? mat_op.GetPermittivityReal()
-                                                    : mat_op.GetPermittivityScalar(),
-                                             omega * omega);
+  MaterialPropertyCoefficient poseps_h1_func(
+      mat_op.GetAttributeToMaterial(),
+      normal ? mat_op.GetPermittivityReal() : mat_op.GetPermittivityScalar(), w2r);
   if (normal)
   {
     poseps_h1_func.NormalProjectedCoefficient(*normal);
```
hughcars, line 235: The complex-ω loss-tangent cross term is missing on the real slot here. Expanding the Ann ε-mass term `+ω²·(i·ε_imag)` with `ω² = w2r + i·w2i`:

- `Im(+ω²·i·ε_imag) = +w2r·ε_imag` → added to `posepsi_h1_func` below ✓
- `Re(+ω²·i·ε_imag) = −w2i·ε_imag` → **not added anywhere**

`AssembleAtt` includes its mirror (`Re(−ω²·i·ε_imag) = +w2i·ε_imag` in `eps_shifted_func`, "included for completeness"), so the two blocks are treated inconsistently and A2(λ) is not quite the exact analytic continuation for lossy cross-sections — same order as the cross terms deliberately kept in Att.

```suggestion
  MaterialPropertyCoefficient poseps_h1_func(
      mat_op.GetAttributeToMaterial(),
      normal ? mat_op.GetPermittivityReal() : mat_op.GetPermittivityScalar(), w2r);
  if (complex_omega && mat_op.HasLossTangent())
  {
    // Cross term: Re(+ω²·(i·ε_imag)) = -w2i·ε_imag — mirror of the +w2i·ε_imag
    // term in AssembleAtt. Accumulate before the (destructive) normal projection
    // below; projection is linear so this equals projecting the sum.
    poseps_h1_func.AddCoefficient(mat_op.GetAttributeToMaterial(),
                                  normal ? mat_op.GetPermittivityImag()
                                         : mat_op.GetPermittivityImagScalar(),
                                  -w2i);
  }
  if (normal)
  {
    poseps_h1_func.NormalProjectedCoefficient(*normal);
```

palace/models/spaceoperator.cpp
```diff
@@ -1039,6 +1073,26 @@ void SpaceOperator::AddExtraSystemBdrCoefficients(double omega,
...
+                                                  MaterialPropertyCoefficient &fbi)
+{
+  // Complex-ω overload for the eigenmode nonlinear solve and its matching preconditioner:
+  // all frequency-dependent boundary terms (2nd-order farfield ABC, surface conductivity,
+  // rational surface impedance, numeric wave ports) are evaluated at the genuinely complex
+  // frequency (ω = -i·λ) so the assembled A2(λ) is the exact analytic continuation. For
+  // real ω the operators' complex overloads reduce to their double overloads. Rational
+  // impedance boundaries are currently rejected outside driven/boundary-mode problems at
+  // configuration time, but the term is stamped here so the operator stays complete if
+  // that restriction is lifted.
```
hughcars: This sentence is now stale — the restriction was lifted by this PR's own "Enable rational impedance boundaries in eigenmode simulations" commit. Also, "reduce to their double overloads" is not exact for the wave-port term (at real ω the complex overload additionally stamps −Im(k_n)·M on the real slot, as documented in waveportoperator.cpp).

```suggestion
  // real ω the operators' complex overloads reduce to their double overloads, except the
  // wave-port term, which additionally carries the line attenuation -Im(k_n)·M on the
  // real slot (the real-ω overload intentionally stamps only Re(k_n)).
```

Same stale count in `GetExtraSystemMatrix(std::complex<double>, ...)` above (~L555): "the three frequency-dependent boundary terms" is now four with rational impedance.

palace/models/spaceoperator.hpp
```diff
@@ -215,6 +223,14 @@ class SpaceOperator
                                                  Operator::DiagonalPolicy diag_policy,
                                                  bool include_wave_ports);
 
+  // Complex-ω overload for the eigenmode nonlinear solve: assembles A2(λ) with all three
+  // frequency-dependent boundary terms evaluated at the genuinely complex frequency
+  // (ω = -i·λ). Always a ComplexOperator (the wave-port / ABC / surf-σ terms acquire a
+  // real-slot contribution at complex ω). For real ω this matches GetExtraSystemMatrix<
+  // ComplexOperator>(double).
```
hughcars: Two nits: "all three" is now four (rational impedance was enabled for eigenmode later in this PR), and "For real ω this matches `GetExtraSystemMatrix<ComplexOperator>(double)`" isn't exact — for a lossy port cross-section the wave-port term additionally carries −Im(k_n)·M on the real slot at real ω (waveportoperator.cpp documents this correctly, and the HYBRID seed relies on precisely this difference).

```suggestion
  // Complex-ω overload for the eigenmode nonlinear solve: assembles A2(λ) with all
  // frequency-dependent boundary terms (2nd-order ABC, surface conductivity, rational
  // impedance, numeric wave ports) evaluated at the genuinely complex frequency
  // (ω = -i·λ). Always a ComplexOperator (these terms acquire a real-slot contribution
  // at complex ω). For real ω this matches GetExtraSystemMatrix<ComplexOperator>(double)
  // up to the wave-port term, which additionally carries the attenuation -Im(k_n)·M on
  // the real slot (the real-ω path intentionally stamps only Re(k_n)).
```

palace/drivers/drivensolver.cpp
```diff
@@ -177,7 +177,7 @@ ErrorIndicator DrivenSolver::SweepUniform(SpaceOperator &space_op) const
       auto A = space_op.GetSystemMatrix(1.0 + 0.0i, 1i * omega, -omega * omega + 0.0i,
                                         K.get(), C.get(), M.get(), A2.get());
       auto P = space_op.GetPreconditionerMatrix<ComplexOperator>(
-          1.0 + 0.0i, 1i * omega, -omega * omega + 0.0i, omega);
+          1.0 + 0.0i, 1i * omega, -omega * omega + 0.0i, omega + 0.0i);
```
hughcars: This `+ 0.0i` does more than change a type: it switches the preconditioner assembly from the real-ω BC path to the new complex-ω one, which has different driven-path behavior. Three effects on driven/ROM solves:

1. **Floquet term dropped from P.** The complex overload doesn't stamp `floquet_port_op`, so P no longer matches A for driven Floquet simulations.
2. **P uses full complex k_n, A uses Re(k_n).** The complex overload calls `SolveKnComplex`, so P picks up a −Im(k_n)·M term the driven system matrix doesn't have.
3. **Extra EVP solve per frequency.** `SolveKnComplex` is deliberately uncached, so each driven frequency point (and each HDM solve in the adaptive/ROM path) re-solves the port cross-section EVP that `Initialize` already solved and cached.

Making the complex overload fall back to the real path when `Im(ω) == 0` would fix this but break the HYBRID seed, which needs the complex stamping at real ω. So the choice can't be inferred from the value — suggest keeping a `double a3` overload for the driven/ROM call sites so real-vs-complex stays an explicit caller decision.

palace/models/modeeigensolver.cpp
```diff
@@ -242,24 +297,33 @@ AssembleAnn(const FiniteElementSpace &h1_fespace, const MaterialOperator &mat_op
 
   std::unique_ptr<mfem::HypreParMatrix> Anni_assembled;
   {
-    const bool has_imag = mat_op.HasLossTangent() || !nn_fbi.empty();
+    const bool has_imag = mat_op.HasLossTangent() || !nn_fbi.empty() || complex_omega;
```
hughcars: Related observation, pre-existing on main rather than introduced here: `AssembleAnn` never carries the domain conductivity term at all. `AssembleAtt` has −ω²ε_c with the full complex permittivity ε_c = ε − iσ/ω (the +iωσ term), but Ann's +ω²ε_c only carries ε and the loss tangent — the −iωσ counterpart (−ω·σ on the imaginary slot at real ω) is missing, and this gate doesn't test `HasConductivity()`.

It's the same asymmetry family as the loss-tangent cross term above, and since this PR is already touching both blocks it might be worth just fixing here — but it changes real-ω results for conductive wave-port cross-sections (driven included), so it needs its own validation. Up to you whether to tackle it in this PR or track it as a follow-up issue.

hughcars (changes requested):
Went through all the complex-ω formulas by hand (Att/Ann splits, BC overloads, wave-port stamping, λ↔ω plumbing, FD Jacobian) — everything checks out except two items, detailed inline:

1. `AssembleAnn` is missing the `−w2i·ε_imag` cross term that `AssembleAtt` includes — the two blocks are inconsistent for lossy cross-sections. Suggestion inline.
2. The `omega + 0.0i` in drivensolver/romoperator silently moves the driven preconditioner onto the complex BC path: Floquet term dropped from P, complex kₙ in P vs Re(kₙ) in A, and an uncached port EVP solve per frequency. Real-vs-complex should stay an explicit caller choice.

Rest is nits and stale comments, inline. Also flagged a pre-existing gap (Ann never carries the −iωσ conductivity term) — fine to fix here or as a follow-up, your call.

palace/drivers/eigensolver.cpp
simlapointe: Made it consistent in [23d97ed](https://github.com/awslabs/palace/pull/778/commits/23d97ed503ce797a82b090ef3a6c4bb9545d2f72)

palace/drivers/eigensolver.cpp
simlapointe: Correct, it's replaced by the new TARGET_IMAGINARY.

palace/linalg/arpack.cpp
simlapointe: `d` was for distance from the target. Changed it to `dist` in [23d97ed](https://github.com/awslabs/palace/pull/778/commits/23d97ed503ce797a82b090ef3a6c4bb9545d2f72)

palace/linalg/arpack.cpp
simlapointe: Yes, added a comment making it explicit 23d97ed

palace/models/postoperator.cpp
simlapointe: Switched back to unicode in 23d97ed

palace/models/modeeigensolver.cpp
simlapointe: You're right. Fixed in 23d97ed

palace/models/spaceoperator.cpp
simlapointe: Fixed in 23d97ed

palace/models/spaceoperator.hpp
simlapointe: Fixed in 23d97ed

palace/drivers/drivensolver.cpp
simlapointe: Fixed in 23d97ed

palace/models/modeeigensolver.cpp
simlapointe: Fixed in 23d97ed and added unit tests in [a820140](https://github.com/awslabs/palace/pull/778/commits/a820140705afbff7d4bf9c3aeab6e4e11c5a928c)

hughcars (approved):

PR #781, opened 2026-06-24
Push containers to ECR/S3

Sbozzolo:
Add a step to push the container images to a private registry/bucket.

This is triggered:
- when a new tag is created
- when pushing to main
- when a PR has the `push-containers` label

neoverse-v2 is built upon tagging only.


.github/workflows/containers.yml
```diff
@@ -6,10 +6,18 @@ name: Containers
...
   pull_request:
   workflow_dispatch:
+    inputs:
+      force_push:
+        description: >-
+          Push the built image/SIF to ECR/S3 under a dev-<shorthash> namespace.
+          Lets maintainers test container publishing from a branch without
+          merging. Off by default; never clobbers the main/release images.
+        type: boolean
+        default: false
```
hughcars: This is a very cool idea, I could see this being a very clean way to deploy a prototype directly off of a PR even, before it gets merged. Do you think we could do this by adding a label to the PR? `cache-prototype` or something, then as the PR progresses we could easily trial things and deploy these prototypes without any developer/user intervention, and with the ease of `main` pushes, we could easily migrate bravehearts to the main straight after.

.github/workflows/containers.yml
```diff
@@ -53,5 +61,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         if: needs.filter.outputs.test == 'true'
+        with:
+          # Fetch tags so the release channel can derive the version from the
+          # git ref (and `git describe` works if ever needed).
+          fetch-depth: 0
       - uses: ./.github/actions/build-container
         if: needs.filter.outputs.test == 'true'
```
hughcars: One small release-path gotcha I think: tag pushes are still downstream of `paths-filter`.

For PRs / main pushes this filter is doing exactly what we want. But for a release tag, especially one created on a commit that is already on `main`, the diff against `main` can be empty and this output can go false. Then the versioned ECR/S3 publish silently never happens.

Could we make release tags bypass the filter, while leaving the normal skip behavior alone? Since the filter output line itself is not in this diff hunk, this is the small local version:

```suggestion
      - uses: actions/checkout@v6
        if: needs.filter.outputs.test == 'true' || startsWith(github.ref, 'refs/tags/v')
        with:
          # Fetch tags so the release channel can derive the version from the
          # git ref (and `git describe` works if ever needed).
          fetch-depth: 0
      - uses: ./.github/actions/build-container
        if: needs.filter.outputs.test == 'true' || startsWith(github.ref, 'refs/tags/v')
```

hughcars (commented):
This is a very cool PR! One future direction question and one possible issue with pushing a tag.

.github/workflows/containers.yml
Sbozzolo: Yes, good idea. I can also change it to be the branch name, making things easier to understand.

.github/actions/build-container/action.yml
```diff
@@ -104,15 +154,66 @@ runs:
...
+        IMAGE_NAME: ${{ steps.name.outputs.image_name }}
+      run: |
+        # Pushes on main/tag builds and push-containers-labelled PRs (see the
+        # metadata step). Auth is ambient via the runner's EC2 instance role.
+        source /tmp/timing_functions.sh
+        start_timer
+
+        ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+        # Redact the account ID from all subsequent log output.
+        echo "::add-mask::${ACCOUNT_ID}"
```
hughcars, line 174: Nice security 👍

.github/workflows/containers.yml
```diff
@@ -1,14 +1,20 @@
 name: Containers
 
 # Build OCI/SIF containers for:
-# - sapphirerapids
-# - neoverse_v1
+# - sapphirerapids   (x64,       c7i)        — every publish event
+# - neoverse_v1      (arm64,     c7g)        — every publish event
+# - neoverse_v2      (arm64-8g,  c8g)        — release tags only
```
hughcars, line 6: Lets not do it in this PR, but as we get the GPU support up, we should maybe ship those images?

hughcars (approved):
Modulo CI, merge away.

.github/workflows/containers.yml
Sbozzolo: Yes, that's easy to do. I didn't do it because we need to make CUDA_ARCH-specific images, and I didn't know what version to pick.

PR #792, opened 2026-06-29
Better handle absorbing and rational impedance BCs in hybrid nonlinear eigensolver

simlapointe:
Better handle the 2nd order absorbing and rational impedance BCs in the hybrid nonlinear eigensolver. For the absorbing BC, this is done by adding a frozen $1/\omega_{\rm{target}}$ term to the quadratic EVP solved for the initial guess instead of trying to interpolate the $1/\omega$ term with a quadratic polynomial. For the rational impedance BC, we determine whether the rational function is better approximated with a quadratic fit or a frozen (at $\omega_{\rm{target}}$) approach and use the most accurate.


palace/linalg/nleps.cpp
```diff
@@ -914,14 +917,146 @@ void NewtonInterpolationOperator::Interpolate(const std::complex<double> sigma_m
...
+  // are members, so they outlive the returned operator — no keepalive needed.
+  std::vector<std::complex<double>> sum_coeffs = {coeffs[order][0], coeffs[order][1],
+                                                  coeffs[order][2]};
+  std::vector<const ComplexOperator *> sum_ops = {ops[0][0].get(), ops[1][0].get(),
+                                                  ops[2][0].get()};
+  for (const auto &term : frozen_terms)
+  {
+    sum_coeffs.push_back(term.corr[order]);
+    sum_ops.push_back(term.M.get());
+  }
```
hughcars: `Mult`/`AddMult(order, ...)` below still evaluate only `Σⱼ coeffs[order][j]·ops[j][0]` and ignore `frozen_terms`, while `GetInterpolationOperator(order)` applies the corrections — two evaluation paths on the same object that disagree after `AddFrozenPole`. Nothing calls the indexed `Mult` in production today, so it's latent, but a future caller would silently get the uncorrected pencil.

I implemented a fix (plus a related dedup of the scalar Newton math) on [`hughcars/nleps-update`](https://github.com/awslabs/palace/tree/hughcars/nleps-update) — two commits you can cherry-pick if you like them, or drop if you don't:

- `d8b60174e` — factor `ComputeLeadingDividedDifferences` + `ComputeMonomialCoefficients` helpers shared by `AddFrozenPole`/`DetermineFrozen` (removes the duplicated divided-difference implementations; `DetermineFrozen` also stops re-reducing the Newton coefficients at each of the 101 sample points).
- `47399524d` — build the per-order (coefficient, operator) pencil lists once in `Interpolate()`, have `AddFrozenPole` append to them, and make `GetInterpolationOperator`/`Mult`/`AddMult` all read that single representation, so they cannot disagree by construction. Also removes the hardcoded 3-term lists and the `FrozenTerm::corr` bookkeeping.

Verified: `rational_impedance_eigen` (freeze branch), `cpw_lumped_eigen`, `cpw_wave_eigen`, `adapter_hybrid` all pass against the unchanged reference data (307 assertions total).

palace/drivers/eigensolver.cpp
```diff
@@ -61,11 +71,46 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
-    A2_1 = interp_op->GetInterpolationOperator(1);
-    A2_2 = interp_op->GetInterpolationOperator(2);
+    auto interp = std::make_unique<NewtonInterpolationOperator>(funcA2, A2->Width());
+    interp->Interpolate(1i * target, 1i * target_max);
+    if (M_ff)
+    {
+      interp->AddFrozenPole(
+          std::move(M_ff), [](std::complex<double> lambda) { return -0.5 / lambda; },
+          1i * target);
+    }
```
hughcars: `M_ff` is only used inside the `has_A2 && HYBRID` branch but is assembled unconditionally for every eigenmode solve (including linear and SLP runs). Hoisting it into the branch keeps the assembly (and its print-header side effect) off the paths that don't need it:

```suggestion
  // Extend K, C, M operators with interpolated A2 operator.
  // K' = K + A2_0, C' = C + A2_1, M' = M + A2_2
  std::unique_ptr<ComplexOperator> Kp, Cp, Mp;
  std::unique_ptr<Interpolation> interp_op;
  std::unique_ptr<ComplexOperator> A2_0, A2_1, A2_2;
  NonlinearEigenSolver nonlinear_type = iodata.solver.eigenmode.nonlinear_type;
  if (has_A2 && nonlinear_type == NonlinearEigenSolver::HYBRID)
  {
    const double target_max = iodata.solver.eigenmode.target_upper;
    auto interp = std::make_unique<NewtonInterpolationOperator>(funcA2, A2->Width());
    interp->Interpolate(1i * target, 1i * target_max);
    // Frozen-ABC seed for the NLEPS HYBRID polynomial seed pencil. The 2nd-order farfield
    // ABC contributes a pole term f(λ)·M_ff to A2(λ), with f(λ) = -0.5/λ, that a
    // polynomial (K' + λC' + λ²M') seed cannot fit accurately. AddFrozenPole removes the
    // pole's interpolated contribution and re-adds it frozen at the target into the
    // K-block.
    auto M_ff =
        space_op.GetFarfieldExtraBoundaryMatrix<ComplexOperator>(Operator::DIAG_ZERO);
    if (M_ff)
    {
      interp->AddFrozenPole(
          std::move(M_ff), [](std::complex<double> lambda) { return -0.5 / lambda; },
          1i * target);
    }
```

palace/models/surfacerationalimpedanceoperator.cpp
```diff
@@ -127,6 +161,13 @@ void SurfaceRationalImpedanceOperator::SetUpBoundaryProperties(
     auto &bdr = boundaries.emplace_back();
     bdr.num = data.num;
     bdr.den = data.den;
+    // Long division of the Robin coefficient g(s) = s·D(s)/N(s) = P(s) + R(s)/N(s): the
+    // polynomial part P is exactly representable in a polynomial eigenvalue pencil, while
+    // the strictly proper remainder R/N carries the poles of g (the zeros of Zs). Used by
+    // the nonlinear eigensolver seed to fit or freeze the pole part.
+    std::vector<double> sD = bdr.den;
+    sD.push_back(0.0);  // s·D(s)
+    PolyDivMod(std::move(sD), bdr.num, bdr.robin_quotient, bdr.robin_remainder);
```
hughcars, line 170: Optional hardening, take it or leave it: an all-zero `Numerator` would make this division abort with the generic "nonzero divisor" message, but it can't get here from a user config — `RationalImpedanceData` rejects empty and all-zero polynomials at parse. If we want `validate-config` to flag it too (without running palace), draft-07 can express it on `Numerator`/`Denominator` in `$defs/RationalImpedance`:

```json
"contains": { "not": { "const": 0 } }
```

Checked against the embedded nlohmann-json-schema-validator: rejects `[0.0, 0.0]` and `[0, 0]` with a pointed error, accepts any nonzero entry. Needs a SchemaVer patch bump; arguably PR 770 follow-up scope rather than this PR.

test/data/regression/input/rational_impedance/series_rlc18_rational_eigen.json
```diff
@@ -0,0 +1,76 @@
...
+      {
+        "Attributes": [5],   // Termination under test (z = L)
+        // Zs(s) = N(s)/D(s), s = i*omega [rad/s], coefficients highest-degree-first.
+        // Series RLC: Zs = R + sL + 1/(sC) = (s^2 LC + s RC + 1)/(sC) with
+        //   R = 376.730313668 Ω/sq, L = 2.0e-8 H/sq, C = 3.908996282497599e-15 F/sq,
+        // a notch resonant at 18 GHz (Zs -> R = eta0, matched absorber) with
+        // Q = w0 L / R = 6. The Robin coefficient s*D/N has a complex pole pair (the
+        // zeros of Zs) inside the interpolation window [Target, 3*Target] = [9, 27] GHz,
+        // which ruins the quadratic NLEPS seed fit and exercises the freeze branch of
+        // the fit-or-freeze seed strategy
```
hughcars: This is a very wordy comment for a test config file.

palace/models/surfacerationalimpedanceoperator.hpp
```diff
@@ -72,6 +78,31 @@ class SurfaceRationalImpedanceOperator
...
+
+  // Stamp a unit coefficient (including the per-attribute crack scaling) for boundary
+  // index idx, so that the assembled boundary mass matrix M_b satisfies: contribution of
+  // boundary idx to A2(λ) = g(λ)·M_b with g = EvalRobinCoef(idx, ·). Used by the nonlinear
+  // eigensolver seed to fit or freeze the boundary's Robin coefficient.
+  void AddUnitBdrCoefficient(int idx, MaterialPropertyCoefficient &fb) const;
+
+  // Evaluate the scalar Robin coefficient g(s) = s·D(s)/N(s) of boundary idx at complex
+  // s = iω = λ (nondimensional).
+  std::complex<double> EvalRobinCoef(int idx, std::complex<double> s) const;
```
hughcars: Two naming nits against the local conventions:

- `EvalRobinCoef` is the only single-f "Coef" in the codebase — "Coefficient(s)" is spelled out everywhere else (including `AddUnitBdrCoefficient` right above) → `EvalRobinCoefficient`.
- `AddUnitBdrCoefficient` is the sole singular in the `Add*BdrCoefficients` family (`AddStiffnessBdrCoefficients`, `AddDampingBdrCoefficients`, `AddBoundaryMassBdrCoefficients`, ...) → `AddUnitBdrCoefficients`.

palace/models/spaceoperator.hpp
```diff
@@ -249,6 +250,27 @@ class SpaceOperator
...
 
+  // Construct the ω-independent boundary curl-curl matrix M_ff for the 2nd-order farfield
+  // (absorbing) BC, with unit coefficient, stored on the real slot. The full A2
+  // contribution at frequency ω is `i·(0.5/ω)·M_ff` (real-ω stamping) or `-0.5/λ·M_ff`
+  // (complex-λ analytic continuation); a caller scales this matrix by the appropriate
+  // complex coefficient. Returns a null pointer if the farfield BC order < 2 or it
+  // contributes no DoFs on this rank. Used by the NLEPS HYBRID frozen-ABC seed strategy.
+  template <typename OperType>
+  std::unique_ptr<OperType>
+  GetFarfieldExtraBoundaryMatrix(Operator::DiagonalPolicy diag_policy);
```
hughcars: Naming: the per-BC ω-independent-factor getters follow `Get<BC>Boundary<Kind>Matrix` — `GetWavePortBoundaryMassMatrix`, and your `GetRationalImpedanceBoundaryMassMatrix` below matches it exactly. This one drops the operator kind and reorders → `GetFarfieldBoundaryCurlCurlMatrix` (also mirrors `AddExtraSystemBoundaryCurlCurlBdrCoefficients`, which assembles it).

hughcars (changes requested):
The math checks out — pole coefficient, `AddFrozenPole` algebra, `DetermineFrozen` error comparison, long division, and the test config values are all consistent. Nice work. A few small suggestions inline.

One real finding: `Mult`/`AddMult` don't apply the frozen corrections that `GetInterpolationOperator` does. Since fixing it properly means restructuring the class internals, I put a fix on a branch instead of asking — two commits, eigen regression suite passes. Cherry-pick if you like them, ignore if not.

Two observations, nothing required:

- The ABC can't just go through `DetermineFrozen` like the rational boundaries: the window-error metric would choose fit (~6% error vs ~200% for freeze), but fitting is exactly what causes the spurious roots — the fictitious λ² the fit puts into M'. So the unconditional freeze is deliberate in a way the code doesn't say; a comment there would protect it from a future "simplification". Relatedly, the fit branch of the rational decision has no eigen test coverage.

- The −0.5/λ in the driver duplicates knowledge the ABC operator already owns (it stamps i·0.5/ω). At some point it might be nice for each frequency-dependent BC to expose its A2 term as an (f(λ), M) pair so the driver can treat them uniformly — follow-up sized, not for this PR.

The branch with both commits is [`hughcars/nleps-update`](https://github.com/awslabs/palace/tree/hughcars/nleps-update), cut from your current head so it cherry-picks cleanly.

palace/models/spaceoperator.hpp
simlapointe: Fixed

palace/models/surfacerationalimpedanceoperator.hpp
simlapointe: Changed to names so they're consistent

test/data/regression/input/rational_impedance/series_rlc18_rational_eigen.json
simlapointe: Trimmed the comment.

palace/models/surfacerationalimpedanceoperator.cpp
simlapointe: Ok, let's punt on this for now but worth doing later.

palace/drivers/eigensolver.cpp
simlapointe: Right, moved it inside the `if`.

palace/linalg/nleps.cpp
simlapointe: Thanks for the fixes! Added your commits to my branch

palace/models/surfacerationalimpedanceoperator.cpp
hughcars: Filed as #834.

hughcars (approved):

Issue #602, opened 2026-01-14
The CPW (lumped ports, adaptive) example is slow on GPUs


Sbozzolo:
(This might be related to #375.)

On CPU, the various CPW tests take approximately the same to complete (within a factor of 2-3). On GPU, the CPW (lumped ports, adaptive) regression test takes more than 20 minutes, much longer than the other tests.

PR #794, opened 2026-06-30
Prune tests

Sbozzolo:
Remove unusual superbuild combinations.

### DIFF
--- a/test/unit/regression/cases.cpp
+++ b/test/unit/regression/cases.cpp
@@ -360,7 +360,7 @@ TEST_CASE("cpw_wave_uniform", "[Serial][Parallel][GPU][Regression]")
 // Adaptive frequency sweeps run as smoke tests: structure (output tree,
 // CSV set, headers, row/column counts) is enforced, numeric comparison
 // is disabled via infinite tolerances.
-TEST_CASE("cpw_lumped_adaptive", "[Serial][Parallel][GPU][Regression]")
+TEST_CASE("cpw_lumped_adaptive", "[Serial][Parallel][Regression]")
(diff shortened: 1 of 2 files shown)


simlapointe (changes requested):
Can we have at least one linux test setup using ARPACK instead of SLEPc? Otherwise LGTM.

test/unit/regression/cases.cpp
```diff
@@ -360,7 +360,7 @@ TEST_CASE("cpw_wave_uniform", "[Serial][Parallel][GPU][Regression]")
 // Adaptive frequency sweeps run as smoke tests: structure (output tree,
 // CSV set, headers, row/column counts) is enforced, numeric comparison
 // is disabled via infinite tolerances.
-TEST_CASE("cpw_lumped_adaptive", "[Serial][Parallel][GPU][Regression]")
+TEST_CASE("cpw_lumped_adaptive", "[Serial][Parallel][Regression]")
```
hughcars, line 363: Is the gpu lumped adaptive performance poor too? I thought it was ok.

hughcars (approved):

hughcars:
Once we're happy with this, we'll have to edit the required list down. I can just delete all the cmake required temporarily, then once this merged add them from the new list.

test/unit/regression/cases.cpp
Sbozzolo: https://github.com/awslabs/palace/issues/602

I think at some point it was fine, but not for a while

simlapointe (approved):

PR #796, opened 2026-07-06
Fix for Spack PR 52627

Sbozzolo:
https://github.com/spack/spack/pull/52627 changes how env is resolved. So, here we add PWD to search for the Dockerfile template in the current directory as well


hughcars:
Looks like the `build-container` jobs are hanging? Does `.` need to be `"."`?

Sbozzolo:
> Looks like the `build-container` jobs are hanging? Does `.` need to be `"."`?

It was green here: https://github.com/awslabs/palace/actions/runs/28828623021

But I changed it to `"."` just to make sure

Sbozzolo:
Included in #797

Issue #507, opened 2025-10-01
Add note on complex normalization to docs


phdum-a:
In Palace, time averages of complex numbers are defined without the factor of 1/2. I.e. the power is $P = VI^*$. This means that complex amplitudes values reported for $V$, $I$ are RMS not peak. We should clarify this in the documentation.

PR #804, opened 2026-07-09
Fix S-parameters when Floquet ports are used with lumped or wave ports

### LINKED ISSUE #802: port-S.csv contains raw un-normalized projections when Floquet and lumped ports coexist
In a driven simulation with both a FloquetPort and lumped ports (each with its own excitation), `MeasureSParameter()` intentionally skips S-parameter post-processing (postoperator.cpp:1329, "Floquet ports use a different modal basis and are only handled when they are the sole port type"). However, `InitializePortS()` in postoperatorcsv.cpp:1108 only checks `nr_ports == 0 || !IsMultipleSimple()` and does not replicate that condition, so `port-S.csv` is still written with the raw projections from `MeasureLumpedPorts()` labeled as S-parameters.

The written values are wrong in two ways:

1. For lumped excitations, the incident subtraction (`vi.S.real() -= 1.0`) is never applied, so the diagonal entries are `1 + S_ii` instead of `S_ii` (reflection coefficients apparently well above 0 dB).

2. For the Floquet excitation, the incident wave is normalized to unit time-averaged power (`p_unit = lambda_eff * A / (2 * omega * mu_r)` in floquetportoperator.cpp:781), while the lumped projection `V / sqrt(R)` follows the peak convention shared with wave ports (`|b|^2 = 2 * P_avg`). The lumped entries in the Floquet-driven column are therefore sqrt(2) too large (+3.01 dB in power).

I ran into this on a periodic RIS unit cell (doubly periodic BCs, FloquetPort with TE normal incidence, two 50 ohm lumped ports on diode gaps, 3-4 GHz sweep, 8 AMR iterations to ~5M DOF). The effect is stable under refinement, so it is not a discretization artifact. Power balance for the Floquet-driven column at the last iteration:

    f (GHz)   sum |S|^2 as written   with lumped power / 2
    3.0       1.207                  0.9991
    3.5       1.417                  0.9981
    4.0       1.620                  0.9966

Dividing the lumped power by exactly 2 closes the balance to 0.997-0.999 over all 101 frequency points (the small residual is substrate loss). Similarly, applying the missing `-1` to the lumped-driven diagonal restores passivity band-wide.

I would suggest either adding the same Floquet-coexistence condition to `InitializePortS()` so the file is simply not written in the unsupported configuration, or supporting the mixed case in `MeasureSParameter()`: the only missing ingredient is the sqrt(2) bridge on Floquet-driven lumped/wave observations (or equivalently normalizing the Floquet incident wave to the peak convention by dropping the factor 2 in p_unit, which leaves Floquet-to-Floquet S invariant but would need a check on absolute-power outputs). The mixed configuration is the natural workflow for metasurface unit-cell characterization with embedded device ports, so I'd be happy to open a PR for either direction, and the power-balance closure above could serve as a regression test.

Observed on main at 9b9d652 by code inspection and with a build of main + unrelated local changes (none touching these files).

simlapointe:
Port S-parameters reported when Floquet ports were used with either lumped or wave ports were previously incorrect, because 1) `MeasureSParameter()` was skipped and 2) the power normalization was incorrect. This fixes them by fully enabling cross-type port S-parameters and applying the correct power normalization.

Closes #802.


palace/models/postoperator.cpp
```diff
@@ -1363,6 +1359,17 @@ void PostOperator<solver_t>::MeasureSParameter() const
                        fem_op->GetWavePortOp().GetPort(drive_port_idx).d_offset)
             : std::complex<double>{1.0, 0.0};
 
+    // Power normalization bridge factor for mixed Floquet + lumped/wave configurations.
+    // Floquet ports normalize incident waves to unit time-averaged power (P_avg = 1),
+    // while lumped/wave ports normalize to unit peak power (∫E×H*·n dS = 1, P_avg = ½).
+    // Cross-type observations require a √2 correction:
+    //   - Floquet drives, lumped/wave observes: lumped |b|² = 2×P_avg, divide by √2
+    //   - Lumped/wave drives, Floquet observes: Floquet |S|² = P_avg, multiply by √2
```
hughcars, line 1367: I think we should use this as extra motivation to fix the convention. Can you link to the issue (does one already exist?) about fixing this?

test/data/regression/ref/floquet_wave/floquet_wave/port-floquet-S.csv
```diff
@@ -0,0 +1,4 @@
...
...        +nan,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan,                         +nan,                       +nan,                         +nan,                       +nan,                        +nan,                       +nan,                        +nan,                       +nan,                        +nan,                       +nan,                        +nan,                       +nan,                        +nan,                       +nan,                        +nan,        -7.363521156122e-02,        +6.953314035467e+01,        -2.565789571148e+02,        +1.181982567837e+02,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan,                        +nan,                       +nan,                        +nan,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan,                       +nan
```
hughcars, line 2: pages of nans like this do make me sad, I wish we had a cleaner solution for this.

CHANGELOG.md
```diff
@@ -12,6 +12,16 @@ The format of this changelog is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/).
```
hughcars, line 13: Tangent to this, but maybe we need to mention the SchemaVer convention here.

hughcars (approved):
I'm not a fan of how we have to mix the power normalization conventions here, but this is the correct solution for now. We should use this as strong motivation to fix it.

palace/models/postoperator.cpp
simlapointe: Only issue I see that's somewhat related is #507 but only relates to the docs. I can open another issue.

CHANGELOG.md
simlapointe: Yeah good point, suggested in #805.

test/data/regression/ref/floquet_wave/floquet_wave/port-floquet-S.csv
simlapointe: Agreed. I'll merge this for now, but can open an issue to nudge us to think of a better way.

PR #810, opened 2026-07-13
Fix issue with adaptively refined periodic meshes

### LINKED ISSUE #809: Adapted mesh saved by SaveAdaptMesh cannot be reloaded when periodic boundaries are used
With nonconformal AMR on a periodic problem and "SaveAdaptMesh": true, the saved mesh cannot be used as input for a subsequent solve. Loading fails at startup:

    Verification failed: (bdr_v_donor.size() == bdr_v_receiver.size()) is false:
     --> Different number of vertices on donor and receiver boundaries. Cannot create periodic mesh.
     ... in function: std::vector<int> palace::mesh::DeterminePeriodicVertexMapping(...)
     ... in file: palace/utils/geodata_impl.cpp:982

The periodic mapping is rebuilt from the config on every load and requires matching donor/receiver discretizations, which the gmsh Periodic constraint guarantees for the initial mesh but the serialized NC-refined mesh does not.

Note also that the saved mesh path is derived from the input mesh stem, so the file is overwritten at every AMR iteration.

The use case is re-using the adapted mesh for fixed-mesh solves with different excitations or port terminations without re-running AMR.

simlapointe:
Fixed a bug where a mesh saved with `"SaveAdaptMesh"` during nonconformal AMR on a periodic problem could not be used as the input mesh for a subsequent solve. Loading now detects the already-periodic mesh and skips the periodicity construction. Also added chunked MPI broadcasting to read large NC meshes. Closes #809.

### DIFF
--- a/palace/utils/geodata_impl.cpp
+++ b/palace/utils/geodata_impl.cpp
@@ -977,13 +977,10 @@ DeterminePeriodicVertexMapping(std::unique_ptr<mfem::Mesh> &mesh,
     }
   }
 
-  // A mesh saved after nonconformal AMR on a periodic mesh has a single boundary
-  // element per periodic seam face carrying only one of the two attributes (NCMesh
-  // stores one attribute per face, so the donor/receiver pair sharing a face after
-  // MakePeriodic collapses to one). Exactly one of the two sets being empty is thus
-  // the signature of an already-periodic mesh: skip MakePeriodic by returning an
-  // empty mapping. Both sets empty means the config names attributes not present in
-  // the mesh at all.
+  // A periodic mesh saved after nonconformal AMR retains only one of the two periodic
+  // boundary ttributes. Exactly one of the two sets being empty is thus the signature of an
+  // already-periodic mesh: skip MakePeriodic by returning an  empty mapping. Both sets
+  // empty means the config names attributes not present in the mesh at all.
   MFEM_VERIFY(!bdr_v_donor.empty() || !bdr_v_receiver.empty(),
               "No boundary elements found for donor or receiver attributes of a periodic "
               "boundary pair!");
--- a/palace/utils/geodata_impl.cpp
+++ b/palace/utils/geodata_impl.cpp
@@ -978,8 +978,8 @@ DeterminePeriodicVertexMapping(std::unique_ptr<mfem::Mesh> &mesh,
   }
 
   // A periodic mesh saved after nonconformal AMR retains only one of the two periodic
-  // boundary ttributes. Exactly one of the two sets being empty is thus the signature of an
-  // already-periodic mesh: skip MakePeriodic by returning an  empty mapping. Both sets
+  // boundary attributes. Exactly one of the two sets being empty is thus the signature of
+  // an already-periodic mesh: skip MakePeriodic by returning an empty mapping. Both sets
   // empty means the config names attributes not present in the mesh at all.
   MFEM_VERIFY(!bdr_v_donor.empty() || !bdr_v_receiver.empty(),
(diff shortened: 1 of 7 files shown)


palace/utils/geodata_impl.cpp
```diff
@@ -977,6 +977,21 @@ DeterminePeriodicVertexMapping(std::unique_ptr<mfem::Mesh> &mesh,
...
+  // an already-periodic mesh: skip MakePeriodic by returning an empty mapping. Both sets
+  // empty means the config names attributes not present in the mesh at all.
+  MFEM_VERIFY(!bdr_v_donor.empty() || !bdr_v_receiver.empty(),
+              "No boundary elements found for donor or receiver attributes of a periodic "
+              "boundary pair!");
+  if (bdr_v_donor.empty() || bdr_v_receiver.empty())
+  {
+    Mpi::Print("Mesh is already periodic for donor/receiver attributes [{}]/[{}], "
+               "skipping MakePeriodic\n",
+               fmt::join(da, ", "), fmt::join(ra, ", "));
```
hughcars: This isn't the only scenario, better to print something like "One donor/receiver attribute is empty, assuming mesh is already periodic", or more neatly worded than that. Basically a user could still send one empty by accident, and what was previously an error we are now assuming means already periodic.

test/unit/test-communication.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <cstdint>
+#include <numeric>
+#include <vector>
+#include <catch2/catch_test_macros.hpp>
+#include "utils/communication.hpp"
+
+using namespace palace;
+
+TEST_CASE("BroadcastLarge", "[communication][Serial][Parallel]")
```
hughcars: Really this is only non trivial when a parallel test, can drop `[Serial]`

palace/utils/geodata_impl.cpp
```diff
@@ -977,6 +977,21 @@ DeterminePeriodicVertexMapping(std::unique_ptr<mfem::Mesh> &mesh,
...
+    return {};
```
hughcars, line 996: Could we extend the existing `PeriodicGmsh` unit test instead of adding new regression data? For example:

```cpp
mesh->EnsureNCMesh(true);
std::stringstream buffer;
mesh->Print(buffer);
buffer.seekg(0);
mesh = std::make_unique<mfem::Mesh>(buffer, false, true, false);
REQUIRE((mesh->bdr_attributes.Find(1) >= 0) !=
        (mesh->bdr_attributes.Find(2) >= 0));
REQUIRE(mesh::DeterminePeriodicVertexMapping(
            mesh, boundary_torus.periodic.boundary_pairs.front()).empty());
```

This also needs `#include <sstream>`.

hughcars (commented):
I reproduced the original periodic `SaveAdaptMesh` reload failure and confirmed this fixes it. I left a few inline comments around focused unit coverage, diagnostics, and minor cleanup.

palace/utils/geodata_impl.cpp
simlapointe: Changed the message a bit and made it a warning in 6c7ec2f

test/unit/test-communication.cpp
simlapointe: Dropped the serial tag in 6c7ec2f

palace/utils/geodata_impl.cpp
simlapointe: Good point, added this unit test in 6c7ec2f

hughcars (approved):

PR #811, opened 2026-07-13
Enable circuit synthesis with frequency-dependent boundary conditions

simlapointe:
This PR enhances the circuit synthesis capability in Palace by:
- Enabling circuit synthesis when (nonlinear) frequency-dependent BCs (e.g. waveports, second-order absorbing, surface conductivity, rational impedance) are present. Nonlinear terms are fit using AAA.
- Computing and printing the eigenvalues of the synthesized circuit (in `rom-eigenvalues.csv` and eigenvectors in `rom-eigenvectors-`) along with their absolute and backward errors in the HDM system.
- Outputting `rom-portload-` and `rom-port-reference` files to help cascading circuits with frequency-dependent BCs.


palace/models/romoperator.cpp
```diff
@@ -1049,46 +2289,380 @@ void RomOperator::PrintPROMMatrices(const Units &units, const fs::path &post_dir
...
+        duplicate = true;
+        break;
+      }
+    }
+    if (!duplicate)
+    {
+      deduped.push_back(m);
+    }
   }
+  return deduped;
```
hughcars: The complex frequency is not a unique mode identifier. Two linearly independent modes may have exactly the same complex eigenfrequency, but this loop keeps only the first, so the synthesized eigenvalue and eigenvector output loses the eigenspace's multiplicity. Each accepted entry already carries the normalized right eigenvector returned by `zggev`; please preserve every accepted eigenpair and return the sorted `modes` list directly.

Frequency-only deduplication also cannot distinguish a redundant numerical representation from genuine degeneracy. If redundant eigenpairs are observed, they need an eigenvector/rank-aware policy rather than unconditional frequency clustering.

```suggestion
  return modes;
```

palace/utils/aaa.hpp
```diff
@@ -0,0 +1,319 @@
...
+  if (m == 0)
+  {
+    return out;
+  }
+  std::complex<double> sum_w = r.wj.sum();
+  std::complex<double> sum_wf = (r.wj.array() * r.fj.array()).sum();
+  // The asymptote d = r(∞) = Σwf/Σw is undefined when the weights sum to (near) zero
+  // (the barycentric rational is strictly proper). Fall back to d = 0 rather than
+  // propagating a NaN through the pole-residue form.
+  out.d = (std::abs(sum_w) > 1.0e-14 * r.wj.cwiseAbs().maxCoeff()) ? sum_wf / sum_w : 0.0;
```
hughcars: A nearly zero `sum(w)` can lower the degree of the cleared denominator and make the barycentric interpolant **improper**; it does not imply that the polynomial part is zero. For example, with `z={0,1}`, `f={0,1}`, and `w={-1,1}`, the barycentric interpolant is exactly `r(z)=z`, even though `sum(w)=0`. This fallback instead produces a zero polynomial part, so the converted pole-residue representation loses the function.

The conversion needs to clear the barycentric denominators, perform polynomial division, retain the full quotient, and reject quotient degrees that cannot fit the quadratic synthesis pencil. A numerically stable implementation is in [`6643bbeb6`](https://github.com/awslabs/palace/commit/6643bbeb6bc95ec12a6f40f621152fd735c32069); it also keeps quotient and pole extraction in an affine-normalized coordinate to avoid cancellation under translated supports.

palace/models/romoperator.cpp
```diff
@@ -944,102 +1295,991 @@ void RomOperator::SolvePROM(int excitation_idx, double omega, ComplexVector &u)
...
+  // lumped-port rows is NumSynthesisPortModes(), NOT the total number of lumped ports.
   const long n_port_modes = static_cast<long>(NumSynthesisPortModes());
   MFEM_ASSERT(n_port_modes <= GetReducedDimension(),
               "More lumped port modes than PROM basis vectors; basis is inconsistent!");
-  for (long j = 0; j < n_port_modes; j++)
+  // Wave-port basis vectors come right after the lumped-port rows; count how many were
+  // actually added (real + imaginary parts each count separately, see
+  // AddWavePortModesForSynthesis → UpdatePROM).
+  long n_waveport_rows = 0;
+  if (!Mwp_p_r.empty())
```
hughcars: `Mwp_p_r` only contains active wave ports because `GetWavePortBoundaryMassMatrix` ultimately returns no mass for `!active`, while `AddWavePortModesForSynthesis` adds every port with `IncludeInSynthesis=true`. Therefore an inactive included port can have `waveport_*` basis rows while this gate leaves `n_waveport_rows=0`, so those rows miss the physical `orth_R` scaling. The same missing mass also prevents its dispersion fit and `rom-port-reference.csv` entry.

`Active` and `IncludeInSynthesis` are independent: this port should remain an unloaded circuit terminal. Simply exposing the mass is not sufficient because it must still be excluded from the physical HDM/PROM Robin term and from both the loaded synthesized pencil and removable port loads. [`45cc11b61`](https://github.com/awslabs/palace/commit/45cc11b619385a0336e4080e095ce91023bf46d5) implements those guards while retaining the mass, normalization, fit, and reference data.

palace/utils/aaa.hpp
```diff
@@ -0,0 +1,319 @@
...
+    B_geig(1 + j, 1 + j) = 1.0;
+  }
+  char jobvl = 'N', jobvr = 'N';
+  int n = static_cast<int>(m + 1);
+  int lwork = 4 * n;
+  std::vector<std::complex<double>> alpha(n), beta(n), work(lwork);
+  std::vector<double> rwork(8 * n);
+  int info = 0;
+  zggev_(&jobvl, &jobvr, &n, A_geig.data(), &n, B_geig.data(), &n, alpha.data(),
+         beta.data(), nullptr, &n, nullptr, &n, work.data(), &lwork, rwork.data(), &info);
```
hughcars: `zggev_` reports invalid arguments and QZ nonconvergence through `info`. If it fails, consuming `alpha` and `beta` below can silently create bogus AAA poles. This is the solve call—not a workspace query—and the existing `lwork = 4*n` satisfies ZGGEV's minimum, so please fail immediately here, consistently with the other `zggev_` call in `romoperator.cpp`.

```suggestion
         beta.data(), nullptr, &n, nullptr, &n, work.data(), &lwork, rwork.data(), &info);
  MFEM_VERIFY(info == 0, "zggev failed with info = " << info);
```

hughcars (commented):
This looks close. I found a few edge cases in the synthesis path that I think are worth considering:

- Frequency-only eigenvalue deduplication can collapse genuinely degenerate modes.
- AAA conversion can lose the polynomial part of an improper barycentric fit.
- An inactive wave port included in synthesis can miss its normalization and reference output.
- The AAA `zggev_` call should check `info` before consuming its results.

I left details inline. For the two changes that do not fit cleanly in suggestions, I put possible fixes in [`6643bbeb6`](https://github.com/awslabs/palace/commit/6643bbeb6bc95ec12a6f40f621152fd735c32069) and [`45cc11b61`](https://github.com/awslabs/palace/commit/45cc11b619385a0336e4080e095ce91023bf46d5). Both are based on the current PR tip and tested through Spack.

One non-blocking test note: the synthesis regression does not currently cover the raw matrices, eigenvectors, HDM errors, or inactive included wave ports.

palace/models/romoperator.cpp
simlapointe: Good point. I never liked the deduplication attempt I had anyway. Removed it and left a comment in [d6b547d](https://github.com/awslabs/palace/pull/811/commits/d6b547d6c405016245e887cdb57f194cdc40f82e)

palace/utils/aaa.hpp
simlapointe: Good catch. Added your commit to my branch

palace/models/romoperator.cpp
simlapointe: Added to my branch.

palace/utils/aaa.hpp
simlapointe: Added the verify in [d6b547d](https://github.com/awslabs/palace/pull/811/commits/d6b547d6c405016245e887cdb57f194cdc40f82e)

simlapointe:
Thanks for the feedback. Regarding the regression not covering raw ROM matrices and eigenvectors; these outputs can vary quite a bit with np, architecture, etc. so that's why I only compare eigenvalues which are stable run-to-run.

palace/utils/aaa.hpp
```diff
@@ -0,0 +1,489 @@
...
+  MFEM_VERIFY(info == 0, "zggev failed with info = " << info);
+  // Collect the finite eigenvalues. Their count is known structurally: the poles are the
+  // roots of the cleared denominator, so exactly deg(D) of the m+1 pencil eigenvalues are
+  // finite and the rest lie at infinity. ZGGEV usually signals infinity with beta = 0
+  // exactly, but QZ roundoff (BLAS/architecture dependent) can instead produce a small
+  // nonzero beta — a bogus "finite" pole at ~1/eps — and no fixed |beta| threshold is
+  // robust when the denominator degree drops (many eigenvalues at infinity, e.g. a purely
+  // polynomial interpolant). Instead keep exactly deg(D) eigenvalues, selecting those
+  // farthest from infinity in the chordal metric |beta| / sqrt(|alpha|² + |beta|²).
+  const long n_finite = static_cast<long>(denominator.size()) - 1;
```
hughcars: Thanks for chasing down the ARM failure. I think this selection now mixes two different denominator representations: `n_finite` comes from the tolerance-trimmed `denominator`, but `alpha`/`beta` still come from the original `xj`/`wj` pencil. If trimming removes a nonzero leading coefficient, that pencil still contains the corresponding finite roots, so retaining only `deg(trimmed D)` eigenvalues can drop one and corrupt the reconstruction.

A small counterexample is:

```cpp
constexpr long m = 3;
const double tau = 64.0 * std::numeric_limits<double>::epsilon() * m;
auto D = [tau](double x)
{
  return 1.0 + 1.01 * tau * x + 0.99 * tau * x * x;
};

r.zj << -1.0, 0.0, 1.0;
r.wj << 0.5 * D(-1.0), -D(0.0), 0.5 * D(1.0);
for (long j = 0; j < m; j++)
{
  r.fj(j) = 1.0 / D(r.zj(j).real());
}
```

The trimming rule retains the linear coefficient but drops the quadratic coefficient, so it requests one finite pole while the original pencil still has two. On the current head I get:

```text
|EvaluatePoleResidue(pr, 0) - EvaluateAAA(r, 0)|
= 2.2762774338974e13
```

Could we instead retain the remainder from the trimmed-polynomial division, solve that same denominator with a companion pencil, and compute residues consistently from it?

```cpp
auto remainder = numerator;
auto quotient = PolynomialDivide(remainder, denominator);
auto x_poles = PolynomialRoots(denominator);

for (const auto x_pole : x_poles)
{
  poles.push_back(center + support_scale * x_pole);
  residues.push_back(
      support_scale * EvaluatePolynomial(remainder, x_pole) /
      EvaluatePolynomialDerivative(denominator, x_pole));
}
```

I tested that form through Spack: all 7 AAA cases pass, including this reproducer and the translated-polynomial case.

hughcars (commented):
One small remaining issue, but essentially done

palace/utils/aaa.hpp
simlapointe: Good catch. Implemented this approach and added the unit test in [93504e8](https://github.com/awslabs/palace/pull/811/commits/93504e8a1ad20628b789538f14606c8925fcd27c)

palace/utils/aaa.hpp
```diff
@@ -0,0 +1,492 @@
...
+      return;
+    }
+    const double tol =
+        64.0 * std::numeric_limits<double>::epsilon() * std::max<long>(m, 1) * scale;
+    while (!p.empty() && std::abs(p.back()) <= tol)
+    {
+      p.pop_back();
+    }
+  };
+  trim_leading_zeros(numerator);
```
hughcars: Thanks—this implements the representation change I suggested. I realized my earlier sketch omitted one stabilization that was present in the version I tested. The new regression checks only `z = 0`, where the large quotient and pole terms cancel exactly. On this commit, evaluating the same case at `z = 0.5 + 0.25i` gives an error of `3.90625e-3`.

When trimming removes leading denominator coefficients, the corresponding `f_center * D` contribution needs to be transferred into the centered numerator before division:

```cpp
const auto untrimmed_denominator = denominator;
trim_leading_zeros(denominator);

for (std::size_t k = denominator.size(); k < untrimmed_denominator.size(); k++)
{
  numerator[k] += f_center * untrimmed_denominator[k];
}
trim_leading_zeros(numerator);
```

Otherwise division creates polynomial and pole terms of approximately `2.27e13`, whose cancellation loses precision away from zero. I tested this minimal addition on top of the current implementation, with the regression also checking `0.5 + 0.25i`; all 264 AAA assertions pass. Sorry—the missing step should have been included in my original sketch.

hughcars (commented):
One more small fix because I failed to make clear in the review

palace/utils/aaa.hpp
simlapointe: My bad, was a bit sloppy. Thanks for catching this!

simlapointe: Fixed in [3ea2ed](https://github.com/awslabs/palace/pull/811/commits/73ea2ed113a658df219eff8aa06e73ea7190e99e)

hughcars (approved):

PR #813, opened 2026-07-13
Add node-local-aware EnsureDirectory helper

Sbozzolo:
Palace created output directories only on the global root rank, which breaks on clusters where each node has its own local filesystem: ranks on non-root nodes write per-rank output into a directory that was never created there.

Add `EnsureDirectory(dir, comm)`: the global root creates the directory, then one rank per shared-memory node (split via MPI_COMM_TYPE_SHARED) ensures it exists on that node's filesystem. On a shared filesystem the extra creates hit an already-existing directory harmlessly.


test/unit/test-outputdir.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+  EnsureDirectory(dir, MPI_COMM_SELF);
+
+  CHECK(fs::is_directory(dir));
+  {
+    std::ofstream f(dir / "written.txt");
+    f << "ok";
+  }
+  CHECK(fs::is_regular_file(dir / "written.txt"));
+}
```
simlapointe: I'm not sure this test does what is intended. I tried putting both a print statement and an abort in the if statement at line 47 of palace/utils/outputdir.hpp and it never triggers. Only the global-root creation fires for all ranks, the node-root creation never triggers. For the test to trigger the node-root creation, you might need to use a fake per-rank "node" communicator built with MPI_Comm_split(color=rank)

simlapointe (commented):
Generally looks good. One small comment below about a test I'm not sure is testing what's intended. Also, am I right that with this change the directory _creation_ is now node-local, but the directory _cleanup_ is still root-only? Could this produce stale/mixed artifacts on node-local filesystems?

Sbozzolo:
> Generally looks good. [...] Could this produce stale/mixed artifacts on node-local filesystems?

Good point. I was mostly thinking of node-local filesytems as ephemeral (e.g., `/tmp/` or a temporary runner), but this would be a problem for persistent node-local filesystems. I changed the PR to also do node-local handling for that.

I also realized that `SaveIteration` should do the same.

test/unit/test-outputdir.cpp
Sbozzolo: Yeah, I just removed the test. It was pretty hacky already.

Sbozzolo:
I verified that paraview and gridfunctions are correctly written and moved around with AMR by running on two nodes, writing to `/tmp/` and ssh-into the nodes and looking at the folders while the simulation was running.

test/unit/test-outputdir.cpp
```diff
@@ -0,0 +1,173 @@
...
+  {
+    fs::create_directories(dir);
+  }
+  Mpi::Barrier(comm);
+  REQUIRE(fs::is_directory(dir));
+
+  // Regression guard: EnsureDirectory must NOT treat an already-existing
+  // directory as an error. It relies on the non-throwing create_directories
+  // overload and ignores the bool return, so a pre-existing directory returns
+  // cleanly rather than aborting.
```
simlapointe: Not sure this comment still applies, EnsureDirectory now uses the throwing overload with an is_directory check

simlapointe (approved):
LGTM! One nit about a possibly stale comment below.

test/unit/test-outputdir.cpp
Sbozzolo: Thank you!

PR #815, opened 2026-07-15
Protect shared Spack buildcache from incompatible writers

hughcars:
## Summary

Protect Palace CI from incompatible Spack buildcache writers by moving to version-isolated, writable GHCR cache packages.

- Replace the old monolithic `ghcr.io/awslabs/palace-develop-testing` cache.
- Use `ghcr.io/awslabs/palace-1.2` for the canonical pinned Spack release (`releases/v1.2`).
- Use `ghcr.io/awslabs/palace-develop` for the Spack `develop` lookahead job, so lookahead remains warm without being able to corrupt the pinned-release cache.
- Derive the cache package name from `inputs.spack-version` in the shared Palace CI action, so future Spack producer lines get isolated cache packages instead of sharing incompatible artifacts.
- Keep both Spack 1.2 and Spack `develop` caches writable, while stopping force-overwrites of existing DAG-hash tags.
- Rebuild the Spack 1.2 `index.spack` once per main Spack workflow, after all nine cache-writing build jobs finish. Regression jobs remain free to start as soon as their corresponding build completes, so index regeneration is not on their critical path.
- Rebuild the isolated Spack `develop` index at the end of its sole Lookahead job, avoiding any concurrent develop-index writers.
- Move documentation and container dependencies into the canonical Spack 1.2 cache and apply the same publishing rules.
- Keep the GHCR cleanup workflow pointed at both new isolated cache packages.
- Restrict `spack buildcache keys` to Spack's public mirror because the GHCR caches are unsigned.

## Motivation

The Intel CI began failing after a Spack-develop job re-published PETSc `3.25.2` (`/5wheh2s`) under the same OCI tag consumed by Spack 1.2. The same Spack 1.2/PETSc combination had passed immediately before that cache overwrite, and locally rebuilding the dependency from source also passes Palace configuration.

Spack DAG hashes do not identify the Spack version or buildcache representation which produced an OCI artifact. Different Spack versions must therefore not write the same GHCR package. At the same time, the develop lookahead jobs still need a warm cache to avoid becoming prohibitively slow, so this PR gives them their own isolated, writable cache instead of making them uncached.

Separately, the old `palace-develop-testing` GHCR package accumulated hundreds of thousands of versions. GitHub's package-level delete endpoint returns server errors for that package, and per-version cleanup is impractically slow. Rather than continue using a pathological cache, this PR moves CI to fresh versioned packages.

## Cache migration

New cache packages:

```text
ghcr.io/awslabs/palace-1.2      # Spack releases/v1.2 jobs
ghcr.io/awslabs/palace-develop  # Spack develop lookahead jobs
```

The old cache package is intentionally abandoned:

```text
ghcr.io/awslabs/palace-develop-testing
```

After this PR merges, `main` will stop reading from or writing to the old cache. The old package can be left unused or purged later by GitHub support; CI no longer depends on clearing it.

## Index management

The main Spack workflow has one `refresh-buildcache-index` job for `palace-1.2`. It runs with Spack 1.2 after all cache-writing build jobs, including when a build fails after banking some dependencies, and performs a full `spack buildcache update-index`.

The index is best-effort. The first run against a fresh package may probe individual OCI tags before `index.spack` exists. On later runs, readers use the previous complete index while the new one is generated; binaries absent from a slightly stale index remain discoverable through Spack's direct OCI-tag fallback. The isolated `palace-develop` package has its own index, refreshed by its sole Spack `develop` Lookahead job after Palace CI completes, and is never read or written by Spack 1.2 jobs.

## Validation

Validated pre-rebase Spack run (identical patch series): https://github.com/awslabs/palace/actions/runs/29551267711

- All edited GitHub workflow/action YAML parses successfully.
- Cache-name derivation maps `releases/v1.2` to `palace-1.2` and `develop` to `palace-develop`.
- All nine Spack 1.2 build jobs fetched and relocated their complete dependency sets from `palace-1.2` (62–84 cached dependencies per job); dependency installation took 51–95 seconds.
- The x64 default regression job fetched 57 dependencies from the cache; only branch-specific Palace was intentionally rebuilt.
- `refresh-buildcache-index` completed successfully in about 85 seconds. `ghcr.io/awslabs/palace-1.2:index.spack` now returns HTTP 200 and contains 754 indexed specs.
- In post-rebase Lookahead run `29581844622`, the Spack `develop` job fetched and relocated all 83 dependencies from `ghcr.io/awslabs/palace-develop`, emitted no index-fetch warning, and successfully refreshed the develop index in 63 seconds.
- Documentation and container workflows successfully consume `palace-1.2`.
- `buildcache push --force` remains removed.

### DIFF
--- a/.github/actions/build-container/spack_env/spack.yaml
+++ b/.github/actions/build-container/spack_env/spack.yaml
@@ -47,7 +47,7 @@ spack:
       url: https://binaries.spack.io/develop
     local-buildcache:
       binary: true
-      url: oci://ghcr.io/awslabs/palace
+      url: oci://ghcr.io/awslabs/palace-1.2
       signed: false
       access_pair:
         id_variable: GITHUB_USER
--- a/.github/actions/build-container/spack_env/spack.yaml
+++ b/.github/actions/build-container/spack_env/spack.yaml
@@ -47,7 +47,7 @@ spack:
       url: https://binaries.spack.io/develop
     local-buildcache:
       binary: true
-      url: oci://ghcr.io/awslabs/palace-develop-testing
+      url: oci://ghcr.io/awslabs/palace
       signed: false
       access_pair:
         id_variable: GITHUB_USER
--- a/.github/actions/palace-ci/action.yml
+++ b/.github/actions/palace-ci/action.yml
@@ -201,7 +201,7 @@ runs:
               url: https://binaries.spack.io/develop
             local-buildcache:
               binary: true
-              url: oci://ghcr.io/awslabs/palace-develop-testing
+              url: oci://ghcr.io/awslabs/palace
               signed: false
               access_pair:
                 id_variable: GITHUB_USER
@@ -231,7 +231,11 @@ runs:
       run: |
         source /tmp/timing_functions.sh
         start_timer
-        spack -e . buildcache keys -y --install --trust
+        # The local GHCR buildcache is unsigned and may not exist yet when a new
+        # cache package is bootstrapped. Only install/trust keys from Spack's
+        # public binary mirror; package installs will still use GHCR binaries
+        # opportunistically and fall back to source builds on cache misses.
+        spack -e . buildcache keys -y --install --trust spack
         end_timer "Configure Binary Mirror Keys"
 
     - name: Bootstrap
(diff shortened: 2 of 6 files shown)


Sbozzolo (approved):
Thank you!

PR #818, opened 2026-07-17
Check for duplicate SchemaVer entries in changelog

### LINKED ISSUE #817: CI check for SchemaVer in CHANGELOG
When merging many PRs with updates to SchemaVer, we have to be careful they don't double count.

Add a small CI check, similar to `check-license`, that scans `CHANGELOG.md` and fails if the same `SchemaVer X-Y-Z` appears more than once.

Something simple is enough, e.g. extract all `SchemaVer [0-9]+-[0-9]+-[0-9]+` entries, sort/count them, and fail on duplicates.

Acceptance:
- `SchemaVer 1-1-0` and `SchemaVer 1-2-0` both present once passes
- duplicate `SchemaVer 1-1-0` fails
- check runs as part of normal PR CI

orangekame3:
Adds a check to the Style workflow that rejects duplicate `SchemaVer` entries in `CHANGELOG.md`.

Tested locally with both unique and duplicate entries.

Closes #817

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -75,6 +75,14 @@ jobs:
           git fetch --no-tags --depth=1 origin "${GITHUB_BASE_REF}"
           ./scripts/check-schema-version "origin/${GITHUB_BASE_REF}"
 
+  check-changelog-schema-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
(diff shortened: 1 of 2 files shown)


hughcars (approved):
Thanks! This is exactly the level of simple I had in mind for this check. I verified locally that unique SchemaVer entries pass and duplicates fail.

PR #819, opened 2026-07-17
CI check for SchemaVer in CHANGELOG

### LINKED ISSUE #817: CI check for SchemaVer in CHANGELOG

hughcars:
Fixes #817.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -75,6 +75,14 @@ jobs:
           git fetch --no-tags --depth=1 origin "${GITHUB_BASE_REF}"
           ./scripts/check-schema-version "origin/${GITHUB_BASE_REF}"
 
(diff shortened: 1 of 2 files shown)


hughcars:
Overtaken by #818

PR #823, opened 2026-07-17
Improve ports documentation in reference.md

simlapointe:
Improve description of lumped and wave ports in `reference.md`. Based mostly on #696.

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -367,7 +367,12 @@ For TEM modes ``Z_{VI} = Z_{PV}``; for other modes they will generally differ.
 
 To Do:
 
-  - Fix dembedding sign and normal sign convention (outward vs inward normal)
+  - Fix dembedding sign and normal sign convention (outward vs inward normal):
+    In the boundarymode-2d branch, the normal sign convention: ``+\hat{\bm{n}}`` is the outward mesh normal.
+    The mode eigenvalue gives ``\text{Re}\{k_n\}\ge 0``, corresponding to forward propagation in the
+    ``+\hat{\bm{n}}`` direction. De-embedding
+    ``\tilde{S}_{ij} = S_{ij}e^{ik_{n,i}d_i}e^{ik_{n,j}d_j}`` uses positive ``d`` to shift the
+    reference plane toward the device (removing waveguide), consistent with [[8]](#References).
   - Update with 1/2 factors if we move to average power definition
   - link to impedance post-processing docs section once available.
 
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -259,7 +259,7 @@ respectively. If ``Z`` is a combination of LRC responses, these add in parallel
 Specifically, ``R_s = \alpha R``, ``L_s = \alpha L``, and ``C_s = C/\alpha``.
 
 In the configuration file, a user specifies either ``L``, ``R``, ``C``, corresponding to ``Z``, or
-``Ls``, ``Rs``, ``Cs`` corresponding to ``Z_s``. For a single-element port, these will just be
+``L_s``, ``R_s``, ``C_s`` corresponding to ``Z_s``. For a single-element port, these will just be
 converted using the single scale factor ``\alpha``. For a multi-element port, we require impedances
 to add in parallel ``{1}/{Z} = \sum_e {1}/{Z_e}``, so that each element sees the same voltage.
 Specifying a lumped ``Z`` in the config means that each element ``e`` can have a different physical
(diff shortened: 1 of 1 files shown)


docs/src/reference.md
```diff
@@ -170,59 +183,105 @@ The eigenmodes are normalized such that they have unit norm and their mean phase
...
-specify are given by ``R_s = \alpha R``, ``L_s = \alpha L``, and ``C_s = C/\alpha``, where
-``\alpha = w/l`` for a rectangular port or ``\alpha = 2\pi / \ln(b/a)`` for a coaxial
-port.
+Here we have allowed for the possibility that a port can be made of disjoint surface elements ``e``,
+which separately contribute to the Poynting vector.
 
-For multielement lumped ports, the effective circuit impedance is given by
+There is an additional normalization related to voltages and currents. For AC simulations of
+microwave networks, circuit ``V`` and ``I`` are not generally well defined and have to be fixed by
+convention [11,12]. This is tantamount to relating the circuit characteristic impedance
```
dnpham23: nit: These refs are written as plaint text here, whereas other citations in the file uses the linked form. Maybe making it [[11,12]](#References) for consistency?

simlapointe: Good catch. Fixed in [cbc5c83](https://github.com/awslabs/palace/pull/823/commits/cbc5c830a3f88c150ea3bc7980ab32e4247d211f)

docs/src/reference.md
```diff
@@ -7,6 +7,19 @@ SPDX-License-Identifier: Apache-2.0
...
+  - Electrostatics: time-independent voltage sources,
+  - Magnetostatics: time-independent current sources,
+  - Driven: monochromatic excitations ``\bm{U}^{inc}(f)``,
+  - Eigenmode: resonances of the system,
+  - Transient: time-dependent excitations ``\bm{U}^{inc}(t)``,
+  - Boundary mode: guided modes on a waveguide cross section.
+
+Here, we will summarize the mathematics of each, provide a description of boundary conditions, and
+specify convention choices. This should be read in conjunction with the "User Guide" and
+"Configuration File" documentation.
```
hughcars: Make these into links

docs/src/reference.md
```diff
@@ -170,59 +183,105 @@ The eigenmodes are normalized such that they have unit norm and their mean phase
...
 ## Lumped ports and wave ports
 
-For lumped port boundaries, the surface impedance can be related to an equivalent circuit
-impedance, ``Z``. There are two common cases:
+Ports have two effects. First, they impose a boundary condition that relates the electric and
+magnetic fields on the surface. Second, for the driven and transient solver, they can excite the
+system with an incident field.
+
+For lumped ports, the boundary condition is an impedance condition as described above — tangential
+fields satisfy ``\bm{E}_t = Z_s \bm{H}_t \times \bm{n}``. Here ``Z_s`` is the physical surface
```
hughcars, line 191: Not related to this PR, but we should be moving away from the double back tick, as I believe this is a Documenter feature rather than markdown. So if we move to a different renderer front end, this would break the math mode.

docs/src/reference.md
```diff
@@ -235,9 +294,98 @@ The incident field ``\bm{E}^{inc}(\bm{x},t)`` is
 \bm{E}^{inc}(\bm{x},t) = p(t)\bm{E}^{inc}(\bm{x})
 ```
 
-where ``\bm{E}^{inc}(\bm{x})`` is identical to the spatial excitation in the frequency
-domain formulation, and ``p(t)`` describes the temporal shape of the excitation. Possible
-options include a sinusoidal, Gaussian, modulated Gaussian, or step excitation.
+where ``\bm{E}^{inc}(\bm{x})`` is identical to the spatial excitation in the frequency domain
+formulation, and ``p(t)`` describes the temporal shape of the excitation. Possible options include a
+sinusoidal, Gaussian, modulated Gaussian, or step excitation.
```
hughcars: Link to the documentation.

docs/src/reference.md
```diff
@@ -235,9 +294,98 @@ The incident field ``\bm{E}^{inc}(\bm{x},t)`` is
...
+
+**Current.** The mode current is defined as a closed-loop path integral of the in-plane magnetic
+field:
+
+```math
+I = \oint_\mathcal{L} \bm{H}_t \cdot d\bm{l}
+```
+
+where ``\mathcal{L}`` is a user-specified closed contour encircling one conductor. The in-plane
+magnetic field ``\bm{H}_t = \mu_r^{-1}\bm{B}_t`` is obtained from the mode fields via
```
hughcars, line 357: This is correct, but we don’t actually compute this right now—I think there’s a bug for discontinuous \(\mu\). I’ve opened #893 to follow up.

hughcars (commented):
Some small nits, and one issue I noticed, but thats no blocker to this

docs/src/reference.md
simlapointe: Fixed in [e61ea53](https://github.com/awslabs/palace/pull/823/commits/e61ea5381773b7198ce518b965586b4474781541)

docs/src/reference.md
simlapointe: Fixed in [e61ea53](https://github.com/awslabs/palace/pull/823/commits/e61ea5381773b7198ce518b965586b4474781541)

docs/src/reference.md
simlapointe: That's a bug/limitation indeed. I left the documentation as is with the correct equation.

hughcars (approved):

PR #824, opened 2026-07-17
Add libCEED output functionals for GPU postprocessing

hughcars:
## Summary

Adds libCEED-backed reducing functionals for GPU-capable postprocessing. Surface, port, and far-field measurements use fixed MFEM-tabulated rules and grouped libCEED operators instead of repeatedly assembling host coefficients and linear forms.

This is the reducing-functional layer of the GPU postprocessing stack. The stacked follow-up #825 adds non-reducing point-field visualization output.

## What changed

- Add grouped libCEED functionals for interface dielectric energy, surface flux, far fields, lumped- and wave-port quantities, and mode overlaps.
- Tabulate mapped face and subface rules once and reuse ordinary fixed libCEED bases for recurring evaluations.
- Support conforming and nonconforming 2D/3D boundaries, including interior interfaces, processor boundaries, ghost values, and ranks with no local contribution.
- Batch disjoint lumped-port power and voltage reductions while retaining per-port fallbacks for unsupported or overlapping attribute sets.
- Preserve existing units, complex conventions, orientation, port normalization, and explicit unsupported-configuration behavior.

## Representative performance

A synthetic surface benchmark on one A100 (8,000 elements, 400 interface elements, order 2) shows the recurring measurement cost after one-time construction:

| Measurement | Existing path | libCEED | One-time setup | Recurring speedup |
|---|---:|---:|---:|---:|
| Interface dielectric energy | 50.0 ms | 0.32 ms | 6.7 ms | 157× |
| Complex port power | 100.1 ms | 1.44 ms | 18.7 ms | 69× |

These are representative microbenchmark timings; the benefit grows when the same functional is evaluated repeatedly.

## Validation

Focused tests compare the device functionals against the existing MFEM/coefficient implementations for serial, MPI, nonconforming, empty-rank, reconstruction, and object-lifetime cases. The final range also passed clean Spack builds and CPU/CUDA test coverage.

## Stack

#825 is based directly on this branch and contains the point-field, ParaView, and GridFunction output layer.


extern/patch/libCEED/libceed_nontensor_atpoints_cuda.diff
```diff
@@ -0,0 +1,3459 @@
...
++        for (CeedInt total = 0; total <= degree; total++) {
++          for (CeedInt i = 0; i <= total; i++) {
++            const CeedInt j = total - i;
++
++            phi[mode++] = CeedBasisPowIntAtPoints(x, i) * CeedBasisPowIntAtPoints(y, j) * CeedBasisPowIntAtPoints(z, k);
++          }
++        }
++      }
++      break;
++    case CEED_TOPOLOGY_PYRAMID:
```
simlapointe, line 1696: I'm not sure the polynomial modal basis used here can represent MFEM’s rational pyramid basis. Using `LinearPyramidFiniteElement`, the exact patched build evaluates a basis function as 0.06 where MFEM gives 0.10; this occurs on CPU, CUDA-ref, and MAGMA. Since this is potentially a silent wrong answer, could we either reject `CEED_TOPOLOGY_PYRAMID` for AtPoints or implement a reconstruction compatible with rational pyramid shape functions?

extern/patch/libCEED/libceed_nontensor_atpoints_cuda.diff
```diff
@@ -0,0 +1,3459 @@
...
++
++  CeedCall(CeedSizeMultiplyAtPoints(ceed, dim, num_elem, &point_length_divisor));
++  CeedCheck(point_length_divisor > 0 && x_length % point_length_divisor == 0, ceed, CEED_ERROR_DIMENSION,
++            "AtPoints coordinate vector is neither compact nor uniformly padded");
++  {
++    const CeedSize points_per_elem = x_length / point_length_divisor;
++    CeedSize       padded_points_size;
++
++    CeedCheck(points_per_elem >= max_num_points, ceed, CEED_ERROR_DIMENSION, "AtPoints coordinate padding is shorter than an element point count");
++    CeedCall(CeedSizeMultiplyAtPoints(ceed, points_per_elem, num_elem, points_stride));
```
simlapointe, line 1443: This accepts uniformly padded point storage, but I'm not sure the existing tensor `AtPoints` implementations use the resulting `points_stride`: it seems the host path indexes by total_num_points and CUDA-ref by max_num_points. A padded-layout reproducer returns success with 17.5, ~0 instead of -1, 1 on CPU and CUDA-ref. Could we either reject padded layouts for tensor bases here or update every tensor implementation to honor this stride before accepting them?

extern/patch/libCEED/libceed_nontensor_atpoints_cuda.diff
```diff
@@ -0,0 +1,3459 @@
...
++  for (CeedInt p = first_degree; p <= last_degree; p++) {
++    const CeedInt modes       = CeedBasisNonTensorNumModesAtPoints(topo, p);
++    CeedScalar   *vandermonde = NULL, *vandermonde_pinv = NULL, *rank_work = NULL, *column_norms = NULL, *candidate = NULL;
++    CeedScalar    max_error = 0.0, tabulation_scale = 1.0;
++    CeedSize      matrix_size, pseudoinverse_square_size, candidate_size;
++    bool          is_valid = true;
++    int           ierr     = CEED_ERROR_SUCCESS;
++
++    if (!modes || modes > num_qpts) break;
++    if (require_overdetermined && num_qpts <= modes) continue;
```
simlapointe, line 1800: Requiring strictly more points than modes rejects valid square, full-rank vector tabulations before the rank check below can evaluate them. On the exact patched build, MFEM ND and RT wedge bases at orders 1 and 2 fail on CPU, CUDA-ref, and MAGMA, while higher orders succeed. Could num_qpts == modes be accepted when the subsequent rank/conditioning test passes, or should prism H(curl)/H(div) support be explicitly excluded from the advertised contract?

simlapointe (commented):
Great improvement! Tested it on a few cases and seems to work well. Identified some possible edge cases below, but they could be false alarms/irrelevant in practice.

palace/fem/facenbrexchange.cpp
```diff
@@ -0,0 +1,571 @@
...
+
+  // Assemble a libCEED point evaluator for each export group, writing the
+  // physical-space field values (space-dimension components per point, point-major)
+  // into the exported vector at the assigned offsets.
+  int max_vsize = 0;
+  for (const auto *fespace : fespaces)
+  {
+    max_vsize = std::max(max_vsize, fespace ? fespace->GetVSize() : 0);
+  }
+  field_staging.SetSize(max_vsize);
```
simlapointe: It looks like `field_staging` is only needed to give the passive field `CeedVector` an initial array during construction, but it remains allocated for the lifetime of this exchange. Since each cached `SurfaceFunctional` that crosses an MPI interface can own a separate `FaceNbrFieldExchange`, this makes memory scale like number of functionals × local solution size rather than with the much smaller exchanged surface data. 
                               
`SurfaceFunctional::Assemble` already handles this by calling `DetachGroupOperatorFieldVectors(groups)` and then f`ield_staging.Destroy()`. Could we do the same with `export_groups` after the loop below finishes? The existing `field_vectors_detached` path should then attach the real source arrays on the first exchange.

palace/fem/interpolator.cpp
```diff
@@ -19,8 +41,191 @@ namespace
...
+    if (mesh.Dimension() != 3 || mesh.SpaceDimension() != 3 ||
+        (map_type != mfem::FiniteElement::H_CURL && map_type != mfem::FiniteElement::H_DIV))
+    {
+      valid = false;
+      return;
+    }
+    const mfem::FiniteElementSpace &mesh_fespace = *mesh.GetNodes()->FESpace();
+    const int npts = op.GetCode().Size();
+    const int rank = Mpi::Rank(comm);
+    field_staging.SetSize(fespace.GetVSize());
```
simlapointe: I think this has the same lifetime issue as the face-neighbor exchange noted above: the full solution-sized staging vector is only construction storage for the passive field vectors, but it remains allocated for as long as the cached probe evaluator. With separate ND and RT probe evaluators this can retain multiple full local FE vectors, including device allocations.                                                     
                                                                                                                                                                                                       
Could we call `DetachGroupOperatorFieldVectors(groups)` and `destroy field_staging` after all point operators have been assembled, following the existing `SurfaceFunctional` pattern?

palace/fem/libceed/basis.cpp
```diff
@@ -38,50 +43,127 @@ void InitTensorBasis(const mfem::FiniteElement &fe, const mfem::IntegrationRule
...
-                                        P, Q, maps.Bt.GetData(), maps.Gt.GetData(),
-                                        qX.GetData(), qW.GetData(), basis));
+    AppendDouble(qW(i));
   }
-  else
+
+  // The cache intentionally has process lifetime, matching registered IntegrationRules
+  // retained for MFEM's pointer-keyed DofToQuad cache. Stored handles keep one reference;
+  // callers receive normal reference-counted copies owned by their operators.
+  static auto *cache = new std::map<BasisKey, CeedBasis>();
```
simlapointe: Could be wrong on this, but it seems this map retains an owning basis reference for every key, and each basis retains its creating Ceed context. As a result, am I right to think that `ceed::Finalize()` only drops Palace's top-level references and does not actually release the contexts or their backend allocations?

If that's the case, could we expose a basis-cache cleanup routine and call it before `CeedDestroy`, while keeping the IntegrationRule registries separate?

palace/fem/output_functionals.hpp
```diff
@@ -0,0 +1,253 @@
...
+  // integrands are quadratic in the field). Collective on the mesh communicator.
+  double Eval(const GridFunction &u) const;
+
+  // Evaluate a surface flux functional for the given fields (either of which may be
+  // nullptr if not required by the flux type). For complex-valued fields, returns the
+  // real and imaginary parts of the flux (ELECTRIC, MAGNETIC), or the stationary real
+  // power (POWER). Collective on the mesh communicator.
+  std::complex<double> EvalFlux(const GridFunction *E, const GridFunction *B) const;
+
+  // Evaluate the complex power P = ∫ E ⋅ (n x H) dS following the conventions of
```
simlapointe: Nit: I think the documented formula has the opposite sign from the implementation and is missing the conjugate. The QFunction and the legacy linear-form path compute P = ∫ (E × H^*) · n dS = -∫ E · (n × H^*) dS.

palace/fem/facenbrexchange.cpp
```diff
@@ -0,0 +1,571 @@
...
+      else
+      {
+        info.apply_qf = f_eval_probe_hdiv_33;
+        info.apply_qf_path = PalaceQFunctionRelativePath(f_eval_probe_hdiv_33_loc);
+      }
+    }
+    CeedOperator op;
+    ceed::AssembleCeedPointEvaluator(info, nullptr, 0, ceed, inputs, value_dim, out_restr,
+                                     &op);
+    export_groups.push_back({ceed, op, std::move(field_sources)});
```
simlapointe, line 505: Nit: I believe that, unlike the surface-functional groups, this point evaluator does not populate `CeedGroupOperator::mesh_nodes` or `mesh_node_fields`, even though it has a passive `grad_x` input backed by the mesh nodes. That means it bypasses the existing per-apply re-pointing/MFEM synchronization path in `ApplyAddGroupOperators`.                                                                                                                                                                                                                

I don't think this causes any issues, but for consistency and safe reuse after host-side coordinate updates, could we set:                  

 ```cpp
export_groups.back().mesh_nodes = pmesh.GetNodes();
export_groups.back().mesh_node_fields = {"grad_x"};
```    
                                                                                                                                                                                                            
and do the analogous thing for `CeedProbeEvaluator` after its `groups.push_back` (palace/fem/interpolator.cpp L197)?

File: palace/fem/output_functionals.cpp [hunk 9]
```diff
@@ -0,0 +1,1914 @@
...
+  {
+    base_ctx.resize(2);
+    base_ctx[0].second = 0.0;
+    base_ctx[1].second = 1.0;
+  }
+
+  return base_ctx;
+}
+
+void SurfaceFunctional::ConfigureGroupQFunction(bool is_2d, bool has_b, double side_scale,
```
simlapointe: Is `side_scale` ever used/varied? It seems all `AddGroup` calls pass 1.0. Could we just remove `FaceGroup::side_scale`?

palace/fem/facenbrexchange.cpp
```diff
@@ -0,0 +1,571 @@
...
+    it = registry.emplace(key, std::move(ir)).first;
+  }
+  else
+  {
+    VerifyRegisteredIr(*it->second, pts);
+  }
+  return it->second.get();
+}
+
+int FieldValueDim(const mfem::ParFiniteElementSpace &fespace, int space_dim)
```
simlapointe: Is this `int FieldValueDim(...)` function ever used?

simlapointe (commented):
This is looking really good! I think having things on the palace side instead of libceed is a big improvement.

A few inline comments below. Some might be non-issues in practice though, so do as you see fit.

palace/fem/output_functionals.cpp
```diff
@@ -0,0 +1,1907 @@
...
+                 plan.elem_b, plan.ghost_b, vol_geom_b, plan.pts_b, plan.point_key_b, 1.0);
+      }
+    }
+  }
+
+  // Initialize the local output vector and field staging vector. Far-field operators
+  // produce 6 values (Re/Im of a 3-vector) per direction per element.
+  const int num_out =
+      (kind == KernelKind::FARFIELD) ? 6 * static_cast<int>(farfield_dirs.size()) : 1;
+  local_out.SetSize(num_marked * num_out);
```
Sbozzolo: This allocates all the target points in one go for all the face elements on the surface of integration.

This can over overflow and lead to OOM on the device. For example, if N = 86400 and order 2, only ~ 5000 faces are needed to overflow `num_out * num_market`.

Even when valid, this can also allocate tens of GBs or more on the device.

palace/fem/output_functionals.cpp
```diff
@@ -0,0 +1,1907 @@
...
+  local_out.UseDevice(true);
+  if (need_field)
+  {
+    const int max_vsize = std::max(nd_fespace ? nd_fespace->GetVSize() : 0,
+                                   rt_fespace ? rt_fespace->GetVSize() : 0);
+    field_staging.SetSize(max_vsize);
```
Sbozzolo: From Fable:

Every passive field `CeedVector` here is created with `length = max(ndV, rtV)` but later re-pointed at buffers of the smaller space, which becomes an out-of-bounds access on backends that
  copy whole vectors.

  The mechanism: `field_staging` is sized `max(nd_fespace->GetVSize(), rt_fespace->GetVSize())`, and `AddFieldInput` creates every field vector from it — `ceed::InitCeedVector(field_staging,
  ceed, &vec)` → `CeedVectorCreate(ceed, v.Size(), cv)` (`libceed/ceed.cpp:101`). A `CeedVector`'s `length` is fixed at creation; the re-point in `ApplyAddGroupOperators`
  (`ceed_group_operator.cpp:153`) uses `init=false`, so it only does `TakeArray`/`SetArray` and never updates `length`.

  For the mixed-space kinds this leaves a permanent mismatch: FARFIELD binds `u_1,u_2` to ND and `u_3,u_4` to RT, and SURFACE_FLUX/POWER alternates ND/RT — i.e. every lumped- and wave-port
  power measurement.

  This is invisible under the default backends because the operator only reads through the element restriction (all offsets `< ndV`), and libCEED's setup check only requires the vector to be
  *at least* as long as the restriction needs. It becomes a real OOB whenever a backend touches `length` entries:

  - `/cpu/self/memcheck/serial`: `memcpy` of `length` entries on every `SetArray` (OOB read), and `TakeArray` → `SyncArray` copies `(rtV−ndV)·8` bytes back past the end of the ND grid
  function (OOB write). `EvalComplexPower` applies 4× per port per frequency, so this fires immediately — and memcheck is exactly the backend one would use to validate this code.
  - `"Device": "CPU"` + `"Backend": "/gpu/cuda/ref"` (a combination this PR explicitly accommodates via the `mem` downgrade at `ceed.cpp:94-97`): `CeedVectorSyncH2D_Cuda` does
  `cudaMemcpy(..., length * sizeof(CeedScalar))` from the borrowed host pointer.

Sbozzolo (commented):
My review was focused on checking performance claims on some realistic cases and verifying whether any big simplification was available.

I found that the PR greatly improves performance in Postprocessing (which was pretty much entirely done on the host before) and no big chunk of it can be removed maintaining comparable level of performance.

I did not look at the details of the code.

Sbozzolo (approved):
My comments were addressed

simlapointe (approved):

PR #825, opened 2026-07-17
Use libCEED point fields for ParaView and GridFunction output

hughcars:
## Summary

Uses libCEED point-field evaluators and device-produced buffers for supported ParaView and GridFunction output. It is stacked directly on #824, which provides the reducing surface-functional substrate.

## What changed

- Add fixed domain and boundary evaluators for `E`, `B`, `V`, `A`, `En`, `Bt`, `Q_s`, `J_s`, `U_e`, `U_m`, `S`, and `Sn` where those fields are available.
- Share immutable face-sampling plans, generation-scoped physical trace caches, and fused derived-field buffers across boundary outputs.
- Support conforming and nonconforming 2D/3D meshes, coarse-face unions, ghost requests, wedge and pyramid faces, and MPI ranks with empty local boundary pieces.
- Write point fields directly from device-produced buffers, including raw-appended VTU payloads, while retaining field names, SI units, component counts, point order, and multi-excitation directory layout.
- Preserve native GridFunction spaces for primary solution fields. Derived `U_e`, `U_m`, `S`, and `Sn` fields use discontinuous interpolatory L2 spaces so material-interface jumps remain element-local.
- Preserve boundary semantics: primary traces are averaged across interfaces, energy/Poynting quantities average their per-side values, and surface charge/current use oriented jumps.
- Write final `Rank` and `Indicator` arrays as cell data and emit surface-current `J_s` fields.
- Delay visualization setup until a solution exists and stage converged eigenvectors before releasing solver-owned storage, avoiding overlap between large eigensolver and visualization allocations.
- Report every converged eigenmode while limiting error-estimator accumulation to the requested `N` modes and field output to `Save` modes, bounded by `N`.

## Performance tradeoff

The libCEED evaluator path has a fixed setup and JIT cost per mesh generation on both CPU and GPU backends. That setup is reused across subsequent saved modes, frequencies, and timesteps. Repeated or large field-output workloads amortize the cost, while very small or infrequent-output jobs may be neutral or slower. The solver algorithm is unchanged; this PR targets postprocessing and visualization output.

## Representative performance

Public Palace examples were run with eight MPI ranks on eight A100 GPUs using `/gpu/cuda/magma` and an identical non-Palace Spack dependency DAG:

| Example | `main` | Final stack | Change |
|---|---:|---:|---:|
| `antenna_short_dipole` | 41.10 s | 40.95 s | -0.4% |
| `cpw_lumped_uniform` | 85.08 s | 75.95 s | -10.7% |
| `cavity2d_driven` | 19.37 s | 22.04 s | +13.8% |
| `transmon_amr` | 1283.65 s | 1011.35 s | -21.2% |

The output-heavy cases show where the time is recovered:

| Example | Output stage | `main` | Final stack | Change |
|---|---|---:|---:|---:|
| CPW | ParaView | 13.26 s | 6.78 s | -48.9% |
| CPW | GridFunction | 8.49 s | 0.88 s | -89.7% |
| Transmon AMR | ParaView | 202.36 s | 73.58 s | -63.6% |
| Transmon AMR | GridFunction | 170.00 s | 8.06 s | -95.3% |

The small cavity demonstrates the expected setup-cost crossover; repeated, output-heavy workloads amortize setup and benefit substantially.

## Compatibility

Raw-appended VTK XML output is supported with ParaView 5.12.1 or newer. Primary and derived field-space conventions are documented in the postprocessing guide.

## Validation

Tests decode the emitted VTU payloads and compare domain and boundary fields against the established coefficient paths across serial, MPI, CUDA, 2D/3D, nonconforming, repeated-save, empty-piece, and object-lifetime cases. The final stack also passed clean Spack builds and paired eight-GPU regression runs.

### DIFF
--- a/palace/drivers/eigensolver.cpp
+++ b/palace/drivers/eigensolver.cpp
@@ -40,6 +40,7 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
   SpaceOperator space_op(iodata, mesh);
   auto K = space_op.GetStiffnessMatrix<ComplexOperator>(Operator::DIAG_ONE);
   auto C = space_op.GetDampingMatrix<ComplexOperator>(Operator::DIAG_ZERO);
+  const bool has_C = (C != nullptr);
   auto M = space_op.GetMassMatrix<ComplexOperator>(Operator::DIAG_ZERO);
 
   // Check if there are nonlinear terms and, if so, setup interpolation operator.
@@ -127,8 +128,7 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
 
   // Configure objects for postprocessing.
   PostOperator<ProblemType::EIGENMODE> post_op(iodata, space_op);
-  ComplexVector E(Curl.Width()), B(Curl.Height());
-  E.UseDevice(true);
+  ComplexVector B(Curl.Height());
   B.UseDevice(true);
 
   // Define and configure the eigensolver to solve the eigenvalue problem:
@@ -447,15 +447,54 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
     eigen->SetBMat(*KM);
     eigen->RescaleEigenvectors(num_conv);
   }
+  std::vector<std::complex<double>> eigenvalues(num_conv);
+  std::vector<double> errors_bkwd(num_conv), errors_abs(num_conv);
+  // Stage finalized eigenvectors in host memory so extracting all converged modes does
+  // not overlap the eigensolver basis with num_conv additional full device vectors. One
+  // device work vector is reused for extraction and later postprocessing.
+  std::vector<std::vector<std::complex<double>>> eigenvectors(
+      num_conv, std::vector<std::complex<double>>(Curl.Width()));
+  ComplexVector E(Curl.Width());
+  E.UseDevice(true);
+  for (int i = 0; i < num_conv; i++)
+  {
+    eigenvalues[i] = eigen->GetEigenvalue(i);
+    errors_bkwd[i] = eigen->GetError(i, EigenvalueSolver::ErrorType::BACKWARD);
+    errors_abs[i] = eigen->GetError(i, EigenvalueSolver::ErrorType::ABSOLUTE);
+    eigen->GetEigenvector(i, E);
+    E.Get(eigenvectors[i].data(), E.Size(), false);
+  }
+
+  // The shifted linear system, eigenvalue solver, and assembled matrices are only needed
+  // through eigenvector extraction. Release their device allocations before error
+  // estimation and field-output postprocessing, where large libCEED visualization
+  // operators may need additional GPU memory on refined meshes.
+  eigen.reset();
+  KM.reset();
+  ksp.reset();
+  P.reset();
+  A.reset();
+  divfree.reset();
+  Kp.reset();
+  Cp.reset();
+  Mp.reset();
+  A2_0.reset();
+  A2_1.reset();
+  A2_2.reset();
+  A2.reset();
+  interp_op.reset();
+  K.reset();
+  C.reset();
+  M.reset();
   Mpi::Print("\n");
 
   for (int i = 0; i < num_conv; i++)
   {
     // Get the eigenvalue and relative error.
-    std::complex<double> omega = eigen->GetEigenvalue(i);
-    double error_bkwd = eigen->GetError(i, EigenvalueSolver::ErrorType::BACKWARD);
-    double error_abs = eigen->GetError(i, EigenvalueSolver::ErrorType::ABSOLUTE);
-    if (!C && !has_A2)
+    std::complex<double> omega = eigenvalues[i];
+    double error_bkwd = errors_bkwd[i];
+    double error_abs = errors_abs[i];
+    if (!has_C && !has_A2)
     {
       // Linear EVP has eigenvalue μ = -λ² = ω².
       omega = std::sqrt(omega);
@@ -468,8 +507,7 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
 
     // Compute B = -1/(iω) ∇ x E on the true dofs, and set the internal GridFunctions in
     // PostOperator for all postprocessing operations.
-    eigen->GetEigenvector(i, E);
-
+    E.Set(eigenvectors[i].data(), E.Size(), false);
     linalg::NormalizePhase(space_op.GetComm(), E);
 
     Curl.Mult(E.Real(), B.Real());
--- a/palace/drivers/eigensolver.cpp
+++ b/palace/drivers/eigensolver.cpp
@@ -439,56 +439,66 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
 
   // Calculate and record the error indicators, and postprocess the results.
   Mpi::Print("\nComputing solution error estimates and performing postprocessing\n");
+  const int num_post = iodata.solver.eigenmode.n;
+  MFEM_VERIFY(num_conv >= num_post, "Eigenmode solve only found "
+                                        << num_conv << " modes when " << num_post
+                                        << " were requested!");
   if (!KM)
   {
-    // Normalize the finalized eigenvectors with respect to mass matrix (unit electric field
+    // Normalize the requested eigenvectors with respect to mass matrix (unit electric field
     // energy) even if they are not computed to be orthogonal with respect to it.
     KM = space_op.GetInnerProductMatrix(0.0, 1.0, nullptr, M.get());
     eigen->SetBMat(*KM);
-    eigen->RescaleEigenvectors(num_conv);
+    eigen->RescaleEigenvectors(num_post);
   }
-  std::vector<std::complex<double>> eigenvalues(num_conv);
-  std::vector<double> errors_bkwd(num_conv), errors_abs(num_conv);
-  // Stage finalized eigenvectors in host memory so extracting all converged modes does
-  // not overlap the eigensolver basis with num_conv additional full device vectors. One
-  // device work vector is reused for extraction and later postprocessing.
-  std::vector<std::vector<std::complex<double>>> eigenvectors(
-      num_conv, std::vector<std::complex<double>>(Curl.Width()));
+  std::vector<std::complex<double>> eigenvalues(num_post);
+  std::vector<double> errors_bkwd(num_post), errors_abs(num_post);
+  const bool stage_eigenvectors = post_op.WillWriteFields();
+  // Field output can construct large libCEED visualization operators. Only in that case,
+  // stage the requested finalized eigenvectors in host memory so the eigensolver and
+  // matrices can release their device allocations before postprocessing.
+  std::vector<std::vector<std::complex<double>>> eigenvectors;
+  if (stage_eigenvectors)
+  {
+    eigenvectors.assign(num_post, std::vector<std::complex<double>>(Curl.Width()));
+  }
   ComplexVector E(Curl.Width());
   E.UseDevice(true);
-  for (int i = 0; i < num_conv; i++)
+  for (int i = 0; i < num_post; i++)
   {
     eigenvalues[i] = eigen->GetEigenvalue(i);
     errors_bkwd[i] = eigen->GetError(i, EigenvalueSolver::ErrorType::BACKWARD);
     errors_abs[i] = eigen->GetError(i, EigenvalueSolver::ErrorType::ABSOLUTE);
-    eigen->GetEigenvector(i, E);
-    E.Get(eigenvectors[i].data(), E.Size(), false);
+    if (stage_eigenvectors)
+    {
+      eigen->GetEigenvector(i, E);
+      E.Get(eigenvectors[i].data(), E.Size(), false);
+    }
   }
 
-  // The shifted linear system, eigenvalue solver, and assembled matrices are only needed
-  // through eigenvector extraction. Release their device allocations before error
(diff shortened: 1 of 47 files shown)


palace/fem/domain_point_field_evaluator.cpp
```diff
@@ -0,0 +1,496 @@
...
+      ceed::AssembleCeedPointEvaluator(info, ctx.empty() ? nullptr : ctx.data(),
+                                       ctx.size() * sizeof(CeedIntScalar), ceed, inputs,
+                                       target_fespace.GetVDim(), out_restr, &op);
+      groups.push_back({ceed, op, std::move(field_sources)});
+      groups.back().mesh_nodes = pmesh.GetNodes();
+      groups.back().mesh_node_fields = {"grad_x"};
+      fem::CacheGroupOperatorFieldVectors(groups.back());
+    }
+
+    // Assemble a second operator for the direct VTU point-buffer path. This evaluates at
```
simlapointe: Should the entire GridFunction and VTU setup blocks be guarded by `build_gridfunction`  and `build_buffer` respectively? Currently only the final operator creation is conditional, so ParaView-only and GridFunction-only configurations still construct unused bases and restrictions.

palace/models/lumpedportoperator.cpp
```diff
@@ -269,10 +269,12 @@ std::complex<double> LumpedPortData::GetPower(GridFunction &E, GridFunction &B)
   // Compute port power, (E x H) ⋅ n = E ⋅ (-n x H), integrated over the port surface using
   // the computed E and H = μ⁻¹ B fields, where +n is the direction of propagation (into the
   // domain). The BdrSurfaceCurrentVectorCoefficient computes -n x H for an outward normal,
-  // so we multiply by -1.
+  // so we multiply by -1. The linear form is reconstructed from scratch each time due to
+  // changing H.
   MFEM_VERIFY((E.HasImag() && B.HasImag()) || (!E.HasImag() && !B.HasImag()),
               "Mismatch between real- and complex-valued E and B fields in port power "
               "calculation!");
+  const bool has_imag = E.HasImag();
```
simlapointe: Is this used anywhere?

palace/fem/output_functionals.cpp
```diff
@@ -210,18 +181,9 @@ std::vector<long long> MakeTraceMapTopologyKey(const mfem::Mesh::FaceInformation
...
-  {
-    mfem::IntegrationPoint fip = mfem::Mesh::TransformBdrElementToFace(
-        face_geom, bdr_to_face_orientation, verts->IntPoint(i));
-    mfem::IntegrationPoint ip;
-    loc.Transform(fip, ip);
-    key.push_back(EncodeReferenceCoordinate(ip.x));
-    key.push_back(EncodeReferenceCoordinate(ip.y));
-    key.push_back(EncodeReferenceCoordinate(ip.z));
-  }
+  key.push_back(static_cast<long long>(face.ncface));
```
simlapointe, line 184: Not actually sure if this is an issue in practice, but it seems `face.ncface` is only an index into the requesting mesh/rank’s local `nc_faces_info`; it is not globally unique. This key is sent to another rank, where requests from different neighbors are grouped together. For non-AtPoints groups, `FaceNbrFieldExchange` retains only the first request’s points, so two neighbors with the same local ncface index but different point matrices can be silently evaluated at the wrong reference points. The process-lifetime rule registry can similarly collide across rebuilt meshes. Could we retain the exact point-map identity, or include sender/exchange identity and verify every grouped point set?

palace/models/postoperator.cpp
```diff
@@ -741,30 +1117,28 @@ void PostOperator<solver_t>::WriteParaviewFieldsFinal(const ErrorIndicator *indi
...
+    paraview->RegisterDomainCellField("Indicator", indicator->Local());
   }
-  paraview->Save();
-  if (rank)
+  for (const char *name :
+       {"E", "E_real", "E_imag", "B", "B_real", "B_imag", "En", "En_real", "En_imag",
+        "Bt_real", "Bt_imag", "V", "A", "U_e", "U_m", "S", "Sn"})
   {
-    paraview->DeregisterField("Rank");
+    paraview->DeregisterDomainPointField(name);
```
simlapointe, line 1130: These domain point fields are permanently deregistered, unlike the legacy field/coefficient maps restored below. The eigenmode driver calls `MeasureFinalize()` after requested mode N, but can continue processing `num_conv > N;` configuration also does not constrain `Save <= N`. A later saved mode can therefore overwrite the metadata cycle without any domain fields? Has this been tested on cases where num_conv > N? Should/Could finalization move after the loop, or could these registrations be restored?

palace/models/postoperator.cpp
```diff
@@ -331,24 +585,33 @@ void PostOperator<solver_t>::SetupFieldCoefficients()
...
+        MakeBdrFieldEvaluator(PointFieldEvaluator::Kind::FIELD_B, *B->ParFESpace(),
+                              B_bdr_eval);
+      }
+      // J_s = n x H = n x μ⁻¹ B.
+      if (use_ceed_boundary_fields)
       {
-        B_si = std::make_unique<BdrFieldVectorCoefficient>(B->Imag());
-        J_si = std::make_unique<BdrSurfaceCurrentVectorCoefficient>(
-            B->Imag(), fem_op->GetMaterialOp(), scaling);
+        MakeBdrCoeffEvaluator(PointFieldEvaluator::Kind::CURRENT_J, *B->ParFESpace(),
```
simlapointe: In 2D, B is scalar, but J_s = n × (H_z ẑ) is still a valid two-component tangential field. This PR adds the 2D f_eval_bdr_current_j_*_21 kernels, but construction is inside VectorDim() > 1, making those kernels unreachable from PostOperator. Should only boundary B be gated here while J_bdr_eval remains enabled in 2D?

palace/drivers/eigensolver.cpp
```diff
@@ -447,15 +447,54 @@ EigenSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
     eigen->SetBMat(*KM);
     eigen->RescaleEigenvectors(num_conv);
   }
+  std::vector<std::complex<double>> eigenvalues(num_conv);
+  std::vector<double> errors_bkwd(num_conv), errors_abs(num_conv);
+  // Stage finalized eigenvectors in host memory so extracting all converged modes does
+  // not overlap the eigensolver basis with num_conv additional full device vectors. One
+  // device work vector is reused for extraction and later postprocessing.
+  std::vector<std::vector<std::complex<double>>> eigenvectors(
```
simlapointe: Nit: This stages 16 × num_conv × local_DOFs bytes per rank in memory, even when field output is disabled. Could this ever be an issue? Should this be conditional/gated?

simlapointe (commented):
Looks really good and works as expected in my tests. Some comments below including potential correctness risks, but unsure if they truly matter in practice so you can do as you see fit.

palace/fem/domain_point_field_evaluator.cpp
hughcars: Addressed in 0655103d9

palace/models/lumpedportoperator.cpp
hughcars: Addressed in 0655103d9

palace/fem/output_functionals.cpp
hughcars: Addressed in 0655103d9

palace/models/postoperator.cpp
hughcars: Addressed in 3a2767968

palace/models/postoperator.cpp
hughcars: Addressed in 0655103d9

palace/drivers/eigensolver.cpp
hughcars: Addressed in 0655103d9

test/unit/test-postoperator-boundary-viz.cpp
```diff
@@ -0,0 +1,1266 @@
...
+{
+  long long count = 0;
+  double max_abs = 0.0;
+  double max_scaled = 0.0;
+  double sum_sq = 0.0;
+
+  void Add(double val, double ref, double rtol, double atol)
+  {
+    const double err = std::abs(val - ref);
+    const double denom = atol + rtol * std::max({1.0, std::abs(val), std::abs(ref)});
```
Sbozzolo: The `1.0` floor turns this into a fixed absolute tolerance of 5.05e-5 for any field whose SI magnitude is below 1.

palace/utils/ceedparaviewdatacollection.cpp
```diff
@@ -0,0 +1,957 @@
...
+  if (domain_cell_fields.empty())
+  {
+    if (cached_mesh_vtu.empty() || cached_mesh_vtu_ref != ref ||
+        cached_mesh_vtu_bdr_output != bdr_output)
+    {
+      std::ostringstream mesh_os;
+      mesh_os.precision(os.precision());
+      mesh->PrintVTU(mesh_os, ref, pv_data_format, high_order_output, GetCompressionLevel(),
+                     bdr_output);
+      cached_mesh_vtu = mesh_os.str();
```
Sbozzolo: `mesh_os.str()` is a copy, so the peak here is ~2x the string (the `ostringstream` buffer plus the copy), and `cached_mesh_vtu` then stays resident for the collection's lifetime. `main` streamed `PrintVTU` straight into the `ofstream` with no retention, and two collections are live (`paraview` and `paraview_bdr`).

The size is roughly `1.334 * (16*np + 9*ne)` bytes per rank, with `np = sum(RefG->RefPts.GetNPoints())`: ~45 MB at 200k local elements / order 2, and ~440 MB steady / ~880 MB transient at 1M local elements / order-3 tets, for the domain collection alone. Caching buys real time across repeated saves since `PrintVTU` plus base64 is O(np) per save, so the trade is defensible — but it is unbounded, not configurable, and lands on exactly the large-mesh runs this stack targets. `std::move(mesh_os).str()` would at least remove the transient doubling.

Worth noting the asymmetry with the `domain_cell_fields` branch below: that path regenerates *and* fully materialises the mesh VTU on every call and never caches it, so the final-step save pays both the CPU and the memory cost.

palace/fem/output_functionals.cpp
```diff
@@ -1371,21 +1752,36 @@ void SurfaceFunctional::AssembleLocal(const Mesh &mesh,
...
-      AddVolGeomInputs("1", group.vol_indices_a, group.vol_geom_a, *group.mapped_ir_a);
-    }
-    if (group.ghost_b)
+    if (trace_cache)
     {
-      AddVolGeomInputsGhost("2", *group.mapped_ir_b);
+      // Trace values have already received their ND/RT Piola map. Feed an identity
+      // volume Jacobian to the legacy derived-output QFunctions so they consume those
+      // physical values directly; face geometry remains the real boundary Jacobian.
+      AddVolGeomInputsPhysical("1", group.vol_indices_a, *group.mapped_ir_a, group.ghost_a);
```
Sbozzolo, line 1818: This is the line that makes `AddVolGeomInputsPhysical`'s identity-Jacobian allocation (`:1723`) apply to all *local* trace-consumer groups — at base that array existed only for ghost (processor-boundary) groups. It is a *constant* identity matrix stored per element per lattice point.

In 3D `geom_comp = 9` and `num_pts` is 6/10/16 for an order-2 tri / order-3 tri / order-3 quad, so 432–1152 B per boundary element per evaluator. An electrostatic run has four trace consumers (`V`, `E`, `Q_s`, `U_e` — `bdr_derived_bundle` is never created without both E and B, `postoperator.cpp:499`), i.e. ~1.7–4.6 KB per boundary element, or roughly 0.9 GB of pure identity matrices on a 200k-boundary-face mesh at `paraview_refine_order = 3`. Magnetostatic (`A`, `B`, `J_s`, `U_m`) is the same. For the two `field_value_kind` consumers `needs_elem_attr` and `needs_face_geom` are both false (`:1491-1494`), so this array is the operator's only per-element input and is entirely redundant.

An indexed `CeedElemRestrictionCreate` whose offsets are identical for every element, with `l_size = geom_comp * num_pts`, would let one element's worth of identity data be shared.

palace/fem/face_sampling_plan.cpp
```diff
@@ -0,0 +1,498 @@
...
+      for (int d = 0; d < J.Width(); d++)
+      {
+        for (int c = 0; c < J.Height(); c++)
+        {
+          entry.face_jacobians[(c + pmesh.SpaceDimension() * d) * nq + q] = J(c, d);
+        }
+      }
+    }
+
+    if (buffer_bases.empty())
```
Sbozzolo: Allocating lazily on the first *marked* boundary element means that on a rank with `GetNBE() > 0` where no boundary element matches the marker, `buffer_bases` stays empty while `GetNBE() > 0`. That breaks the invariant asserted at `ceedparaviewdatacollection.cpp:462` (`field.bases->size() == NumPointFieldEntities(location)`), so such a rank would abort there rather than produce wrong data.

Unreachable today — every production caller passes an all-ones marker (`postoperator.cpp:425-433, 439-447, 460-465, 477-484`), so any rank with `GetNBE() > 0` allocates, and `GetNBE() == 0` is fine since `0 == 0`. Still, `buffer_bases.assign(pmesh.GetNBE(), -1)` before the loop makes it unconditional and costs nothing.

Sbozzolo (commented):
I tried to break down this PRs in self-contained chunks, hoping that most of the performance benefits would come from a small improvement. In particular, I was hoping that we could get away with not having to implement our own ParaView writer. 
Instead, I found that all the components of this PR are needed for good performance improvements on realistic cases. 

With this PR, both Postprocessing and Paraview are now subdominant in the execution of realistic simulations.

Sbozzolo (approved):
My comments were addressed

simlapointe (approved):

PR #826, opened 2026-07-17
Build the cuDSS communication layer for the selected MPI

hughcars:
Follow-up to #717.

- Add a `CUDSS_COMM_LIB` superbuild option and forward its absolute path to MFEM as `MFEM_CUDSS_COMM_LIB`.
- Use NVIDIA's bundled Open MPI communication layer by default only when the selected MPI is positively identified as Open MPI; any other MPI must supply a matching layer via `CUDSS_COMM_LIB`.
- Spack: depend on `cudss+mpi`, which builds the communication layer against the concrete MPI provider. The `cudss+mpi` variant and MFEM's consumption of `libcudss_commlayer_mpi` are now upstream in spack-packages, so the recipe patches this PR originally carried have been dropped.

Validation (pre-rebase): configure probe for Open MPI auto-selection, MPICH missing-path rejection, and MPICH explicit-path selection; MFEM and Palace built through Spack with MPICH + CUDA + cuDSS; two-rank GPU solve with MPICH and `CUDSS_COMM_LIB` unset at runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,8 @@ set(PALACE_WITH_SUPERLU ON CACHE BOOL "Build with SuperLU_DIST sparse direct sol
 set(PALACE_WITH_STRUMPACK OFF CACHE BOOL "Build with STRUMPACK sparse direct solver")
 set(PALACE_WITH_MUMPS OFF CACHE BOOL "Build with MUMPS sparse direct solver")
 set(PALACE_WITH_CUDSS OFF CACHE BOOL "Build with cuDSS sparse direct solver")
+set(CUDSS_COMM_LIB "" CACHE FILEPATH
+  "Full path to a cuDSS communication layer built for the selected MPI")
 set(PALACE_WITH_SLEPC ON CACHE BOOL "Build with SLEPc eigenvalue solver")
 set(PALACE_WITH_ARPACK OFF CACHE BOOL "Build with ARPACK eigenvalue solver")
 set(PALACE_WITH_LIBXSMM ON CACHE BOOL "Build with LIBXSMM backend for libCEED")
--- a/cmake/ExternalMFEM.cmake
+++ b/cmake/ExternalMFEM.cmake
@@ -128,6 +128,49 @@ if(PALACE_WITH_CUDA)
       "-DCUDSS_DIR=${CUDSS_DIR}"
     )
   endif()
+  if(PALACE_WITH_CUDSS)
+    if("${CUDSS_COMM_LIB}" STREQUAL "")
+      # NVIDIA ships an Open MPI communication layer with cuDSS. Use it only after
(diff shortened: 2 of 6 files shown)


Sbozzolo (commented):
Why are we treating Open MPI differently from everything else? We could just unconditionally compile against whatever MPI is provided and never use the version shipped with cuDSS. We already have MPI in the stack, so we don't need to add a new CUDSS_COMM_LIB variable. 

```cmake
if(PALACE_WITH_CUDSS)
    # cuDSS loads its MPI calls from a small plugin that must be compiled against the
    # same MPI as Palace. Build it from the source NVIDIA ships with cuDSS.
    set(PALACE_CUDSS_COMM_SRC "${CUDSS_DIR}/src/cudss_commlayer_openmpi.cu")
    if(NOT EXISTS "${PALACE_CUDSS_COMM_SRC}")
      message(FATAL_ERROR
        "Could not find ${PALACE_CUDSS_COMM_SRC}. The cuDSS tarball archives include the "
        "communication layer source under src/; the pip/conda wheels do not. Install the "
        "archive and set CUDSS_DIR to it."
      )
    endif()

    find_package(cudss CONFIG REQUIRED HINTS "${CUDSS_DIR}/lib/cmake/cudss")
    find_package(CUDAToolkit REQUIRED)

    add_library(cudss_commlayer_mpi SHARED "${PALACE_CUDSS_COMM_SRC}")
    target_link_libraries(cudss_commlayer_mpi PRIVATE cudss CUDA::cudart MPI::MPI_CXX)
    set_target_properties(cudss_commlayer_mpi PROPERTIES
      CUDA_ARCHITECTURES "${CMAKE_CUDA_ARCHITECTURES}"
      LIBRARY_OUTPUT_DIRECTORY "${CMAKE_INSTALL_PREFIX}/lib"
      BUILD_WITH_INSTALL_RPATH ON
      INSTALL_RPATH_USE_LINK_PATH ON
    )
    set(PALACE_CUDSS_COMM_LIB
      "${CMAKE_INSTALL_PREFIX}/lib/${CMAKE_SHARED_LIBRARY_PREFIX}cudss_commlayer_mpi${CMAKE_SHARED_LIBRARY_SUFFIX}"
    )
    message(STATUS "Building cuDSS communication layer for the selected MPI: ${PALACE_CUDSS_COMM_LIB}")
    list(APPEND MFEM_DEPENDENCIES cudss_commlayer_mpi)
    list(APPEND MFEM_OPTIONS "-DMFEM_CUDSS_COMM_LIB=${PALACE_CUDSS_COMM_LIB}")
  endif()
```

hughcars:
Agreed, one path is better. Took your block as-is in 463780f6, dropped `CUDSS_COMM_LIB` and the Open MPI probe.

Sbozzolo (approved):

PR #828, opened 2026-07-18
Add macOS Spack build/test workflow

### LINKED ISSUE #766: Add Spack container build for macOS
Currently, we only build containers for neoverse-v1 and sapphirerapids. It would be useful to add a spack container build for macOS as well. This would also double as build test for spack on macOS.

Sbozzolo:
Adds `.github/workflows/spack-macos.yml`: builds Palace on macOS through Spack and runs its unit tests. This is the Spack analogue of `build-and-test-macos.yml` (which uses the Homebrew + raw CMake path) and doubles as a build test for the Spack recipe on macOS.

Closes #766


Sbozzolo:
Needs: https://github.com/spack/spack-packages/pull/5661

spack_repo/patches/pr5661_libxsmm.diff
```diff
@@ -0,0 +1,34 @@
...
+-    depends_on("binutils+ld+gas@2.33:", type="build")
++    # Only needed to build libxsmm < 1.17 (per the note above); scope it so
++    # newer versions build with the system assembler. On macOS binutils+gas+ld
++    # is unbuildable (recipe fails against the current SDK) and unavailable from
++    # Homebrew (no gas/ld), so an unconditional dep breaks builds that do not
++    # need it.
++    depends_on("binutils+ld+gas@2.33:", type="build", when="@:1.16")
+ 
+     # Version 2.0 supports both x86_64 and aarch64
+     requires("target=x86_64:", "target=aarch64:")
```
bernhardkaindl, line 18: Hi @Sbozzolo, I merged this fix into the live [repos/spack_repo/builtin/packages/libxsmm/package.py](https://github.com/spack/spack-packages/pull/5661/changes#diff-66e497099672c8b3abbc4185263c807ea16ebd9c6d170cd3ee75651af997a0fa) 15 hours ago.

The part which isn't merged in repos/spack_repo/builtin/packages/libxsmm/package.py is the explanation why the fix was especially needed for MacOS:

```py
+    # Only needed to build libxsmm < 1.17 (per the note above); scope it so
+    # newer versions build with the system assembler. On macOS binutils+gas+ld
+    # is unbuildable (recipe fails against the current SDK) and unavailable from
+    # Homebrew (no gas/ld), so an unconditional dep breaks builds that do not
+    # need it.
```

However, I think we'd normally want this explanation in the actual spack recipe, not hide in in the patch, which would fail to apply as soon the recipe changes in at those lines:

```suggestion
```

Instead, I think the explanation why the fix was needed especially for MacOS should be applied directly as a change repos/spack_repo/builtin/packages/libxsmm/package.py, not by adding a patch as a file.

spack_repo/patches/pr5661_libxsmm.diff
```diff
@@ -0,0 +1,34 @@
...
++        # after install and fail to load. Rewrite each id to its final path.
++        if spec.satisfies("platform=darwin +shared"):
++            from glob import glob as _glob
++            for lib in _glob(join_path(prefix.lib, "libxsmm*.dylib")):
++                if not os.path.islink(lib):
++                    which("install_name_tool")("-id", lib, lib)
++
+         if spec.satisfies("+header-only"):
+             install_tree("src", prefix.src)
+
```
bernhardkaindl, line 34: Hm, I've to admit, I don't understand why to apply this change only using a patch file when running CI.

If it is a fix for building `libxsmm` on darwin, I assume these changes should be applied as a direct commit to update the actual recipe itself and not patch the recipe in CI.

```suggestion
```

bernhardkaindl (commented):
@Sbozzolo, please see my suggestions. AFAICS, I think those changes should be applied to [repos/spack_repo/builtin/packages/libxsmm/package.py](https://github.com/spack/spack-packages/pull/5661/changes#diff-66e497099672c8b3abbc4185263c807ea16ebd9c6d170cd3ee75651af997a0fa) as a direct commit of changes to it, not as a patch on top of it, I assume.

Sbozzolo:
> @Sbozzolo, please see my suggestions. AFAICS, I think those changes should be applied to [repos/spack_repo/builtin/packages/libxsmm/package.py](https://github.com/spack/spack-packages/pull/5661/changes#diff-66e497099672c8b3abbc4185263c807ea16ebd9c6d170cd3ee75651af997a0fa) as a direct commit of changes to it, not as a patch on top of it, I assume.

Thank you very much for your comments!

I was experimenting (and still am) to figure out the best way to build with Spack. I'll contribute back once I figure out the cleanest solution.

spack_repo/patches/pr5661_libxsmm.diff
Sbozzolo: Upstreamed here: https://github.com/spack/spack-packages/pull/4000

docs/src/developer/spack.md
```diff
@@ -105,6 +104,81 @@ to the `develop` section of your `spack.yaml` pointing to your local MFEM
...
+
+```@eval
+import Markdown
+path = joinpath("spack", "setup-macos.sh")
+Markdown.parse("```bash\n$(read(path, String))\n```")
+```
+
+```sh
+~/repos/palace/docs/src/developer/spack/setup-macos.sh
+spack -e . develop --path=~/repos/palace palace@develop
```
hughcars: The documented path needs the same package preparation as CI before concretization. This environment enables `+libxsmm`, so the current macOS build needs `pr5661_libxsmm.diff`; the local Palace recipe also needs the MFEM symlink workaround. Without these steps the documented setup can fail while CI passes.

```suggestion
PALACE_DIR=~/repos/palace
"$PALACE_DIR"/docs/src/developer/spack/setup-macos.sh

SPACK_BUILTIN="$(spack location --repo builtin)"
for patch in "$PALACE_DIR"/spack_repo/patches/*.diff; do
  if [ -f "$patch" ]; then
    if ! git -C "$SPACK_BUILTIN" apply --check "$patch" 2>/dev/null; then
      echo "Cannot apply $patch (PR may have already been merged)"
      echo "Please remove $patch from the Palace repo if the PR was merged"
    else
      git -C "$SPACK_BUILTIN" apply "$patch"
    fi
  fi
done
ln -sf "$SPACK_BUILTIN"/packages/mfem \
  "$PALACE_DIR"/spack_repo/local/packages/

spack -e . repo add --scope="env:$(pwd)" "$PALACE_DIR"/spack_repo/local
spack -e . develop --path="$PALACE_DIR" local.palace@develop
```

hughcars: The developer path only needs the libxsmm package fix here. With the checked-in `+libxsmm` spec and Spack's `releases/v2026.06` package repository, the builtin recipe still selects `binutils@2.33.1+gas+ld` on macOS and does not repair shared-library install IDs. CI avoids both with `pr5661_libxsmm.diff`; please apply that patch before concretization while leaving this path on `builtin.palace`.

```suggestion
PALACE_DIR=~/repos/palace
"$PALACE_DIR"/docs/src/developer/spack/setup-macos.sh

SPACK_BUILTIN="$(spack location --repo builtin)"
LIBXSMM_PATCH="${PALACE_DIR}/spack_repo/patches/pr5661_libxsmm.diff"
if git -C "$SPACK_BUILTIN" apply --check "$LIBXSMM_PATCH" 2>/dev/null; then
  git -C "$SPACK_BUILTIN" apply "$LIBXSMM_PATCH"
elif ! git -C "$SPACK_BUILTIN" apply --reverse --check "$LIBXSMM_PATCH" 2>/dev/null; then
  echo "Error: cannot apply $LIBXSMM_PATCH" >&2
  exit 1
fi

spack -e . develop --path="$PALACE_DIR" palace@develop
```

hughcars, line 174: Nit: Is the implicit `builtin.palace` intentional here? These instructions resolve the builtin recipe, while CI tests `local.palace`, so CI does not validate the documented setup.

hughcars (commented):
One tiny documentation nit, but lgtm.

docs/src/developer/spack.md
Sbozzolo: I added instructions on using `local.palace`. It was semi-intentional (in trying to keep things simple).

PR #830, opened 2026-07-20
Build full microarchitecture container matrix on release tags

Sbozzolo:
On release tags, build a matrix of microarchitecture-specific containers instead of only the runner's native arch:
  - x86_64_v3, x86_64_v4, sapphirerapids   
  - aarch64, neoverse_v1, neoverse_v2       

Passing here: https://github.com/awslabs/palace/actions/runs/29653166956

### DIFF
--- a/.github/actions/build-container/action.yml
+++ b/.github/actions/build-container/action.yml
@@ -1,6 +1,16 @@
 name: Build Palace Containers
 description: Build a Palace container image from the stored Spack environment
 
+inputs:
+  target:
+    description: >-
+      Spack microarchitecture to build for (e.g. x86_64_v3, neoverse_v2).
+      Empty means build for the runner's native microarchitecture
+      (spack arch -t). Must be at or below the runner's native ISA so
+      build-time binaries remain executable on the host.
+    required: false
+    default: ''
+
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


.github/actions/build-container/action.yml
```diff
@@ -89,6 +104,29 @@ runs:
...
+      # system curl), which are fixed at the runner's own microarch. Forcing them
+      # to a different target is unsatisfiable ("Cannot select a single
+      # node_target for package glibc/gcc").
+      shell: bash
+      env:
+        TARGET: ${{ inputs.target }}
+      run: |
+        if [ -n "$TARGET" ]; then
+          cd .github/actions/build-container/spack_env
+          spack -e . config add "packages:all:target:[${TARGET}]"
```
hughcars, line 123: `packages:all:target` is only a preference, and the cited validation run shows Spack ignored it for the `neoverse_v1` job. Despite requesting `neoverse_v1`, the concrete root spec was:

```text
- uzuoeio palace@develop+arpack~asan~coverage~cuda~cudss+gslib+int64~ipo+libxsmm+mumps~openmp~rocm+shared+slepc+strumpack+sundials+superlu-dist build_system=cmake build_type=Release cxxstd=17 dev_path=/opt/palace generator=make platform=linux os=ubuntu24.04 target=neoverse_v2 %c,cxx,fortran=gcc@16.0.1
```

Because `ARCH_LABEL` still comes from the requested input, this would publish a Neoverse V2 binary under the `neoverse_v1` ECR tag and S3 key.

Validation job: https://github.com/awslabs/palace/actions/runs/29653166956/job/88102856413

.github/workflows/containers.yml
```diff
@@ -49,34 +58,45 @@ jobs:
...
-    runs-on: [self-hosted, arm64-8g]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - {runner: x64,      target: x86_64_v3}
+          - {runner: x64,      target: x86_64_v4}
+          - {runner: x64,      target: sapphirerapids}
+          - {runner: arm64-8g, target: aarch64}
+          - {runner: arm64-8g, target: neoverse_v1}
```
hughcars: The `neoverse_v1` target was not selected from the V2 runner. Using the existing V1 runner for this matrix entry makes the requested target native rather than relying on the soft preference:

```suggestion
          - {runner: arm64,    target: neoverse_v1}
```

hughcars (commented):
One small issue with a neoverse v2 being published as neoverse v1. Can either mark it as actually a neo v2, or pick the neo v1 runner, or figure a way to make spack stricter in its adherence, your call.

.github/workflows/containers.yml
Sbozzolo: Thank you! Applied this suggestion.

PR #829, opened 2026-07-18
Fix CI permissions for external fork pull requests

hughcars:
## Summary

Fix external-fork CI failures by separating read-only builds from trusted publication.

- Fork PRs may read Spack caches but cannot publish caches, indices, or containers.
- Trusted jobs publish dependencies separately and serialize index updates.
- Long Tests run in the read-only PR workflow; a metadata-only trusted workflow reports their status without executing fork code.

## Validation

All applicable checks pass: 55 passed, with 2 expected skips.

PR #831, opened 2026-07-21
Add InactivePorts boundary condition option for magnetostatic sweeps

dnpham23:
## Summary

Adds user control over how inactive surface current ports are treated during a
magnetostatic inductance sweep. Previously, when one port was excited, every other
port was implicitly left **open** (a natural, zero-surface-current boundary
condition). This PR makes that behavior configurable, both globally and per-port, so
inactive ports can instead be **shorted** (treated as a perfect conductor/PEC),
allowing induced screening currents to flow across them and changing the extracted
inductance matrix accordingly.

## Motivation

The open-vs-short treatment of neighboring conductors materially affects the
extracted inductances. Shorting an inactive ring lets it carry induced screening
current, which is the physically correct model for some device geometries (e.g.
superconducting loops, guard rings). Exposing this as a first-class option lets
users pick the model that matches their problem instead of being locked into the
open assumption.

## Changes

**New configuration**
- Global default: `config["Solver"]["Magnetostatic"]["InactivePorts"]` - `"Open"`
  (default) or `"Short"`.
- Per-port override: `config["Boundaries"]["SurfaceCurrent"][...]["InactiveMode"]`,
  which takes precedence over the global default for that port.

**Solver (`magnetostaticsolver.cpp`)**
- Resolves the effective mode per inactive port (per-port override → global default).
- For each excitation step, collects the boundary attributes of inactive ports that
  resolve to `Short` and, when any exist, rebuilds the curl-curl operator with those
  attributes added as essential (PEC) boundaries for that step. Steps with no shorted
  ports reuse the shared solver, so the common all-open case is unchanged.

**Operator (`curlcurloperator.{cpp,hpp}`)**
- `CurlCurlOperator` accepts an optional `additional_dbc` attribute list; adds a
  `GetStiffnessMatrix(extra_dbc_attr)` overload that reassembles with extra essential
  boundaries.
- Skips the "attribute assigned multiple BCs" check when shorted ports are present,
  since an inactive port is intentionally in both the PEC and surface-current lists
  for that step.

**Config/labels**
- New `InactivePortMode` enum (`Open`/`Short`) with string mapping and JSON
  serialization; `SurfaceCurrentData` gains an optional `inactive_port_mode`.
- JSON schema (`config-schema.json`) updated with a shared `InactivePortMode` def,
  the global `InactivePorts` field, and the per-port `InactiveMode` field.

**Examples & docs**
- `examples/rings/rings.json`: global `InactivePorts` set explicitly to `"Open"`.
- New `examples/rings/rings_inactive_ports.json`: demonstrates the per-port override
  (global `"Open"`, outer ring shorted when inactive).
- Guide (`problem.md`) and rings example (`rings.md`) prose document both the global
  option and the per-port override; the config reference is regenerated from the schema.

## Testing

- Full CMake build passes.
- Verified the three scenarios (all-open, all-short, mixed) produce the expected
  distinct inductance matrices.
- `rings.json` (Open) output matches the existing regression reference within
  `rtol=1e-4`, so the `rings` regression test passes.

## Notes

- Default behavior is unchanged (`"Open"`), so existing configs produce identical
  results.


docs/src/guide/problem.md
```diff
@@ -147,12 +147,21 @@ boundaries.
...
 
+The treatment of the *inactive* current ports during each step of the sweep is controlled
+by [`config["Solver"]["Magnetostatic"]["InactivePorts"]`](../config/reference.md#config-solver-magnetostatic).
+The default, `"Open"`, applies a natural (zero surface current) boundary condition so that
+no current flows across an inactive port. Setting `"Short"` instead treats inactive ports as
+perfect conductors (PEC), allowing induced screening currents to flow across them. The mode
+can be overridden on a per-port basis using
+[`config["Boundaries"]["SurfaceCurrent"]["InactiveMode"]`](../config/reference.md#config-boundaries-surfacecurrent),
+which takes precedence over the global default for that port.
+
```
simlapointe, line 164: Should we discuss (here or elsewhere) what the off-diagonal entries of the inductance matrix mean/represent when we have a mixed configuration (some ports open and some short).

palace/drivers/magnetostaticsolver.cpp
```diff
@@ -104,6 +113,39 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+        // Rebuild the operator with the shorted inactive ports as essential (PEC)
+        // boundaries for this excitation step.
+        CurlCurlOperator curlcurl_step(iodata.boundaries, iodata.solver,
+                                       iodata.domains.materials, iodata.problem.type, mesh,
+                                       short_attrs);
+        auto K_step = curlcurl_step.GetStiffnessMatrix();
+        KspSolver ksp_step(iodata, curlcurl_step.GetNDSpaces(),
+                           &curlcurl_step.GetH1Spaces());
+        ksp_step.SetOperators(*K_step, *K_step);
+        ksp_step.Mult(RHS, A[step]);
```
simlapointe: I could be wrong here, but if `RHS` is assembled at line 113 by the base operator, whose `GetCurrentExcitationVector` only zeroes the _base_ Dirichlet tdofs, then the short DOFs could carry nonzero `RHS` entries that `DIAG_ONE` turns into solution values when ports share edges? Probably only matters in the edge case where an active port shares mesh edges with a shorted port, but if assemble the RHS from curlcurl_step (its tdof list includes the short attributes) or explicitly zero the merged list I think we remove this risk.

palace/models/curlcurloperator.cpp
```diff
@@ -232,6 +245,37 @@ std::unique_ptr<Operator> CurlCurlOperator::GetStiffnessMatrix()
   return K;
 }
 
+std::unique_ptr<Operator>
+CurlCurlOperator::GetStiffnessMatrix(const mfem::Array<int> &extra_dbc_attr)
```
simlapointe, line 236: Is this new function called anywhere? palace/drivers/magnetostaticsolver.cpp doesn't seem to call it

palace/drivers/magnetostaticsolver.cpp
simlapointe, line 217: If we used `ksp_step` for short ports, its solver stats will not get recorded in the metadata

palace/drivers/magnetostaticsolver.cpp
```diff
@@ -32,7 +32,16 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+  // current allowed). The mode is resolved per port: a port's own "InactiveMode" overrides
+  // the global "/Solver/Magnetostatic/InactivePorts" default when set.
+  const InactivePortMode global_mode = iodata.solver.magnetostatic.inactive_port_mode;
+  auto port_is_short = [&](int port_idx)
+  {
+    const auto &port_data = iodata.boundaries.current.at(port_idx);
+    return port_data.inactive_port_mode.value_or(global_mode) == InactivePortMode::SHORT;
+  };
   KspSolver ksp(iodata, curlcurl_op.GetNDSpaces(), &curlcurl_op.GetH1Spaces());
   ksp.SetOperators(*K, *K);
```
simlapointe: For an all-short config this is unnecessary, right? I see an intermediate commit d72c784 seemed to address that but the work was later reverted. I guess there's a good reason for it but just checking

simlapointe (commented):
Looks really good. A few small things inline below. One thing missing is some test coverage. I don't know if it's worth adding a regression test, but we should at least have a `test-config.cpp` case for parsing InactivePorts/InactiveMode (including the override precedence), and probably also a schema unit test.

docs/src/guide/problem.md
dnpham23: Good point. I added a paragraph. Short version: with all ports Open, M_ij is a geometry-only quantity, symmetric by reciprocity. Once ports are shorted, they carry induced screening currents, so M_ij becomes the effective mutual inductance in the presence of those currents. In a mixed config the shorted set (and hence screening currents) can differ between the solve exciting i and the one exciting j, so the cross-energies aren't guaranteed reciprocal. We symmetrize from A_j^T K A_i and it should be read as the effective coupling for that specific inactive-port configuration, not a geometry-only value.

palace/drivers/magnetostaticsolver.cpp
dnpham23: Great catch! `GetCurrentExcitationVector `only zeros the base Dirichlet tdofs, so a shorted port sharing edges with the active port could carry nonzero RHS entries that DIAG_ONE injects as solution values. Fixed by adding `CurlCurlOperator::ZeroEssentialTrueDofs(short_attrs, RHS)`, which zeros the merged essential set (base Dirichlet + shorted attributes) before the solve.

palace/drivers/magnetostaticsolver.cpp
dnpham23: Fixed. There's no longer a separate ksp_step. A single ksp is reused for every step (base and short), so SaveMetadata(ksp) now captures all solves.

palace/drivers/magnetostaticsolver.cpp
dnpham23: Right, and that's what [d72c784](https://github.com/awslabs/palace/commit/d72c7847f8da38c1c65d8593b43d35689771ee6d) was trying to address before reverted. Rather than a special-case short_mode flag, the operator binding is now lazy: a set_operator helper calls ksp.SetOperators only when the bound operator actually changes. An all-short config never binds the base K, so it skips that preconditioner setup, while a mixed/open config still binds it exactly once.

palace/models/curlcurloperator.cpp
dnpham23: It wasn't, thanks. The old code rebuilt a whole CurlCurlOperator + throwaway KspSolver per step instead. I now reworked the driver to actually use this overload: it now reassembles just the stiffness matrix with the shorted ports added as essential BCs, and the single reused ksp swaps to it. So the overload is now the mechanism for Short mode rather than dead code.
While wiring it up I also found the overload had a memory bug: ParOperator::SetEssentialTrueDofs stores a shallow reference to the DOF list, but the overload passed loop-local Array<int> that were destroyed each iteration. The returned operator held dangling references, causing an intermittent parallel segfault/bus error in EliminateBC (serial masked it). Now held in a member (extra_dbc_tdof_lists) that outlives the operator. Also switched the marker sizing to BdrAttrToMarker (global-max) so a rank whose partition lacks the shorted-port boundary can't build an undersized marker. 20× parallel stress runs are now clean.

dnpham23:
With regards to test coverage, I added the following:
- `test-config.cpp:` a "Config Magnetostatic InactivePorts" case (global default, Open/Short parsing, and per-port override precedence via `value_or(global)`) plus a SurfaceCurrent `InactiveMode` parsing section.
- `test-schema.cpp`: enum-constraint validation for both `SurfaceCurrent/InactiveMode` and `Magnetostatic/InactivePorts `(valid values accepted, invalid rejected). Invalid-enum rejection lives at the schema layer since the C++ parser aborts via MFEM_VERIFY rather than throwing.

hughcars:
I would like to hold off on merging this until https://github.com/awslabs/palace/pull/829 is in, because this PR is coming from an external fork so we can use it as an opportunity to test it.

palace/drivers/magnetostaticsolver.cpp
```diff
@@ -104,6 +130,38 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+        }
+      }
+
+      if (short_attrs.Size() > 0)
+      {
+        // Reassemble the stiffness matrix with the shorted inactive ports added as
+        // essential (PEC) boundaries for this excitation step, and zero the excitation on
+        // those DOFs so the DIAG_ONE elimination does not inject spurious values on edges
+        // shared with the active port. The reused solver records its stats for metadata.
+        K_step = curlcurl_op.GetStiffnessMatrix(short_attrs);
```
simlapointe, line 155: In this case each excitation is solved as a different boundary-value problem, with a different essential-DOF set and operator, rather than using one common operator with different right-hand sides. Therefore, I'm not sure if we can use the standard reciprocal cross-energy formula, but `PostprocessTerminals` still computes M_ij = A_j^T K A_i / (I_i I_j) and mirrors it as a reciprocal inductance matrix.

With excitation-dependent Short boundaries, that cross-energy is not either directional flux linkage Phi_i / I_j. In the mixed rings case, the directional values are L_21 = -6.51e-14 H and L_12 = 1.96e-12 H, while terminal-M.csv reports both as approximately -9.05e-23 H. This also affects global all-Short configurations. Consequently, terminal-M.csv, terminal-Minv.csv, and terminal-Mm.csv do not represent the simulated system.

So I'm wondering if we should still print these matrices for excitation-dependent Short mode or not. I'm not sure, but maybe Short mode should either:
1. report only the per-excitation effective self-inductance A_i^T K_i A_i / I_i^2, without producing M, Minv, or Mm; or
2. use a fixed set of passive shorted boundaries for every excitation, so all solutions share one operator and a reciprocal matrix is well-defined.

What do you think? Whatever we choose should be clearly reflected in the docs, currently the “symmetrized effective mutual inductance” might be confusing.

simlapointe (changes requested):
Thanks for the changes, looks good! Thought about it a bit more and have one question/concern remaining, see below.

palace/drivers/magnetostaticsolver.cpp
dnpham23: You're right, and thanks for pushing on this! A Short inactive port is added to the essential-DOF set on every step except the one that excites it. So exciting a Short port uses a different operator than the other excitations, and reciprocity doesn't hold. This is true even for a global all-Short config, since the shorted set changes per excitation. The directional M12/M21 are genuinely asymmetric because they're measurements on two different boundary value problems.

So rather than option 1 (self-only) or option 2 (fixed screens/no sweep), I went with a hybrid that reduces to both in the limits: report self-inductances for all ports, but mutual inductances only between ports that are both Open when inactive. Short-involving off-diagonals are written as NaN, and terminal-Minv.csv/terminal-Mm.csv are computed over the reciprocal Open–Open sub-block only.

Behavior in the three regimes:
- All Open (default): full reciprocal geometry-only matrix unchanged. Verified bit-identical to the rings.json regression output.
- All Short: diagonal self-inductances, off-diagonals NaN (your option 1).
- Mixed: self-inductances for all, reciprocal mutuals among the Open sub-block, NaN elsewhere.
- On the rings mixed case this now gives M[1][1]=4.27e-11, M[2][2]=7.14e-10, off-diagonals NaN. Verified identical across np = 1, 2, 4. I also rewrote the doc to explain the open-vs-short reciprocity distinction.

Please let me know what you think.

docs/src/guide/problem.md
```diff
@@ -147,12 +147,38 @@ boundaries.
...
+*except* the one that excites it, so exciting a `"Short"` port uses a different operator than
+the other excitations; its cross-energy with another column mixes two different
+boundary-value problems and does not correspond to a reciprocal mutual inductance.
+
+Palace therefore reports the self-inductance (diagonal) for every port, but writes a mutual
+inductance ``M_{ij}`` only between ports that are both `"Open"` when inactive. Off-diagonal
+entries involving a `"Short"` port are written as `NaN`, and `terminal-Minv.csv` and
+`terminal-Mm.csv` are computed over the reciprocal `"Open"`-`"Open"` sub-block only (their
+`"Short"`-port rows and columns are likewise `NaN`). With all ports `"Open"` (the default)
+this reduces to the usual full, symmetric, geometry-only inductance matrix.
```
simlapointe: What does "geometry-only" mean here? Do you mean that the resulting inductance matrix is only geometry-dependent (if so, it's also material-dependent I guess)? Or am I misunderstanding?

simlapointe (approved):
LGTM! One small comment below.

docs/src/guide/problem.md
dnpham23: "geometry-only" was sloppy. I meant to contrast with the screened case (where Short ports add artificial PEC boundaries that aren't part of the device geometry), but the inductance is obviously material-dependent too. I replaced it with "full, symmetric inductance matrix (no screening boundaries)" to make the distinction clear. Thanks!

palace/drivers/magnetostaticsolver.cpp
```diff
@@ -178,14 +239,49 @@ void MagnetostaticSolver::PostprocessTerminals(
...
+  // Flux loops always share one operator, so they are all reciprocal.
+  const InactivePortMode global_mode = iodata.solver.magnetostatic.inactive_port_mode;
+  std::vector<bool> reciprocal(n, true);
+  {
+    int col = 0;
+    for (const auto &[idx, data] : surf_j_op)
+    {
+      const auto &port_cfg = iodata.boundaries.current.at(idx);
+      reciprocal[col++] =
+          port_cfg.inactive_port_mode.value_or(global_mode) == InactivePortMode::OPEN;
```
hughcars: Is this correct for a single surface-current port? That is already a supported case (`cavity2d_magnetostatic.json`). The sole port is never inactive, so choosing `Short` should have no effect on the solve. I ran that case with `Open` and `Short`: the other CSVs, including `terminal-M.csv`, were identical, but `Short` changed `Minv[1][1]` and `Mm[1][1]` to `NaN`.

I think the single-port case should remain reciprocal:

```suggestion
      reciprocal[col++] =
          n_current == 1 ||
          port_cfg.inactive_port_mode.value_or(global_mode) == InactivePortMode::OPEN;
```

For coverage, we can set `"InactivePorts": "Short"` in the existing `cavity2d_magnetostatic.json` regression. After this fix, its current finite references should pass unchanged.

palace/utils/configfile.cpp
```diff
@@ -1301,6 +1306,7 @@ ElectrostaticSolverData::ElectrostaticSolverData(const json &electrostatic)
 MagnetostaticSolverData::MagnetostaticSolverData(const json &magnetostatic)
 {
   n_post = magnetostatic.value("Save", n_post);
+  inactive_port_mode = magnetostatic.value("InactivePorts", inactive_port_mode);
```
hughcars, line 1309: Could we add the companion `InactivePorts` entry to `ConcretizeMagnetostatic`? An omitted value resolves to `Open`, but the dry-run/resolved JSON currently contains only `Save` under `Solver.Magnetostatic`.

```cpp
ApplyEntries(j_magnetostatic,
             {{"Save", magnetostatic.n_post},
              {"InactivePorts", ToString(magnetostatic.inactive_port_mode)}});
```

palace/drivers/magnetostaticsolver.cpp
```diff
@@ -104,6 +131,38 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+        }
+      }
+
+      if (short_attrs.Size() > 0)
+      {
+        // Reassemble the stiffness matrix with the shorted inactive ports added as
+        // essential (PEC) boundaries for this excitation step, and zero the excitation on
+        // those DOFs so the DIAG_ONE elimination does not inject spurious values on edges
+        // shared with the active port. The reused solver records its stats for metadata.
+        K_step = curlcurl_op.GetStiffnessMatrix(short_attrs);
```
hughcars, line 155: **Non-blocking:** with multiple Open ports and at least one Short port, the Open-port excitations appear to rebuild the same screened operator and KSP setup. I don't think we should complicate this PR with caching before profiling it, but let's track a follow-up after merge: **"Reuse screened magnetostatic operators"**. Any cache will also need to own the corresponding essential-DOF data for the operator lifetime.

hughcars (changes requested):
One test-coverage note: `test-schema.cpp` already covers the valid and invalid values for the new keys. The missing resolved-config behavior belongs in the existing `ConcretizeDefaults` coverage in `test-config.cpp`, since JSON Schema defaults do not populate the resolved JSON. I think the focused addition there is enough:

```cpp
config = IoData::ConcretizeDefaults(iodata, config);
CHECK(config["Solver"]["Magnetostatic"]["InactivePorts"] == "Open");
```

Also, do you think it is worth having a config and data for the two rings example with shorts and/or one open and one short? I don't know how physically relevant, but it would re-use the existing mesh, and magnetostatics solves are fairly quick, so it would keep this feature tested going forward.

palace/drivers/magnetostaticsolver.cpp
dnpham23: Good catch, fixed. As you pointed out, single current port is never inactive during its own and only solve. I added `n_current == 1 || `to the predicate, as you suggested.

I also took your coverage suggestion: `cavity2d_magnetostatic.json` now sets `"InactivePorts": "Short"`. After the fix its references are bit-identical to before (verified `terminal-M/Minv/Mm.csv`, `terminal-I.csv`, domain-E.csv all unchanged), so it guards this single-port behavior going forward without new reference data.

palace/utils/configfile.cpp
dnpham23: Done. `ConcretizeMagnetostatic` now emits `InactivePorts` alongside `Save`, so an omitted value resolves to `"Open"` in the dry-run/resolved JSON:

```
ApplyEntries(j_magnetostatic,
             {{"Save", magnetostatic.n_post},
              {"InactivePorts", ToString(magnetostatic.inactive_port_mode)}});
```

palace/drivers/magnetostaticsolver.cpp
dnpham23: Agreed, let's not add caching before profiling. I'll open a post-merge follow-up "Reuse screened magnetostatic operators" and note that any cache must own the corresponding essential-DOF data for the operator's lifetime (same lifetime constraint as `extra_dbc_tdof_lists` today).

dnpham23:
Thanks you! I added the focused `ConcretizeDefaults` assertion you suggested in `test-config.cpp`:

`CHECK(config["Solver"]["Magnetostatic"]["InactivePorts"] == "Open");`
For ongoing coverage I went with the single-port `cavity2d_magnetostatic.json` route (now `"InactivePorts": "Short"`, references unchanged) rather than a two-rings shorts/mixed case. The reason: in Short/mixed mode the off-diagonals are `NaN` by design, and the regression comparator uses Catch2 `WithinRel/WithinAbs`, which fail on any `NaN`, so a rings-Short reference can't be diffed without extending the harness. The cavity2d case exercises the Short code path end-to-end while keeping all references finite.

hughcars:
> Thanks you! I added the focused `ConcretizeDefaults` assertion you suggested in `test-config.cpp`:
> 
> `CHECK(config["Solver"]["Magnetostatic"]["InactivePorts"] == "Open");` For ongoing coverage I went with the single-port `cavity2d_magnetostatic.json` route (now `"InactivePorts": "Short"`, references unchanged) rather than a two-rings shorts/mixed case. The reason: in Short/mixed mode the off-diagonals are `NaN` by design, and the regression comparator uses Catch2 `WithinRel/WithinAbs`, which fail on any `NaN`, so a rings-Short reference can't be diffed without extending the harness. The cavity2d case exercises the Short code path end-to-end while keeping all references finite.

Ok, we should cut another issue after this to extend the test harness to allow for nan to nan comparisons (that fails if either side is nan alone), then once that is done we can add that regression case.

hughcars (approved):

PR #832, opened 2026-07-21
Allow publishing containers without a PR

Sbozzolo:
The only way to publish a dev-<branch> prototype container was to open a PR and add the `push-containers` label; workflow_dispatch existed but was hard-wired to build-only. This PR adds a boolean `push` input to the dispatch trigger so a container can be published straight from a branch without creating a PR first.

### DIFF
--- a/.github/actions/build-container/action.yml
+++ b/.github/actions/build-container/action.yml
@@ -40,6 +40,11 @@ runs:
         # resolve head.repo to the fork, so this stays false for them — they must
         # never reach the instance-role push credentials on the public repo.
         LABEL_PUSH: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'push-containers') && github.event.pull_request.head.repo.full_name == github.repository }}
+        # 'true' iff a manual dispatch requested a push. GitHub only lets users
+        # with write access dispatch a workflow, and dispatch always runs a
+        # branch of this repo (never a fork), so this path is push-safe without
+        # the head.repo guard the label path needs.
+        DISPATCH_PUSH: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.push == 'true' }}
       run: |
         ARCH_LABEL=$(spack arch -t)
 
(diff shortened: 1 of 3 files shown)


Sbozzolo:
To publish a `dev-<branch>` prototype straight from a branch, trigger the workflow manually with the `push` input enabled. Only users with write access can dispatch the workflow, and dispatch always runs a branch of this repository (never a fork), so this path is push-safe.

Using the [GitHub CLI](https://cli.github.com/):

```bash
# Publish a dev-<branch> prototype from the current branch.
gh workflow run containers.yml --ref "$(git branch --show-current)" -f push=true

# Follow the run that was just started.
gh run watch "$(gh run list --workflow=containers.yml --limit=1 --json databaseId --jq '.[0].databaseId')"
```

Omit `-f push=true` (or set `push=false`) for a build-only run that verifies the image builds without publishing. From the web UI, the same controls live under Actions → Containers → Run workflow, where you pick the branch and tick push.

Sbozzolo:
Example: https://github.com/awslabs/palace/actions/runs/29865223835

.github/actions/build-container/action.yml
```diff
@@ -54,15 +59,17 @@ runs:
...
-        #                              (plain PRs, manual dispatch re-runs).
+        #                              (plain PRs, build-only dispatch re-runs).
         # ECR tags can't contain '/', so VERSION uses '-' while the S3 prefix
         # keeps '/' for a tidy bucket layout.
         if [ "$EVENT_NAME" = "push" ] && [ "$GITHUB_REF" = "refs/heads/main" ]; then
           VERSION="main"; S3_PREFIX="main"; PUSH=true
         elif [ "$EVENT_NAME" = "push" ] && [ "${GITHUB_REF#refs/tags/v}" != "$GITHUB_REF" ]; then
           VERSION="${GITHUB_REF#refs/tags/v}"; S3_PREFIX="$VERSION"; PUSH=true
-        elif [ "$LABEL_PUSH" = "true" ]; then
+        elif [ "$LABEL_PUSH" = "true" ] || [ "$DISPATCH_PUSH" = "true" ]; then
```
hughcars, line 72: One small concurrency gotcha I think: the lock is keyed by event ref, while the published destination is keyed by source branch. A labeled PR uses `refs/pull/...`, and a dispatch uses `refs/heads/...`, so both can publish the same `dev-<branch>` artifacts concurrently. Could we key publication concurrency from the source repo + branch instead, so the two paths serialize?

.github/workflows/containers.yml
```diff
@@ -13,9 +13,16 @@ on:
     # 'labeled' so adding the `push-containers` label triggers a prototype
     # deploy; the rest are the usual PR-build triggers.
     types: [opened, reopened, synchronize, labeled]
-  # Manual build-only re-run (e.g. after a transient runner failure). Never
-  # publishes — prototype deploys go through the `push-containers` PR label.
+  # Manual run from any branch. Leaving `push` unchecked is a build-only
+  # re-run (e.g. after a transient runner failure); checking it publishes a
+  # `dev-<branch>` prototype without opening a PR. GitHub restricts dispatch
```
hughcars, line 18: Small guard/docs mismatch: `workflow_dispatch --ref` can target a tag too. That would publish `dev-v<tag>` and also trigger the ref-only Neoverse V2 job. Could we either reject dispatches unless `github.ref_type == 'branch'`, or explicitly support and document the tag behavior?

docs/src/developer/notes.md
```diff
@@ -492,6 +492,27 @@ spack spec local.palace@develop
...
+(never a fork), so this path is push-safe.
+
+Using the [GitHub CLI](https://cli.github.com/):
+
+```bash
+# Publish a dev-<branch> prototype from the current branch.
+gh workflow run containers.yml --ref "$(git branch --show-current)" -f push=true
+
+# Follow the run that was just started.
+gh run watch "$(gh run list --workflow=containers.yml --limit=1 --json databaseId --jq '.[0].databaseId')"
```
hughcars, line 509: Small race here: `--limit=1` is repository-wide, so a PR or main run landing between these commands can make us watch the wrong run. Current `gh workflow run` returns the created run URL/ID when available—could we use that exact run instead?

hughcars (approved):
Some tiny nits, your call if you want to adjust.

PR #837, opened 2026-07-26
Use Hermitian similarity for smoother spectral estimates

hughcars:
## Summary

- Estimate real and exactly-real complex Chebyshev and automatically damped Jacobi spectra using the Hermitian similarity operator `D^-1/2 A D^-1/2`.
- Extend `ProductOperator` and `ComplexProductOperator` to compose three or more factors.
- Leave the general-complex non-Hermitian path unchanged.

## Motivation

For nonconstant positive diagonal `D`, `D^-1 A` is generally not Euclidean-Hermitian even when `A` is Hermitian. Passing it to SLEPc as a Hermitian eigenproblem violates the solver contract.

`D^-1/2 A D^-1/2` is Hermitian under the existing SPD assumption and is similar to `D^-1 A`, so it provides the required spectrum while satisfying that contract.

## Tests

- The focused regression fails with the previous estimator and passes with this change.
- Full serial and two-rank MPI unit suites pass.
- Required CI and long tests pass.

### DIFF
--- a/palace/linalg/chebyshev.cpp
+++ b/palace/linalg/chebyshev.cpp
@@ -13,18 +13,28 @@ namespace
 
 double GetLambdaMax(MPI_Comm comm, const Operator &A, const Vector &dinv)
 {
-  // Assumes A SPD (diag(A) > 0) to use Hermitian eigenvalue solver.
-  DiagonalOperator Dinv(dinv);
-  ProductOperator DinvA(Dinv, A);
-  return linalg::SpectralNorm(comm, DinvA, true);
+  // D⁻¹A is generally not Hermitian, but is similar to the Hermitian operator
+  // D⁻¹ᐟ²AD⁻¹ᐟ² under the existing SPD assumption.
+  Vector dinv_sqrt(dinv);
+  linalg::Sqrt(dinv_sqrt);
+  DiagonalOperator DinvSqrt(dinv_sqrt);
+  ProductOperator S(DinvSqrt, A, DinvSqrt);
+  return linalg::SpectralNorm(comm, S, true);
 }
 
 double GetLambdaMax(MPI_Comm comm, const ComplexOperator &A, const ComplexVector &dinv)
 {
-  // Assumes A SPD (diag(A) > 0) to use Hermitian eigenvalue solver.
+  if (A.IsReal())
+  {
+    ComplexVector dinv_sqrt(dinv);
+    linalg::Sqrt(dinv_sqrt.Real());
+    ComplexDiagonalOperator DinvSqrt(dinv_sqrt);
+    ComplexProductOperator S(DinvSqrt, A, DinvSqrt);
+    return linalg::SpectralNorm(comm, S, true);
+  }
   ComplexDiagonalOperator Dinv(dinv);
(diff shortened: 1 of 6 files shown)


palace/linalg/chebyshev.cpp
```diff
@@ -13,18 +13,28 @@ namespace
...
 double GetLambdaMax(MPI_Comm comm, const Operator &A, const Vector &dinv)
 {
-  // Assumes A SPD (diag(A) > 0) to use Hermitian eigenvalue solver.
-  DiagonalOperator Dinv(dinv);
-  ProductOperator DinvA(Dinv, A);
-  return linalg::SpectralNorm(comm, DinvA, true);
+  // D⁻¹A is generally not Hermitian, but is similar to the Hermitian operator
+  // D⁻¹ᐟ²AD⁻¹ᐟ² under the existing SPD assumption.
+  Vector dinv_sqrt(dinv);
+  linalg::Sqrt(dinv_sqrt);
```
simlapointe, line 19: Both the old and new code assume diag(A) > 0, but the consequence of violating it changes from "finite, possibly poor estimate" to a hard crash. With A = diag(1, -1) (indefinite diagonal, otherwise well-behaved):                                                                                                                                                                         
                                                                                                                                                                                                
- Old code: D⁻¹A = I, returns λ = 1, smoothers proceed.                                                                                                                                          
- This PR: Sqrt produces NaN → SLEPc aborts the whole job from inside orthogonalization, before the MFEM_VERIFY(lambda_max > 0.0) guard can fire, with nothing pointing at the smoother diagonal:                                                                                                                                                                                      
                                                                                                                                                                                                
```                                                                                                                                                                                              
[0]PETSC ERROR: The inner product is not well defined: nonzero imaginary part nan.
[0]PETSC ERROR: #1 BV_SafeSqrt() ...
[0]PETSC ERROR: #7 EPSSolve_KrylovSchur_Default() ...
[0]PETSC ERROR: #9 GetMaxSingularValue() at palace/linalg/slepc.cpp:275
MPI_ABORT was invoked on rank 0 ...                                             
```                                                                                                                                                                                              
                                                                                                                                                                                              
This is potentially reachable: pc_mat_shifted defaults to 0 except for driven+AMS and boundary-mode, so an unshifted K − ω²M-type preconditioner with "PCMatReal": true (or a lossless eigenmode problem, where ComplexParOperator has a null imaginary part and takes the IsReal() branch) can hit negative diagonal entries on under-resolved meshes / coarse MG levels at high frequency.               
                                                                                                                                                                                                
Suggestion: verify positivity of the diagonal before Sqrt with an actionable message, or fall back to the non-Hermitian SVD estimate when negative entries are detected. `JacobiSmoother::SetOperator` in particular has no `lambda_max` guard at all, so it deserves one either way.

Also applies to the duplicate in palace/linalg/jacobi.cpp:19 and the complex branches at chebyshev.cpp:30 / jacobi.cpp:30.

simlapointe (commented):
Good catch! One potentially-worthy check for the sqrt positivity assumption but otherwise looks good.

simlapointe (approved):

PR #839, opened 2026-07-26
Fix MPI output-directory test deadlocks

### LINKED ISSUE #838: MPI output-directory tests can deadlock
# MPI output-directory tests can deadlock

The parallel output-directory tests added in #813 can deadlock because they make rank-local fatal assertions about filesystem state immediately before a collective helper mutates that state.

For example, rank 0 can pass this check and create the directory inside `EnsureDirectory` before rank 1 evaluates the same check:

```cpp
REQUIRE(!fs::exists(dir));
EnsureDirectory(dir, comm);
```

Rank 1 then fails `REQUIRE` and unwinds into the `SharedTempDir` destructor barrier while rank 0 enters the helper barrier. Those barriers cross-match; rank 0 continues into `MPI_Comm_split_type`, which rank 1 skipped, and hangs.

This test forces the race:

```cpp
#include <thread>

TEST_CASE_METHOD(palace::test::SharedTempDir,
                 "EnsureDirectory rank-skew deadlock reproducer", "[Parallel]")
{
  MPI_Comm comm = MPI_COMM_WORLD;
  auto dir = temp_dir / "postpro";

  if (!Mpi::Root(comm))
  {
    while (!fs::exists(dir))
    {
      std::this_thread::yield();
    }
  }

  REQUIRE(!fs::exists(dir));
  EnsureDirectory(dir, comm);
}
```

```console
mpirun -np 2 palace-unit-tests \
  "EnsureDirectory rank-skew deadlock reproducer"
```

Existing CI evidence:

- [Pre-merge #813](https://github.com/awslabs/palace/actions/runs/30117115678/job/89560572628): `mpi-EnsureDirectory creates a new directory` never completed.
- [Current main, Linux](https://github.com/awslabs/palace/actions/runs/30128681826/job/89598127066): `mpi-RemovePreviousOutput removes an existing directory and its contents` never completed; 156/157 tests finished.
- [Current main, macOS/MPICH](https://github.com/awslabs/palace/actions/runs/30128681854/job/89598131522): the `EnsureDirectory` case never completed; 156/157 tests finished.

The affected tests normally finish in under two seconds. The jobs instead time out with `mpiexec` and both test ranks still alive.

The fix is to cache and globally reduce each precondition before the fatal assertion:

```cpp
bool directory_absent = !fs::exists(dir);
Mpi::GlobalAnd(1, &directory_absent, comm);
REQUIRE(directory_absent);

EnsureDirectory(dir, comm);
```

The same pattern should be applied to the `is_directory`, `is_symlink`, and `!exists` checks before `RemovePreviousOutput`. The `SaveIteration is MPI collective` test already does this.

hughcars:
Fix [#838](https://github.com/awslabs/palace/issues/838)


Sbozzolo (approved):
Oops! Thank you for fixing this!

PR #841, opened 2026-07-27
Allow exciting reactive lumped ports

simlapointe:
Previously, excited lumped ports were required to be purely resistive. This PR removes this restriction and excited lumped ports may now carry reactance (`"L"` and/or `"C"`, including purely reactive ports with `"R": 0`). The reactance enters the system matrix as a physical termination; the incident drive is normalized to the port `"R"` when positive, or to an internal unit reference impedance for a purely reactive port.

### DIFF
--- a/docs/src/guide/boundaries.md
+++ b/docs/src/guide/boundaries.md
@@ -197,7 +197,13 @@ for implementation details.
     The medium adjacent to the port must be homogeneous and isotropic.
 
 For each port, the excitation is normalized to have unit incident power over the port boundary
-surface.
+surface. Lumped ports with reactive elements (`"L"` and/or `"C"`) may also be excited, including
+purely reactive ports with `"R": 0`. This is useful for driving a structure through a Josephson
+junction or other lumped reactive element. The port reactance enters the system matrix as a
+physical termination, and the fields, synthesized circuit matrices, and port voltages/currents are
+all valid. For a purely reactive port (`R = 0`) the power normalization references an internal unit
+impedance; see the [mathematical reference](../reference.md#Lumped-ports-and-wave-ports) for the
+S-parameter interpretation in this case.
 
 The presence of an incident excitation at a port is controlled by the settings
 [`config["Boundaries"]["LumpedPort"][]["Excitation"]`](../config/reference.md#config-boundaries-lumpedport)
--- a/docs/src/guide/boundaries.md
+++ b/docs/src/guide/boundaries.md
@@ -198,12 +198,9 @@ for implementation details.
 
 For each port, the excitation is normalized to have unit incident power over the port boundary
 surface. Lumped ports with reactive elements (`"L"` and/or `"C"`) may also be excited, including
-purely reactive ports with `"R": 0`. This is useful for driving a structure through a Josephson
-junction or other lumped reactive element. The port reactance enters the system matrix as a
-physical termination, and the fields, synthesized circuit matrices, and port voltages/currents are
-all valid. For a purely reactive port (`R = 0`) the power normalization references an internal unit
-impedance; see the [mathematical reference](../reference.md#Lumped-ports-and-wave-ports) for the
-S-parameter interpretation in this case.
+purely reactive ports with `"R": 0`. For a purely reactive port (`R = 0`) the power normalization
+references an internal unit impedance; see the [mathematical reference](../reference.md#Lumped-ports-and-wave-ports)
+for the S-parameter interpretation in this case.
 
 The presence of an incident excitation at a port is controlled by the settings
 [`config["Boundaries"]["LumpedPort"][]["Excitation"]`](../config/reference.md#config-boundaries-lumpedport)
--- a/docs/src/guide/boundaries.md
+++ b/docs/src/guide/boundaries.md
@@ -197,8 +197,9 @@ for implementation details.
     The medium adjacent to the port must be homogeneous and isotropic.
 
 For each port, the excitation is normalized to have unit incident power over the port boundary
-surface. Lumped ports with reactive elements (`"L"` and/or `"C"`) may also be excited, including
-purely reactive ports with `"R": 0`. For a purely reactive port (`R = 0`) the power normalization
+surface. Lumped ports with reactive elements (`"L"` and/or `"C"`, or the surface-parameter
+equivalents `"Ls"` and/or `"Cs"`) may also be excited, including purely reactive ports with
+`"R": 0` (`"Rs": 0`). For a purely reactive port (`R = 0`) the power normalization
 references an internal unit impedance; see the [mathematical reference](../reference.md#Lumped-ports-and-wave-ports)
 for the S-parameter interpretation in this case.
 
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -277,7 +277,11 @@ fixed ``\sqrt{R_j/Z_0}`` conversion factor separates them from a voltage-transfe
 They satisfy reciprocity (``S_{ji} = S_{ij}``). The same unit-reference projection applies
 to *passive* purely reactive lumped ports (``R = 0`` with ``L`` and/or ``C``) appearing as
 observation ports: their S-parameter rows are finite and referenced to ``Z_0``, rather
-than identically zero.
+than identically zero. Likewise, the incident voltage and current columns of the port
+postprocessing output (``V^{inc}``, ``I^{inc}``) at a purely reactive excited port are
+finite and referenced to ``Z_0`` — like the S-parameter row, they are bookkeeping
+quantities consistent with the assembled drive normalization, not physical incident
+waves (an ``R = 0`` port supports no incident traveling wave).
 
 In the time domain, the time histories of the port voltages can be Fourier-transformed to
 get their frequency domain representation for scattering parameter calculation.
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -250,9 +250,12 @@ S_{ij} = \frac{\displaystyle\int_{\Gamma_i}\bm{E}\cdot\bm{E}^{inc}_i\,dS}
 For a resistive excited port, the incident field ``\bm{E}^{inc}`` is normalized so that the
 power integrated over the port boundary is unity, referenced to the port resistance ``R``.
 A lumped port with nonzero inductance and/or capacitance may also be excited. In this case
-the port's ``R``, ``L``, and ``C`` all enter the system matrix as a physical parallel
-termination, and the incident field is normalized to a real reference resistance: ``R`` when
-``R > 0``, or the free-space impedance ``Z_0 \approx 376.7~\Omega`` when ``R = 0``. The
+the *termination* and *excitation* coefficients differ: the port's full ``R``, ``L``, and
+``C`` enter the system matrix as a physical parallel termination through the Robin
+coefficient ``\gamma = i\omega/Z_s`` with the complex ``Z_s(\omega)``, while the source
+term ``\bm{U}^{inc}`` is assembled with the real drive coefficient ``i\omega/R_{ref}``,
+where ``R_{ref}`` is a real reference resistance: ``R`` when ``R > 0``, or the free-space
+impedance ``Z_0 \approx 376.7~\Omega`` when ``R = 0`` (and equivalently ``Z_s^{-1} \to R_{ref}^{-1}`` in the time-domain source term). The
 scattering parameters at resistive ports (observation or drive) retain their standard power-wave
 interpretation. For a purely reactive drive port (``R = 0``), the reported ``S_{ii}`` is related
 to the loaded port impedance by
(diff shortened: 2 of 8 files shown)


palace/models/lumpedportoperator.cpp
```diff
@@ -37,15 +37,11 @@ LumpedPortData::LumpedPortData(const config::LumpedPortData &data,
   {
     if (has_circ)
     {
-      MFEM_VERIFY(data.R > 0.0, "Excited lumped port must have nonzero resistance!");
-      MFEM_VERIFY(data.C == 0.0 && data.L == 0.0,
-                  "Lumped port excitations do not support nonzero reactance!");
+      MFEM_VERIFY(data.R >= 0.0, "Excited lumped port must have non-negative resistance!");
```
hughcars, line 40: Relaxing this also admits negative `L`/`C` on excited ports: the old `== 0` check rejected them, while they now reach the matrix assembly as negative stiffness/mass terms. If non-Foster elements are not intentional, can we validate `L >= 0` and `C >= 0` here too, with the equivalent check for `Ls`/`Cs`?

palace/models/lumpedportoperator.hpp
```diff
@@ -91,6 +91,15 @@ class LumpedPortData
...
   constexpr bool HasExcitation() const { return excitation != 0; }
 
+  // Real reference resistance used to normalize the incident-field amplitude of an excited
+  // port. For a resistive port this is R, so the incident/reflected-wave normalization and
+  // the port's own termination impedance coincide and behaviour is unchanged. For a purely
+  // reactive excited port (R == 0) there is no real port resistance, so we fall back to the
+  // unit reference impedance |Z_R| = 1 in internal units (= Z_freespace) purely to define a
+  // finite drive amplitude; the reactance acts physically through the system-matrix
+  // termination, and that port's own S column is referenced to this unit impedance.
+  double GetExcitationRefResistance() const { return (std::abs(R) > 0.0) ? R : 1.0; }
```
hughcars, line 101: `rom-port-reference.csv` still uses `GetCharacteristicImpedance(omega)`, while direct Palace S now uses this real `R`/`Z₀` reference. The ROM table is documented as the reference for converting synthesized admittance to Palace/Kurokawa S, so reactive ports now have two definitions. The physical RLC load already lives in `rom-portload-*`; should this table use the real reference too, or should the two concepts be split and named explicitly?

palace/models/lumpedportoperator.cpp
```diff
@@ -182,7 +182,13 @@ void LumpedPortData::InitializeLinearForms(mfem::ParFiniteElementSpace &nd_fespa
...
     for (const auto &elem : elems)
     {
-      const double Rs = R * GetToSquare(*elem);
+      // Reference the S-parameter projection to the same real resistance used to normalize
+      // the incident drive (R for a resistive port; the unit reference for a purely
+      // reactive R == 0 port, so this does not divide by zero). The reactance is already
+      // present in the system matrix, so the projected field is the physical response; a
+      // purely reactive port's own S is not a meaningful traveling-wave quantity
+      // regardless.
+      const double Rs = GetExcitationRefResistance() * GetToSquare(*elem);
```
hughcars, line 195: This fallback applies to every lumped port, not only excited ones. Existing passive `R = 0` LC ports therefore change from an identically zero S row to a finite unit-reference projection under the same drive; the transmon example's junction port is one case. That is probably the right reciprocity-preserving behavior, but it is an output-contract change and should be called out in the changelog/reference.

docs/src/guide/boundaries.md
```diff
@@ -197,7 +197,10 @@ for implementation details.
     The medium adjacent to the port must be homogeneous and isotropic.
 
 For each port, the excitation is normalized to have unit incident power over the port boundary
-surface.
+surface. Lumped ports with reactive elements (`"L"` and/or `"C"`) may also be excited, including
```
hughcars: The same change enables surface-parameter ports (`Rs = 0` with `Ls` and/or `Cs`), but the guide and changelog only mention `R`/`L`/`C`. Could we mention both forms?

docs/src/reference.md
```diff
@@ -247,6 +247,28 @@ S_{ij} = \frac{\displaystyle\int_{\Gamma_i}\bm{E}\cdot\bm{E}^{inc}_i\,dS}
     {\displaystyle\int_{\Gamma_i}\bm{E}^{inc}_i\cdot\bm{E}^{inc}_i\,dS} - \delta_{ij} \,.
 ```
 
+For a resistive excited port, the incident field ``\bm{E}^{inc}`` is normalized so that the
+power integrated over the port boundary is unity, referenced to the port resistance ``R``.
+A lumped port with nonzero inductance and/or capacitance may also be excited. In this case
+the port's ``R``, ``L``, and ``C`` all enter the system matrix as a physical parallel
+termination, and the incident field is normalized to a real reference resistance: ``R`` when
```
hughcars: The source equations above still use the physical `γ = iω/Z_s` (and `Z_s^{-1}` in time domain), but the implementation now drives with `iω/R_ref` and leaves L/C in the system matrix. Please distinguish the termination and excitation coefficients here. The comments in `lumpedportoperator.cpp:627` and `postoperator.cpp:1312` are stale for the same reason.

docs/src/reference.md
```diff
@@ -247,6 +247,28 @@ S_{ij} = \frac{\displaystyle\int_{\Gamma_i}\bm{E}\cdot\bm{E}^{inc}_i\,dS}
...
+scattering parameters at resistive ports (observation or drive) retain their standard power-wave
+interpretation. For a purely reactive drive port (``R = 0``), the reported ``S_{ii}`` is related
+to the loaded port impedance by
+
+```math
+S_{ii} + 1 = \frac{2\,Z_{\text{loaded}}}{Z_0}
+```
+
+where ``Z_{\text{loaded}} = Z_p \parallel Z_{\text{struct}}`` is the parallel combination of
+the port impedance ``Z_p = (1/R + 1/(i\omega L) + i\omega C)^{-1}`` and the structure's
```
hughcars: For `R = 0`, this reads `1/0`, making `Z_p = 0` (a short) instead of omitting the resistor. Could we define the conductance as `1/R` for `R > 0` and zero otherwise, or state that zero-valued branches are omitted?

docs/src/reference.md
```diff
@@ -247,6 +247,28 @@ S_{ij} = \frac{\displaystyle\int_{\Gamma_i}\bm{E}\cdot\bm{E}^{inc}_i\,dS}
...
+input impedance at the port plane. This is a reciprocal, frequency-dependent
+quantity — but it is not bounded by unity and should not be interpreted as a power
+reflection coefficient. Transmission parameters ``S_{ji}`` (``j \ne i``) between a reactive
+drive and a resistive observation port are physical voltage-transfer ratios and satisfy
```
hughcars: `S_ji` is not generally a raw voltage-transfer ratio when the two references differ. A reactive drive uses `Z₀` while a resistive observation port uses `R_j`, leaving a `sqrt(R_j/Z₀)` conversion factor. “Real-reference-normalized transmission amplitude” would be more accurate.

test/unit/test-lumpedportintegration.cpp
```diff
@@ -785,3 +785,122 @@ TEST_CASE("LumpedPort_BasicTests_3ElementPort_Cube321", "[lumped_port][Serial][P
...
+  else
+  {
+    CHECK_THAT(port_1.GetExcitationRefResistance(), WithinRel(1.0));
+  }
+
+  // (3) Characteristic impedance Z_ref(w) = 1 / (1/R + 1/(iwL) + iwC). Check against a
+  // direct evaluation at a representative (nondimensional) frequency.
+  const double f_GHz = 10.0;
+  const double omega =
+      iodata.units.Nondimensionalize<VT::FREQUENCY>(2.0 * M_PI * f_GHz * 1.0e9);
```
hughcars: `Nondimensionalize<FREQUENCY>` takes GHz here, not Hz, so this evaluates at `10^10` GHz rather than 10 GHz. Other call sites use `2π * Nondimensionalize(f_GHz)`. This should be:

```cpp
const double omega =
    2.0 * M_PI * iodata.units.Nondimensionalize<VT::FREQUENCY>(f_GHz);
```

test/unit/test-lumpedportintegration.cpp
```diff
@@ -785,3 +785,122 @@ TEST_CASE("LumpedPort_BasicTests_3ElementPort_Cube321", "[lumped_port][Serial][P
...
+  space_op.GetExcitationVector1(1, RHS);
+  double max_abs = 0.0;
+  for (int i = 0; i < RHS.Real().Size(); i++)
+  {
+    CHECK(std::isfinite(RHS.Real()[i]));
+    max_abs = std::max(max_abs, std::abs(RHS.Real()[i]));
+  }
+  Mpi::GlobalMax(1, &max_abs, world_comm);
+  // The drive is nonzero (a genuine incident field was assembled).
+  CHECK(max_abs > 0.0);
```
hughcars, line 904: Could this cover the output paths too? Neither `GetExcitationVoltage()` nor `GetSParameter()` is exercised, so reverting either path to use `R` would still pass. An `R = 0` assertion for excitation voltage and S projection would catch that.

hughcars (commented):
A few edge cases and issues, but generally looking good. Mostly doc mismatches etc.

test/unit/test-lumpedportintegration.cpp
simlapointe: Added that to the test in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

test/unit/test-lumpedportintegration.cpp
simlapointe: Fixed in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

docs/src/reference.md
simlapointe: Update the documentation in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

docs/src/reference.md
simlapointe: Yes, done in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

docs/src/reference.md
simlapointe: Updated the reference.md and corresponding comments in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

docs/src/guide/boundaries.md
simlapointe: Both forms mentioned in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

palace/models/lumpedportoperator.cpp
simlapointe: Mentioned in changelog and reference md files in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

palace/models/lumpedportoperator.hpp
simlapointe: Good catch. Now in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac) the `rom-port-reference.csv` uses the `GetExcitationRefResistance()`

palace/models/lumpedportoperator.cpp
simlapointe: Added checks for `L`, `C`, `Ls` and `Cs >= 0` in [3cd7977](https://github.com/awslabs/palace/pull/841/commits/3cd7977653e9d5954ab3979ebe41382ff3dfb0ac)

palace/models/lumpedportoperator.cpp
dnpham23: Minor: 
The changelog note covers port-S.csv, but I think one more output path changed and isn't yet mentioned? `GetExcitationVoltage()` switched from R to `GetExcitationRefResistance() `(line 150). That feeds the `inc{idx}_{ex} `columns of `port-V.csv` and `port-I.csv` via `postoperatorcsv.cpp:1064`:

```
double V_inc = data.GetExcitationVoltage() * Jcoeff;
double I_inc = (std::abs(V_inc) > 0.0)
                   ? data.GetExcitationPower() * Jcoeff * Jcoeff / V_inc
                   : 0.0;
```
Previously, a purely reactive drive gave `Rs = R * GetToSquare(...) = 0`, so `E_inc = sqrt(0) = 0` and `V_inc = 0;` the `std::abs(V_inc) > 0.0` guard then zeroed `I_inc` too, and both columns read as a visible "not applicable." Now `V_inc` is finite and `I_inc = 1/V_inc` (since `GetExcitationPower() ` is hardcoded to 1.0), so both are emitted dimensionalized to volts and amps, describing an incident wave that doesn't physically exist at an `R = 0` port.

I think that's the right choice for consistency with the assembled drive, but it's the same class of change as the `port-S.csv` row and a user has no way to tell those columns are `Z_0`-referenced bookkeeping rather than physical incident quantities. Could the changelog note and the new `reference.md` paragraph extend to `Vinc/Iinc` alongside the S-parameter row?

dnpham23 (approved):
LGTM. Maybe just one minor optional comment (see below)

palace/models/lumpedportoperator.cpp
hughcars: Follow up to this point, why did that change not light up in the regression suite? Are these inc values not checked?

simlapointe: They are checked but not produced by any existing test case. The only case with L/C on a lumped port at all is the single transmon and it's an eigenmode problem so no excitation, no inc columns in their port-V.csv

docs/src/reference.md
```diff
@@ -247,6 +247,42 @@ S_{ij} = \frac{\displaystyle\int_{\Gamma_i}\bm{E}\cdot\bm{E}^{inc}_i\,dS}
     {\displaystyle\int_{\Gamma_i}\bm{E}^{inc}_i\cdot\bm{E}^{inc}_i\,dS} - \delta_{ij} \,.
 ```
 
+For a resistive excited port, the incident field ``\bm{E}^{inc}`` is normalized so that the
+power integrated over the port boundary is unity, referenced to the port resistance ``R``.
+A lumped port with nonzero inductance and/or capacitance may also be excited. In this case
+the *termination* and *excitation* coefficients differ: the port's full ``R``, ``L``, and
+``C`` enter the system matrix as a physical parallel termination through the Robin
+coefficient ``\gamma = i\omega/Z_s`` with the complex ``Z_s(\omega)``, while the source
```
hughcars: The prose now distinguishes the coefficients, but the equations above still define `Uinc = -2γ(...)` with `γ = iω/Zs` and use `Zs^{-1}` in time domain. Could we update those equations directly to use `γ_exc = iω/R_ref` and `R_ref^{-1}`? As written, the reference still gives two definitions for the source term.

hughcars (approved):
Thanks, the review fixes look good. One remaining documentation mismatch inline; non-blocking.

docs/src/reference.md
simlapointe: Good point. Updated the reference equations in [7835dbd](https://github.com/awslabs/palace/pull/841/commits/7835dbd1c1a88386e39d686de70e1ca3f249091c)

PR #842, opened 2026-07-28
Change approach in dpkg lock

Sbozzolo:
A freshly booted runner races us: systemd's apt-daily / unattended-upgrade units kick off their own apt-get on boot and hold the dpkg lock, which makes the `apt-get install` steps below die with exit code 100.

We prevent that by:
  1. `systemctl stop` on the *services* blocks until the unit's whole cgroup (children included) has drained, so nothing is left orphaned.
  2. `DPkg::Lock::Timeout` tells apt to *wait* for the lock rather than failing. It uses the same lock apt/dpkg do, so it is reliable where `flock`/`rm` are not, and it covers every apt call in the job -- here and in downstream actions (e.g. build-container installing podman).

### DIFF
--- a/.github/actions/setup-runner/action.yml
+++ b/.github/actions/setup-runner/action.yml
@@ -66,9 +66,9 @@ runs:
     - name: Disable unattended upgrades
       shell: bash
       run: |
-        # Stop the periodic apt units and unattended-upgrades. `systemctl stop`
-        # waits for the cgroup to drain, so any in-flight apt-get is gone (not
-        # merely orphaned) by the time this returns.
+        # Stop the periodic APT units and ask unattended-upgrades to shut down
+        # cleanly. Any remaining dpkg lock contention is handled by the timeout
+        # below.
         sudo systemctl stop apt-daily.timer apt-daily-upgrade.timer || true
         sudo systemctl stop apt-daily.service apt-daily-upgrade.service || true
         sudo systemctl stop unattended-upgrades.service || true
--- a/.github/actions/setup-runner/action.yml
+++ b/.github/actions/setup-runner/action.yml
@@ -55,30 +55,39 @@ runs:
         }
         EOF
 
-    # This can sometimes interfere with sudo apt operations below. We don't
-    # really care about these updates anyway, as we spin up new instances all
-    # the time.
+    # A freshly booted runner races us: systemd's apt-daily / unattended-upgrade
+    # units kick off their own apt-get on boot and hold the dpkg lock, which
+    # makes the `apt-get install` steps below die with exit code 100.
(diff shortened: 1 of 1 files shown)


.github/actions/setup-runner/action.yml
```diff
@@ -55,30 +55,39 @@ runs:
...
+    # units kick off their own apt-get on boot and hold the dpkg lock, which
+    # makes the `apt-get install` steps below die with exit code 100.
+    #
+    # We prevent that by:
+    #   1. `systemctl stop` on the *services* blocks until the unit's whole
+    #      cgroup (children included) has drained, so nothing is left orphaned.
+    #   2. `DPkg::Lock::Timeout` tells apt to *wait* for the lock rather than
+    #      failing. It uses the same lock apt/dpkg do, so it is reliable where
+    #      `flock`/`rm` are not, and it covers every apt call in the job -- here
+    #      and in downstream actions (e.g. build-container installing podman).
```
hughcars: `apt-daily-upgrade.service` uses `KillMode=process` on Ubuntu, so the whole-cgroup guarantee is stronger than what systemd provides here. The timeout below still covers the practical dpkg contention. Consider narrowing this description to the invariant we actually enforce:

```suggestion
    # We prevent that by:
    #   1. Stopping the periodic APT units and their timers.
    #   2. Configuring later apt/apt-get package operations to wait for dpkg's
    #      locks instead of deleting lock files or failing immediately.
```

.github/actions/setup-runner/action.yml
```diff
@@ -55,30 +55,39 @@ runs:
...
-        sudo systemctl disable apt-daily-upgrade.timer || true
-        sudo systemctl stop apt-daily.service || true
-        sudo systemctl stop apt-daily-upgrade.service || true
-        
-        # Remove lock files if they exist
-        sudo rm -f /var/lib/dpkg/lock-frontend
-        sudo rm -f /var/lib/dpkg/lock
+        # Stop the periodic apt units and unattended-upgrades. `systemctl stop`
+        # waits for the cgroup to drain, so any in-flight apt-get is gone (not
+        # merely orphaned) by the time this returns.
```
hughcars: Same thing here: systemd does not strictly drain the upgrade service's child cgroup, while the timeout below is enough for the practical race. Consider avoiding the stronger cgroup-drain guarantee:

```suggestion
        # Stop the periodic APT units and ask unattended-upgrades to shut down
        # cleanly. Any remaining dpkg lock contention is handled by the timeout
        # below.
```

hughcars (approved):
Two documentation nits for you to consider, but functionally good to go.

PR #843, opened 2026-07-28
Centralize container publishing in a trusted workflow

Sbozzolo:
Container publishing now runs in a dedicated workflow, separate from the build. The build produces artifacts only; a workflow_run-triggered publish workflow (always loaded from main) authorizes the publish from GitHub-verified event data and holds the only AWS credentials, obtained via a least-privilege OIDC role.

- build-container: emits OCI/SIF + build-metadata artifacts; no AWS access.
- publish-containers (new): authorizes by event (forks excluded, tag/branch resolved via git-refs, PR label + head SHA checked), assumes the OIDC role, and publishes to ECR + S3.
- containers: workflow_dispatch is a publish request.

Note that a drawback of this approach is that making changes to the publishing code becomes harder, because the code that is run is always the one on main, so testing is not easy.

I tested it in my Palace fork (as much as I could).

Supersedes #832 
Closes #832

### DIFF
--- a/.github/publish_containers/authorize.py
+++ b/.github/publish_containers/authorize.py
@@ -110,17 +110,35 @@ class GitHubApi:
         self.repo = repo
 
     def _get(self, path: str) -> object | None:
+        """Decoded JSON body, or None if the resource does not exist (404).
+
+        Only an explicit 404 means "absent". Every other failure — auth, rate
+        limit, 5xx, transport, non-JSON body — raises, because silently reading
+        it as "absent" would turn an outage into an unauthorized publish: a
+        release could then finish green having published nothing.
+        """
+        # --include prints the status line and headers before the body, which is
+        # the only way to tell 404 apart from the rest (gh exits 1 for all of
+        # them).
         proc = subprocess.run(
-            ["gh", "api", path],
+            ["gh", "api", "--include", path],
             capture_output=True,
             text=True,
         )
-        if proc.returncode != 0:
-            return None  # 404 / not found → treat as absent, never as an error blob
-        try:
-            return json.loads(proc.stdout)
-        except json.JSONDecodeError:
+        headers, _, body = proc.stdout.replace("\r\n", "\n").partition("\n\n")
+        match = re.match(r"^HTTP/\S+ ([0-9]{3})(?:\s|\Z)", headers)
+        if match is None:
+            detail = proc.stderr.strip() or "missing HTTP status"
+            raise RuntimeError(f"GitHub API request failed for {path}: {detail}")
+        status = int(match.group(1))
+        if status == 404:
             return None
+        if proc.returncode != 0 or not 200 <= status < 300:
+            raise RuntimeError(f"GitHub API request failed for {path}: HTTP {status}")
+        try:
+            return json.loads(body)
+        except json.JSONDecodeError as exc:
+            raise RuntimeError(f"GitHub API returned invalid JSON for {path}") from exc
 
     def ref_sha(self, kind: str, name: str) -> str | None:
         """Commit sha a ref points at, or None. Peels annotated tags.
--- a/.github/publish_containers/authorize.py
+++ b/.github/publish_containers/authorize.py
@@ -1,0 +1,282 @@
+#!/usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Authorization decision for the trusted container publish workflow.
+
+The publish workflow runs from the default branch (via ``workflow_run``), so
+this code — not the PR-controlled build — decides whether a build may publish
+and to which selector. The decision uses only trusted inputs: the event facts
+GitHub reports for the build run, plus live lookups against the GitHub API. The
+build passes no authorization data of its own.
+
+Two entry points, one decision function:
+
+- default: emit ``authorized=`` / ``selector=`` to ``$GITHUB_OUTPUT``.
+- ``--reverify EXPECTED``: recompute the decision immediately before the writes
+  (TOCTOU guard) and fail unless it still authorizes the same selector.
+
+A build is authorized to publish only when, checked against the live GitHub
+API:
+- it did NOT come from a fork (head repo == base repo); and
+- the triggering event is one of:
+  - push to `main`, and `main` still points at the built commit
+    -> publishes the `main` channel;
+  - push of a `vX.Y.Z` release tag that still points at the built commit
+    -> publishes the `X.Y.Z` channel;
+  - manual `workflow_dispatch` on a branch that still points at the built
+    commit -> publishes `dev-<branch>`;
+  - a same-repo `pull_request` whose head (matched by sha AND ref) is an OPEN
+    PR currently carrying the `push-containers` label -> publishes
+    `dev-<branch>`.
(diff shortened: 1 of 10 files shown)


docs/src/developer/notes.md
```diff
@@ -492,6 +492,56 @@ spack spec local.palace@develop
...
+
+Release tags (`vX.Y.Z`) build the full microarchitecture matrix instead, so
+a release publishes a tuned image for each target:
+
+  - `x86_64_v3`, `x86_64_v4`, `sapphirerapids` (x64)
+  - `aarch64`, `neoverse_v1`, `neoverse_v2` (arm64)
+
+### Publishing a `dev-<branch>` prototype from a branch
+
+To publish a `dev-<branch>` container, just trigger the `Containers` workflow.
```
hughcars: Could we document the branch-name restriction here? `DEV_BRANCH_RE` only accepts the fairly narrow `prefix/name` shape below. Otherwise someone can follow these instructions, wait for the build to finish successfully, and still get no container because publication rejects the branch name.

```suggestion
To publish a `dev-<branch>` container, trigger the `Containers` workflow. The
branch must have the form `prefix/name`: the prefix may contain only letters,
numbers, `_`, or `.`, and the name must start alphanumeric and may additionally
contain `_`, `.`, or `-`. Exactly one `/` is supported.
```

.github/workflows/containers.yml
```diff
@@ -73,10 +76,13 @@ jobs:
         arch: [arm64, x64]
     runs-on: [self-hosted, '${{ matrix.arch }}']
     steps:
+      # A manual dispatch bypasses the paths-filter: dispatching is an explicit
+      # publish request, so it must build regardless of what changed. (Release
+      # tags are handled by build-containers-release, which has no filter.)
       - uses: actions/checkout@v6
-        if: needs.filter.outputs.test == 'true'
+        if: needs.filter.outputs.test == 'true' || github.event_name == 'workflow_dispatch'
```
hughcars: I think there is a race here. A docs-only push to `main` still makes the `Containers` workflow succeed and triggers this publisher. Suppose A is a real container change and B is docs-only: if A's publisher is still running, B enters the same `publish-containers-main` group and cancels A before discovering that B has no artifacts. If they do not overlap, A can instead fail the ref check after B advances `main`. Either way, A never gets published and the rolling channel stays stale until another container change lands.

The same thing can already happen at the build level because a newer `main` run can cancel an active build. Could we consider building on every `main` push? That keeps the exact-SHA policy, closes both windows, and makes an actual build failure show up as a red `Containers` run. Together with the matching change on the build action below:

```suggestion
        if: needs.filter.outputs.test == 'true' || github.event_name == 'workflow_dispatch' || github.ref == 'refs/heads/main'
```

This does spend two cached native builds on docs-only pushes. If that is too expensive, the other option would be to define `main` as the latest successfully published ancestor and attach an explicit GitHub status to the source SHA.

.github/workflows/containers.yml
```diff
@@ -73,10 +76,13 @@ jobs:
...
       - uses: ./.github/actions/build-container
-        if: needs.filter.outputs.test == 'true'
+        if: needs.filter.outputs.test == 'true' || github.event_name == 'workflow_dispatch'
```
hughcars: Same change here; otherwise the checkout runs but the actual build still gets skipped.

```suggestion
        if: needs.filter.outputs.test == 'true' || github.event_name == 'workflow_dispatch' || github.ref == 'refs/heads/main'
```

.github/workflows/publish-containers.yml
```diff
@@ -0,0 +1,154 @@
...
+    needs: authorize
+    if: needs.authorize.outputs.authorized == 'true'
+    runs-on: ubuntu-latest
+    # Concurrency is keyed on the authorized selector (only known now), so a
+    # newer publish for the same channel cancels an older in-flight one, while an
+    # unauthorized run — which never reaches this job — can never enter the
+    # group. Releases use distinct selectors, so they never cancel each other.
+    concurrency:
+      group: publish-containers-${{ needs.authorize.outputs.selector }}
+      cancel-in-progress: true
```
hughcars: I don't think we want to cancel a publisher halfway through updating the ECR tags and S3 objects, since that can leave one channel containing artifacts from different commits. Could we serialize publishers for each selector instead?

```suggestion
      cancel-in-progress: false
```

The live ref check below prevents a queued stale publisher from overwriting newer results. If the older publisher has already started, the newer one runs afterward and becomes final. This complements the `main` fix above rather than replacing it.

.github/publish_containers/authorize.py
```diff
@@ -0,0 +1,282 @@
...
+
+class GitHubApi:
+    """Live GitHub API access via the ``gh`` CLI (authenticated by GH_TOKEN)."""
+
+    def __init__(self, repo: str) -> None:
+        self.repo = repo
+
+    def _get(self, path: str) -> object | None:
+        proc = subprocess.run(
+            ["gh", "api", path],
```
hughcars: Could we distinguish a missing ref from an actual `gh api` failure here? At the moment a 401/403, rate limit, network error, 5xx, or malformed response all become normal unauthorization, so even a release publication can exit successfully without publishing anything. One way to expose the HTTP status is:

```suggestion
            ["gh", "api", "--include", path],
```

.github/publish_containers/authorize.py
```diff
@@ -0,0 +1,282 @@
...
+            capture_output=True,
+            text=True,
+        )
+        if proc.returncode != 0:
+            return None  # 404 / not found → treat as absent, never as an error blob
+        try:
+            return json.loads(proc.stdout)
+        except json.JSONDecodeError:
+            return None
```
hughcars: Then we can treat only an explicit 404 as "not found" and let operational or decoding failures fail the job:

```suggestion
        headers, _, body = proc.stdout.replace("\r\n", "\n").partition("\n\n")
        match = re.match(r"^HTTP/\S+ ([0-9]{3})(?:\s|$)", headers)
        if match is None:
            detail = proc.stderr.strip() or "missing HTTP status"
            raise RuntimeError(f"GitHub API request failed for {path}: {detail}")
        status = int(match.group(1))
        if status == 404:
            return None
        if proc.returncode != 0 or not 200 <= status < 300:
            raise RuntimeError(f"GitHub API request failed for {path}: HTTP {status}")
        try:
            return json.loads(body)
        except json.JSONDecodeError as exc:
            raise RuntimeError(f"GitHub API returned invalid JSON for {path}") from exc
```

I would also add cases for a real 404, 403/429/5xx, a transport failure, and a non-JSON success at the subprocess boundary. I tested the combined change against the existing suite plus those four cases (56 tests total).

hughcars (commented):
Overall, I think this is a good way to split building from publishing. I assume the AWS changes will happen at the same time: the build runners lose their old publish access and the OIDC role gets only the ECR and S3 permissions it needs. With that in place, I only found two things I think are worth addressing: the `main` publication race and the way GitHub API failures are handled. I also left a small documentation suggestion about the supported branch names.

.github/workflows/publish-containers.yml
Sbozzolo: This would remove the only automatic way to cancel a job that was stuck, so I added a timeout of 45 minutes

hughcars (approved):

PR #861, opened 2026-07-31
Support magnetostatic inductance extraction with mixed current/flux excitation

### LINKED ISSUE #821: Inductance extraction with mixed current-flux ports
The magnetostatic solver currently supports inductance extraction for current-only and flux-only excitations. It would be helpful to be able to extract inductance when there are both current and flux sources.

dnpham23:
Adds support for magnetostatic inductance extraction with combined `SurfaceCurrent` and
`FluxLoop` excitations, which was previously rejected by config validation, and reports a
single inductance matrix in henries covering both port types.

With both excitation types present the normalized cross-energies form a hybrid
(mixed impedance/admittance) matrix $H$ rather than an inductance matrix, because each column
is normalized by its own excitation: a current column carries units of H and a flux column
1/H. The two step types also impose complementary constraints. A current step holds every flux
loop at zero flux (the loops' PEC surfaces are in the base essential set), while a flux step
leaves every current port open ($I_c = 0$). That gives

```
H_cc = M_cc - M_cf*M_ff^-1*M_fc      (screened, a Schur complement)
H_ff = M_ff^-1                       (flux-loop reluctance)
```

The current-flux cross-energy, however, **vanishes identically**. Pairing a current state with
a flux state gives $E=\sum_c I_c\Phi_c + \sum_f I_f \Phi_f$, in which the first sum is zero because the
flux state carries no port current and the second is zero because the current state links no
loop flux. Both sums vanish term by term, so the two excitation types are energy-orthogonal
and `H_cf ≡ 0` for any geometry.

Importantly, the physical mutual inductance is **not** zero (it is 2.95e-14 H in the example
below); it is merely unobservable through cross-energies. So it has to be measured directly.

### How the coupling is recovered

During a flux excitation the current ports are open, so the response is $\Phi_c= M_{cf}M_{ff}^{-1}\Phi_f$. 
Integrating `B·n` over a surface spanning each current port's loop
therefore yields $G=M_{cf}M_{ff}^{-1}$ directly, and the blocks convert in closed form:

```
M_ff = H_ff^-1
M_cf = G*M_ff
M_cc = H_cc + G*M_ff*G^T
```

The last term un-screens the current-port block; substituting $G$ shows it is exactly the Schur
term removed above. Only $H_{ff}$ is inverted, exactly as in the existing flux-only path, so this
introduces no new linear algebra or failure modes.

The loop-spanning surface cannot be inferred from a port's terminal attributes, so it is given
by a new optional field:

```json
"SurfaceCurrent":
[
  {
    "Index": 1,
    "Attributes": [4],
    "Direction": [0.0, 1.0, 0.0],
    "ApertureAttributes": [5, 6, 7]
  }
]
```

By Stokes' theorem any surface spanning the same loop gives the same answer, so the choice is
not unique. If a port has no `ApertureAttributes`, its mutual inductances are reported `NaN`,
its self-inductance is left screened rather than reported as a bare value, `Minv`/`Mm` are
not computed, and a warning names the port and states both consequences.

### Also included

Fixes a NaN-detection bug in `InvertOrMarkSingular`. It used `DenseMatrix::MaxMaxNorm`, whose
running maximum never updates on a NaN comparison (`if (norm < abs_entry)` is false for NaN),
so it reports a finite norm for a matrix containing NaNs. Entries are now scanned explicitly
with `std::isfinite`. This is what initially caused the no-aperture warning above to never
fire.

`ComputeFluxThroughSurface` is extracted from `VerifyFluxThroughHoles` in
`surfacecurlsolver.cpp` so that the flux verification and the new measurement share one code
path; it takes a set of attributes rather than a single hole.

### Validation

New example `examples/ring_disk/`: a wire ring (radius 100 μm) encircling a PEC disk
(radius 50 μm) with a 10 μm hole at its center. The ring is the current port, the hole the flux
loop. Three configs share the mesh: mixed, current-only, and flux-only, the latter two serving
as regression baselines.

```
M    = [[+7.009202666798e-10, +2.950908491389e-14],
        [+2.950908491389e-14, +1.921248752285e-11]]
```

| Check | Result |
|---|---|
| Regression, current-only and flux-only | both unchanged (7.009202213570e-10, 1.921248723948e-11) |
| Solver tolerance 1e-8 -> 1e-11 | `M_cf` stable to 8 significant figures |
| Aperture `[5,6,7]` vs `[7]` (Stokes) | identical to 12 digits |
| `M_cf` vs. independent hand computation | agrees to 1.7e-7 |
| `M_cc` vs. current-only run | agrees to 13 digits; un-screening correction 4.53e-17 H |
| Independent path `L = Φ_gap/I` vs. cross-energy | 1.2e-4 |
| `Minv` multiplied out, `Mm` row sums | identity to ~1e-13; row sums exact |
| No-aperture fallback | `M_cf` NaN, `M_cc` screened (not poisoned), warning fires |
| All example configs vs. bumped schema | validate |

The tolerance sweep is the load-bearing check: the cross-energy route swings four orders of
magnitude and flips sign as the tolerance tightens (the signature of an analytically zero
quantity), whereas the measured value is stable to 8 figures. The aperture-independence check
verifies orientation and integration against a theorem rather than against another Palace run.

### Test coverage
`ring_disk` is registered as a regression case (`ring_disk_mixed_current_flux`,
`[Serial][Parallel][GPU][Regression]`).

### Scope

The zero-flux clamp during current steps is the physically correct model for a closed
superconducting loop, which conserves fluxoid and sources whatever persistent current holds its
flux fixed (equivalently, a normal conductor as frequency -> infinity). A normal-metal loop at DC is a different
situation: it does not screen at all, so it should be modeled as a `SurfaceCurrent` port on
non-PEC material rather than as a flux loop, and needs no new mode. Finite-frequency eddy screening is
outside the magnetostatic solver's scope.


palace/drivers/magnetostaticsolver.cpp
```diff
@@ -200,6 +281,22 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+      // what makes the current-flux mutual inductance recoverable.
+      int col = 0;
+      for (const auto &[j_idx, j_data] : curlcurl_op.GetSurfaceCurrentOp())
+      {
+        const auto &aperture = iodata.boundaries.current.at(j_idx).aperture_attributes;
+        if (!aperture.empty())
+        {
+          linked_flux(col, flux_idx) = ComputeFluxThroughSurface(
+              B_gf, aperture, curlcurl_op.GetMesh(), curlcurl_op.GetMaterialOp(),
+              flux_direction, curlcurl_op.GetComm());
```
hughcars: Use the current port's aperture polarity here, not `flux_direction`. In direction-based mode, `BdrSurfaceFluxCoefficient` evaluates `B · flux_direction`, and this argument belongs to the excited flux loop. Reversing `SurfaceCurrent.Direction` therefore leaves `M_cf` unchanged; reversing `FluxLoop.Direction` reverses both `B` and the projection, so it also leaves `M_cf` unchanged. The reported mutual cannot follow either port's polarity, and nonparallel or curved apertures get the wrong magnitude. Give each aperture an orientation tied to its current-port polarity, and add regressions that reverse each port independently.

hughcars: `SurfaceCurrentOperator` divides the unit current by `data.elems.size()` because these are parallel elements of one port. This sums one flat aperture union without the same weighting: two identical elements produce twice the correct `M_cf` and four times the correction to `M_cc`. The full fix needs one weighted aperture per element. If that is outside this PR, reject the unsupported case rather than return the wrong matrix.

```suggestion
        const auto &aperture = iodata.boundaries.current.at(j_idx).aperture_attributes;
        if (!aperture.empty())
        {
          MFEM_VERIFY(
              j_data.elems.size() == 1,
              "Mixed SurfaceCurrent/FluxLoop inductance extraction does not yet support "
              "multielement SurfaceCurrent ports!");
          linked_flux(col, flux_idx) = ComputeFluxThroughSurface(
              B_gf, aperture, curlcurl_op.GetMesh(), curlcurl_op.GetMaterialOp(),
              flux_direction, curlcurl_op.GetComm());
```

palace/models/surfacecurlsolver.cpp
```diff
@@ -359,6 +360,67 @@ void SolveSurfaceCurlProblem(const SurfaceFluxData &flux_data, const IoData &iod
...
+                                 const std::vector<int> &attributes, const Mesh &mesh,
+                                 const MaterialOperator &mat_op,
+                                 const mfem::Vector &flux_direction, MPI_Comm comm)
+{
+  if (attributes.empty())
+  {
+    return 0.0;
+  }
+
+  // Ensure face neighbor data is available for shared boundary evaluation.
```
hughcars, line 374: Validate every aperture attribute against the final mesh. The loop below silently skips an unknown value, so a typo such as `ApertureAttributes: [999]` becomes a finite zero mutual and an uncorrected current block. This belongs in boundary setup; a guard here is the smallest safe fix.

```suggestion
  const auto &bdr_attributes = mesh.Get().bdr_attributes;
  for (auto attr : attributes)
  {
    MFEM_VERIFY(bdr_attributes.Find(attr) >= 0,
                "Unknown flux surface boundary attribute " << attr << "!");
  }

  // Ensure face neighbor data is available for shared boundary evaluation.
```

palace/drivers/magnetostaticsolver.cpp
```diff
@@ -321,6 +443,134 @@ void MagnetostaticSolver::PostprocessTerminals(
...
+        {
+          correction += M(i, n_current + b) * G(j, b);
+        }
+        // A port with no aperture yields a NaN correction; report the screened H_cc rather
+        // than poisoning the whole current block.
+        if (!std::isfinite(correction))
+        {
+          correction = 0.0;
+        }
+        M(i, j) = M(j, i) = flux_invertible ? H(i, j) + correction : nan;
```
hughcars: `H_cc(i,j)` is a screened Schur-complement entry when either aperture row is missing, not the bare mutual advertised by `terminal-M.csv`. Keep the documented screened fallback for the affected diagonal only; the corresponding off-diagonals are unknown and should remain `NaN`.

```suggestion
        // A missing aperture leaves only the documented screened diagonal. The
        // corresponding off-diagonals are not bare mutual inductances.
        if (!std::isfinite(correction))
        {
          if (i != j)
          {
            M(i, j) = M(j, i) = nan;
            continue;
          }
          correction = 0.0;
        }
        M(i, j) = M(j, i) = flux_invertible ? H(i, j) + correction : nan;
```

hughcars (changes requested):
Mixed current/flux extraction needs four corrections before merge: current-port aperture polarity, per-element weighting, final-mesh attribute validation, and a complete bare-matrix output contract.

I implemented these changes on [`hughcars/mixed-current-flux-inductance`](https://github.com/awslabs/palace/tree/hughcars/mixed-current-flux-inductance) at [`f0827b4848`](https://github.com/awslabs/palace/commit/f0827b48484b769cad241d577a7cd636bc1889dd). The branch replaces port-level `ApertureAttributes` with an element-owned `Aperture { Attributes, Direction }`, uses `Direction` to orient the geometric $B\cdot n$ flux, applies the same element weights as current assembly, validates aperture attributes against the final mesh, requires mixed current ports to remain open, and removes the partial screened-matrix fallback.

The Spack build and full serial unit suite pass. `ring_disk_mixed_current_flux` passes in serial and with two MPI ranks. The remaining coverage gap is a full-solve multielement or curved-aperture regression.

dnpham23:
Thanks @hughcars, for the redesign and the reference branch! The element-owned oriented `Aperture` is a cleaner model than the port-level `ApertureAttributes` I had, and orienting the geometric normal (rather than projecting onto the reference direction) is a nice fix for polarity. I've merged your `hughcars/mixed-current-flux-inductance` branch, so all four corrections are in. Just made a few small changes on top of that.

hughcars (approved):

PR #863, opened 2026-08-03
Serialize Spack buildcache index writers (peel of 829)

Sbozzolo:
I extracted a non-controversial change from #829 (the rest of it is more invasive).

The other two parts of 829 are (1) long-runs on forks, (2) prevent publication of buildcache from forks. Part (2) is already clean on main: forks get a read-only token and spack fails gracefully when it cannot publish to the buildcache. I am working on (1) in a different PR.

For this PR:

`buildcache update-index` is a read-modify-write of the mirror's single `index.spack` tag: it lists every spec tag, rebuilds the index, and overwrites. Individual spec tags are immutable and safe to push concurrently, but two concurrent index writers are last-writer-wins, and specs banked by the loser can drop out of the index.

The workflow-level concurrency groups are keyed on github.ref, so they serialize runs within a branch or PR but not across them. The index is one shared object per mirror, so two different PRs -- or a PR and a push to main -- could refresh it at the same time.

Give each mirror's refresh a job-level concurrency group that deliberately omits github.ref, with cancel-in-progress: false so a second writer queues instead of clobbering. In lookahead.yml this requires promoting the refresh step to its own job, since steps cannot carry `concurrency`; it rebuilds a minimal mirror-only Spack environment the same way spack.yml already does. Its previous comment claimed no writer could race it, which held for jobs (nothing else builds against Spack develop) but not for concurrent runs.

### DIFF
--- a/.github/workflows/lookahead.yml
+++ b/.github/workflows/lookahead.yml
@@ -123,6 +123,7 @@ jobs:
     concurrency:
       group: palace-develop-buildcache-index
       cancel-in-progress: false
+      queue: max
     steps:
       - name: Install Spack develop
         uses: actions/checkout@v6
--- a/.github/workflows/lookahead.yml
+++ b/.github/workflows/lookahead.yml
@@ -70,11 +70,11 @@ jobs:
           run-regression-tests: 'true'
           use-develop-deps: 'true'
 
-  # Build against the tip of Spack (develop) rather than the pinned 1.2 release
+  # Build against the tip of Spack (develop) rather than the pinned release
   # series, to catch upstream Spack/spack-packages changes that break Palace.
   # Dependencies stay at their pinned versions so a failure points at Spack.
   # This job writes only to the isolated palace-develop buildcache, never the
-  # pinned Spack 1.2 cache used by required CI.
+  # pinned Spack cache used by required CI.
   build-test-x64-gcc-spack-develop:
     needs: filter
     if: needs.filter.outputs.test == 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/lookahead.yml
```diff
@@ -90,20 +90,68 @@ jobs:
...
+        github.event_name != 'pull_request' ||
+        github.event.pull_request.head.repo.full_name == github.repository
+      )
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    concurrency:
+      group: palace-develop-buildcache-index
+      cancel-in-progress: false
```
hughcars, line 125: `cancel-in-progress: false` protects the running writer, but GitHub only keeps one pending job by default. A third refresh replaces the second; if the outer workflow then cancels it, the index can remain stale.

```suggestion
      cancel-in-progress: false
      queue: max
```

.github/workflows/spack.yml
```diff
@@ -171,6 +171,16 @@ jobs:
...
+    # Serialize index writers across all refs. Individual spec tags are
+    # immutable and may be pushed concurrently, but `update-index` is a
+    # read-modify-write of the mirror's single `index.spack` tag, so two
+    # concurrent writers are last-writer-wins and the loser's specs can drop
+    # out of the index. The workflow-level group is keyed on github.ref and so
+    # does not serialize across branches/PRs; this group deliberately omits it.
+    # cancel-in-progress: false queues the newer writer instead of clobbering.
+    concurrency:
+      group: palace-1.2-buildcache-index
+      cancel-in-progress: false
```
hughcars, line 183: Same here for the 1.2 mirror.

```suggestion
      cancel-in-progress: false
      queue: max
```

hughcars (approved):
Looks good. The per-mirror concurrency groups are the right scope, and both refresh jobs passed in CI. One non-blocking queueing edge case inline.

Issue #862, opened 2026-08-01
Support frequency-dependent materials for 3D eigenmode simulations.


WEIXUEP:
Hi, 

I want to use frequency-dependent materials in 3D eigen mode simulation, the material may be defined as follows:
```
"PermittivityEqn": "2.08 + 0.0287629*ln((2.53303e22 + f*f)/(8.22935e9 + f*f))",
"ConductivityEqn": "1e-12 + 3.20031e-12*f*(atan(f/90715.7) - atan(f/1.59155e11))"
```

The variable `f` denotes frequency in Hz, consistent with the unit convention adopted in Ansys HFSS.

In most engineering scenarios, only permittivity and conductivity exhibit frequency dependence. For this iteration, only PermittivityEqn and ConductivityEqn are enabled. These equation-based definitions are mutually exclusive with the original fixed constant material parameters to avoid ambiguous configuration conflicts.

I have already implemented this feature in https://github.com/WEIXUEP/palace/tree/feature/permittivity-conductivity-eqn, and I’m wondering if this can be merged.

hughcars:
Hi @WEIXUEP, 

Thank you for your interest in palace, and interest in contributing to palace. That sounds like a feature that could be of interest to the broader user base. If you would like to open a PR we can start the process for reviewing and iterating on it.

PR #864, opened 2026-08-04
Support frequency-dependent materials for 3D eigenmode simulations.  

WEIXUEP:
See issue https://github.com/awslabs/palace/issues/862.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 
 #### New Features
 
+  - Added frequency-dependent scalar material permittivity and electrical conductivity
+    expressions for eigenmode simulations via `PermittivityEqn` and `ConductivityEqn`.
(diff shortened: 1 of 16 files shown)


hughcars:
Hi @WEIXUEP,

We won't be accepting this PR at this time as there are a few issues: 1) we do not want a string parser in the config reading apparatus, as it will be difficult from a maintenance perspective, and 2) a truly robust and general implementation of it is very complicated to maintain. Given the particular model you are asking to implement is actually a fairly recognized one, I actually went ahead and just implemented that directly as a nonlinear material model in https://github.com/awslabs/palace/pull/877. Doing it in that fashion also allows us to enforce that only well-posed models are being allowed in: with a general interface we would face the prospect of users asking why their particular model does not work, when really it was their particular model parameterization that failed in obscure and difficult to express fashion. So for the sake of supporting our highest goal of model performance, whilst maintaining usability, we will not be accepting this PR, and I invite you to comment on https://github.com/awslabs/palace/pull/877 instead if there are other models you would like in addition (within reason).

PR #865, opened 2026-08-05
Add layer caching to container workflow and support for specifying a spack-package commit

### LINKED ISSUE #847: Update spack recipe to use libxsmm releases
Palace was tracking `libxsmm@main` because the project hadn't made a release in 3 yars. However, libxsmm has started releasing versions again, so we should probably switch to pinning to a release.

Requires a release of spack-packages after 2026.06

Sbozzolo:
We often need newer versions of `spack-packages` because of fixes in recipes. Unfortunately, the spack container system does not allow specifying a custom sha for `spack-packages`, so we need to use directly a base container image (ubuntu) and manually install spack/spack-packages. This adds a little bit of additional cost so that the workflow takes ~ 10-12 minutes. To mitigate that, I also support for caching the base layer with all the palace dependencies. This way, when only palace needs to be recompiled (ie, dependencies have not changed), the entire workflow completes in less than 5 minutes.

This will be important in an upcoming PR that adds GPU containers, which are much more expensive to build, even when the spack cache is warm.

Closes #847


.github/actions/build-container/spack_env/PalaceDockerfile
```diff
@@ -1,31 +1,27 @@
...
-ARG GITHUB_TOKEN=""
+ARG SPACK_PACKAGES_COMMIT
 
 # FIXME(awslabs/palace#731): workaround for a Spack spec-groups bug that
 # prevents the gcc compiler from being installed. Build-stage only, so the
 # runtime image is unaffected. Remove with the spack.yaml `compilers` group
 # fix.
 #
-# noble's default archive does not carry gcc-16 yet; pull it from the
+# noble's default archive does not carry gcc-15 yet; pull it from the
```
Sbozzolo, line 10: I moved to gcc 15 because CUDA does not support GCC 16 yet

.github/actions/build-container/spack_env/spack.yaml
```diff
@@ -14,6 +14,9 @@ spack:
     targets:         # Compile for the local microarchitecture
       granularity: microarchitectures
   config:       
+    # Some upstream source mirrors are selected through redirects and can fail
+    # transiently. Let curl retry 5xx responses before failing the full build.
+    url_fetch_method: "curl --retry 10 --retry-delay 2 --retry-max-time 120"
```
hughcars, line 19: Nice! Didn't know this exists, but this was very frustrating in the past.

.github/actions/build-container/action.yml
```diff
@@ -2,6 +2,10 @@ name: Build Palace Containers
 description: Build a Palace container image from the stored Spack environment
 
 inputs:
+  spack-packages-commit:
+    description: 'spack-packages commit used by the builtin Spack repository'
+    required: false
+    default: '3dcad5bb860184c91c17b97a9f3b13252d26d309'
```
hughcars: Does this pin the global recipes? I assume this is partly to avoid the cache constantly missing/for stability?

hughcars (approved):

.github/actions/build-container/action.yml
Sbozzolo: Yes, exactly, for those two reasons + being able to move spack-packages to versions beyond the ones released.

PR #866, opened 2026-08-05
Add multi-architecture CUDA container build

Sbozzolo:
This PR adds a workflow to compile CUDA-enabled containers. 

Containers are built for x86_64_v3 and various CUDA architectures in a single executable. I could not build for targets newer than sm_90 because the job would take forever (probably the same problem as https://github.com/mfem/mfem/issues/5363).

CUDA containers are huge (> 4 GB), so I spent considerable effort slimming them down. This is also needed to make CI faster by reducing upload time. 

I upstreamed all the PRs:
- https://github.com/spack/spack-packages/pull/5888
- https://github.com/spack/spack-packages/pull/5889
- https://github.com/spack/spack-packages/pull/5890
- https://github.com/spack/spack-packages/pull/5891
- https://github.com/spack/spack-packages/pull/5892
- https://github.com/spack/spack-packages/pull/5894
- https://github.com/spack/spack-packages/pull/5895
- https://github.com/mfem/mfem/pull/5440
- https://github.com/CEED/libCEED/pull/2001


.github/actions/build-container/action.yml
```diff
@@ -27,6 +27,22 @@ runs:
...
+        # architectures themselves are fixed in spack.yaml (cuda_constraints);
+        # this step only selects which profile spack.yaml concretizes. Exported
+        # via $GITHUB_ENV so every later step (and the podman build-arg) sees it.
+        if nvidia-smi >/dev/null 2>&1; then
+          echo "PALACE_CUDA=1" >> "$GITHUB_ENV"
+          echo "NVIDIA GPU detected: building the CUDA profile."
+        else
+          echo "PALACE_CUDA=" >> "$GITHUB_ENV"
+          echo "No NVIDIA GPU detected: building CPU-only."
+        fi
```
hughcars: We already know from the matrix whether this is meant to be a GPU build. Could we use that here, and fail if `nvidia-smi` doesn't work on the GPU runner? Otherwise a broken GPU environment silently turns this into a CPU build.

```suggestion
    - name: Select and validate build profile
      shell: bash
      run: |
        if [[ "${{ matrix.runner }}" == "gpu" ]]; then
          if ! nvidia-smi >/dev/null 2>&1; then
            echo "::error::GPU runner selected, but nvidia-smi failed."
            exit 1
          fi
          echo "PALACE_CUDA=1" >> "$GITHUB_ENV"
          echo "NVIDIA GPU detected: building the CUDA profile."
        else
          echo "PALACE_CUDA=" >> "$GITHUB_ENV"
          echo "Building the CPU-only profile."
        fi
```

hughcars (commented):
Really nice extension, one small plausible failure mode I think can be fixed, then good to go.

.github/actions/build-container/action.yml
Sbozzolo: Applied, with the small difference that I also match `gpu-4xlarge`

hughcars (approved):

PR #868, opened 2026-08-05
Fetch MFEM patches from immutable URLs

Sbozzolo:
Spack maintainers seem to prefer we do not host patches when they are directly available online. This also removes the need to maintain links locally and reduces potential problems in compiling with spack locally.

This is quite nice for older versions of Palace, so that we don't have to keep files around when patches get merged.


dnpham23 (commented):
Nice cleanup. Just one question/comment below

spack_repo/local/packages/palace/package.py
```diff
@@ -199,19 +199,60 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
+                    sha256="530532da3ae8815d004bb6ce19f6f08a1248c3d585503551c90d0eeae7fb3f87",
+                ),
+                # https://github.com/mfem/mfem/pull/5124
+                patch(
+                    "https://raw.githubusercontent.com/awslabs/palace/"
+                    "b22f654ab36fe01f1f3176349c60626efed1a6a2/extern/patch/mfem/"
+                    "mfem_pr5124_cudss.diff",
+                    sha256="d0b5893ec7925cbc8a70cc5eba2abe037754b99b5bb366c4bae90f0583d22290",
+                    when="@:4.9 +cudss",
+                ),
```
dnpham23, line 254: Non-blocking: since the four `raw.githubusercontent.com` URLs pin b22f654ab, `extern/patch/mfem/ ` now doubles as an immutable content host for spack while remaining the live source for the CMake build.  If someone later re-curates one of those snapshots for CMake, spack would keep fetching the old pinned bytes and the two builds would diverge silently. Maybe add a short note in `extern/patch/mfem/` or a comment near the pin, so the pin and sha256 need updating when a snapshot is regenerated?

Sbozzolo: Good catch. I changed the superbuild to use the same patch-downloading mechanism.

cmake/ExternalMFEM.cmake
```diff
@@ -401,15 +401,67 @@ endif()
...
+)
+# https://github.com/mfem/mfem/pull/4983
+download_mfem_patch(
+  mfem_pr4983.diff
+  "https://raw.githubusercontent.com/awslabs/palace/b22f654ab36fe01f1f3176349c60626efed1a6a2/extern/patch/mfem/mfem_pr4983.diff"
+  530532da3ae8815d004bb6ce19f6f08a1248c3d585503551c90d0eeae7fb3f87
+)
+if(PALACE_WITH_CUDSS)
+  # https://github.com/mfem/mfem/pull/5124
+  # https://github.com/mfem/mfem/pull/5389
```
hughcars: 5389 isn't here?

hughcars (approved):
Nice fix. One comment debris I think.

PR #873, opened 2026-08-07
Support long tests for fork pull requests

Sbozzolo:
Simplify long-tests to support running from forks.

Previously, long tests had three states:
- pending,
- success,
- failure

I did this so that developers could have a PR open with "pending long tests", opt in only when they were ready, but prevent merging the PR unless the developer added the "no-long-tests" label or if the tests passed (to certify the intent).

Maintaining this while supporting forks adds substantial complexity (see, e.g., #829). If we drop the "pending" state, everything becomes much simpler.

Without pending, long tests is always either success of failure. It is success if:
- there is no "trigger-long-run" label
- the long tests success

It if failure if there is the "trigger-long-run" label AND tests fail.

So, long-runs become truly opt-in and there is no longer a requirement to specify the intent with one of the two labels.

Closes #829


.github/workflows/long-tests.yml
```diff
@@ -1,79 +1,93 @@
-# This workflow runs long tests on a PR when the 'trigger-long-tests' label is
-# added. Long tests are a required check for merging non-trivial PRs.
+# Runs long, expensive tests on self-hosted large runners.
```
hughcars, line 4:
```suggestion
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0

# Runs long, expensive tests on self-hosted large runners.
```

.github/workflows/long-tests.yml
```diff
@@ -1,79 +1,93 @@
...
+# The `long-tests` job succeeds either when testing is deferred (there is no
+# `trigger-long-tests` label), or after the tests pass. New commits are tested
+# automatically while the label remains present.
 name: Long Tests
 
 on:
-  workflow_dispatch:
   pull_request:
-    types: [labeled, synchronize]
+    types: [labeled, opened, synchronize]
```
hughcars, line 13: Since `unlabeled` isn't a trigger, removing the label after a failure leaves the red check on that SHA until the next push. Good, maybe worth a sentence in `testing.md`.

hughcars (approved):
Job results instead of API-written statuses is the right fix for forks, and the ignored-event concurrency group avoids clobbering a running result. Rebase then good to go.

PR #874, opened 2026-08-07
Reject all-zero rational impedance polynomials in schema

### LINKED ISSUE #834: Schema: require a nonzero coefficient in `RationalImpedance` polynomial arrays
The schema only enforces `minItems: 1` on `Numerator`/`Denominator`, so `validate-config` accepts an all-zero polynomial (e.g. `"Numerator": [0.0, 0.0]`) that the config parser then rejects at solve time. Add to both arrays in `$defs/RationalImpedance`:

```json
"contains": { "not": { "const": 0 } }
```

The embedded nlohmann-json-schema-validator supports `contains`: rejects `[0.0, 0.0]` and `[0, 0]`, accepts any array with a nonzero entry. Needs a SchemaVer patch bump. Follow-up to #770, punted from #792.

danielgaskins:
## Description

Reject all-zero `Numerator` and `Denominator` arrays for `RationalImpedance` during JSON schema validation. This aligns early configuration validation with the existing runtime checks and gives users feedback before a solve starts.

The schema version is bumped from `1-5-0` to `1-5-1`, and the schema descriptions now state that each polynomial requires at least one nonzero coefficient.

## Testing

- Added schema unit coverage for:
  - a valid polynomial containing internal zero coefficients
  - all-zero integer and floating-point numerator arrays
  - all-zero integer and floating-point denominator arrays
- Validated all 45 example configurations against the updated Draft-07 schema
- Ran `check-schema-version`, `check-changelog-schema-versions`, `check-test-tags`, and `git diff --check`

The compiled C++ suite was not run locally because this checkout has no configured Palace build; CI will run the authoritative unit test.

Closes #834

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,9 +70,9 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 #### Bug Fixes
 
   - Reject all-zero numerator or denominator polynomial coefficients in
-    `config["Boundaries"]["RationalImpedance"]` during schema validation, instead of
-    accepting the configuration only to reject it when parsing the solve. SchemaVer 1-5-1
-    [PR 874](https://github.com/awslabs/palace/pull/874).
+    `config["Boundaries"]["RationalImpedance"]` during schema validation, matching the
+    existing checks in the configuration parser and providing users with earlier feedback.
+    SchemaVer 1-5-1 [PR 874](https://github.com/awslabs/palace/pull/874).
   - Fixed validation of domain material coverage so postprocessing and retained mesh
     attributes without a corresponding `config["Domains"]["Materials"]` entry are rejected
     instead of silently assigning zero material coefficients to retained volumes. [PR
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,10 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 
 #### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -69,6 +69,10 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 
 #### Bug Fixes
 
+  - Reject all-zero numerator or denominator polynomial coefficients in
```
hughcars, line 72: Could we frame this explicitly as aligning schema validation with an existing parser invariant? That makes clear no previously usable configuration is invalidated and avoids the slightly unclear phrase “parsing the solve.”

```suggestion
  - Reject all-zero numerator or denominator polynomial coefficients in
    `config["Boundaries"]["RationalImpedance"]` during schema validation, matching the
    existing checks in the configuration parser and providing users with earlier feedback.
    SchemaVer 1-5-1 [PR 874](https://github.com/awslabs/palace/pull/874).
```

hughcars (commented):
Great little fix, thank you. One tiny nit in the changelog entry.

CHANGELOG.md
danielgaskins: Updated with the suggested wording. Thanks for the clarification!

hughcars (approved):

hughcars:
Bypass merge due to issue running CI from forks.

danielgaskins:
Thanks for reviewing and merging this, Hugh. I appreciate the helpful feedback!

Issue #816, opened 2026-07-16
Fix long-tests status manager for fork PRs


hughcars:
The `Long Test Status Manager` workflow runs on `pull_request` and tries to set the synthetic `long-tests` status:

```bash
gh api "repos/$REPO/statuses/$HEAD_SHA" --method POST ...
```

For fork PRs, the `GITHUB_TOKEN` from `pull_request` does not get `statuses: write`, even though the workflow requests it, so this fails with:

```text
Resource not accessible by integration (HTTP 403)
```

Example: https://github.com/awslabs/palace/actions/runs/29533185814/job/87738407630

We should move the status-setting workflow to `pull_request_target`, since it only needs PR metadata / changed paths and should not run PR code.

Acceptance:
- fork PRs can set/update the `long-tests` status
- `trigger-long-tests` / `no-long-tests` behavior stays the same
- no PR-head code is executed from the `pull_request_target` workflow

Need to also sanity-check `long-tests.yml`, since it writes the same status after long tests complete.

Sbozzolo:
This is fixed

PR #875, opened 2026-08-07
Enforce license headers for Python and YAML files

### LINKED ISSUE #846: Amazon license header for Python files is not checked
The `check-license` action checks that every file comes with the correct license, but this is not applied to python or yaml files (e.g., GitHub actions).

danielgaskins:
## Description

Extend the license-header check to cover Python, YAML, and YML files. This closes a gap where new scripts and GitHub Actions could be added without the required license notice.

Backfill the Amazon Apache-2.0 header on existing first-party YAML files and rename the shared `#`-comment template from `jl.txt` to `hash.txt` now that it covers Julia, CMake, Python, and YAML. The vendored Spack recipe retains its existing upstream `(Apache-2.0 OR MIT)` header and is checked against a dedicated template rather than being excluded or relicensed.

## Testing

- Ran the exact `viperproject/check-license-header@v2` action implementation: 0 errors
- Ran the v1 local license-check CLI: passed
- Parsed every YAML/YML file with PyYAML
- Compiled all Python sources with `compileall`
- Ran all 64 `.github/publish_containers` unit tests
- Ran `git diff --check`

Closes #846


danielgaskins:
CI note: all substantive checks passed, including `check-license`, the Linux/macOS matrices, GPU tests, docs, and publishing tests. The sole failure is `Long Test Status Manager / handle-non-trivial-changes`: this fork PR is classified as non-trivial because it adds license-header comments to `long-test*.yml`, then the fork token cannot write the pending commit status (the known fork-permission problem tracked in #816, with #873 addressing it).

Since the workflow changes in this PR are comments only, could a maintainer please apply the `no-long-tests` label?

.github/license-check/spack.txt
```diff
@@ -0,0 +1,3 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
```
hughcars: I don't think we should have this, needing it feels like a bit of an anti-pattern. I suspect we probably need to change the notice on the package.py to match the amazon license instead.

hughcars (commented):
One change, don't have the spack.txt special license, change the license in package.py, that should be covered by the same license as other files.

.github/license-check/spack.txt
danielgaskins: Good point. I removed the separate `spack.txt` rule and changed `package.py` to use the repository’s Amazon/Apache header, so it is now covered by the normal Python/YAML license check.

hughcars (approved):
Nice QoL addition thanks!

danielgaskins:
Thanks for the approval. I checked the remaining container failures. Both the arm64 and x64 jobs received a 403 while uploading build-cache blobs to the project’s GHCR namespace. The other build and test jobs passed. Could a maintainer rerun these checks after restoring the registry permission?

hughcars:
> Thanks for the approval. I checked the remaining container failures. Both the arm64 and x64 jobs received a 403 while uploading build-cache blobs to the project’s GHCR namespace. The other build and test jobs passed. Could a maintainer rerun these checks after restoring the registry permission?

Hi, this is a side effect of a security measure we have in place, a fix will be around eventually but in the meantime I will make the merge manually.

PR #876, opened 2026-08-09
Avoid assembling exact-zero operator terms

hughcars:
Palace currently treats a coefficient as present when it has allocated storage, even if every stored value is zero. This can repeatedly apply fine-level libCEED terms that contribute nothing mathematically.

This PR omits exactly zero terms before assembling multigrid smoothing levels. Tiny nonzero values are unaffected.

The configured coarse structure is retained for symbolic reuse. This matters for `ComplexCoarseSolve`: a nonlinear seed with `λ = iω` has `Im(λ²) = 0`, but that full-domain imaginary mass term becomes nonzero once `Re(λ) != 0`, adding off-diagonal block structure. The same coarse assembly rule is used for real-approximate solves to keep the implementation independent of the selected sparse representation.

### Performance

Matched runs used the Order-4 eigenmode stress case with no AMR (`Model.Refinement.MaxIts = 0`), 8 MPI ranks on 8 A100 GPUs, `/gpu/cuda/magma`, exact-complex cuDSS, and GPU-aware MPI disabled. Configuration, mesh, dependencies, solver tolerances, output, and precision were identical. Main is the mean of three runs; this PR is one final-design run.

| Metric | Main | This PR | Change |
|---|---:|---:|---:|
| Total time | 922.5 s | 779.6 s | **-15.5%** |
| Preconditioner | 454.8 s | 314.1 s | **-31.0%** |
| Peak GPU memory | 103,661 MiB | 99,789 MiB | **-3.7%** |

Setup time was unchanged. Every run used 661 linear iterations, 677 preconditioner applications, and 1,354 coarse solves. The final candidate passed the control-derived numeric CSV gate.

### Testing

- CPU and CUDA/cuDSS Spack builds
- Full SLEPc and ARPACK unit suites, serial and MPI-2
- 32/32 ARPACK+SuperLU regression cases passed at 5 MPI ranks
- Repeated-frequency and nonlinear eigenvalue cases passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,12 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
     quantities referenced to the internal unit impedance, not physical incident waves.
     [PR 841](https://github.com/awslabs/palace/pull/841).
 
+#### Performance Improvements
+
+  - Omit material terms with mathematically exact-zero coefficients from fine-level partial
+    assembly while retaining the configured coarse sparse structure for symbolic reuse.
+    [PR 876](https://github.com/awslabs/palace/pull/876).
+
 #### Bug Fixes
 
   - Reject all-zero numerator or denominator polynomial coefficients in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,12 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
     quantities referenced to the internal unit impedance, not physical incident waves.
     [PR 841](https://github.com/awslabs/palace/pull/841).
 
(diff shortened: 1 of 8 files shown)


simlapointe (approved):
LGTM! 

If it's not too complicated, it might be nice to add unit test(s) for the spaceoperator.cpp changes, since right now only the materialoperator.cpp changes are covered. Maybe something like a test comparing exact-zero and tiny-nonzero real/imaginary slots, and/or  something like a exact-complex preconditioner test that assembles at λ = iω, updates to λ = ε + iω, and verifies the coarse imaginary CSR pattern is unchanged while the fine zero term is initially omitted.

PR #879, opened 2026-08-11
Reuse screened magnetostatic operators

### LINKED ISSUE #845: Reuse screened magnetostatic operators across Open-port excitations
Follow-up from [PR #831](https://github.com/awslabs/palace/pull/831) ([review comment](https://github.com/awslabs/palace/pull/831#discussion_r3661986625)).

### Context

PR #831 added `Short` treatment for inactive surface-current ports in the magnetostatic
inductance sweep. When a port resolves to `Short` while inactive, its boundary is added to
the essential-DOF set for every excitation step except the one that excites it, so the
stiffness matrix is reassembled per step (`GetStiffnessMatrix(short_attrs)`), and the reused
`KspSolver` rebinds/re-sets-up its preconditioner for that operator.

### Observation

With multiple `Open` ports and at least one `Short` port, every `Open`-port excitation
shorts the same set of inactive ports, so they all rebuild an identical screened operator
and repeat the same KSP setup. That work could be done once and shared across those steps.

### Proposal

Cache screened operators keyed by their shorted-attribute set (or, more coarsely, cache the
single all-`Short`-ports-except-active operator when it recurs) so the assembly and
preconditioner setup are reused across `Open`-port excitations.

### Constraint

`ParOperator::SetEssentialTrueDofs` stores a shallow reference to the essential-DOF list, so
any cache must also own the corresponding essential-DOF data for the operator's lifetime —
the same lifetime issue currently handled by `CurlCurlOperator::extra_dbc_tdof_lists`.

### Note

This should be profiled before adding caching complexity, since
the assembly + setup cost may be small relative to the solves for typical problem sizes.

dnpham23:
Follow-up to [#831](https://github.com/awslabs/palace/pull/831). Reuses screened magnetostatic stiffness operators across Open-port excitations so a multi-port inductance sweep no longer reassembles and repreconditions an identical operator once per step. Closes [#845](https://github.com/awslabs/palace/issues/845).

### Description 
When a sweep has multiple `Open` ports and at least one `Short` port, every Open-port excitation shorts the identical set of inactive ports. Previously each step reassembled the same screened operator and repeated the same AMS preconditioner setup.

### Main changes

- **Operator cache in `CurlCurlOperator`.** keyed on the sorted, deduplicated shorted-attribute set; each entry owns its essential-true-DOF lists (shallow-reference lifetime handled via `std::map` node stability).
- **Two-pass sweep**: pass 1 solves grouped by shorted-attribute set (first-appearance order, leaving already-grouped orderings untouched); pass 2 postprocesses in canonical order so output ordering is unchanged.

### Testing
New `rings_multiring_inactive_ports` regression case (2 Open + 1 Short, isolated multiring case dir), NaN-vs-NaN comparator fix, and verified rings / multiring / circular_hole_flux_loop / spheres all pass at 1, 2, and 4 ranks.

### Additional local validation
Stress-tested the cache on a 5-ring geometry (one port per ring) with ports 3 & 5 `Short`, ports 1/2/4 `Open`: 3 distinct shorted sets, the shared `{3,5}` operator reused by three non-contiguous Open steps.

- Grouped solve order `[1,2,4,3,5]`, identical at 1/2/4 ranks.
- Open–Open reciprocity exactly symmetric at all rank counts.
- M partition-invariant to ~2e-10 ($\ll$ 1e-8 tol); NaN structure identical.
- 3 preconditioner setups for 5 steps (2 saved).


palace/models/curlcurloperator.cpp
```diff
@@ -250,19 +260,24 @@ CurlCurlOperator::GetStiffnessMatrix(const mfem::Array<int> &extra_dbc_attr)
   BilinearForm k(GetNDSpace());
   k.AddDomainIntegrator<CurlCurlIntegrator>(muinv_func);
   auto k_vec = k.Assemble(GetNDSpaces(), skip_zeros);
+
+  // Insert the cache entry before binding essential DOFs: SetEssentialTrueDofs stores a
+  // shallow reference to the tdof list, so the list must live at its final address.
+  // std::map nodes are address-stable across later insertions, so the operator's references
+  // stay valid for the cache's lifetime.
+  auto &entry = screened_stiffness_cache[key];
```
hughcars: This keeps every miss. With `S` Short ports, all `S` keys are unique, so we retain `O(S)` hierarchies for no reuse. Can we bound this to the current operator group? Since equal keys are already adjacent, an operator-owned two-entry rolling cache should preserve all reuse while keeping the previously bound operator alive:

```cpp
if (!screened_stiffness_cache.empty() &&
    screened_stiffness_cache.back().key == key)
{
  return *screened_stiffness_cache.back().K;
}
if (screened_stiffness_cache.size() == 2)
{
  screened_stiffness_cache.pop_front();
}
auto &entry = screened_stiffness_cache.emplace_back();
entry.key = std::move(key);
```

examples/rings/multiring_inactive_ports.json
```diff
@@ -0,0 +1,71 @@
...
+      {
+        "Index": 3,
+        "Attributes": [6],  // Outer ring
+        "Direction": [-0.5, 0.8660254037844386, 0.0],
+        // Per-port override: this port is shorted (PEC) when inactive. Ports 1 and 2 use the
+        // global "Open" default below. With two Open ports and one Short port, exciting port
+        // 1 and exciting port 2 both short the same set {port 3}, so they share an identical
+        // screened operator -- the redundancy the operator cache eliminates.
+        "InactiveMode": "Short"
+      }
```
hughcars, line 47: Can we make this Open/Short/Open? `A,B,A` exercises both regrouping and reuse; current `A,A,B` never reorders. The refs will need regenerating.

```suggestion
      {
        "Index": 1,
        "Attributes": [4],  // Inner ring
        "Direction": [-0.5, 0.8660254037844386, 0.0]  // "+Y" rotated π/6 around (0,0,1)
      },
      {
        "Index": 2,
        "Attributes": [5],  // Middle ring
        "Direction": [-0.5, 0.8660254037844386, 0.0],
        // Port 2 is Short while ports 1 and 3 use the global Open default. This gives the
        // operator-key order A, B, A, exercising both cache reuse and solve reordering.
        "InactiveMode": "Short"
      },
      {
        "Index": 3,
        "Attributes": [6],  // Outer ring
        "Direction": [-0.5, 0.8660254037844386, 0.0]
      }
```

hughcars (commented):
Two inline comments.

palace/models/curlcurloperator.cpp
dnpham23: Thanks! Addressed in [41131f5](https://github.com/awslabs/palace/pull/879/commits/41131f5722336f1df97b6d3d40db0253950dc52f)

examples/rings/multiring_inactive_ports.json
dnpham23: Addressed in [41131f5](https://github.com/awslabs/palace/pull/879/commits/41131f5722336f1df97b6d3d40db0253950dc52f)

hughcars (approved):
Both comments addressed, thanks!

PR #880, opened 2026-08-11
Surface PETSc and SLEPc probe failures in CMake output

### LINKED ISSUE #649: try_compile hides useful information when it fails
Sometimes, builds fail for errors like
```
-- PETSc test program - Failed
-- PETSc test program with static linkage - Failed
CMake Error at CMakeLists.txt:252 (message):
  PETSc could not be found, be sure to set PETSC_DIR
```

This error message is not useful. The real error can be found in `palace-build/CMakeFiles/CMakeConfigureLog.yaml`. It would be useful to surface this error to the main log so that we can easily diagnose what is going on (e.g., in CI).

danielgaskins:
## Description

Preserve the diagnostics captured by the PETSc and SLEPc `try_run` probes and print them when dependency detection ultimately fails. Compile failures now include the compiler/build output; probes that compile but fail at runtime include their exit code and program output.

Diagnostics from the initial pkg-config attempt are deferred while Palace tries the existing `--static` fallback. If that fallback succeeds, configuration remains quiet. If it also fails, both attempts are reported so the final `PETSc/SLEPc could not be found` error has actionable context in the main configure log.

## Testing

Used a standalone CMake harness with fake imported pkg-config targets to exercise:

- a PETSc compile failure caused by a missing header, including the full compiler error
- PETSc and SLEPc runtime failures with exit codes 7 and 9
- failure of both default and `--static` pkg-config attempts
- successful PETSc and SLEPc probes, which remain quiet

Also ran `git diff --check`.

Closes #649


palace/cmake/PkgConfigHelpers.cmake
```diff
@@ -152,8 +152,16 @@ int main()
   endif()
   if(PETSC_TEST_SUCCESS)
     set(${_petsc_test_success} TRUE PARENT_SCOPE)
+    set(${_petsc_test_diagnostic} "" PARENT_SCOPE)
   else()
+    if(PETSC_TEST_COMPILED)
+      set(PETSC_TEST_DIAGNOSTIC
+          "Exited with code ${PETSC_TEST_EXITCODE}\n${PETSC_TEST_OUTPUT}")
+    else()
+      set(PETSC_TEST_DIAGNOSTIC "Compile output:\n${PETSC_TEST_COMPILE_OUTPUT}")
```
hughcars, line 161: This still reports CMake’s cross-compile sentinel as if it were a real program exit code. With a normal CPU cross configure, `try_run()` can compile the probe but set `PETSC_TEST_EXITCODE` to `PLEASE_FILL_OUT-FAILED_TO_RUN`, which currently becomes `Exited with code PLEASE_FILL_OUT-FAILED_TO_RUN`.

Can we distinguish that from an actual nonzero exit?

```suggestion
    if(PETSC_TEST_COMPILED)
      if(PETSC_TEST_EXITCODE MATCHES "^PLEASE_FILL_OUT-")
        set(PETSC_TEST_DIAGNOSTIC
            "The PETSc test program compiled, but CMake did not run it while "
            "cross-compiling. Set CMAKE_CROSSCOMPILING_EMULATOR or provide the "
            "try_run cache results requested above.")
      elseif(PETSC_TEST_EXITCODE STREQUAL "FAILED_TO_RUN")
        set(PETSC_TEST_DIAGNOSTIC
            "The PETSc test program compiled, but CMake could not execute it.\n"
            "${PETSC_TEST_OUTPUT}")
      else()
        set(PETSC_TEST_DIAGNOSTIC
            "Exited with code ${PETSC_TEST_EXITCODE}\n${PETSC_TEST_OUTPUT}")
      endif()
    else()
      set(PETSC_TEST_DIAGNOSTIC "Compile output:\n${PETSC_TEST_COMPILE_OUTPUT}")
    endif()
```

palace/cmake/PkgConfigHelpers.cmake
```diff
@@ -247,9 +263,16 @@ int main()
...
     # message(STATUS "SLEPc test program - Successful")
     set(${_slepc_test_success} TRUE PARENT_SCOPE)
+    set(${_slepc_test_diagnostic} "" PARENT_SCOPE)
   else()
-    # message(STATUS "SLEPc test program - Failed")
+    if(SLEPC_TEST_COMPILED)
+      set(SLEPC_TEST_DIAGNOSTIC
+          "Exited with code ${SLEPC_TEST_EXITCODE}\n${SLEPC_TEST_OUTPUT}")
+    else()
+      set(SLEPC_TEST_DIAGNOSTIC "Compile output:\n${SLEPC_TEST_COMPILE_OUTPUT}")
```
hughcars, line 272: Same issue in the SLEPc path: a cross-compile sentinel is not a program exit code. Can we mirror the PETSc handling here?

```suggestion
    if(SLEPC_TEST_COMPILED)
      if(SLEPC_TEST_EXITCODE MATCHES "^PLEASE_FILL_OUT-")
        set(SLEPC_TEST_DIAGNOSTIC
            "The SLEPc test program compiled, but CMake did not run it while "
            "cross-compiling. Set CMAKE_CROSSCOMPILING_EMULATOR or provide the "
            "try_run cache results requested above.")
      elseif(SLEPC_TEST_EXITCODE STREQUAL "FAILED_TO_RUN")
        set(SLEPC_TEST_DIAGNOSTIC
            "The SLEPc test program compiled, but CMake could not execute it.\n"
            "${SLEPC_TEST_OUTPUT}")
      else()
        set(SLEPC_TEST_DIAGNOSTIC
            "Exited with code ${SLEPC_TEST_EXITCODE}\n${SLEPC_TEST_OUTPUT}")
      endif()
    else()
      set(SLEPC_TEST_DIAGNOSTIC "Compile output:\n${SLEPC_TEST_COMPILE_OUTPUT}")
    endif()
```

hughcars (commented):
Thanks for the PR @danielgaskins, surfacing the actual probe output is definitely useful. The normal compile and runtime failures and the static fallback all look good. I found one cross-compile case where CMake’s sentinel gets printed as an exit code; I’ve left suggestions for PETSc and SLEPc.

palace/cmake/PkgConfigHelpers.cmake
danielgaskins: Updated the PETSc path to distinguish the cross-compilation placeholder, `FAILED_TO_RUN`, and real nonzero exit codes. I used `string(CONCAT)` so the multi-part diagnostic remains a single string.

palace/cmake/PkgConfigHelpers.cmake
danielgaskins: Mirrored the same handling for SLEPc. I also exercised success, both sentinel cases, a real exit code, and a compile failure for each dependency with a standalone CMake script.

hughcars (approved):

hughcars:
Please add a CHANGELOG entry under `In progress` as well.

hughcars (approved):

hughcars:
Bypass merging because pipelines are being slow and CI has run many times.

PR #881, opened 2026-08-11
Publish schema compatibility data

### LINKED ISSUE #833: Distribute json schema compatibility table in machine-readable file
The [documentation](https://awslabs.github.io/palace/stable/developer/notes/#Relationship-to-the-*Palace*-version) provides a table to map Palace versions to schema versions. It would be useful if this table were also exposed as a machine-readable file that other tools can reference without having to hardcode versions.

Then, the table in the docs could be generated from the file.

danielgaskins:
Add a machine-readable `schema-compatibility.json` alongside the configuration schema and generate the developer-documentation compatibility table from it. The installed schema directory therefore exposes the mapping to downstream tools without requiring them to scrape Markdown.

The documentation workflow now watches schema inputs, and the runtime embed step explicitly includes only the validation schema rather than the new release metadata. This also updates the schema-version check’s stale instruction to point contributors to the changelog.

Tests:
- `julia --project=docs docs/generate_schema_compatibility.jl` (including idempotence and malformed-entry checks)
- `julia --project=docs docs/make.jl`
- `./scripts/format-source -jl --dry-run`
- generated-schema CMake script check

Closes #833

### DIFF
--- a/docs/generate_schema_compatibility.jl
+++ b/docs/generate_schema_compatibility.jl
@@ -9,8 +9,8 @@ using JSON
 const COMPATIBILITY_PATH =
     joinpath(@__DIR__, "..", "scripts", "schema", "schema-compatibility.json")
 const NOTES_PATH = joinpath(@__DIR__, "src", "developer", "notes.md")
-const TABLE_BEGIN = "<!-- BEGIN GENERATED SCHEMA COMPATIBILITY TABLE -->"
-const TABLE_END = "<!-- END GENERATED SCHEMA COMPATIBILITY TABLE -->"
+const TABLE_BEGIN = "```@raw html\n<!-- BEGIN GENERATED SCHEMA COMPATIBILITY TABLE -->\n```"
+const TABLE_END = "```@raw html\n<!-- END GENERATED SCHEMA COMPATIBILITY TABLE -->\n```"
 
 function markdown_row(values::Vector{String}, widths::Vector{Int})::String
     return "| " *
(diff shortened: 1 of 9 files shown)


docs/generate_schema_compatibility.jl
```diff
@@ -0,0 +1,77 @@
...
+    data["schema_versions"] isa Vector ||
+        error("Compatibility data field \"schema_versions\" must be an array")
+
+    source = read(notes_path, String)
+    length(findall(TABLE_BEGIN, source)) == 1 ||
+        error("Expected exactly one compatibility table start marker in $notes_path")
+    length(findall(TABLE_END, source)) == 1 ||
+        error("Expected exactly one compatibility table end marker in $notes_path")
+
+    generated = "$TABLE_BEGIN\n$(compatibility_table(data["schema_versions"]))\n$TABLE_END"
```
hughcars: This doesn’t actually render as a table. I built the docs locally and added a few fake entries; Documenter wraps the whole block in a `<p>` because the HTML comments are directly adjacent to the Markdown table.

Can we put the markers in `@raw html` blocks and leave blank lines around the table?

```julia
const TABLE_BEGIN = "```@raw html\n<!-- BEGIN GENERATED SCHEMA COMPATIBILITY TABLE -->\n```"
const TABLE_END = "```@raw html\n<!-- END GENERATED SCHEMA COMPATIBILITY TABLE -->\n```"
```

```julia
generated = "$TABLE_BEGIN\n\n$(compatibility_table(data[\"schema_versions\"]))\n\n$TABLE_END"
```

I tried this locally and it produces an actual `<table>`.

docs/generate_schema_compatibility.jl
```diff
@@ -0,0 +1,77 @@
...
+            ]
+        )
+    end
+
+    headers = ["Schema version", "First *Palace* release", "Notes"]
+    widths = [maximum(length(row[i]) for row in [headers, rows...]) for i = 1:3]
+    separators = [":" * repeat("-", width - 2) * ":" for width in widths]
+    separators[3] = ":" * repeat("-", widths[3] - 1)
+
+    lines = [markdown_row(headers, widths), markdown_row(separators, widths)]
```
hughcars: This currently fights JuliaFormatter. Running `./scripts/format-source -jl` rewrites the separator row, then the next docs build changes it back, so the generated file can’t stay clean.

Can we generate the form JuliaFormatter expects directly?

```suggestion
    separators = [":" * repeat("-", width) * ":" for width in widths]
    separators[3] = ":" * repeat("-", widths[3]) * " "

    lines = [markdown_row(headers, widths), "|" * join(separators, "|") * "|"]
```

I tried this locally as well; regeneration stays idempotent and the formatter leaves it alone.

hughcars (changes requested):
Thanks for the PR @danielgaskins, I tried it out and found a few small issues but the basic idea is a nice way to keep things together.

Sbozzolo:
LGTM if @hughcars' comments are addressed

docs/generate_schema_compatibility.jl
danielgaskins: Sure thing, I moved both markers into `@raw html` blocks and added blank lines around the generated table. I also reran the generator twice to confirm the output is stable.

docs/generate_schema_compatibility.jl
danielgaskins: Updated the separator row to match JuliaFormatter's output directly. The pinned JuliaFormatter 2.5.0 check now leaves both changed files unchanged.

danielgaskins:
Updated this branch from current main and added the Palace 0.18.0 / SchemaVer 1-6-0 release boundary to the compatibility data. The generated table matches the JSON, and the CMake embedding check confirms that the metadata is installed but not embedded as a runtime schema.

hughcars (approved):

hughcars:
Please add a CHANGELOG entry under `In progress` as well.

hughcars (approved):

PR #882, opened 2026-08-12
Consolidate orthogonalization dispatch

### LINKED ISSUE #617: Merge OrthogonalizeColumn routines
There are OrthogonalizeColumn routines in both the iterative solves and rom operator that should be merged, perhaps as part of a larger linalg sping-cleaning.

See discussion here in PR review:

> Yes. I just followed the same structure as already existed. As far as I can tell the difference between OrthogonalizeColumn & OrthogonalizeIteration is a j / j+1 indexing. But I must admin that I did not want to change the `iterative.cpp` code. There seems to be a lot of generic linear algebra routines that could be pulled out and refactored. Maybe I can open an issue for this?

_Originally posted by @phdum-a in https://github.com/awslabs/palace/pull/326#discussion_r2674326173_

danielgaskins:
## Description

Centralize the selection of modified and classical Gram-Schmidt algorithms in a shared `linalg::OrthogonalizeColumn` dispatcher.

This removes the duplicate dispatch logic from the iterative solvers and ROM operator while preserving their existing basis-size semantics. The orthogonalization unit tests now call the production dispatcher directly instead of maintaining a separate test-only copy of the same switch.

## Testing

- Ran the project C++ formatter with clang-format 19
- Ran `git diff --check`
- Updated the existing serial, parallel, real, complex, weighted, and unweighted orthogonalization tests to exercise the shared dispatcher

Closes #617

### DIFF
--- a/palace/linalg/iterative.cpp
+++ b/palace/linalg/iterative.cpp
@@ -304,26 +304,6 @@ inline void ApplyBA(PreconditionerSide side, const OperType *A, const Solver<Ope
   }
 }
 
-template <typename VecType, typename ScalarType>
-inline void OrthogonalizeIteration(Orthogonalization type, MPI_Comm comm,
-                                   const std::vector<VecType> &V, VecType &w,
-                                   ScalarType *Hj, std::size_t j)
-{
-  // Orthogonalize w against the leading j + 1 columns of V.
-  switch (type)
-  {
-    case Orthogonalization::MGS:
-      linalg::OrthogonalizeColumnMGS(comm, V, w, Hj, j + 1);
-      break;
-    case Orthogonalization::CGS:
-      linalg::OrthogonalizeColumnCGS(comm, V, w, Hj, j + 1);
-      break;
-    case Orthogonalization::CGS2:
-      linalg::OrthogonalizeColumnCGS(comm, V, w, Hj, j + 1, true);
-      break;
-  }
-}
-
 }  // namespace
 
 template <typename OperType>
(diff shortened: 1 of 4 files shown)


danielgaskins:
CI note: I checked the failed x64 and GPU jobs. Both stop before this change is built because the existing `mfem_pr4983.diff` patch no longer applies to MFEM `develop`. The jobs also receive 403 responses when they try to upload build-cache objects from the fork. `handle-non-trivial-changes` is the existing fork status-permission failure. I have left the branch unchanged because none of these failures comes from the orthogonalization changes.

hughcars:
> CI note: I checked the failed x64 and GPU jobs. Both stop before this change is built because the existing `mfem_pr4983.diff` patch no longer applies to MFEM `develop`. The jobs also receive 403 responses when they try to upload build-cache objects from the fork. `handle-non-trivial-changes` is the existing fork status-permission failure. I have left the branch unchanged because none of these failures comes from the orthogonalization changes.

Yes the lookahead branches failures right now are known, because a big patch we had been maintaining for mfem was merged, once they cut their next version, we'll bump ours and release it all together. So the lookahead failures are non-blocking for PRs right now.

hughcars (approved):
Thanks for the PR @danielgaskins, this looks like a nice cleanup. I traced the real, complex, and weighted paths, and the consolidated dispatch preserves the existing basis-count and conjugation behavior across GMRES, FGMRES, and the ROM code. The failing checks appear to be the existing MFEM patch and fork buildcache issues rather than anything from this change.

PR #883, opened 2026-08-15
Accept configuration enum values case-insensitively

### LINKED ISSUE #586: Palace is strict with the capitalization of solvers in the configuration file
Palace requires `config["Solver"]["Linear"]["Type"]` to be spelled with very specific capitalizations (e.g., it's `SuperLU`, not `superLU`). I think that this is a little too pedantic (hence, a little annoying--the correct spelling is hard to remember). The user intent is clearly expressed, so I think Palace should accept any capitalization.

SpookyJumpyBeans:
Palace required exact capitalization for configuration enum values (e.g. "SuperLU", not "superlu"). Since the user intent is unambiguous, accept any capitalization for these values.

Two layers enforced the exact case, so both are addressed:
- FromString (utils/enum_string.cpp) now compares case-insensitively, so the enum <-> string mapping resolves any capitalization. Enum spellings are unique when compared case-insensitively, so this cannot be ambiguous.
- Schema validation (utils/jsonschema.cpp) matches enum values case-sensitively and aborts before parsing on a mismatch. A new CanonicalizeEnumCase pass rewrites enum-valued strings to their canonical schema spelling before validation, reusing the existing schema walk. It leaves exact matches and genuinely invalid values untouched so validation still reports real errors.

Wired into IoData::ParseAndValidate before validation, added unit tests, and a changelog entry.

Fixes #586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,11 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 
 #### Bug Fixes
 
+  - Configuration enum values (e.g. solver names such as `"SuperLU"`) are now matched
+    case-insensitively, so any capitalization (`"superlu"`, `"SUPERLU"`, ...) is accepted
+    both during schema validation and parsing, and is normalized to the canonical spelling.
+    Previously an incorrect case caused validation to abort.
+    [Issue 586](https://github.com/awslabs/palace/issues/586).
   - Reject all-zero numerator or denominator polynomial coefficients in
     `config["Boundaries"]["RationalImpedance"]` during schema validation, matching the
     existing checks in the configuration parser and providing users with earlier feedback.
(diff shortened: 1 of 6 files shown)


SpookyJumpyBeans:
Thanks for looking at this. #890 is a better fit if the schema should stay case-sensitive, so I’ll close this in favor of that

hughcars:
> Thanks for looking at this. #890 is a better fit if the schema should stay case-sensitive, so I’ll close this in favor of that

Yes thank you for bringing the issue to our attention @SpookyJumpyBeans, I decided that given we are hoisting more onto the schema we shouldn't relax the requirements, but instead if we could make the failure a bit more user friendly, that would make it easier to correct when encountered. Unfortunately our support for rich documentation makes it a bit more involved, but hopefully that branch tidies it up some.

PR #890, opened 2026-08-20
Improve diagnostics for invalid configuration enum values

hughcars:
## Summary

- Recursively recover valid enum values at the exact failing configuration path through arrays, `$ref`, `oneOf`, `anyOf`, and other schema composition.
- List the valid values when schema validation rejects an enum-like entry.
- Suggest the canonical capitalization for case-only mismatches.
- Preserve the canonical, case-sensitive configuration contract.

## Motivation

Addresses #586.

This is intended to supersede #883 with a diagnostics-only approach. Rather than canonicalizing configuration values before validation and making runtime acceptance differ from the published schema, Palace reports the canonical options and a targeted correction:

```text
valid values: ..., "SuperLU", ... Did you mean "SuperLU"?
```

The input configuration and embedded schema remain unchanged. Thanks to @SpookyJumpyBeans for the work in #883 that prompted and informed this alternative.

## Implementation

Enum lookup follows the exact JSON instance path and recursively traverses arrays, internal references, and schema composition. Values are collected only from the corresponding sub-schema path, deduplicated in schema order, and used solely to enrich validation errors.

The current Palace schema's 33 enum-like paths are covered, including nested forms such as:

- `Solver.Driven.Samples[].Type`
- `Boundaries.LumpedPort[].Elements[].CoordinateSystem`
- `Boundaries.SurfaceCurrent[].InactiveMode`
- direction enums referenced inside `anyOf`

### DIFF
--- a/palace/utils/jsonschema.cpp
+++ b/palace/utils/jsonschema.cpp
@@ -261,11 +261,9 @@ json FindEnumInSchema(const json &schema, const std::string &ptr)
   return values.empty() ? json() : values;
 }
 
-// Find a unique allowed string which differs from the input only by ASCII case. Returns
-// null when there is no match or when multiple spellings match case-insensitively.
+// Find an allowed string which differs from the input only by ASCII case.
 const json *FindCaseInsensitiveMatch(const json &allowed, std::string_view input)
 {
-  const json *match = nullptr;
   for (const auto &candidate : allowed)
   {
     if (!candidate.is_string())
@@ -279,14 +277,10 @@ const json *FindCaseInsensitiveMatch(const json &allowed, std::string_view input
                               { return std::tolower(a) == std::tolower(b); });
     if (matches)
     {
-      if (match != nullptr)
-      {
-        return nullptr;
-      }
-      match = &candidate;
+      return &candidate;
     }
   }
-  return match;
+  return nullptr;
 }
 
 }  // namespace
--- a/palace/utils/jsonschema.cpp
+++ b/palace/utils/jsonschema.cpp
@@ -333,6 +333,8 @@ public:
              const std::string &message) override
   {
     const std::string path = ptr.to_string();
+    // These message strings are implementation details of json-schema-validator 2.4.0;
+    // update them when upgrading the dependency.
     const bool is_enum_failure =
         message.find("instance not found in required enum") != std::string::npos ||
         message.find("instance not const") != std::string::npos ||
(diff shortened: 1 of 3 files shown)


dnpham23 (approved):
Nice improvement! I pulled the branch, rebuilt, and ran `[schema]` locally. All 12 cases / 265 assertions pass. Just a few notes, none blocking.

palace/utils/jsonschema.cpp
```diff
@@ -140,120 +142,145 @@ json ResolveRef(const json &node, const json &defs)
...
-  while (pos < ptr.size())
+// Collect enum-like values at an exact instance path. Composition branches are searched
+// with the same unconsumed path, so a property is never matched merely because it has the
+// same name elsewhere in the schema.
+void CollectEnumValues(const json &schema, const json &defs,
+                       const std::vector<std::string> &tokens, std::size_t token_index,
+                       json &values, int depth = 0)
+{
+  constexpr int max_depth = 32;
+  if (!schema.is_object() || depth > max_depth)
```
dnpham23: `FindAllSchemasByKey` emits an `Mpi::Warning` when it exceeds `max_depth`, while `CollectEnumValues` just returns. A matching warning would keep the two consistent and avoid silently truncating on a malformed schema.

palace/utils/jsonschema.cpp
```diff
@@ -304,44 +332,51 @@ class SchemaErrorHandler : public error_handler
...
+        message.find("instance not const") != std::string::npos ||
+        message.find("no subschema has succeeded") != std::string::npos;
+    json enum_values;
+    if ((schema != nullptr) && is_enum_failure)
     {
-      json enum_values = FindEnumInSchema(*schema, ptr.to_string());
-      if (!enum_values.is_null() && !enum_values.empty())
+      enum_values = FindEnumInSchema(*schema, path);
+      if (!enum_values.is_null() && enum_values.is_array() && !enum_values.empty() &&
+          !reported_enum_paths.insert(path).second)
```
dnpham23: The path dedup keeps the first error per path and drops the rest, so clean output depends on 2.4.0 emitting the aggregate before the per-branch messages. Maybe note that ordering in the version-coupling comment, or suppress `[combination:` lines by content once a path's enum is reported (order-independent)?

palace/utils/jsonschema.cpp
```diff
@@ -304,44 +332,51 @@ class SchemaErrorHandler : public error_handler
...
+        // message already contains the union of values at this exact instance path.
         return;
       }
     }
 
-    errors << "At " << FormatPath(ptr.to_string()) << ": " << message;
-    // Enhance type mismatch errors with actual type. These message strings are
-    // implementation details of json-schema-validator 2.4.0; update if upgrading.
-    // Use find() so the enhancement also fires for oneOf/anyOf-wrapped messages
+    errors << "At " << FormatPath(path) << ": " << message;
```
dnpham23: When an enum sits inside a branch of an outer `oneOf`, the enriched line carries the internal prefix, for example:

`At ["Solver"]["Driven"]["Samples"][0]["Type"]: [combination: oneOf / case#0] instance not const; valid values: "Point", "Linear", "Log". Did you mean "Linear"?`

The `valid values` + `Did you mean `payload is correct, but` [combination: oneOf / case#0] instance not const` is noise the end user doesn't have to see. Since we already compute `enum_values` at this point, maybe stripping a leading `[combination: …]`  fragment from message when `enum_values` is populated would give a fully clean line? Optional though.

dnpham23:
All comments addressed. Thanks!

PR #889, opened 2026-08-19
Prevent SaveAdaptMesh from overwriting archived meshes

alexhad6:
Fixes #887.

This removes the top-level mesh symlink left by `SaveIteration` before writing the next adapted mesh, preventing `SaveAdaptMesh` from overwriting a mesh archived by `SaveAdaptIterations`. It also adds regression coverage for consecutive adapted mesh writes.

TODO: Verify that tests pass and run a Palace simulation from this dev branch.

Issue #860, opened 2026-07-31
Electric field visualization: field values not consistent across different parts of the model


VolkerMuehlhaus:
Hi all,

apologies from the title, I wasn't sure how to properly describe this effect: 
Below is the electric field result within an RFIC interdigital capacitor, visualized at some cutting plane between the multi-layer fingers. It is expected that field pattern spans across the entire square finger area, but the **composite** files excitation_1.pvd and data.pvtu show some region with no (or low) field.

**S-parameter results seem to be correct, and from my understanding it's only an issue with field data for visualization.**

<img width="1651" height="955" alt="Image" src="https://github.com/user-attachments/assets/c7c836b5-ad76-4f03-b7a7-8778b4255b33" />

Visualized is the z component of Ereal at z=293.5. To simplify visualization in that plane, I had placed a high resistance surface in the mesh at that location. 

What I believe goes wrong: 
The model was simulated using 16 threads, dividing the model into 16 domains (is that the right word?). By comparing the overall field view to the data for the individual pieces, I see that the missing region in the plot belongs to 2 of the 16 domains, at least that's the major part of the missing/zero field between the fingers.

<img width="1744" height="2176" alt="Image" src="https://github.com/user-attachments/assets/1f75e8be-d9e0-4edf-9ebd-be056e409aed" />

https://github.com/user-attachments/assets/dbc882f5-188c-41f9-8793-69ac1b3a118c

Attached is mesh and config.json, the complete output data is too large to upload here. 
[testbench_small_mesh0u25_dump1ghz.zip](https://github.com/user-attachments/files/30580334/testbench_small_mesh0u25_dump1ghz.zip)

Simulation is was a few minutes only, so I hope you can reproduce the issue. All help is much appreciated!

Best regards
Volker

``` 
 "GitTag": "v0.16.0-34-gea2e7b23",
  "LinearSolver": {
    "TotalIts": 192,
    "TotalSolves": 1
  },
  "Problem": {
    "DegreesOfFreedom": 1872888,
    "MPISize": 16,
    "MeshElements": 256803,
    "MultigridDegreesOfFreedom": [
      371699,
      1872888
    ],
    "OpenMPThreads": 1
  }
```

VolkerMuehlhaus:
@hughcars If you already had a chance to look at this: do you need any more material/data from me to reproduce or investigate?

hughcars:
Hi @VolkerMuehlhaus,

[...]

PR #892, opened 2026-08-21
Fix output archiving and symlink handling on shared filesystems

### LINKED ISSUE #887: SaveAdaptMesh overwrites the mesh in iteration2/ during AMR
### Description

With `SaveAdaptMesh` and `SaveAdaptIterations` enabled, the first adapted mesh is moved into `iteration2/` and replaced by a symlink. Subsequent mesh writes follow that symlink, repeatedly overwriting `iteration2/<mesh>.mesh`.

Each `iterationN/` directory should retain the mesh associated with that iteration’s solve, while the top-level mesh should reference the final adapted mesh.

It appears this was introduced by #673 (released in Palace v0.16.1), which changed AMR iteration archiving from copying files to moving them and leaving top-level symlinks.

### To Reproduce

Modify `examples/transmon/transmon_amr.json`:

```diff
{
  "Model": {
    "Refinement": {
-      "MaxIts": 2
+      "MaxIts": 3,
+      "SaveAdaptIterations": true,
+      "SaveAdaptMesh": true
    }
  }
}
```

Run the example and inspect the iteration directories. The updated mesh is stored in `iteration2/` instead of the corresponding iteration directory.

hughcars:
Fixes #887. Supersedes #889.

This PR builds on #889 from @alexhad6 and preserves that commit and authorship.

It fixes three related output-filesystem problems:

- Prevent `SaveAdaptMesh` from following a top-level symlink and overwriting a mesh archived by `SaveAdaptIterations`.
- Preserve ParaView output from earlier driven excitations instead of deleting the entire ParaView root when switching excitations.
- Make adaptive iteration archiving safe for shared NFS/SMB filesystems by snapshotting directory entries before modifying them and using an atomic claim so only one node archives each shared filesystem view.

The ParaView fix closes the previous MFEM data collection before switching paths, avoiding deletion of files that may still be open on SMB.

This addresses the filesystem failure discussed in #860, but not the separate field-visualization issue reported there.

Closes #889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,9 +70,9 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 #### Bug Fixes
 
   - Fixed adaptive iteration output archiving overwriting earlier meshes and made its
-    filesystem updates more robust.
+    filesystem updates more robust. [PR 892](https://github.com/awslabs/palace/pull/892).
   - Fixed ParaView output for multiple driven excitations deleting fields from earlier
-    excitations.
+    excitations. [PR 892](https://github.com/awslabs/palace/pull/892).
   - Fixed a false-positive in the domain material-coverage check that could abort numeric
     wave-port simulations under nonconformal AMR on multiple MPI ranks. The check now
     validates only domains owned by a local volume element, ignoring ghost/neighbor
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,10 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 
 #### Bug Fixes
 
+  - Fixed adaptive iteration output archiving overwriting earlier meshes and made its
+    filesystem updates more robust.
+  - Fixed ParaView output for multiple driven excitations deleting fields from earlier
+    excitations.
   - Fixed a false-positive in the domain material-coverage check that could abort numeric
     wave-port simulations under nonconformal AMR on multiple MPI ranks. The check now
     validates only domains owned by a local volume element, ignoring ghost/neighbor
--- a/palace/utils/outputdir.hpp
+++ b/palace/utils/outputdir.hpp
@@ -43,6 +43,30 @@ inline void AbortOnNodeFilesystemError(MPI_Comm comm, const std::string &operati
 
 }  // namespace internal
 
+// Apply a filesystem operation on the global root and propagate any error collectively so
+// other ranks do not continue into a mismatched MPI control path.
+template <typename Func>
+inline void ApplyOnRootFilesystem(MPI_Comm comm, const std::string &operation, Func &&func)
+{
+  std::string local_error;
+  if (Mpi::Root(comm))
+  {
+    try
+    {
+      func();
+    }
+    catch (const std::exception &e)
+    {
+      local_error = e.what();
+    }
+    catch (...)
+    {
+      local_error = "Unknown filesystem error";
+    }
+  }
+  internal::AbortOnNodeFilesystemError(comm, operation, local_error);
+}
+
 // Apply an operation once to each node's view of the filesystem. The global root runs
 // first so that shared filesystems are updated before the remaining node roots inspect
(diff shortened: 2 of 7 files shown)


palace/utils/outputdir.hpp
```diff
@@ -123,6 +203,35 @@ inline void RemovePreviousOutput(const fs::path &dir, MPI_Comm comm)
...
+      [&]()
+      {
+        fs::remove_all(path);
+        std::ofstream stream(path);
+        if (!stream.is_open())
+        {
+          throw fs::filesystem_error("Could not open output file", path,
+                                     std::make_error_code(std::errc::io_error));
+        }
+        write(stream);
```
Sbozzolo: Tiny suggestion: I would suggest changing the name of `write` to make it clearer that it is a user-supplied function and not the default `write` of a stream.

palace/utils/outputdir.hpp
```diff
@@ -86,6 +110,62 @@ inline void ApplyOnEachNodeFilesystem(MPI_Comm comm, const std::string &operatio
...
+  {
+    apply_here();
+  }
+  MPI_Comm_free(&node_comm);
+
+  internal::AbortOnNodeFilesystemError(comm, operation, local_error);
+  ApplyOnEachNodeFilesystem(comm,
+                            fmt::format("Cleaning up filesystem operation: {}", operation),
+                            [&]() { fs::remove_all(claim); });
+}
```
Sbozzolo, line 139:
```suggestion
  template <typename Func>
  inline void ApplyOnceOnEachNodeFilesystem(MPI_Comm comm, const fs::path &claim,
                                     const std::string &operation, Func &&func)
  {
        ApplyOnEachNodeFilesystem(
        comm, "Removing stale claims for " + operation,
        [&]()
        {
          fs::create_directories(claim.parent_path()); 
          fs::remove_all(claim);
        });

    ApplyOnEachNodeFilesystem(
        comm, operation,
        [&]()
        {
         // Only the rank that successfully creates the `claim` directory runs the function, if the directory already exists, the rank does nothing.
        
          // We use directories for claims because `create_directory` provides a portable atomic create-if-absent operation that does not required POSIX-specific APIs and is simpler for network-filesystems than proper file locks (e.g., `open` with `O_CREAT |   O_EXCL`)
          auto acquired_claim = fs::create_directory(claim);
          if (acquired_claim)
          {
            func();
          }
        });

    ApplyOnEachNodeFilesystem(
        comm, "Cleaning up " + operation,
        [&]() { fs::remove_all(claim); });
  }
```

I would also make `fs::create_directories(claim.parent_path()); ` responsability of the caller, so that you can remove the `fs::create_directory`. You can just call `ApplyOnEachNodeFilesystem` with `create_directories` `step_output`, which is safe to call in the caller.

CHANGELOG.md
```diff
@@ -69,6 +69,10 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 
 #### Bug Fixes
 
+  - Fixed adaptive iteration output archiving overwriting earlier meshes and made its
+    filesystem updates more robust.
+  - Fixed ParaView output for multiple driven excitations deleting fields from earlier
+    excitations.
```
Sbozzolo: Missing MR links here

palace/utils/outputdir.hpp
Sbozzolo: Maybe also add a short description/comment on how this function works. I had to stare at it for a while to understand it.

palace/utils/outputdir.hpp
hughcars: Fixed in 452237cedf

palace/utils/outputdir.hpp
hughcars: Fixed in 452237cedf

hughcars: Fixed in 452237cedf

CHANGELOG.md
hughcars: Fixed in 452237cedf

Sbozzolo (approved):

PR #895, opened 2026-08-21
Limit MFEM #4983 patch to version 4.9

### LINKED ISSUE #894: MFEM develop lookahead fails after #4983 merged upstream
MFEM [#4983](https://github.com/mfem/mfem/pull/4983) merged on August 20. Palace's Spack package still applies its backport to every MFEM version, including `@develop`.

The x64 and GPU lookahead jobs now stop while building MFEM because the patch tries to create files that already exist and several hunks no longer apply:

- [x64 job](https://github.com/awslabs/palace/actions/runs/32469341809/job/96732777259)
- [GPU job](https://github.com/awslabs/palace/actions/runs/32469341809/job/96732777255)

MFEM 4.9 still needs the backport. The CMake superbuild also uses a pinned MFEM commit that predates the merge. The Spack patch can be limited with `when="@:4.9"`, matching the other release-only MFEM patches. This would let `@develop` use the upstream implementation without changing pinned builds.

danielgaskins:
MFEM #4983 is now part of MFEM develop. Applying Palace's backport to that version causes the x64 and GPU lookahead jobs to fail before Palace builds.

Limit the Spack dependency patch to MFEM 4.9, where it still applies and is still needed. The CMake superbuild stays unchanged because its pinned MFEM commit predates the upstream merge.

Testing:

- `python3 -m py_compile spack_repo/local/packages/palace/package.py`
- `git diff --check`
- Verified that the backport applies cleanly to MFEM 4.9
- Verified that the backport is rejected by current MFEM master

Closes #894

### DIFF
--- a/spack_repo/local/packages/palace/package.py
+++ b/spack_repo/local/packages/palace/package.py
@@ -243,6 +243,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
                     "b22f654ab36fe01f1f3176349c60626efed1a6a2/extern/patch/mfem/"
                     "mfem_pr4983.diff",
                     sha256="530532da3ae8815d004bb6ce19f6f08a1248c3d585503551c90d0eeae7fb3f87",
+                    when="@:4.9",
                 ),
                 # https://github.com/mfem/mfem/pull/5124
                 patch(


hughcars (approved):

hughcars:
So this won't actually fully fix the lookahead, given there was a change in the api for the patch, but it's still worth having in the meantime (and will be the future solution).

PR #900, opened 2026-08-25
Retry transient PPA setup in container CI

### LINKED ISSUE #899: Retry transient Apptainer PPA failures in container CI
## Problem

The container workflow can fail before building Palace when Launchpad has a transient problem serving the Apptainer PPA signing key.

In PR #890, both container matrix legs failed in the `Install Podman` step of run [32844332239](https://github.com/awslabs/palace/actions/runs/32844332239):

- [build-containers (x64)](https://github.com/awslabs/palace/actions/runs/32844332239/job/97790576273)
- [build-containers (arm64)](https://github.com/awslabs/palace/actions/runs/32844332239/job/97790576283)

The failing command was:

```bash
sudo add-apt-repository -y ppa:apptainer/ppa
```

Launchpad returned HTTP 500 while `add-apt-repository` requested the PPA signing key:

```text
lazr.restfulclient.errors.ServerError: HTTP Error 500: Internal Server Error
b'GPGKeyTemporarilyNotFoundError'
```

The two architecture jobs received separate Launchpad request/OOPS IDs and failed at essentially the same time. No Palace source, Dockerfile, OCI image, or SIF build was reached. A manual rerun was required.

`GPGKeyTemporarilyNotFoundError` is specifically a temporary Launchpad/key-retrieval failure, so failing the complete job after one request makes container CI unnecessarily flaky.

## Location

`.github/actions/build-container/action.yml`, step `Install Podman`.

## Proposed direction

Add bounded retry with backoff around the idempotent external package-repository setup, at minimum:

```bash
sudo add-apt-repository -y ppa:apptainer/ppa
sudo apt update
sudo apt install -y podman apptainer
```

The implementation should:

- retry transient network/HTTP failures within the same job;
- log the attempt number and delay;
- preserve the final command output and exit status after the retry budget is exhausted;
- remain bounded so permanent configuration/package errors still fail promptly;
- avoid retrying Palace build or test failures.

A small shared shell retry helper in the runner/container setup actions may be preferable to duplicating loops. As an alternative, investigate configuring the official PPA through a pinned keyring/source entry without depending on Launchpad's API during every job, while retaining support for both amd64 and arm64.

## Acceptance criteria

- A transient Launchpad 5xx/signing-key retrieval error is retried automatically.
- Container CI no longer requires a manual workflow rerun for this failure mode.
- Both x64 and arm64 use the hardened setup path.
- Permanent setup failures still terminate after a bounded number of attempts with useful logs.

hughcars:
Fixes #899.

Container builds currently fail before reaching Palace when Launchpad transiently returns `GPGKeyTemporarilyNotFoundError` from `add-apt-repository`.

This adds bounded retry with backoff around the Apptainer and Ubuntu toolchain PPA setup. Transient repository, update, and package-install failures retry in place; permanent failures still terminate after four attempts with the final command output intact.

### DIFF
--- a/.github/actions/build-container/action.yml
+++ b/.github/actions/build-container/action.yml
@@ -29,10 +29,20 @@ runs:
         source /tmp/timing_functions.sh
         start_timer
 
-        sudo add-apt-repository -y ppa:apptainer/ppa
-        sudo apt update
-
-        sudo apt install -y podman apptainer
+        for attempt in 1 2 3 4; do
+          if sudo add-apt-repository -y ppa:apptainer/ppa && \
+             sudo apt update && \
+             sudo apt install -y podman apptainer; then
+            break
+          fi
+          if [ "$attempt" -eq 4 ]; then
+            echo "Apptainer PPA setup failed after ${attempt} attempts" >&2
+            exit 1
+          fi
+          delay=$((15 * attempt))
+          echo "Apptainer PPA setup attempt ${attempt} failed; retrying in ${delay}s" >&2
+          sleep "$delay"
+        done
 
         # Configure podman to resolve short image names from Docker Hub
         echo 'unqualified-search-registries = ["docker.io"]' | sudo tee /etc/containers/registries.conf.d/docker.conf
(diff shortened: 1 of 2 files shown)


Sbozzolo:
As far as I can tell, the problem with the PPA only happened today. I would wait to see if the problem occurs more frequently before adding this type of relatively heavy intervention.

PR #901, opened 2026-08-25
Add electrostatic MMS verification with optional Enzyme support

laylagi:
Adds the Method of Manufactured Solutions (MMS) for the electrostatic solver, with an optional Enzyme-AD (Automatic Differentiation) backend.

The tests exercise the production solve path and verify:

  - Homogeneous and non-homogeneous Dirichlet conditions.
  - Neumann flux and anisotropic permittivity.
  - Potential and recovered electric-field errors.
  - Polynomial exactness on affine hexahedra and tetrahedra.
  - Optimal convergence on Cartesian and curved tetrahedral meshes.
  - Serial and distributed-memory execution.

  #### Enzyme support

Enzyme support is disabled by default with `PALACE_WITH_ENZYME=OFF`. When enabled, the build requires matching upstream Clang and LLVM versions and can either build Enzyme or use an existing installation through `ENZYME_DIR`.

Enzyme-specific sources are compiled separately so the compiler plugin is not applied to unrelated Palace sources. CI enables Enzyme in the Clang sanitizer configuration while the remaining configurations continue to test the default Enzyme-disabled path.

#### Building and testing with Enzyme

Enzyme requires upstream Clang and a matching LLVM development package. Apple Clang is not supported because it does not provide a matching upstream LLVM package.

On Linux, configure from the repository root with:

  ```bash
  cmake -S . -B build-enzyme \
    -DCMAKE_C_COMPILER=clang \
    -DCMAKE_CXX_COMPILER=clang++ \
    -DLLVM_DIR="$(llvm-config --cmakedir)" \
    -DPALACE_MFEM_USE_EXCEPTIONS=ON \
    -DPALACE_WITH_ENZYME=ON

  cmake --build build-enzyme --target palace-tests --parallel
```
On macOS with Homebrew LLVM:

  ```bash
  LLVM_PREFIX="$(brew --prefix llvm)"

  cmake -S . -B build-enzyme \
    -DCMAKE_C_COMPILER="$LLVM_PREFIX/bin/clang" \
    -DCMAKE_CXX_COMPILER="$LLVM_PREFIX/bin/clang++" \
    -DLLVM_DIR="$("$LLVM_PREFIX/bin/llvm-config" --cmakedir)" \
    -DPALACE_MFEM_USE_EXCEPTIONS=ON \
    -DPALACE_WITH_ENZYME=ON

  cmake --build build-enzyme --target palace-tests --parallel
```

The superbuild builds Enzyme automatically. To use an existing compatible installation, also pass `-DENZYME_DIR=<path-to-EnzymeConfig.cmake-directory>`.

Run the analytic and Enzyme-backed MMS tests with:

```bash
  ./build-enzyme/bin/palace-unit-tests "[electrostaticmms]" --skip-benchmarks
  mpirun -np 2 ./build-enzyme/bin/palace-unit-tests \
    "[electrostaticmms]" --skip-benchmarks

  ./build-enzyme/bin/palace-unit-tests "[enzyme]" --skip-benchmarks
  mpirun -np 2 ./build-enzyme/bin/palace-unit-tests \
    "[enzyme]" --skip-benchmarks
```

#### macOS build note

The Homebrew LLVM workflow also requires passing the macOS SDK sysroot to LIBXSMM. LIBXSMM is a libCEED backend, not an Enzyme dependency; this is a general Darwin/Homebrew-Clang superbuild fix encountered while validating the Enzyme workflow.

#### Documentation

Adds a numerical-verification guide, reproducible convergence data and plots, and make targets for regenerating the documentation artifacts.


Sbozzolo:
Flyby comment: this PR should also add support for spack.

Here are some notes to develop with spack: https://awslabs.github.io/palace/dev/developer/spack/

palace/models/laplaceoperator.hpp
```diff
@@ -61,6 +61,20 @@ class LaplaceOperator
...
+
+  // Optional prescribed Dirichlet values on the essential boundary (not owned). Null means
+  // homogeneous (V = 0). Set only by manufactured-solution verification tests.
+  mfem::Coefficient *dbc_source = nullptr;
+
+  // Optional prescribed Neumann flux g = ε ∂V/∂n on non-essential boundaries (not owned),
+  // applied over neumann_attr. Null means the natural homogeneous-Neumann BC. Set only by
+  // manufactured-solution verification tests.
+  mfem::Coefficient *neumann_source = nullptr;
+  mfem::Array<int> neumann_attr;
```
hughcars: These exist only for the verification tests (as the comments say, nothing in the config can set them), so the production `LaplaceOperator` and `ElectrostaticSolver` carry a volumetric RHS assembly, three non-owning pointers and a new solve branch that no simulation runs. Everything the tests need from production is public (`GetStiffnessMatrix()`, `ParOperator::EliminateRHS`, `KspSolver`, the discrete gradient), so the harness could assemble the source and Neumann terms and project the Dirichlet data itself under `test/unit/`, leaving the solver untouched. Is this meant as the start of a volume-charge feature? If so I'd say that in the PR description; otherwise I'd move it into the tests.

palace/drivers/electrostaticsolver.cpp
```diff
@@ -18,39 +18,50 @@
...
+  // aliases). A manufactured-solution verification problem instead has no terminals and is
+  // driven by a volumetric source, handled by a single source-driven solve below. Combining
+  // these modes would require a separate particular source solution in addition to the
+  // terminal basis solutions, which is not currently implemented.
+  const int n_step = static_cast<int>(laplace_op.GetSources().size());
+  const bool has_terminals = n_step > 0;
+  const bool has_rhs_source = laplace_op.HasRhsSource();
+  MFEM_VERIFY(has_terminals != has_rhs_source,
+              "Electrostatic simulation currently supports either terminal boundaries or "
+              "a volumetric source, but not both!");
```
hughcars: If the hooks stay: with no terminals, the only failure a config can reach, this now reports "not both" of a volumetric source no config can set, instead of main's "No terminal boundaries specified":

```suggestion
  MFEM_VERIFY(has_terminals || has_rhs_source,
              "No terminal boundaries specified for electrostatic simulation!");
  MFEM_VERIFY(!(has_terminals && has_rhs_source),
              "Electrostatic simulation currently supports either terminal boundaries or "
              "a volumetric source, but not both!");
```

spack_repo/local/packages/palace/package.py
```diff
@@ -77,6 +77,12 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
         description="Measure code coverage when running (leads to severe loss of performance)",
         when="@0.16:",
     )
+    variant(
+        "enzyme",
+        default=False,
+        description="Build Enzyme-backed automatic-differentiation unit tests",
+        when="@0.18:",
```
hughcars: 0.18.0 and 0.18.1 shipped without `PALACE_WITH_ENZYME`, so `palace@0.18.0+enzyme` would build Enzyme and pass a flag those releases ignore. Gate on the next release, here and in the `depends_on` below:

```suggestion
        when="@0.19:",
```

spack_repo/local/packages/palace/package.py
```diff
@@ -105,6 +111,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("scnlib~shared", when="~shared@0.14:")
     depends_on("eigen", type="build")
     depends_on("lcov@1.15:", when="+coverage@0.16:", type="run")
+    depends_on("enzyme@0.0.206:", when="+enzyme@0.18:")
```
hughcars: Same:

```suggestion
    depends_on("enzyme@0.0.206:", when="+enzyme@0.19:")
```

cmake/ExternalLIBXSMM.cmake
```diff
@@ -21,6 +21,34 @@ set(LIBXSMM_OPTIONS
   "VERBOSE=1"
 )
 
+# Homebrew Clang loses its implicit macOS SDK when LIBXSMM adds an explicit
+# Apple target triple. Pass the SDK as a compiler flag so system headers remain
+# available without requiring callers to export SDKROOT.
+if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
```
hughcars: This is a separate Homebrew-Clang macOS fix rather than part of MMS/Enzyme (the Enzyme CI path is Linux-only), and it adds an `xcrun` probe that can `FATAL_ERROR` on every Darwin superbuild. I'd split it into its own PR so it's reviewed, and if needed reverted, on its own.

palace/drivers/electrostaticsolver.cpp
```diff
@@ -67,13 +78,42 @@ ElectrostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+  BlockTimer bt0(Timer::CONSTRUCT);
+  LaplaceOperator laplace_op(iodata, mesh);
+  SaveMetadata(laplace_op.GetH1Spaces());
+  std::vector<Vector> V;
+  Solve(V, laplace_op);
+
+  // Post-processing: recover E = -∇V per terminal, measure, and estimate the error.
+  BlockTimer bt1(Timer::POSTPRO);
+  const auto &Grad = laplace_op.GetGradMatrix();
+  PostOperator<ProblemType::ELECTROSTATIC> post_op(iodata, laplace_op);
```
hughcars: This split changes every electrostatic run, not just MMS: all terminal solves now finish before the first `MeasureAndPrintAll`, so the per-terminal CSV rows and ParaView output that main writes as each solve completes only appear after the last one, and `PostOperator` and `GradFluxErrorEstimator` setup moves from the Construction to the Postprocessing timer. If the hooks move into the tests, main's single loop can stay as it was.

spack_repo/patches/enzyme_llvm_config_cmakedir.diff
```diff
@@ -0,0 +1,47 @@
+diff --git a/repos/spack_repo/builtin/packages/enzyme/package.py b/repos/spack_repo/builtin/packages/enzyme/package.py
```
hughcars: Is there an upstream spack-packages PR for this? `spack_repo/patches/README.md` asks for the PR number in the name so the patch can be dropped when it lands; without one it's carried indefinitely.

hughcars (commented):
Needs a rebase before I re-review: the Linux workflow, `CHANGELOG.md` and the `Makefile` conflict with main (all mechanical), and the changelog entry has to move to a new In progress section now that 0.18.0 and 0.18.1 have shipped. While there, it says optimal convergence on affine and curved tetrahedra; the convergence sweep is on affine hexahedra and curved tetrahedra, with affine tets only checked for polynomial exactness.

The MMS setup holds up: I rederived the sources, the Neumann flux sign on the x = 1 face and the rate tolerances. No config keys are added, so nothing is needed in the schema or SchemaVer.

Keeping Enzyme off by default and in its own test object library is the right shape. The scope question I'd settle before the details is inline: whether the verification hooks belong in the production operator and driver.

PR #902, opened 2026-08-26
Add test verifying schema defaults match their subschemas

Sbozzolo:
JSON Schema treats "default" as an annotation and does not validate it against the subschema that declares it, so a typo like {"type": "integer", "default": "many"} passes the meta-schema and only misbehaves once the default is applied.

Add a [schema] unit test that walks the embedded config schema and validates every "default" against its declaring subschema (attaching the root $defs so internal "#/$defs/..." refs resolve).

### DIFF
--- a/test/unit/test-schema.cpp
+++ b/test/unit/test-schema.cpp
@@ -2,11 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <array>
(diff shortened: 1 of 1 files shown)


hughcars (approved):
Merges cleanly and passes on the current schema. Rebase on main then good to go.

PR #906, opened 2026-08-26
Add AMR iteration index to palace.json metadata, make writemetadata atomic

Sbozzolo:
Record the most recently completed AMR iteration as "Problem"/"Iteration" in palace.json, giving external tooling a machine-readable signal that an AMR iteration has finished without parsing the stdout log. The index is 1-based and aligned with the "iterationXX" archive subdirectories and palace_parser (the initial solve is iteration 1).

On this, I personally find the indexing confusing. The base solve ends up in the `iteration01` folder and carries the `iteration: 1` marker. I think this is inconsistent with how we define iterations elsewhere. For instance, `MaxIts = 0` means only base solve and no AMR iterations.

However, fixing the indexing in the `iteration` folders to be 0 based would be a breaking change.

In this, I also change `WriteMetadata` to be atomic (by first writing a temporary file), so there is never a case with a partial palace.json file written to disk.


CHANGELOG.md
```diff
@@ -56,6 +56,9 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
   - Enable circuit synthesis with frequency-dependent boundary conditions, including
     wave ports with a new optional `IncludeInSynthesis` wave-port flag. SchemaVer 1-4-0
     [PR 811](https://github.com/awslabs/palace/pull/811).
+  - Added a `"Problem"`/`"Iteration"` field to the `palace.json` metadata file, recording the
+    1-based index of the most recently completed adaptive mesh refinement (AMR) iteration
+    (`1` for the initial solve, matching the `iterationXX` archive subdirectory).
```
simlapointe, line 61: PR link is missing

simlapointe (approved):
LGTM (one nit inline comment). 

I agree the iteration numbering is slightly confusing, we might want to revisit that later.

CHANGELOG.md
Sbozzolo: Thank you!

PR #908, opened 2026-08-26
Require CMake 3.24 for standalone builds

Sbozzolo:
Cmake had an incorrect lower bound for the minimal version required. Now, it matches the one defined in spack and in the documentation.


hughcars (approved):

hughcars:
One question, is this modeled in the spack recipe too?

Sbozzolo:
> One question, is this modeled in the spack recipe too?

Yes, it was already fixed in spack

PR #909, opened 2026-08-27
Accelerate online phase of adaptive sweeps 

### LINKED ISSUE #903: Avoid the full 2D wave-port eigensolve at every driven output frequency.
## Problem

In driven simulations with numeric wave ports, each port's 2D boundary eigenproblem is re-solved at every output frequency. The adaptive fast sweep reduces the 3D solves, but not the port solves: the online phase remains O(number of output frequencies), as the reference documentation notes ("the projected non-quadratic contribution must be updated at each output frequency").

Concrete example: a CPW with two numeric wave ports, 1–100 GHz, 300 output frequencies, finite-conductivity boundaries, `AdaptiveTol: 0.02`. Each port-mode solve takes ~12 s, so the online sweep alone is roughly 2 × 300 × 12 s ≈ 2 hours, dwarfing the PROM construction. Fine frequency grids with wave ports are effectively impractical even when the 3D problem is cheap.

## Suggestion

One possible approach: apply the same reduced-order idea to the port eigenproblem itself. Collect converged port eigenvectors at a few sampled frequencies (e.g. the PROM snapshot frequencies, where full port solves already happen), orthonormalize into a small basis V, and solve the tiny projected eigenproblem V†A(ω)V at output frequencies, with the reduced residual as an error indicator and a full solve as fallback.

Simpler mitigations (warm-starting the eigensolver from the previous frequency's mode, reusing the shifted factorization as a preconditioner) would also help, but only reduce the per-frequency cost rather than remove it.

simlapointe:
Accelerate the online phase of adaptive driven sweeps by:

- Replacing the exact 2D wave-port EVP solve in the adaptive driven sweep online phase with a per-port PROM trained on the offline modes. Closes #903. 
- ~~Improving csv file IO.~~ Will remove from this PR and do in #938 instead.
- Doing more post-processing in the reduced PROM basis when possible.


palace/models/modeoperatorassembly.cpp
```diff
@@ -0,0 +1,437 @@
...
+      else
+      {
+        const auto &ild = mat_op.GetInvLondonDepth();
+        mfem::DenseTensor ild_scalar(1, 1, ild.SizeK());
+        for (int k = 0; k < ild.SizeK(); k++)
+        {
+          ild_scalar(0, 0, k) = ild(0, 0, k);
+        }
+        london_n.AddCoefficient(attr_to_mat, ild_scalar);
+        london_n.NormalProjectedCoefficient(*normal);
```
hughcars: The 1×1 copy was only needed on main because the London term was added into an already-projected coefficient; that second projection also zeroed Ann for non-x normals. `london_n` is separate now, so project the full tensor.
```suggestion
        london_n.AddCoefficient(attr_to_mat, mat_op.GetInvLondonDepth());
        london_n.NormalProjectedCoefficient(*normal);
```

hughcars (commented):
Went through the reduced eigenproblem, the 3D coupling, and the MPI paths. Reduced and exact select modes the same way, so they can only differ where exact would itself switch modes between two training points, which is a reasonable trade. The 3D coupling is exact and the collectives line up in every branch. Ran the 4-excitation `cpw_wave` case on 4 ranks: 301/301 reduced port solves per port, no fallbacks, worst residual 1e-7, online phase 48 s against a 202 s offline.

A few things before I approve:
- `reference.md` "Non-quadratic boundary terms" is now stale, and it would be good to have a paragraph on the port PROM, its guard and fallback, and the stats printout.
- CSV deferral: see #938.
- London term inline.

Rebase on main, three files conflict. Things to keep from main when resolving: #921's `direct_pc_op` block (between the exact assembly and the non-port return), #930's orthogonalization, the lazy ParaView init, and the batched `GetVoltages` for the reduced lumped-port path. The MaxIts/KSPTol changelog line is already in via #921.

simlapointe:
Rebased, fixed a GPU performance issue, and re-enabled the cpw_lumped_adaptive and cpw_wave_adaptive test cases on GPU now that they are on par/faster than their uniform-sweep counterparts.

scripts/schema/config-schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "urn:palace:schema:1-7-0",
+  "$id": "urn:palace:schema:1-6-0",
```
hughcars: This reverts the released 1-7-0 to 1-6-0: the `AdaptiveSParameterOnly` bump and its removal cancelled out on the old base, and after the rebase the net diff is a downgrade. The PR adds no schema fields, so the file should be untouched; `check-schema-version` only checks that the id changed, which is why CI didn't catch it.

```suggestion
  "$id": "urn:palace:schema:1-7-0",
```

test/unit/regression/cases.cpp
```diff
@@ -1317,6 +1391,10 @@ TEST_CASE("adapter_driven_synth", "[Serial][Parallel][Regression]")
   opts.custom_checks["port-S.csv"] = TestWavePortSRoundTrip(1.0e-6);
   // No field output is requested in the config.
   opts.paraview_fields = false;
+  // AdaptiveTol is 1e-6 for this case. Compare the physical complex S-parameters over all
+  // 61 output frequencies much more tightly than the generic 2% synthesis/eigenvalue
+  // tolerance while retaining a small absolute allowance near zeros.
+  opts.custom_checks["port-S.csv"] = ComparePortSParameters(1.0e-4, 1.0e-7);
```
hughcars: `custom_checks` is a map, so this replaces main's `TestWavePortSRoundTrip(1.0e-6)` from line 1391, dropping the de-embed/reconnect round trip (the only check on the `rom-coupled-{S,G,H}` outputs) for exactly the synthesis path this PR changes. Running both:

```suggestion
  // AdaptiveTol is 1e-6 for this case, so also compare the physical complex S-parameters
  // over all 61 output frequencies, much more tightly than the generic 2%
  // synthesis/eigenvalue tolerance while retaining a small absolute allowance near zeros.
  opts.custom_checks["port-S.csv"] = [round_trip = TestWavePortSRoundTrip(1.0e-6),
                                      compare_s = ComparePortSParameters(1.0e-4, 1.0e-7)](
                                         palace::Table &actual, palace::Table &reference,
                                         const std::filesystem::path &actual_path)
  {
    round_trip(actual, reference, actual_path);
    compare_s(actual, reference, actual_path);
  };
  // No field output is requested in the config.
  opts.paraview_fields = false;
```

palace/models/romoperator.cpp
```diff
@@ -1104,257 +1295,271 @@ void RomOperator::SolvePROM(int excitation_idx, double omega, ComplexVector &u)
...
     {
-      Ar.noalias() += g * rm.vk_V * rm.Vh_cvk.transpose();
+      if (port_data.excitation != excitation_idx)
+      {
+        continue;
+      }
+      const auto it = sV_wp_full.find(port_idx);
+      MFEM_VERIFY(it != sV_wp_full.end(),
+                  "Missing wave port mode pairing for excited port " << port_idx << "!");
+      RHSr += std::complex<double>(0.0, -2.0 * omega) * it->second;
```
hughcars: `sV_wp_full` only holds active ports, but a port with `"Active": false` can still carry an `Excitation` (nothing rejects it, and main's paths handle it), so this aborts at the first online frequency. Falling back to the assembled excitation covers it; `ProjectVecInternal` with `n0 = 0` overwrites `RHSr`, and `RHS1` is added after the loop:

```suggestion
      const auto it = sV_wp_full.find(port_idx);
      if (it == sV_wp_full.end())
      {
        // Excited but inactive port: no port-space pairing is cached, so use the assembled
        // excitation vector for this excitation instead.
        space_op.GetExcitationVector2(excitation_idx, omega, RHS2);
        ProjectVecInternal(space_op.GetComm(), V, RHS2, RHSr, 0);
        break;
      }
      RHSr += std::complex<double>(0.0, -2.0 * omega) * it->second;
```

palace/models/postoperator.cpp
```diff
@@ -2077,6 +2078,215 @@ auto PostOperator<solver_t>::MeasureAndPrintAll(int ex_idx, int step,
...
+    }
+  }
+
+  const auto &basis = rom_op.GetBasis();
+  const std::size_t n = basis.size();
+  if (n == 0)
+  {
+    return;
+  }
+  std::vector<Vector> basis_E, basis_B;
```
hughcars: Non-blocking: this materialises the whole PROM basis twice as L-vectors (ND and RT) before projecting, a transient of about 2x `V` on top of `V` itself, on the GPU runs this PR is speeding up. The same projection only needs one scratch L-vector per basis column, as `ProjectMatInternal` does.

palace/models/postoperator.cpp
```diff
@@ -2077,6 +2078,215 @@ auto PostOperator<solver_t>::MeasureAndPrintAll(int ex_idx, int step,
...
+      measurement_cache.domain_E_field_energy_all = exact_E;
+      measurement_cache.domain_H_field_energy_all = exact_H;
+      measurement_cache.domain_E_field_energy_i = std::move(exact_domain_E);
+      measurement_cache.domain_H_field_energy_i = std::move(exact_domain_H);
+      Mpi::Warning("Reduced domain-energy postprocessing disagrees with full evaluation "
+                   "(relative error {:.3e}); reverting to full postprocessing.\n",
+                   max_error / std::max(max_reference, 1.0e-300));
+    }
+  }
+  const auto port_voltages = fem_op->GetLumpedPortOp().GetVoltages(*E);
```
hughcars: Non-blocking: this repeats `MeasureLumpedPorts`' `I_RLC`/`I`/`S`/energy formulas minus `GetPowers`, so a later change to the S or energy convention in one place would silently desynchronise the adaptive CSVs from the uniform ones. A shared helper (or `MeasureLumpedPorts(bool with_power)`) would keep them together.

hughcars (changes requested):
Everything from my last pass is in. Two rebase artifacts need fixing (schema id, dropped round-trip check), plus an inactive-port abort, all inline. The Default, MUMPS and STRUMPACK GPU jobs never got a runner and were cancelled, so they need a re-run before merge.

PR #911, opened 2026-08-28
Add mesh topology metadata to palace.json for SaveAdaptMesh

Sbozzolo:
When Model.Refinement.SaveAdaptMesh is enabled, emit a `"SaveAdaptedMesh"` block to `postpro/palace.json `recording the true (conforming) topological entity counts (vertices, edges, faces), the element count, and the sorted domain and boundary attributes of the final adapted mesh. This lets external tooling size a re-run from the saved mesh without re-reading it.

### DIFF
--- a/test/unit/test-basesolver.cpp
+++ b/test/unit/test-basesolver.cpp
@@ -45,12 +45,9 @@ public:
   }
 };
 
-// Reconstruct the exact conforming DOF count (GetTrueVSize) of an FE space from the
-// per-geometry true entity counts: each true entity contributes fec.DofForGeometry(geom)
-// dofs of its geometry. This is the reconstruction oracle used by the nonconforming test
-// below -- an earlier NC-list bucketing method silently mis-counted (over-counted edges,
-// mis-split faces) in 3D, and reproducing GetTrueVSize this way catches that whole class
-// of bug. Mirrors the DofForGeometry usage validated exact in the mfem_dofcheck_nc harness.
+// Reconstruct the exact conforming DOF count (`GetTrueVSize`) from the per-geometry
+// true entity counts. Agreement with `GetTrueVSize` verifies that constrained entities
+// are excluded and face geometries are classified correctly on nonconforming meshes.
 long long Reconstruct(const mesh::MeshEntityCounts &c, mfem::FiniteElementCollection &fec)
 {
   long long n = c.true_vertices * fec.DofForGeometry(mfem::Geometry::POINT) +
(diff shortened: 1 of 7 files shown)


test/unit/test-basesolver.cpp
```diff
@@ -31,6 +32,109 @@ std::string ReadFile(const fs::path &path)
...
+    return {};
+  }
+};
+
+// Reconstruct the exact conforming DOF count (GetTrueVSize) of an FE space from the
+// per-geometry true entity counts: each true entity contributes fec.DofForGeometry(geom)
+// dofs of its geometry. This is the reconstruction oracle used by the nonconforming test
+// below -- an earlier NC-list bucketing method silently mis-counted (over-counted edges,
+// mis-split faces) in 3D, and reproducing GetTrueVSize this way catches that whole class
+// of bug. Mirrors the DofForGeometry usage validated exact in the mfem_dofcheck_nc harness.
```
hughcars: `mfem_dofcheck_nc` is not in the repository, so this reference cannot be followed. State the invariant here instead:

```suggestion
// Reconstruct the exact conforming DOF count (`GetTrueVSize`) from the per-geometry
// true entity counts. Agreement with `GetTrueVSize` verifies that constrained entities
// are excluded and face geometries are classified correctly on nonconforming meshes.
```

test/unit/test-basesolver.cpp
```diff
@@ -223,6 +327,454 @@ TEST_CASE_METHOD(palace::test::SharedTempDir,
...
+}
+
+TEST_CASE_METHOD(palace::test::SharedTempDir,
+                 "RebalanceMesh reports exact true counts on a nonconforming hex mesh",
+                 "[basesolver][Serial]")
+{
+  MPI_Comm comm = MPI_COMM_WORLD;
+
+  // Pure-hexahedral nonconforming mesh: only quadrilateral faces, so this is the !tri_faced
+  // path (SQUARE present, TRIANGLE absent).
```
hughcars: `!tri_faced` no longer exists. Describe the behavior this test protects:

```suggestion
  // A pure-hexahedral nonconforming mesh must report quadrilateral true faces only.
```

hughcars (approved):
Looks good. I checked the AMR lifecycle, MPI/root ownership, nonconforming face counting, and the pairing between each saved mesh and its metadata. I don’t see any correctness issues, and the tests cover the important nonconforming and mixed-geometry cases well. The failing Lookahead jobs appear unrelated to this PR.

I left two small inline suggestions to remove references to intermediate development code. One other nit: the PR description still says `SaveAdaptMesh` rather than `SavedAdaptedMesh` and describes only the final mesh; it would be worth updating to match the implementation.

PR #912, opened 2026-08-28
Fix MatrixFunction for fully anisotropic material tensors

### LINKED ISSUE #907: Explicit eigensolver in densematrix is incorrect in the most general case
https://github.com/awslabs/palace/blob/8b9f21ccfeee3c08b71ca3ecf0c996214cd8b157/palace/linalg/densematrix.cpp#L153

I noticed that this is not symmetric with respect to a, b, c (the term `-ac` is missing).

As a result, `MatrixSqrt` and `MatrixPow` return silently-wrong results for symmetric 3×3 matrices whose  three off-diagonal entries are all nonzero (i.e. fully anisotropic material tensor). There is no crash or warning: the returned matrix is symmetric and plausible-looking but numerically wrong by tens of percent.

When hit, the wrong MatrixSqrt/MatrixPow propagates into:

 - models/materialoperator.cpp: `mat_invz0 = √(μ⁻¹ε)` (far-field/absorbing boundary impedance, 2D mode-solver impedance), `mat_c0 = (με)^(−1/2)`
    (second-order ABC term, Stratton–Chu far-field integrals, Floquet ports, light-speed min/max), and `mat_epsilon_abs`.
- linalg/errorestimator.cpp: √ε, ε(−1/2) and the μ variants used as error-indicator weights (affects AMR).

My first attempt at fixing this by simply fixing the term uncovered more issues:

1. The discriminant 4·x1^3− x2^2 is >= 0 in exact arithmetic (zero at repeated eigenvalues), but goes slightly
 negative in floating point there; it is passed to std::sqrt unclamped, producing NaN. This shows up for the near-isotropic case once x1 is fixed.
2. With a repeated eigenvalue the `SafeDivide` ratios collapse to 0 and the reconstructed f(M) is wrong even when the eigenvalues themselves are correct. This is a limitation of the explicit formula implemented here. This affects sapphire.

To fully fix this problem, we need to be able to fallback to a general eigensolver.

Sbozzolo:
The general (all off-diagonal nonzero) branch of `linalg::MatrixFunction` used a closed-form 3x3 eigendecomposition with an incorrect first invariant (missing an a*c term) and eigenvector expressions that assume distinct eigenvalues.

As a result `MatrixSqrt`/`MatrixPow` returned silently-wrong results for fully anisotropic material tensors  and broke entirely for degenerate spectra such as rotated uniaxial tensors (e.g. sapphire). This fed wrong values into the material operator's sqrt(mu^-1 eps), (mu eps)^-1/2 and eps*sqrt(I + tand tand^T), and into the flux error-estimator weights.

This PR replaces the explicit formula with `mfem::DenseMatrix::CalcEigenvalues`. The diagonal and single-off-diagonal special-case branches are correct and left unchanged. This is a little slower than the explicit formula, but this function is never used in performance-sensitive regions.

Closes #907

### DIFF
--- a/palace/linalg/densematrix.cpp
+++ b/palace/linalg/densematrix.cpp
@@ -14,16 +14,17 @@ namespace palace
 namespace
 {
 
-// Compute matrix functions for symmetric real-valued 2x2 or 3x3 matrices. Returns the
-// matrix U * f(Λ) * U' for input U * Λ * U'.
-// Reference: Deledalle et al., Closed-form expressions of the eigen decomposition of 2x2
-//            and 3x3 Hermitian matrices, HAL hal-01501221 (2017).
+// Compute matrix functions for symmetric real-valued 1x1, 2x2, or 3x3 matrices. Returns
+// the matrix U * f(Λ) * U' for input U * Λ * U', evaluated from the eigendecomposition
+// computed by mfem::DenseMatrix::CalcEigenvalues.
 mfem::DenseMatrix MatrixFunction(const mfem::DenseMatrix &M,
                                  const std::function<double(const double &)> &functor)
 {
   MFEM_ASSERT(M.Height() == M.Width(),
               "MatrixFunction only available for square matrices!");
   const auto N = M.Height();
+  MFEM_VERIFY(1 <= N && N <= 3,
+              "MatrixFunction only supports 1x1, 2x2, or 3x3 matrices, N: " << N << "!");
   constexpr auto tol = 10.0 * std::numeric_limits<double>::epsilon();
   for (int i = 0; i < N; i++)
   {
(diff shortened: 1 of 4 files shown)


simlapointe (approved):
Good catch!

hughcars (changes requested):
Could we keep the trivial 1×1 early return and use `CalcEigenvalues` for everything else? This removes the sparsity dispatch, the unsupported two-coupling case, and the scale-dependent zero tolerance. The implementation becomes roughly:

```cpp
if (N == 1)
{
  Mout(0, 0) = functor(M(0, 0));
  return Mout;
}

double lambda[3], vec[9];
M.CalcEigenvalues(lambda, vec);
for (int i = 0; i < N; i++)
{
  mfem::Vector v(vec + N * i, N);
  AddMult_a_VVt(functor(lambda[i]), v, Mout);
}
return Mout;
```

With the existing size check tightened to `1 <= N && N <= 3`, `N` is guaranteed to be 2 or 3 after the early return. MFEM returns orthonormal eigenvectors, so no additional normalization should be necessary.

palace/linalg/densematrix.cpp
```diff
@@ -41,175 +42,17 @@ mfem::DenseMatrix MatrixFunction(const mfem::DenseMatrix &M,
...
+  // every nonzero pattern and returns an orthonormal eigenbasis even for degenerate spectra
+  // (a rotated uniaxial material tensor is a common degenerate input). The eigenvectors are
+  // stored consecutively, N doubles each; wrap each as a non-owning Vector (no allocation).
+  double lambda[3], vec[9];
+  M.CalcEigenvalues(lambda, vec);
+  for (int i = 0; i < N; i++)
   {
-    MFEM_ABORT("MatrixFunction only supports 2x2 or 3x3 matrices, N: " << N << "!");
+    const mfem::Vector v(vec + N * i, N);
+    AddMult_a_VVt(functor(lambda[i]) / (v * v), v, Mout);
```
hughcars, line 55: Nit: MFEM returns normalized eigenvectors, so `v * v` should already be one. Keeping this as extra safety is fine.

hughcars (approved):
Thanks, this addresses my feedback.

PR #914, opened 2026-08-30
Surface PETSc and SLEPc test program output on failure

### LINKED ISSUE #649: try_compile hides useful information when it fails

SpookyJumpyBeans:
Closes #649.

## Problem

When a configure-time dependency test fails, the log shows only

```
-- PETSc test program - Failed
-- PETSc test program with static linkage - Failed
CMake Error at CMakeLists.txt:252 (message):
  PETSc could not be found, be sure to set PETSC_DIR
```

The real compile or link error is written to `palace-build/CMakeFiles/CMakeConfigureLog.yaml`. That file is rarely published from CI, so in practice the actual cause is invisible where it is needed most.

## Cause

`check_petsc_build` and `check_slepc_build` already capture the output — `COMPILE_OUTPUT_VARIABLE PETSC_TEST_COMPILE_OUTPUT` and `RUN_OUTPUT_VARIABLE PETSC_TEST_OUTPUT` — and then never use it. In `check_slepc_build` the corresponding `message()` calls are present but commented out.

## Change

Print the captured output when the test fails, choosing based on how it failed:

- test program did not build → `*_TEST_COMPILE_OUTPUT`
- built but exited non-zero → `*_TEST_RUN_OUTPUT`

This is the behaviour `check_libceed_build` already has, so the three dependency checks now report failures the same way. The dead commented-out `message()` lines in `check_slepc_build` are replaced by the same conditional.

Both the `try_run` path and the `try_compile` path used when `PALACE_WITH_CUDA`/`PALACE_WITH_HIP` is set are covered; in the latter, `*_TEST_COMPILED` is false on failure and the compile output is printed.

A successful configure is unchanged and prints nothing extra. Only the failure path is affected, so this does not add noise to normal builds.

## Testing

- `cmake -P palace/cmake/PkgConfigHelpers.cmake` parses the modified file (execution then stops at the pre-existing `find_package(PkgConfig REQUIRED)`, as expected off-Linux).
- Exercised the new conditional directly with all three states — compile failure, run failure, and success — and confirmed it selects the compile output, the run output, and prints nothing, respectively.

I do not have a machine with PETSc/SLEPc available to trigger a real failing configure, so the end-to-end failure output has not been observed on a live build. The logic is a direct parallel of the existing libCEED path.


hughcars:
Thanks for this — it addresses the same issue (#649) as #880, which has already been through a review round and additionally handles the cross-compile `try_run` sentinel and the deferred `--static` fallback diagnostics. Closing this one as a duplicate in favour of #880; the approach here (mirroring `check_libceed_build`) is the same one that lands there.

PR #915, opened 2026-08-30
Fix parenthesis in test

Sbozzolo:
The `energy_participation` argument uses `(1 + randd(9999) / 10000)`, giving values always larger than 1.


dnpham23 (approved):
Nice catch!

hughcars (approved):

PR #918, opened 2026-09-02
Bump MFEM to v4.10 and drop patches merged upstream

dnpham23:
## Summary

MFEM **v4.10** is released and now includes several PRs Palace was carrying as local backport patches. This bumps `EXTERN_MFEM_GIT_TAG` to v4.10 and cleans up the patch set accordingly.

## Patches dropped (now merged upstream in v4.10)

- `mfem_pr4983.diff`, `mfem_pr5246.diff`, `mfem_pr5415.diff`, `mfem_pr5124_cudss.diff`
- `mfem_gmsh_parser_performance.diff`: v4.10 rewrote the Gmsh reader into a new `mesh/gmsh.cpp` (signature changed to `ReadGmshMesh(istream&)`) and already adopts its main optimization (`unordered_map` vertex map). The patch targeted the old `mesh_readers.cpp` code that no longer exists, so it can't be rebased and is largely superseded. Only the `reserve()` preallocations are lost. 

## Patches kept (still unmerged upstream)

- `mfem_pr3847.diff`, `mfem_pr5353.diff`

## Code changes

PR 4983's public API changed between our backport and the merged version, so `palace/models/surfacecurlsolver.cpp` is updated:

- `GetBoundaryElementsByAttribute` now returns a position-indexed `std::vector<Array<int>>` (was a `std::unordered_map<int, Array<int>>` keyed by attribute value).
- `GetBoundaryEdgeDoFs` → `GetBoundaryLoopEdgeDofs`, with reordered/retyped outputs (`dof_edges` is now `Array<int>*`). The hole DoF-to-edge map is rebuilt from the two parallel output arrays; only its edge-id values are consumed downstream by `mesh::MatchBoundaryEdges`.

### DIFF
--- a/cmake/ExternalGitTags.cmake
+++ b/cmake/ExternalGitTags.cmake
@@ -133,7 +133,7 @@ set(EXTERN_MFEM_GIT_BRANCH
   "Git branch for external MFEM build"
 )
 set(EXTERN_MFEM_GIT_TAG
-  "d9d6526cc1749980a2ba1da16e2c1ca1e07d82ec" CACHE STRING
+  "d964264cdb9a13e94a201b6c236c7721e0c8765f" CACHE STRING
   "Git tag for external MFEM build"
 )
 
--- a/cmake/ExternalMFEM.cmake
+++ b/cmake/ExternalMFEM.cmake
@@ -419,43 +419,11 @@ download_mfem_patch(
   "https://github.com/mfem/mfem/compare/2d574015756711029556c14d096ca52c15d5b663...50ead1a9a785e3273b2a72ff59ac8ed8a496b498.diff"
   e9be1a0d4b2642ed1b72f31c36065cf0aacbe342b6594d5d943782c73a6177f4
 )
-# https://github.com/mfem/mfem/commit/e4a2b9568c40f20e24612066d155cc6a9973b247
-download_mfem_patch(
-  mfem_gmsh_parser_performance.diff
-  "https://github.com/mfem/mfem/commit/e4a2b9568c40f20e24612066d155cc6a9973b247.diff"
-  6ced66f487780af66fb8184d329b9aad4b694e711b65830391e8c6d0c898713e
-)
-# https://github.com/mfem/mfem/pull/5246
-download_mfem_patch(
-  mfem_pr5246.diff
-  "https://raw.githubusercontent.com/awslabs/palace/b22f654ab36fe01f1f3176349c60626efed1a6a2/extern/patch/mfem/mfem_pr5246.diff"
-  d5227c18768369b8fa3a20f4457dd378a360346850329ab1970d18ed5a73b0d6
-)
 # https://github.com/mfem/mfem/pull/5353
 download_mfem_patch(
   mfem_pr5353.diff
   "https://raw.githubusercontent.com/awslabs/palace/b22f654ab36fe01f1f3176349c60626efed1a6a2/extern/patch/mfem/mfem_pr5353.diff"
   c35f584090f97c84c12fc80e6d5c068512911d192132e18f5aa4254f507c5e4f
-)
-# https://github.com/mfem/mfem/pull/4983
-download_mfem_patch(
-  mfem_pr4983.diff
-  "https://raw.githubusercontent.com/awslabs/palace/b22f654ab36fe01f1f3176349c60626efed1a6a2/extern/patch/mfem/mfem_pr4983.diff"
-  530532da3ae8815d004bb6ce19f6f08a1248c3d585503551c90d0eeae7fb3f87
-)
-if(PALACE_WITH_CUDSS)
-  # https://github.com/mfem/mfem/pull/5124
-  download_mfem_patch(
-    mfem_pr5124_cudss.diff
-    "https://raw.githubusercontent.com/awslabs/palace/b22f654ab36fe01f1f3176349c60626efed1a6a2/extern/patch/mfem/mfem_pr5124_cudss.diff"
-    d0b5893ec7925cbc8a70cc5eba2abe037754b99b5bb366c4bae90f0583d22290
-  )
-endif()
-# https://github.com/mfem/mfem/pull/5415
-download_mfem_patch(
-  mfem_pr5415.diff
-  "https://github.com/mfem/mfem/commit/9d1438d8a2502cc927c63e093cf8c855ff17918e.diff"
-  482655b6b740b880713d67bcca843571244b7d383c95e0cef3d3102b3327ff2f
 )
 
 include(ExternalProject)
--- a/palace/models/surfacecurlsolver.cpp
+++ b/palace/models/surfacecurlsolver.cpp
@@ -65,20 +65,24 @@ void SolveSurfaceCurlProblem(const SurfaceFluxData &flux_data, const IoData &iod
   int num_holes = hole_surface_attrs.Size();
 
   // Extract hole boundary DOFs
-  std::unordered_map<int, mfem::Array<int>> attr_to_elements;
+  std::vector<mfem::Array<int>> attr_to_elements;
   std::vector<std::unordered_map<int, int>> hole_dof_to_edge_maps(num_holes);
-  std::vector<mfem::Array<int>> hole_ess_tdof_lists(num_holes);
-  std::vector<mfem::Array<int>> hole_ldof_markers(num_holes);
-  std::vector<std::unordered_set<int>> hole_boundary_edge_ldofs(num_holes);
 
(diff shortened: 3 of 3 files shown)


hughcars:
Can probably use the same trick as https://github.com/llnl/axom/pull/1981#discussion_r3971888074 before 4.10 mfem hits their spack recipe if need be. I'd prefer to put that in and cut an to remove it later, than have our CI blocked by them not updating their recipe.

dnpham23:
Thanks, @hughcars ! I took that route and vendored [spack-packages#6314](https://github.com/spack/spack-packages/pull/6314) as `spack_repo/patches/pr6314_mfem_4_10_0.diff`. It adds `mfem@4.10.0` to the pinned snapshot and auto-skips once it lands upstream, so no snapshot bump is needed and it's easy to drop later.
Two follow-ons surfaced from the bump:

1. CUDA build: MFEM 4.10 merged the multi-CUDA-arch makefile change (mfem#5440), so our vendored `pr5895_mfem_multi_cuda_arch.diff` was re-applying that source patch on 4.10 and failing with "Reversed patch detected". I capped that patch to `@4.9.0` (the `@4.9.0`: recipe logic is unchanged since 4.10 ships the same makefile). Verified locally with spack 1.2 on the pinned snapshot: palace concretizes with `mfem@4.10.0` + its four retained patches, and the CUDA patch no longer applies to 4.10.
2. Lookahead: both jobs fail at concretization, not build because develop-deps injects `mfem: require: "@develop"`, which conflicts with our exact `depends_on mfem@4.10.0`. Maybe we can loosen the pin to `mfem@4.10.0`: lets Lookahead concretize against `mfem@develop` while normal builds still resolve to `4.10.0` (the @4.10.0-gated patches are unaffected)? I can make that change if you'd prefer, or leave the exact pin as-is.

spack_repo/local/packages/palace/package.py
```diff
@@ -203,7 +203,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
     with when("@0.16:"):
         # +lapack means: use external lapack
         depends_on(
-            "mfem+mpi+metis+lapack@4.9.0",
+            "mfem+mpi+metis+lapack@4.10:",
```
hughcars: This moves palace 0.16–0.18.1 onto MFEM 4.10 and drops the 4.9 patches they need, so Spack can no longer build those releases. Split by Palace version: keep a `with when("@0.16:0.18")` block with `mfem@4.9.0` and main's 4.9 patch list unchanged, and put this `mfem@4.10:` dependency with the four remaining patches under `with when("@0.19:")`. `develop` satisfies `@0.19:`.

spack_repo/local/packages/palace/package.py
```diff
@@ -237,38 +220,16 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
...
-                    "https://raw.githubusercontent.com/awslabs/palace/"
-                    "1382ca5e9f72369b33c0ff5e8e0a244ac6597f6a/extern/patch/mfem/"
-                    "mfem_nc_partition_fixes.diff",
-                    sha256="a28bf879ecf197856d24ef5427d493f84a159224c22e7cd17ec977e06a222ffb",
-                    when="@4.9.0",
+                    "https://github.com/mfem/mfem/compare/"
+                    "10b0b596dbc26dca384b9f25f23026b1a6403592..."
+                    "399d72c2b7607e25d196dc19fd70b8840eaf1cc8.diff",
+                    sha256="771d758ce461c7353d2b33ff6b5c9fd5404cae77a0340c1a67c3935ce606fd4c",
+                    when="@4.10.0",
```
hughcars: `@4.10:` so this also applies to develop. When mfem#5494 lands upstream it stops applying, Lookahead fails, and we narrow it to `@4.10.0`. It applies cleanly to MFEM master (45799ebe) today.
```suggestion
                    when="@4.10:",
```

spack_repo/local/packages/palace/package.py
```diff
@@ -278,7 +239,7 @@ class Palace(CMakePackage, CudaPackage, ROCmPackage):
                     "https://github.com/mfem/mfem/commit/"
                     "3091ba40b238c4008b67216314bb26da6738b833.diff",
                     sha256="52ccf3332f87aaf7ebc84674448226201c04343a3e298006bea5fd8be8e92533",
-                    when="@4.9.0",
+                    when="@4.10.0",
```
hughcars: Same as #5494.
```suggestion
                    when="@4.10:",
```

hughcars (commented):
Keeping `@4.10:` is fine. Two things inline so older releases still build and Lookahead tells us when our patches land upstream.

spack_repo/local/packages/palace/package.py
dnpham23: Done. Split into `depends_on("mfem@4.9.0", when="@0.16:0.18", …) `with the unchanged 4.9 patch list and `depends_on("mfem@4.10:", when="@0.19:", …)` with the four remaining patches. One tweak: I used the `when= arg` on each `depends_on` rather than a with `when("@0.16:0.18")` block. The context-manager form folds the Palace range into the MFEM patch `when=` conditions, so `@0.16:0.18` makes `mfem@4.9.0`'s patches drop/error (`0.16:0.18` does not satisfy `4.9.0`). `main`'s with `when("@0.16:") `only worked because mfem 4.x satisfies the open bound. Verified `palace@0.18.0` -> `mfem@4.9.0` with the same 9 patches as main.

spack_repo/local/packages/palace/package.py
dnpham23: Applied. Confirmed it still applies cleanly to mfem master (45799ebe), so `mfem@develop` picks it up in Lookahead and we'll get the signal to narrow to `@4.10.0` when it merges.

spack_repo/local/packages/palace/package.py
dnpham23: Applied, same as #5494. Applies cleanly to mfem master today and rides along on `mfem@develop.`

hughcars (approved):
Needs a changelog line under Build system for the MFEM 4.10 bump, then good to go. Nice to see Lookahead green again.

hughcars:
https://github.com/spack/spack-packages/pull/6314 was merged, so should be able to use the direct tag now

Sbozzolo:
> [spack/spack-packages#6314](https://github.com/spack/spack-packages/pull/6314) was merged, so should be able to use the direct tag now

This requires bumping the version of spack-packages in our repo. I am doing that in https://github.com/awslabs/palace/pull/969

spack_repo/patches/pr5895_mfem_multi_cuda_arch.diff
Sbozzolo: If you are changing this, you should also change PR5895 in spack-packages

hughcars:
Ok, we can either merge this as is, then follow up after #969 goes in with another, or wait, rebase on top of that and use the proper tag, your choice @dnpham23.

spack_repo/patches/pr5895_mfem_multi_cuda_arch.diff
dnpham23: Right, #5895 is yours so I can't push to it. FYI for #969: I gated our copy to `@4.9.0` since 4.10 has the makefile change (mfem#5440) merged, but the Python `CUDA_ARCH=sm_%s` is still single-arch and breaks multi-arch on 4.10. Yours to fold in, or I can PR against your branch.

dnpham23:
Let's go with merging this as-is. It's self-contained (pr6314 auto-skips once the snapshot has 4.10.0), so no need to block it on #969. I'll follow up after #969 lands to drop pr6314 and reconcile the rest.

spack_repo/patches/pr5895_mfem_multi_cuda_arch.diff
Sbozzolo: Does this mean that MFEM 4.10 does not support multi cuda arch?

Sbozzolo:
I can include dropping pr6314 in #969.

spack_repo/patches/pr5895_mfem_multi_cuda_arch.diff
dnpham23: 4.10 itself supports it i think.  The makefile change (mfem#5440) that accepts a comma-separated CUDA_ARCH is merged in 4.10. The gap is only in the spack recipe: it emits single-arch CUDA_ARCH=sm_%s, so it never feeds the arch list to 4.10's makefile. So for #5895 the source patch is no longer needed on 4.10, but the Python comma-join still is.

PR #919, opened 2026-09-02
Fix nonconformal AMR issues in simulations with wave ports

simlapointe:
3D simulations using wave ports and nonconformal AMR sometimes encountered issues when NC refinement affected the wave port mesh:

- Hanging edges with the wave port boundary attribute appeared in the wave port mesh, corrupting the 2D EVP solution.
- The `VoltagePath` line integral was affected by new DOFs introduced by the refinement.

This PR fixes both issues by repairing the submesh extraction and parent-submesh DOF transfers.


palace/utils/geodata.cpp
```diff
@@ -1527,90 +1526,375 @@ void RemapSubMeshBdrAttributes(SubMeshT &submesh, const mfem::Array<int> &surfac
...
+  {
+    const int q = first_local_query + i;
+    int min_attr = match_min[q], max_attr = match_max[q];
+    if (min_attr == no_min_match)
+    {
+      min_attr = face_match_min[q];
+      max_attr = face_match_max[q];
+    }
+    MFEM_VERIFY(
+        min_attr == no_min_match || min_attr == max_attr,
```
dnpham23: With this, a port perimeter edge that lies along a PEC/Impedance junction (or coincides with a parent vertex shared by two non-surface attrs) produces a midpoint the geometric fallback matches to two distinct non-surface attributes -> hard abort. Would it be better to have a deterministic tie-break with a warning over aborting?

palace/utils/geodata.cpp
```diff
@@ -1527,90 +1526,375 @@ void RemapSubMeshBdrAttributes(SubMeshT &submesh, const mfem::Array<int> &surfac
...
-                  recv_counts.data(), displs.data(), comm);
+  std::vector<int> all_edge_attrs(attr_total);
+  Mpi::Allgatherv(local_attr_count, local_edge_attrs.data(), all_edge_attrs.data(),
+                  attr_counts.data(), attr_displs.data(), comm);
 
-  // Build resolved global vertex pair → attribute map. For edges shared by multiple parent
-  // boundary faces, prefer the face that is NOT part of the mode analysis surface.
+  // For edges shared by multiple parent boundary faces, prefer a face outside the selected
+  // mode surface. This preserves the existing remapping convention for conforming edges.
   std::map<std::pair<int, int>, int> gvpair_to_attr;
```
dnpham23: Conforming map (1568-1582) picks last-seen non-surface among duplicates; geometric fallback aborts on multiple distinct non-surface. Aligning them (e.g. min_attr) removes the abort in line 1876 for free.

palace/utils/geodata.cpp
```diff
@@ -1527,90 +1526,375 @@ void RemapSubMeshBdrAttributes(SubMeshT &submesh, const mfem::Array<int> &surfac
...
-    int gv0 = static_cast<int>(pvert_gi[ev[0]]);
-    int gv1 = static_cast<int>(pvert_gi[ev[1]]);
-    auto key = std::make_pair(std::min(gv0, gv1), std::max(gv0, gv1));
+    parent.GetEdgeVertices(parent_edge_map[submesh_edge], vertices);
+    const int gv0 = static_cast<int>(pvert_gi[vertices[0]]);
+    const int gv1 = static_cast<int>(pvert_gi[vertices[1]]);
+    const auto key = std::make_pair(std::min(gv0, gv1), std::max(gv0, gv1));
     auto it = gvpair_to_attr.find(key);
-    if (it != gvpair_to_attr.end())
+    if (it != gvpair_to_attr.end() && surface_attr_set.count(it->second) == 0)
```
dnpham23: This now requires the topologically-resolved attr be non-surface, else routes to geometric fallback. A genuinely-interior perimeter edge (only adjacent parent face is the surface itself) now stays unresolved, which leads to warning + default attr. Just wanted to confirm this is intended?

palace/utils/geodata.cpp
```diff
@@ -1527,90 +1526,375 @@ void RemapSubMeshBdrAttributes(SubMeshT &submesh, const mfem::Array<int> &surfac
...
+    }
+  }
+
+  int global_remapped = remapped;
+  Mpi::GlobalSum(1, &global_remapped, comm);
+  if (global_remapped < global_unresolved_count)
+  {
+    Mpi::Warning(
+        fmt::format("Could not remap {:d}/{:d} NC child edges on the submesh perimeter!",
+                    global_unresolved_count - global_remapped, global_unresolved_count));
```
dnpham23: "Could not remap N/M NC child edges" is a `Mpi::Warning`; those edges silently keep their default attr, which feeds back into the essential-DOF extraction fix in `waveportoperator.cpp` (the new `GetEssentialTrueDofs`, ~L51-67). Might worth considering whether an unresolved perimeter edge should be fatal for wave ports specifically.

palace/models/waveportoperator.cpp
```diff
@@ -1013,25 +996,26 @@ std::complex<double> WavePortData::GetExcitationVoltage() const
...
-  *parent_E0t = 0.0;
-  port_nd_transfer_reverse->Transfer(port_E0t->Real(), parent_E0t->Real());
-  port_nd_transfer_reverse->Transfer(port_E0t->Imag(), parent_E0t->Imag());
+
+  // The flattened companion has identical topology, orientation, and ND ordering. Copying
+  // all local DoFs preserves NC slave values and avoids the lossy submesh-to-parent
+  // transfer.
+  MFEM_VERIFY(voltage_port_E0t->Real().Size() == port_E0t->Real().Size(),
+              "Flattened wave port field size mismatch!");
+  voltage_port_E0t->Real() = port_E0t->Real();
```
dnpham23: The direct ND-dof copy assumes the port face is planar (orthogonal projection = isometry). Non-planar/curved port faces would distort |V|. Maybe document the planarity assumption or adding a guard?

test/unit/test-2d-submesh.cpp
```diff
@@ -208,6 +209,132 @@ TEST_CASE("RemapSubMeshBdrAttributes", "[geodata][Serial]")
...
+  {
+    transferred_essential += transferred_true[i] != 0.0;
+  }
+  // Parent-marker transfer over-constrains two interior H1 DoFs on this NC surface.
+  // Essential DoFs must instead be extracted directly from the remapped submesh boundary.
+  CHECK(direct_essential.Size() == 6);
+  CHECK(transferred_essential == 8);
+}
+
+TEST_CASE("Flattened NC port mesh preserves ND DoFs", "[geodata][Serial]")
```
dnpham23: This is `[Serial]` only; the DOF-copy relies on identical L-vector ordering between the SpaceDim-3 and SpaceDim-2 ND spaces, which is a parallel concern. Maybe add a `[Parallel]` companion case?

dnpham23 (commented):
Thanks so much @simlapointe for this fix! I have just a few inline comments.

test/unit/test-2d-submesh.cpp
simlapointe: Added [Parallel] coverage to the same regression and made the GSLIB segment selection collective-safe for ranks with no local elements. It passes locally with 1, 2, and 6 MPI ranks, as well as the MFEM debug-device backend (95b695490a).

palace/utils/geodata.cpp
simlapointe: I checked this against the NC regression fixture: 5/9 unmatched child edges are legitimate artificial NC submesh boundaries with no adjacent non-surface parent face, and they must remain unconstrained. Making all unresolved edges fatal would reject that valid case. I kept the warning (since the same outcome can indicate missing physical geometry) and documented why the generated attribute is retained.

palace/utils/geodata.cpp
simlapointe: Agreed on aligning the two paths. The conforming map is now deterministic (minimum distinct non-surface attribute) rather than last-seen, and the geometric fallback uses the same rule. Both warn on ambiguity instead of aborting (95b695490a).

palace/models/waveportoperator.cpp
simlapointe: Yes, planarity is an existing assumption of the wave-port formulation as well as this ND-DoF copy. I documented explicitly in the wave-port guide that the complete port face must be planar and that the boundary-mode solve and VoltagePath use one tangent frame/normal (95b695490a).

palace/utils/geodata.cpp
simlapointe: Good catch. I verified that multiple non-surface matches can occur in the existing conforming lookup, so aborting only in the geometric fallback was too strict and inconsistent. Both paths now choose the smallest attribute deterministically and emit a warning when an ambiguous edge is actually used by the submesh perimeter (95b695490a).

palace/utils/geodata.cpp
simlapointe: This routing is intentional. For an NC child edge, a surface-only topological parent can be a coarse edge of which the child is a strict subset while the child physically lies on a nonmatching adjacent boundary face. Such cases need the geometric fallback. I added a comment explaining this. If the fallback finds no non-surface geometry, the generated submesh attribute is retained.

simlapointe:
> Thanks so much @simlapointe for this fix! I have just a few inline comments.

Thanks for the useful feedback, @dnpham23. See my responses above

dnpham23 (approved):

palace/utils/geodata.cpp
```diff
@@ -1527,90 +1526,411 @@ void RemapSubMeshBdrAttributes(SubMeshT &submesh, const mfem::Array<int> &surfac
...
+  }
+
+  // NC child edges can be strict subsets of a coarser physical parent edge, or can fall
+  // inside a nonmatching adjacent parent boundary face. Gather only the query points;
+  // every rank tests them against its local parent geometry and reductions combine the
+  // answers. This keeps the geometric work distributed instead of replicating an
+  // O(N_query * N_global_boundary) search on every rank.
+  mfem::IntegrationPoint center;
+  center.Set1w(0.5, 1.0);
+  mfem::Vector physical_midpoint(submesh.SpaceDimension());
```
hughcars: This generic remapper is also called by `BoundaryModeSolver` after its extracted submesh has been projected to 2D. In that path this midpoint has only two coordinates, gets padded with `z = 0`, and is then compared with the original 3D parent geometry. An unresolved NC perimeter edge will usually fail to match. Please remap the boundary attributes before projection, and add a serial extracted-NC regression that exercises that call order.

palace/utils/geodata.cpp
```diff
@@ -1527,90 +1526,411 @@ void RemapSubMeshBdrAttributes(SubMeshT &submesh, const mfem::Array<int> &surfac
...
+    parent.GetBdrElementTransformation(be, &transformation);
+    const auto *vertex_rule =
+        mfem::Geometries.GetVertices(transformation.GetGeometryType());
+    const int nv = vertex_rule->GetNPoints();
+    MFEM_VERIFY((nv == 3 || nv == 4) && edges.Size() == nv,
+                "Expected triangular or quadrilateral parent boundary element!");
+    double corner_coordinates[4][3] = {};
+    for (int j = 0; j < nv; j++)
+    {
+      transformation.Transform(vertex_rule->IntPoint(j), point);
```
hughcars: This samples only the transformed corner vertices, then treats every edge as a straight segment and every face as planar corner triangles. That is not the geometry of a high-order curved boundary. A planar port can meet a curved conductor wall, and an NC child edge on that wall will not lie on the corner chord. Please test containment against the actual isoparametric geometry and add a curved high-order case.

CHANGELOG.md
```diff
@@ -143,6 +143,8 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
   - Fixed intermittent MPI deadlocks in output-directory unit tests when ranks observed
     filesystem changes at different times. [PR
     839](https://github.com/awslabs/palace/pull/839).
+  - Fixed issues with nonconformal AMR corrupting the wave port mesh and VoltagePath line
```
hughcars: This branch predates the 0.18.0 release cut, so rebasing it as-is puts this fix in the already released 0.18.0 notes. Please move it under the current `In progress` bug-fix section for v0.18.1.

palace/models/waveportoperator.cpp
```diff
@@ -603,23 +562,47 @@ WavePortData::WavePortData(const config::WavePortData &data,
...
+    mfem::Vector origin(voltage_path[0]), e1(voltage_path[1]), e2(3);
+    e1 -= origin;
+    e1.Add(-(e1 * port_normal), port_normal);
+    MFEM_VERIFY(e1.Norml2() > 0.0,
+                "Wave port VoltagePath must contain a nonzero tangential segment!");
+    e1 /= e1.Norml2();
+    e2(0) = port_normal(1) * e1(2) - port_normal(2) * e1(1);
+    e2(1) = port_normal(2) * e1(0) - port_normal(0) * e1(2);
+    e2(2) = port_normal(0) * e1(1) - port_normal(1) * e1(0);
+    e2 /= e2.Norml2();
```
hughcars: This is technically modifying the mesh, can we not assume that it was ok to begin with? This seems more like something to measure than to modify. How big were the deviations you were observing?

palace/utils/geodata.cpp
```diff
@@ -1527,90 +1526,411 @@ void RemapSubMeshBdrAttributes(SubMeshT &submesh, const mfem::Array<int> &surfac
...
+  Mpi::Allgatherv(local_attr_count, local_edge_attrs.data(), all_edge_attrs.data(),
+                  attr_counts.data(), attr_displs.data(), comm);
 
-  // Build resolved global vertex pair → attribute map. For edges shared by multiple parent
-  // boundary faces, prefer the face that is NOT part of the mode analysis surface.
+  // For edges shared by multiple parent boundary faces, prefer a face outside the selected
+  // mode surface. If multiple distinct non-surface attributes share the edge, choose the
+  // smallest deterministically and warn below if that edge is actually used by the
+  // submesh perimeter.
   std::map<std::pair<int, int>, int> gvpair_to_attr;
```
hughcars: global vertex pair?

CHANGELOG.md
hughcars: This branch predates the 0.18.0 release cut, so rebasing it as-is puts this fix in the already released notes. Please move it under the current `In progress` bug-fix section when updating the branch.

hughcars (commented):
The two fixes here look right, but the implementation is larger than it needs to be. I pushed a single simplification commit directly on top of this branch: [hughcars/pr919-simplify](https://github.com/awslabs/palace/compare/simlapointe/waveport-order-amr...hughcars/pr919-simplify), commit [`ff9c0163b6`](https://github.com/awslabs/palace/commit/ff9c0163b67c27b6f23e3d5cab909fcc880cc1fe).

It uses MFEM's NC slave/master edge ancestry instead of the distributed midpoint/edge/face geometry search. It also evaluates the existing `port_E0t` field directly from cached element/reference locations, removing the duplicate mesh and finite-element representation. This removes 253 lines.

I built it through Spack and ran the NC essential-DoF and end-to-end NC TE10 `Z_PV` tests in serial and with two MPI ranks. Formatting and `git diff --check` pass.

MFEM [PR #5259](https://github.com/mfem/mfem/pull/5259), included in MFEM 4.10, adds `FindPointsGSLIB` support for triangular surfaces embedded in 3D. Once this PR is merged, please open an issue to replace the one-time exhaustive location setup with direct GSLIB lookup on the existing port submesh after Palace is bumped to MFEM 4.10.

simlapointe:
Thanks for the simplification, @hughcars. I added your commit, fixed a small issue, and rebased on top of main.

hughcars (approved):

PR #921, opened 2026-09-02
Use complex-aware preconditioner in BoundaryMode solver

### LINKED ISSUE #920: BoundaryMode/WavePort solve converges slowly in cases with lossy BCs
In BoundaryMode/WavePort cases with lossy BCs, the solve can take a large number (e.g. >~100) of GMRES iterations, while it usually converges in <5 iterations with PEC boundary conditions, as mentioned in https://github.com/awslabs/palace/issues/903#issuecomment-5515793022. This is because the sparse-direct preconditioner used in the BoundaryMode solve only uses the real part of the system matrix, so convergence is degraded when there are significant imaginary contributions. This issue is to investigate this and improve the BoundaryMode preconditioner so it handles those cases better.

simlapointe:
Improve BoundaryMode linear solver convergence when lossy boundary conditions are present by including complex terms in the real-valued preconditioner ($P = A_r + A_i$, like in the 3D case) or using an exact complex-valued preconditioner when `"WavePorts.ComplexCoarseSolve"` is true. Closes #920.


palace/models/modeeigensolver.cpp
```diff
@@ -532,6 +532,54 @@ ModeEigenSolver::SolveResult ModeEigenSolver::Solve(std::complex<double> omega,
   // Frequency-dependent matrices assemble on the FE space communicator.
   AssembleFrequencyDependent(omega, sigma);
 
+  // The sparse matrix sum and doubled-real block construction below are MPI-collective on
+  // the FE space communicator. Wave-port sparse-direct solvers run on a subcommunicator,
+  // so assemble the complex-aware preconditioner on all ranks before non-port ranks
+  // return. The resulting real operator is already in the form expected by the wrapped
+  // sparse solver, avoiding another collective assembly in MfemWrapperSolver.
```
hughcars:
```suggestion
  // the FE space communicator. Wave-port sparse-direct solvers run on a subcommunicator,
  // so assemble the complex-aware preconditioner on all ranks before non-port ranks
  // return. direct_pc_op only carries the assembled real matrix (2N x 2N in the
  // doubled-real case) into MfemWrapperSolver::SetOperator; its Mult is not meaningful.
```

palace/models/waveportoperator.cpp
```diff
@@ -366,8 +366,14 @@ WavePortData::WavePortData(const config::WavePortData &data,
...
                            const mfem::Array<int> &dbc_attr)
   : mat_op(mat_op), excitation(data.excitation), active(data.active),
-    include_in_synthesis(data.include_in_synthesis)
+    include_in_synthesis(data.include_in_synthesis), port_linear(linear)
 {
+  port_linear.max_it = data.ksp_max_its;
+  port_linear.tol = data.ksp_tol;
+  if (data.complex_coarse_solve)
+  {
+    port_linear.complex_coarse_solve = *data.complex_coarse_solve;
```
hughcars: With global `"PCMatReal": true`, `port_linear.pc_mat_real` is inherited and the `!linear.pc_mat_real` guard in `ModeEigenSolver::Solve` skips this entirely, so an explicit per-port `"ComplexCoarseSolve": true` is silently ignored. Give the explicit override precedence:
```suggestion
    port_linear.complex_coarse_solve = *data.complex_coarse_solve;
    port_linear.pc_mat_real = false;
```

CHANGELOG.md
```diff
@@ -60,6 +60,13 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
     1-based index of the most recently completed adaptive mesh refinement (AMR) iteration
     (`1` for the initial solve, matching the `iterationXX` archive subdirectory).
     [PR 902](https://github.com/awslabs/palace/pull/902)
+  - Improve BoundaryMode linear solver convergence when lossy boundary conditions are present
```
hughcars: The `MaxIts`/`KSPTol` fix is a separate bug, give it its own bullet. This will also need to move to `In progress` when you rebase.

palace/models/modeeigensolver.cpp
hughcars:
```suggestion
  // the FE space communicator. Wave-port sparse-direct solvers run on a subcommunicator,
  // so assemble the complex-aware preconditioner on all ranks before non-port ranks
  // return. direct_pc_op only carries the assembled real matrix (2N x 2N in the
  // doubled-real case) into MfemWrapperSolver::SetOperator; its Mult is not meaningful.
```

palace/models/waveportoperator.cpp
hughcars, line 340: With global `"PCMatReal": true`, `port_linear.pc_mat_real` is inherited and the `!linear.pc_mat_real` guard in `ModeEigenSolver::Solve` skips this entirely, so an explicit per-port `"ComplexCoarseSolve": true` is silently ignored. Give the explicit override precedence:
```suggestion
    port_linear.complex_coarse_solve = *data.complex_coarse_solve;
    port_linear.pc_mat_real = false;
```
Note this also means an explicit per-port `false` gives `Ar + Ai` rather than `Ar`. I think that is the right reading of a per-port choice, but if you would rather preserve `PCMatReal` for that case, only clear it when the override is `true`.

CHANGELOG.md
hughcars: The `MaxIts`/`KSPTol` fix is a separate bug, give it its own bullet. This will also need to move to `In progress` when you rebase.

hughcars (approved):
Confirmed the block assembly has to happen before the non-port ranks return: `HypreParMatrixFromBlocks` allgathers on the FE-space comm. benvial's numbers on #920 are the evidence this works. Nits inline, then rebase onto main once #919 is in and merge.

PR #922, opened 2026-09-04
Fix segfault in NLEPS random-guess fallback

Sbozzolo:
`QuasiNewtonSolver::Solve` dereferenced `GetEssentialTrueDofs()` unconditionally when seeding a random vector after exhausting the initial guesses. That accessor returns `nullptr` on ranks whose partition owns no essential (Dirichlet/PEC) true dofs, so on a distributed mesh those ranks segfaulted. With the crash gone, an unconverged eigenmode solve now returns cleanly and the existing `num_conv` verification reports the failure instead of crashing.

### DIFF
--- a/palace/linalg/nleps.cpp
+++ b/palace/linalg/nleps.cpp
@@ -455,10 +455,16 @@ int QuasiNewtonSolver::Solve()
       eig = sigma;
       if (num_init_guess < 3)
       {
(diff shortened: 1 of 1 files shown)


simlapointe (approved):
LGTM! Thanks for catching and fixing this!

PR #923, opened 2026-09-04
Cleanup zlib usage

Sbozzolo:
Palace had a zlib internal module that was never used. Part of the intended use seemed to be to add support for compressing the SavedAdaptedMesh. However, this can be done natively by MFEM, so I switched to doing that.

Gzipped meshes take 1/3 of the space as plain text ones.

We still need to link zlib to Palace, but once https://github.com/spack/spack-packages/pull/143 lands, we can drop zlib from Palace (because CMake will find the transitive link)


CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ See the [developer notes on schema versioning](https://awslabs.github.io/palace/
 
 #### New Features
 
+  - The adapted mesh saved by `config["Model"]["Refinement"]["SaveAdaptMesh"]` is now
```
hughcars, line 54: This is under the released 0.18.0 section. Move it to `In progress` under `#### Interface Changes`.

palace/utils/outputdir.hpp
```diff
@@ -179,15 +179,31 @@ inline void RemovePreviousOutput(const fs::path &dir, MPI_Comm comm)
...
       {
         fs::remove_all(path);
+        if (compress)
+        {
+          mfem::ofgzstream stream(path.string(), true);
+          if (!stream.good())
+          {
+            throw fs::filesystem_error("Could not open output file", path,
+                                       std::make_error_code(std::errc::io_error));
+          }
```
hughcars: Dead: this `ofgzstream` constructor never propagates `fs_.rdstate()`, and a failed open already throws `strict_fstream::Exception`.
```suggestion
```

test/unit/test-basesolver.cpp
```diff
@@ -321,7 +330,55 @@ TEST_CASE_METHOD(palace::test::SharedTempDir,
...
+  {
+    std::ofstream os(plain_path);
+    serial_mesh.Print(os);
+  }
+
+  Units units(1.0, 1.0);
+  auto load = [&](const fs::path &p)
+  {
+    IoData iodata(units);
+    iodata.model.mesh = p.string();
```
hughcars, line 362: `clean_unused_elements` defaults to true with no domain attributes, so both meshes are emptied and the checks below compare 0 == 0.
```suggestion
    IoData iodata(units);
    iodata.model.mesh = p.string();
    iodata.model.clean_unused_elements = false;
```

hughcars (commented):
This changes the name and format of an existing output, so it's an interface change rather than a cleanup. Fine for 0.18.1 given `SaveAdaptMesh` is opt-in. Fix the changelog placement and the test below, then I'll approve.

Sbozzolo:
> This changes the name and format of an existing output, so it's an interface change rather than a cleanup. Fine for 0.18.1 given `SaveAdaptMesh` is opt-in. Fix the changelog placement and the test below, then I'll approve.

Fixed in d7a9f9b60633358c808b7d04cd01537624fb1b0a

hughcars (approved):

PR #929, opened 2026-09-10
Total inductance extraction for thin-film superconductors

dnpham23:
## Summary
Adds magnetostatic extraction of the total loop inductance (geometric plus London kinetic) for thin-film superconducting structures with holes and current terminals. Flux is imposed as a gauge-invariant fluxoid constraint on each hole, and finite-penetration-depth films carry a London sheet penalty, so the extracted inductance matrix includes the kinetic contribution. Also adds non-conformal AMR support for this path, worked examples, documentation, and regression coverage.

## Motivation
Palace's magnetostatics previously produced only geometric inductance. Thin-film superconducting circuits have $d\lesssim\lambda$, where the kinetic inductance is a first-class circuit element. This PR makes Palace extract L = L_geom + L_kin directly.

## Examples & docs

- New `examples/circular_hole/` (square washer, single/double circular holes) and `examples/microstrip/` (kinetic-inductance validation vs the analytic parallel-plate  $\mu_0\lambda_\perp\ell/w$).
- New docs page *Flux Trapping and Kinetic Inductance* documenting the formulation and config.

## Validation

- Square washer $(\lambda\rightarrow 0)$: $L\approx 3.04$ pH, matching the Ketchen–Jaycox $1.25 \mu_0 d$ ($\approx 3.14$ pH) formula.
- Microstrip: kinetic contribution matches the analytic 2-sheet $2\mu_0\lambda_\perp\ell/w$ to ~5%.
- Regression cases added: `square_hole_flux_loop`, `circular_hole_london_flux`, `double_hole_london_flux`, `circular_hole_london_nc_amr`, `ring_disk_mixed_current_flux`; rank-invariant at np=1/2/4/8.


palace/drivers/magnetostaticsolver.cpp
```diff
@@ -283,16 +311,106 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
+        Mpi::Warning(curlcurl_op.GetComm(),
+                     "London fluxoid mode carries ~zero hole flux for flux loop {:d}; "
+                     "halting adaptation!\n",
+                     idx);
+        solve_converged_ = false;
+        return {indicator, curlcurl_op.GlobalTrueVSize()};
+      }
+      double alpha = (data.GetExcitationFlux() - phi_p) / phi_h;
+      A[step] = A_p_london;
+      A[step].Add(alpha, A_h_london);
```
hughcars: At finite λ this constrains the wrong quantity. The shifted penalty already gives K_s = −(A_t − a_h)/L_ksq, so the solve driven by `a_h` of circulation Φ enforces the fluxoid ∮(A + L_ksq K_s)·dl = Φ exactly (the normalization comment in `CurlCurlOperator` already calls cᵀa_h the fluxoid). Forcing cᵀA = ∮A·dl = Φ on top of that fixes the flux instead, and α·A_h acts like a fictitious line current along the hole edge.

The results can't be physical. For the documented annulus (r = 1–3 µm, d = 0.1 µm, λ = 0.4 µm), L_ksq = μ₀λ coth(d/λ) = 2.05 pH/sq, and any sheet current I around the hole carries kinetic energy of at least ½·L_ksq·(2π/ln 3)·I², so L ≥ 11.7 pH from the kinetic term alone, against a documented total of 5.17 pH. They're also mesh-dependent: the `nc_amr` reference moves from 5.17 to 5.34 pH after one refinement.

I'd drop the second solve and α. With `a_h` normalized so cᵀa_h = Φ, the single solve is the fluxoid-Φ state and R = 2E/Φ² follows directly. Report ∮A·dl as the flux alongside the fluxoid, and say in the docs and schema that `FluxAmounts` is the fluxoid. λ→0 is unaffected, since flux and fluxoid coincide there.

palace/models/domainpostoperator.cpp
```diff
@@ -152,6 +154,18 @@ DomainPostOperator::DomainPostOperator(const config::DomainPostData &postpro,
...
+      // Add the kinetic sheet term so AᵀM_mag·A = 2·E_mag includes the London kinetic
+      // inductance, yielding total (geometric + kinetic) inductance. The coefficient must
+      // outlive PartialAssemble(), which stores a pointer to it.
+      MaterialPropertyCoefficient fbr(mat_op.MaxCeedBdrAttribute());
+      if (sc_sheet_op && !sc_sheet_op->empty())
+      {
+        sc_sheet_op->AddStiffnessBdrCoefficients(1.0, fbr);
+        if (!fbr.empty())
+        {
+          m.AddBoundaryIntegrator<VectorFEMassIntegrator>(fbr);
```
hughcars: This makes `GetMagneticFieldEnergy(A)` on a flux step return the unshifted S(A, A) rather than the kinetic energy S(A − a_h, A − a_h). The committed refs show it: in `square_hole`, the λ→0 limit where the kinetic energy should vanish, `E_mag` is 3.2e-7 J while `E_mag[1]`, the whole volume, is 4.0e-13 J (`p_mag[1]` = 1.2e-6). The same energy normalizes the flux step's error indicators, so AMR marking and `Refinement.Tol` effectively ignore flux steps. I'd keep `M_mag` volume-only and add the kinetic energy of A − a_h explicitly where it's needed (the driver already has `ApplySheetMass`).

palace/models/surfacecurlsolver.cpp
```diff
@@ -5,360 +5,219 @@
...
+      csum[0] += x[0];
+      csum[1] += x[1];
+      csum[2] += x[2];
+      cnt += 1.0;
+    }
   }
+  Mpi::GlobalSum(3, csum, comm);
+  Mpi::GlobalSum(1, &cnt, comm);
+  MFEM_VERIFY(cnt > 0.0, "No hole boundary elements found for London cut generator!");
+  double cx = csum[0] / cnt, cy = csum[1] / cnt;
```
hughcars: This builds the flux line from x and y only, so it assumes the hole axis is ±z; the loop's `Direction` isn't consulted. For a film in the xz or yz plane the line lies in the film, cᵀa_h ≈ 0, the `|cᵀa_h| > 1e-30` guard in `CurlCurlOperator` silently skips the rescale, and the run carries on with a wrong generator and no error. Either `MFEM_VERIFY` that the direction is ±z, or build the two in-plane coordinates from a basis of the plane normal to `Direction`.

docs/src/examples/circular_hole.md
```diff
@@ -76,153 +112,116 @@ settings are:
...
+Sweeping ``\lambda`` at fixed ``d = 0.1\,\mu\text{m}`` shows the kinetic inductance adding to the
+geometric baseline as the film penetration grows, and the ``\lambda\to0`` value reproducing the
+perfect-conductor limit:
+
+| ``\lambda`` (μm) | ``L_{\mathrm{ksq}} = \lambda\coth(d/\lambda)`` (μm) | ``L`` (pH) |
+|:----------------:|:--------------------------------------------------:|:----------:|
+| ``0`` (PEC)      | ``0``                                              | ``2.853``  |
+| ``0.1``          | ``0.131``                                          | ``3.388``  |
+| ``0.2``          | ``0.433``                                          | ``4.075``  |
+| ``0.4``          | ``1.633``                                          | ``5.173``  |
```
hughcars: This table is the check-style failure: the header's second cell is one character wider than the separator and rows, so JuliaFormatter re-pads the other five lines.

```suggestion
|:----------------:|:---------------------------------------------------:|:----------:|
| ``0`` (PEC)      | ``0``                                               | ``2.853``  |
| ``0.1``          | ``0.131``                                           | ``3.388``  |
| ``0.2``          | ``0.433``                                           | ``4.075``  |
| ``0.4``          | ``1.633``                                           | ``5.173``  |
```

palace/drivers/magnetostaticsolver.cpp
```diff
@@ -264,15 +283,24 @@ MagnetostaticSolver::Solve(const std::vector<std::unique_ptr<Mesh>> &mesh) const
...
         ksp.Mult(RHS, A[step]);
       }
+      if (!ksp.GetConverged())
+      {
+        // Halt rather than write an unreliable inductance from a non-converged solve.
+        Mpi::Warning(curlcurl_op.GetComm(),
+                     "Current source solve did not converge for port {:d}!\n", idx);
+        solve_converged_ = false;
+        return {indicator, curlcurl_op.GlobalTrueVSize()};
+      }
```
hughcars: This applies to every magnetostatic run, not just the London path. A current-only solve that stops at `MaxIts` used to warn in `BaseKspSolver::Mult` and still write its results; now it returns early and trips the new `MFEM_VERIFY(solve_converged_)` in `SolveEstimateMarkRefine`, so the run aborts with no output. The other drivers still warn and continue. I'd scope the early exit to the London path, or make it a deliberate, documented change across drivers.

palace/drivers/basesolver.cpp
```diff
@@ -349,6 +351,18 @@ void BaseSolver::SolveEstimateMarkRefine(std::vector<std::unique_ptr<Mesh>> &mes
...
+      // Solver could not converge on the refined mesh (e.g. the fluxoid solve as the mesh
+      // grows). Halt adaptation and retain the previous, converged iteration's output
+      // rather than aborting the run or reporting an unreliable result.
+      Mpi::Warning(
+          comm,
+          "Solve did not converge after refinement iteration {:d}; halting AMR and "
+          "keeping the last converged iteration!\n",
+          it);
+      break;
+    }
```
hughcars: The `break` keeps the previous iteration's CSVs, but not the rest of the output. `RebalanceMesh` has already written the refined mesh for `SaveAdaptMesh`, and the failed `Solve` has already written its `MeshElements`/`DegreesOfFreedom` to `palace.json`, so both describe the mesh whose solve failed. The final summary also prints that mesh's `ntdof` next to the previous `err`. Either restore them or have the warning say which outputs come from which mesh.

scripts/schema/config-schema.json
```diff
@@ -300,6 +300,14 @@
             "$ref": "#/$defs/FluxLoop"
           }
         },
+        "Superconductor": {
+          "type": "array",
+          "description": "Array of thin-film superconductor sheet boundaries for magnetostatic simulations. Each sheet contributes a London kinetic-inductance term to the extracted total inductance, modeled by the finite-thickness London kinetic sheet inductance L_ksq = mu0 * lambda * coth(d/lambda), specified either via the penetration depth and film thickness or directly (as the thin-film limit lambda^2/d when d << lambda). Only available for magnetostatic simulations.",
```
hughcars: μ₀λ coth(d/λ) is the one-sided-penetration result. For a free-standing film with field on both faces, as in the washers, it's (μ₀λ/2) coth(d/2λ): same Pearl limit, and within 2% at d = 0.1 µm, λ = 0.4 µm, but a factor of 2 apart for d ≫ λ, so this should say which geometry it assumes.

hughcars (changes requested):
The λ→0 flux-loop rework looks like a real correction over 0.18 and should stay: the `circular_hole` PEC limit moves from 2.00 to 2.85 pH and the square washer lands within 5% of Ketchen–Jaycox. At finite λ there are two blocking problems, inline: the second solve constrains the flux rather than the fluxoid, and `E_mag` includes the unshifted sheet energy. The cut generator also ignores `Direction`.

This needs a CHANGELOG entry: New Features for `Superconductor` and `LondonPCShift`; Interface Changes with the `SchemaVer 2-0-0` tag for `FluxLoopPEC` → `FilmAttributes`, `Regularization` → `PecPenetrationDepth`, and the new one-hole-per-`FluxLoop` limit (the schema still accepts arrays for `HoleAttributes`/`FluxAmounts`), since every existing `FluxLoop` config is now rejected; and a Bug Fixes note that 0.18 `FluxLoop` inductances were ~30% low, citing the re-baselines (`ring_disk` M₂₂ +38%, M₁₂ ×300) so users know to re-run. #909 and #952 also edit the schema `$id`; for #909 that line is the only conflict, so whichever lands later re-bumps.

Cleanup: `examples/microstrip/run_{pec,sc,sc1}.log` are local run logs (with `/Users/...` paths), and `examples/circular_hole/mesh/rect_hole.msh` (1.5 MB) and its generator have no consumers. The example and regression JSON comments and the `SuperconductorData` comments in `configfile.{hpp,cpp}` still give L_ksq = μ₀λ²/d, and `docs/src/guide/boundaries.md` still describes the removed 2D surface-curl solve and says nonconformal adaptation is unsupported.

PR #930, opened 2026-09-10
Reduce weighted CGS and CPU multigrid operator costs

Sbozzolo:
Applies two optimizations:

First, `RomOperator::UpdatePROM` orthogonalizes against a `W`-weighted inner product. It passed a dot functor to `linalg::OrthogonalizeColumn`, which applied `W` once per basis vector (`m` applications per pass). Since `w` is unchanged while its projections are computed, one application suffices: `OrthogonalizeColumnWeightedCGS`
computes `W·w` into a caller-owned workspace and then takes plain `LocalDot` against each basis vector. `W` is reapplied after the first update when refining (CGS2), so the cost is `m` → 1 per pass. This accelerates PROM constructions by a factor of 2 in the cases considered (PROM construction is typically not dominant, so the overall effect is relatively modest). 

Second, `SpaceOperator::AssemblePreconditioner` hard-wired `assemble_q_data = false`. Caching the geometry/material tensors pays off for the repeated smoother applications on CPU, so it is now enabled whenever no device backend is active. This speeds up the single transmon case by 10 % at the cost of a small amount of extra memory.


simlapointe (commented):
Looks good! Only issue I see is the new `TEMPLATE_TEST_CASE` triggers a fatal coverage error in the GCC/OpenMP coverage job.

PR #931, opened 2026-09-10
Fix the quadrature data offset in the 32 H(div) mass build QFunction

Sbozzolo:
`f_build_hdivmass_32` wrote its second quadrature-data block one component too early, so the cached curl-curl block of a 2D-in-3D H(div)/curl-curl + mass operator was never written and the mass block was partially overwritten. 

Found while working on #930.


simlapointe (approved):
Good catch!

PR #932, opened 2026-09-11
Fuse compatible CPU complex operator applications

Sbozzolo:
Complex partial-assembly applications repeat basis work for the real and imaginary matrices. With this PR, we fuse compatible QData-assembled H(curl)/H(div) mass terms using a two-component libCEED operator (and preserve the original operators for diagonal, transpose, and full assembly).

This optimization matters only for simulations with lossy materials (because they have imaginary components). In my experiments (single transmon, and some other realistic simulations in the range of 1-10M DOFs, with one run with 150M DOFs), I've seen a typical ~ 10 % reduction in time-to-solution, with up to 25 % reduction in preconditioner time.

Combined with #930 and https://github.com/spack/spack-packages/pull/6425, I've measured improvements in Palace between 3 and 27 % compared to 0.18.


palace/fem/libceed/operator.cpp
```diff
@@ -242,6 +247,525 @@ void Operator::AddMultTranspose(const Vector &x, Vector &y, const double a) cons
...
+    // result only after both actions finish, including when their x and y alias.
+    input_view = x;
+    packed->Mult(input, output);
+    if (remainder)
+    {
+      // The packed action has finished reading input. Reuse it for the remainder,
+      // which still reads the caller's x; output stays separate until accumulation.
+      remainder->Mult(x, input_view);
+      output_view.AXPY(1.0, input_view);
+    }
```
simlapointe, line 576: If my understanding is correct, whenever unmatched terms remain, the implementation:

 1. Stages the input.
 2. Computes the packed contribution, zeroing its output first.
 3. Computes the remainder into the input scratch.
 4. Adds the remainder through a full complex AXPY.

 Even a small unmatched boundary contribution therefore introduces passes over all local DOFs.

 A simpler order is:

 ```cpp
   input_view = x;
   if (remainder)
   {
     remainder->Mult(input_view, output_view);
     packed->AddMult(input, output);
   }
   else
   {
     packed->Mult(input, output);
   }
 ```

This preserves private staging and caller alias safety. The packed AddMult uses exactly the supported +1 coefficient. It removes the packed-output zeroing and the final complex AXPY:

palace/fem/libceed/operator.cpp
```diff
@@ -242,6 +247,525 @@ void Operator::AddMultTranspose(const Vector &x, Vector &y, const double a) cons
...
+  PalaceCeedCall(ceed, CeedElemRestrictionGetOffsets(original, CEED_MEM_HOST, &offsets));
+  const CeedSize length = 2 * static_cast<CeedSize>(size);
+  if (type == CEED_RESTRICTION_CURL_ORIENTED)
+  {
+    const CeedInt8 *orientations;
+    PalaceCeedCall(ceed, CeedElemRestrictionGetCurlOrientations(original, CEED_MEM_HOST,
+                                                                &orientations));
+    PalaceCeedCall(ceed,
+                   CeedElemRestrictionCreateCurlOriented(
+                       ceed, leaf.elements, leaf.nodes, 2, size, length, CEED_MEM_HOST,
```
simlapointe, line 454: One backend detail worth checking: changing `num_comp` from 1 to 2 here loses a restriction specialization in our pinned libCEED version. Its reference restriction dispatcher (https://github.com/CEED/libCEED/blob/39f259f89332e936122f7e02d6088a1dae3fb628/backends/ref/ceed-ref-restriction.c#L834-L893) has specialized routines for component counts 1, 3, 4, and 5, but none for 2, so these packed restrictions fall back to `CeedElemRestrictionApply_Ref_Core`. This also affects opt/XSMM, which use the reference restriction implementation.

This may not be worth it, but wondering if we could/shoud add two-component specializations upstream, matching the existing block-size-1 and block-size-8 variants?

palace/fem/libceed/operator.cpp
```diff
@@ -242,6 +247,525 @@ void Operator::AddMultTranspose(const Vector &x, Vector &y, const double a) cons
...
+      PalaceCeedCall(ceed, CeedOperatorSetField(op.value, field, restriction.value,
+                                                basis.value, CEED_VECTOR_ACTIVE));
+    }
+  }
+  PalaceCeedCall(ceed, CeedOperatorCheckReady(op.value));
+  packed.AddSubOperator(std::exchange(op.value, nullptr));
+}
+
+// The base owns the finalized originals for diagonal, transpose, full assembly, and
+// coarsening. Only forward applications use the packed composite and its remainder.
```
simlapointe, line 553: Potential follow-up speedup: the adjoint application is also used in the preconditioner. VCycle (palace/linalg/gmg.cpp#L204-L206) performs transpose post-smoothing, and complex Chebyshev smoothing (palace/linalg/chebyshev.cpp#L78-L106) dispatches that through MultHermitianTranspose/AddMultHermitianTranspose. Those applications currently retain the original four-real path.

Could we consider a packed adjoint implementation as a follow-up, and measure its incremental benefit on a full V-cycle? This would extend fusion to post-smoothing rather than only the forward applications. It would need to preserve the actual transpose/conjugation semantics, including for nonsymmetric tensors; simply reusing the forward action would not be sufficient.

Not suggesting this necessarily needs to go in this PR, but maybe worth considering?

palace/fem/libceed/operator.cpp
```diff
@@ -242,6 +247,525 @@ void Operator::AddMultTranspose(const Vector &x, Vector &y, const double a) cons
...
+  PalaceCeedCall(
+      ceed, CeedBasisGetNumQuadratureComponents(basis, CEED_EVAL_INTERP, &qcomponents));
+  PalaceCeedCall(ceed, CeedElemRestrictionGetType(restriction, &type));
+  PalaceCeedCall(ceed, CeedElemRestrictionGetNumComponents(restriction, &rcomponents));
+  PalaceCeedCall(ceed, CeedElemRestrictionGetNumElements(restriction, &leaf->elements));
+  PalaceCeedCall(ceed, CeedElemRestrictionGetBlockSize(restriction, &block));
+  PalaceCeedCall(ceed, CeedElemRestrictionGetLVectorSize(restriction, &length));
+  PalaceCeedCall(ceed, CeedOperatorGetActiveVectorLengths(op, &in_length, &out_length));
+  if (tensor || dim != 3 || components != 1 || qcomponents != 3 || rcomponents != 1 ||
+      (space != CEED_FE_SPACE_HCURL && space != CEED_FE_SPACE_HDIV) ||
```
simlapointe, line 402: The auxiliary-space hierarchy also repeats real/imaginary basis work, but the current recognizer deliberately excludes H1 operators. Could/should we extend this fusion to the complex H1 auxiliary diffusion operators used by distributive relaxation, or is time spent in the auxiliary smoothing negligible so it's not worth it?

simlapointe (commented):
This looks quite good. A few comments below about some potential further improvements. Not sure if they're all worth it though.

palace/fem/libceed/operator.cpp
Sbozzolo: I am opening a PR for this, but my microbenchmarks show essentially no improvement with the specialized code for the specific paths Palace takes.

palace/fem/libceed/operator.cpp
Sbozzolo: Yes, you are right.

palace/fem/libceed/operator.cpp
Sbozzolo: Yes, I focused on the more expensive part first. I think that the benefit for fusing on H1 would be smaller and would add complexity, but it is worth benchmarking and following up.

palace/fem/libceed/operator.cpp
Sbozzolo: Yes, that's worth looking into. I opened an issue to keep track of this.

simlapointe: Thanks for opening the issue, we can implement it in a future PR.

palace/fem/libceed/operator.cpp
simlapointe: Thanks for opening the issue, we can investigate/implement in a future PR.

simlapointe (commented):
Thanks for the opening #933 and for the libceed PR. Let's maybe improve the `remainder` treatment I noted above, otherwise LGTM.

Sbozzolo:
> Thanks for the opening #933 and for the libceed PR. Let's maybe improve the `remainder` treatment I noted above, otherwise LGTM.

Sorry, I forgot to push!

Sbozzolo:
> Thanks for the opening #933 and for the libceed PR. Let's maybe improve the `remainder` treatment I noted above, otherwise LGTM.

Fixed in 0c3d758c960d1d682b3bbc0a02aedf15b9bec3c9

simlapointe (approved):

hughcars:
Please add a CHANGELOG entry under `In progress` for this.

PR #936, opened 2026-09-16
Fix farfield rE dimensionalization

### LINKED ISSUE #935: `farfield-rE.csv` values depend on `Lc`
`Measurement::Dimensionalize` (`postoperatorcsv.cpp:131`, and the reload path at `:273`) converts `farfield.E_field` with `Nondimensionalize<FIELD_E>`, but `rE_∞` is a voltage, so the written values are off by `Lc/Z0` (~31× low for the antenna example). Reproducer: run `examples/antenna/antenna_halfwave_dipole.json` as-is (auto `Lc = 12 m`) and with `"Lc": 1.0` added to `"Model"` (`L0` unchanged, so the physical model is identical). `S11` and the energies agree to 1e-10, but every entry of `farfield-rE.csv` changes by exactly 12×, which a physical quantity shouldn't do. With the `VOLTAGE` scale instead, `∮|rE|²/2Z0 dΩ` matches `(1−|S11|²)·P_inc` to 0.3 %. The kernel and normalized patterns are fine (`test-strattonchu.cpp` tests the raw kernel with the right scale, so the writer was never covered). Fix + test + rescaled references on `hughcars/fix-farfield-units`.

hughcars:
Fixes https://github.com/awslabs/palace/issues/935


Sbozzolo (approved):
Thank you!

PR #886, opened 2026-08-18
Wave-port modal correction (n×H) for longitudinal-field modes across driven, PROM, eigenmode, and circuit synthesis

dnpham23:
Wave ports produce non-physical S-parameters for modes with a longitudinal electric field ($E_n\neq 0$, e.g. TM/hybrid modes): a lossless shorted guide can report $|S_{11}| > 1$, so the S-matrix is neither unitary nor reciprocal.

### Root cause
The exact in-plane port relation is $n\times H = -(1/i\omega\mu)(ik_nE_t + \nabla_tE_n)$. The excitation, mode normalization, and S-parameter projection use this full modal $n\times H$ (including $\nabla_tE_n$), but the assembled LHS Robin operator only stamped the scalar surface admittance $ik_nM$ and dropped $\nabla_tE_n$. For $E_n \neq 0$ the two disagree, so a port is not matched to its own mode and $\sum_i|S_{ij}|^2 \neq 1$.

### Fix
A matrix-free complex-symmetric modal correction $W = \sum(W_{\text{full}} - W_{\text{scalar}})$ is added onto the sparse local boundary mass, supplying exactly the $\nabla_t E_n$ term the mass drops. On the port's own mode $(ik_nM + W)e = -i\omega s_{\text{full}}$, matching the excitation, so the port is reflectionless. For TEM/TE modes $(E_n \approx 0)$ $W\equiv 0$ and the baseline is recovered exactly.

$W$ is applied consistently across the full scope:

- Driven / PROM: stamped in `GetExtraSystemOperator` and projected into the reduced-order operator (full solve + adaptive fast-frequency sweep).
- Eigenmode: evaluated at the complex eigenfrequency.
- Circuit synthesis: folded into the synthesized L/R/C pencil; frequency-dependent coupling maps G(ω)/H(ω) and reconstructed S exported (`rom-coupled-{G,H,S}.csv`) with per-port de-embedding, so the synthesized network reproduces the solver's own S-parameters. Mixed lumped+wave and inactive included ports supported.

### Generic fixes included here
- SLP nonlinear eigensolver robustness: fall back to Hybrid when SLEPc is unavailable or an ARPACK backend is selected (resolved before the Hybrid interpolation is built), and handle a null damping matrix / missing nonlinear term.
- Synthesized eigenvalue-root reporting: $Q=|\omega|/(2|Im \{\omega|\})$ convention matching the eigenmode postprocessor; filter spurious roots via reported HDM residuals rather than fixed cutoffs.

[...]

PR #937, opened 2026-09-17
Time-averaged power, energy, and port normalization

hughcars:
**Draft — want comments on the convention before polishing.**

Our frequency-domain fields are peak phasors, but the quadratic outputs (energies, powers, Poynting) were the raw phasor product, i.e. twice the time average. Lumped and wave ports were normalized to `∫E×H*·n dS = 1`, which is 0.5 W time-averaged, while Floquet ports carried 1 W time-averaged and mixed-port S-parameters were patched with `√2` bridges. Ratios (Q, EPR, S) were all fine, so nobody noticed; absolute energies were 2× and driven fields were those of a 0.5 W drive.

This switches to the standard convention: quadratic outputs are time averages, every port family carries 1 W time-averaged incident power (50 Ω lumped port → `V_inc = 10 V`, `I_inc = 0.2 A` peak), time-domain and static outputs untouched. The factors live in one place (`electromagnetics::TimeAverageWeight` / `UnitPowerAmplitude`) and are applied by the owners of the physical quantities, never in norms or mode overlaps.

**Visible changes**
- Eigenmode, boundary mode, Floquet/current-driven: `E_elec`, `E_mag`, `E_cap`, `E_ind`, `Φ_pow`, `U_e`/`U_m`/`S` ×½.
- Lumped/wave-driven: `V`, `I`, `V_inc`/`I_inc`, probes, far field, `Φ_elec`/`Φ_mag` ×√2. Energies and powers numerically unchanged.
- Unchanged: `S`, `Z`, `Z_PV`, `f`, `Q`, `κ`, EPR, participation ratios, error indicators, ROM synthesis, everything transient/static.

**Verification**
- New `test-timeaverage.cpp` pins the production reducers and the lumped-port drive to values written from the physics (`U_e = ε₀/2`, `U_m = 13μ₀/4`, 10 V / 0.2 A / 1 W). Fails on main by exactly 2×.
- Transient vs driven on the same coax: `|V_driven| = √2·peak V(t)`, `E_driven = 2⟨E(t)⟩` to 0.2 %. The transient solver has no phasor convention, so this is the cleanest check we have.
- COMSOL's coax formulas (`C' = 4We/|V|²`, `L' = 4Wm/|I|²`) hit the analytic line parameters to 0.3 %; Floquet plane wave, TM010 probe-vs-energy and dipole radiated-vs-accepted power all within 1 %.
- 50 regression references rescaled by explicit ½ / √2 factors; the suite passes against them, which also proves the protected columns didn't move.

**Open questions**
1. Time average (not RMS, not peak product) — I'm confident this is what people expect, but it changes absolute outputs.
2. Driven field amplitudes grow by √2; anything assuming `V_inc = √R` will move.
3. Time-domain excitations stay at unit *instantaneous* power (`V_inc(t) = √R·g(t)`). Should they follow 1 W-average too?
4. Lossy wave-port modes are still normalized by `|∫E×H*·n|`, as before. Deferred on purpose.


dnpham23:
Nice, @hughcars, I took a quick look and I think your approach and reasoning make sense. On the questions you posed, here are my two cents:
1. Time average (not RMS, not peak product): I think this is the correct choice.
2. I think the $\sqrt{2}$ growth in driven amplitudes is correct and unavoidable, and can be treated as the headline breaking change. 1 W time-avg into a matched 50 ohm port -> V_rms = √(P·R) = 7.07 V => V_peak = √2·7.07 = 10 V. So V_inc,peak = √(2R) is the physically right amplitude for a peak-phasor solver; the old √R = 7.07 V was a 0.5 W drive.
3. I think you should keep time-domain at unit instantaneous power and not switch it to 1-W-average. A transient pulse has no single period, so "time-averaged power" is ill-defined without picking an arbitrary averaging window; the natural transient convention is instantaneous power at unit pulse amplitude (√R·g(t)). Moreover, your own cross-check |V_driven| = √2·peak V(t) is exactly the expected relationship between a 1-W-time-avg CW drive and a unit-instantaneous transient. That √2 is actually consistency proof and not a discrepancy. Switching TD to  1-W-average would remove that clean relationship and force an artificial window.
4. I think deferring lossy wave-port |∫E×H*·n| normalization is fine, we just need to document the caveat.

hughcars:
Agree on all four. Time domain stays at unit instantaneous power, now stated in the docs. The lossy wave port normalization is O(tanδ²) in power but the conjugate projection is O(tanδ) in S phase, so it's a mode model problem not a ½/√2 one — split out as #960. Rebased on 0.18.1, #886 references migrated, suite passes.

PR #941, opened 2026-09-18
Fix NaN from iterative solvers on zero residual

nikosavola:
eps = max(rel_tol * initial_res, abs_tol) collapses to 0 when both the initial residual and abs_tol are 0, so the strict res < eps test is 0 < 0 and never fires. The loop then runs on a zero residual: CG divides 0 / 0 at alpha = beta / denom, and GMRES and FGMRES divide by beta when normalizing V[0].


test/unit/test-iterative.cpp
```diff
@@ -0,0 +1,167 @@
...
+  SpdTestOperator() : Operator(2) {}
+
+  void Mult(const Vector &x, Vector &y) const override
+  {
+    y.SetSize(2);
+    y[0] = 4.0 * x[0] + x[1];
+    y[1] = x[0] + 3.0 * x[1];
+  }
+
+  void MultTranspose(const Vector &x, Vector &y) const override { Mult(x, y); }
```
hughcars: `FgmresSolver zero residual` aborts before it reaches the solver: `JacobiSmoother::SetOperator` calls `op.AssembleDiagonal`, which `SpdTestOperator` doesn't override, so MFEM aborts with "Not relevant or not implemented for this Operator", serial and at np=2. With the override all four cases pass, and the FGMRES one is red on the base like the others:

```suggestion
  void MultTranspose(const Vector &x, Vector &y) const override { Mult(x, y); }

  void AssembleDiagonal(Vector &diag) const override
  {
    diag.SetSize(2);
    diag[0] = 4.0;
    diag[1] = 3.0;
  }
```

hughcars (changes requested):
The FGMRES test aborts as written (inline), and this needs a CHANGELOG entry, something like: Fixed CG, GMRES, and FGMRES returning NaN when the initial residual is exactly zero (e.g. a zero right-hand side) and no absolute tolerance is set.

PR #938, opened 2026-09-17
Defer adaptive sweep CSV output to end of online phase

simlapointe:
Defer adaptive sweep CSV output to end of online phase. This can substantially accelerate the online phase when sampling a large number of frequencies.

PR #942, opened 2026-09-18
Write CSV tables incrementally instead of rewriting them

nikosavola:
Every measurement step removed and rewrote each CSV table in full, re-rendering every row already on disk, so the bytes written grew with the square of the step count. Tables whose rows are final as they are written, which is the long transient case, now append only the new rows.

- Rows a later excitation pass can still add columns to keep the existing whole-file write, so multi-excitation sweeps behave exactly as before and restart still works.
- Appending is byte-exact: a row renders the same regardless of which other rows are in the table.
- Transient runs also defer writes, which cuts how often the file is opened and closed.

## Benchmarks

Driving the real writer: old path (`WriteFullTableTrunc`, what main did for transients) against new (`WriteTableIncremental`). M5 arm64, single run each, so the absolute numbers are indicative rather than rigorous.

Scaling, one table with 5 data columns:

| steps | bytes written (before) | bytes written (after) | time (before) | time (after) |
| --- | --- | --- | --- | --- |
| 1,000 | 84.3 MB | 168 KB | 1.18 s | 24 ms |
| 2,000 | 336.5 MB | 336 KB | 4.53 s | 54 ms |
| 4,000 | 1.35 GB | 672 KB | 19.9 s | 96 ms |

Both bytes and time grow 4x per doubling before and 2x per doubling after, i.e. quadratic to linear.

Multi-excitation driven sweeps are unaffected. They still write the whole file for now.

## Testing

Final files are byte-identical to the previous behavior in every case above. A probe drives the real writer through the sweep fill order and reloads at every step, checking the file still matches the fill pattern a restart expects; it fails against the naive append-only version without the whole-file fallback, which is how a restart regression was caught during development.


palace/models/postoperatorcsv.cpp
```diff
@@ -1881,14 +1881,21 @@ PostOperatorCSV<solver_t>::PostOperatorCSV(const config::ProblemData &problem,
   {
     nr_expected_measurement_rows = solver.driven.sample_f.size();
     reload_table = (solver.driven.restart != 1);
-    defer_table_writes = (solver.driven.adaptive_tol > 0.0);
+    defer_table_writes = true;
```
hughcars: #938 limited the 10 s buffering to adaptive sweeps because every measurement rewrote the whole file. With `WriteTableIncremental` the non-deferred path is an append per table per step (your benchmark: 4,000 steps in 96 ms), so extending the buffering to every driven and transient run buys nothing for this PR's goal. What it costs is durability: an interrupted non-adaptive sweep or transient run now loses up to 10 s of rows, where on main every row is on disk after its step, and resuming a killed driven sweep with `"Restart"` (once #944 lands) depends on those rows being there. I'd keep #938's scope:

```suggestion
    defer_table_writes = (solver.driven.adaptive_tol > 0.0);
```

palace/models/postoperatorcsv.cpp
```diff
@@ -1881,14 +1881,21 @@ PostOperatorCSV<solver_t>::PostOperatorCSV(const config::ProblemData &problem,
...
     m_ex_idx = ex_idx_v_all.at(ex_idx_i);
   }
 
+  // Transient solver: defer writes too, else every step rewrites the tables in full.
+  if constexpr (solver_t == ProblemType::TRANSIENT)
+  {
+    defer_table_writes = true;
+    last_deferred_flush = std::chrono::steady_clock::now();
+  }
+
```
hughcars: And drop this block; the comment describes the pre-PR path, since a non-deferred write now appends:

```suggestion
```

palace/models/postoperatorcsv.hpp
```diff
@@ -242,8 +242,9 @@ class PostOperatorCSV
   // Copy savepath from PostOperator for simpler dependencies.
   fs::path post_dir;
   bool reload_table = false;  // Driven simulation with non-default restart.
-  // Adaptive driven output is buffered and rewritten at most once per flush interval (and
-  // at finalization), so an interrupted sweep still leaves usable tables.
+  // Driven and transient output is written at most once per flush interval instead of on
+  // every measurement. Tables holding partially filled rows are still written whole, so an
+  // interrupted run leaves tables a restart can read the fill state back from.
```
hughcars: Comments to match:

```suggestion
  // Adaptive driven output is buffered and written at most once per flush interval (and
  // at finalization), so an interrupted sweep still leaves usable tables.
```

palace/models/postoperatorcsv.hpp
```diff
@@ -425,8 +426,8 @@ class PostOperatorCSV
   auto PrintEigPortQ() -> std::enable_if_t<U == ProblemType::EIGENMODE, void>;
 
 public:
-  // Flush tables buffered during an adaptive driven sweep. Other solver modes write eagerly
-  // and this is a no-op for them.
+  // Flush rows still buffered by the flush interval. Other solver modes write on every
+  // measurement, so this is a no-op for them.
```
hughcars: Same:

```suggestion
  // Flush tables buffered during an adaptive driven sweep. Other solver modes write eagerly
  // and this is a no-op for them.
```

hughcars (changes requested):
The deferral extension should come out (inline), and this needs a Performance Improvements entry in the CHANGELOG, e.g. CSV output tables are now appended row by row instead of rewritten at every measurement, so long transient and single-excitation driven runs write O(N) rather than O(N²) bytes. This and #944 both append tests to `test-tablecsv.cpp`, so whichever lands second needs a trivial rebase.

Separately, the 10 s flush from #938 is checked per table, so a flush can land with `domain-E` a row ahead of the port tables. I pushed a fix as a single commit on `hughcars/csv-flush-step-boundary` (632dd023), on top of your head; can you cherry-pick it? It moves the check to the end of `PrintAllCSVData` so every flush is on a step boundary.

PR #944, opened 2026-09-18
Fix column name lookup after a restart

### LINKED ISSUE #943: Driven restart throws on the first measurement after reloading the CSV tables
**What happens**

Setting `"Restart"` to a value greater than 1 in a driven simulation loads the existing CSV tables and then dies on the first measurement:

```
libc++abi: terminating due to uncaught exception of type std::out_of_range: Column Ee_2 not found in table
```

**Why**

The CSV loader names the columns it parses positionally (`col_0`, `col_1`, ...) and indexes them under those names. `PostOperatorCSV::MoveTableValidateReload` then restores the real names from the reference table onto the `Column` objects, but leaves `Table::name_to_index` holding the placeholder names. Every `Print*` looks a column up by name (`table["idx"]`, `table[format("Ee_{}", m_ex_idx)]`), so the first lookup after a restart throws. `Table::has()` is stale the same way and reports real columns as absent, which would also make the wave-port impedance columns be skipped.

**Where it came from**

PR #393 added the CSV parsing, and with it the `col_i` naming in the loader and the rename-without-reindex in `MoveTableValidateReload`. Before that PR restart failed earlier, with a row-count mismatch in `CheckAppendIndex` (#376); this is the next failure in the same path.

**Why it is not caught**

- `test/unit/test-postoperatorcsv.cpp` stops at the `Initialize*` calls and never prints after a reload.
- No regression input uses `"Restart"`, so CI never exercises a driven restart.

**Proposed fix**

Make the name index self-healing on lookup, or rebuild it where the names are reassigned. A PR follows.

nikosavola:
Closes #943

*Description of changes:*

Restarting a sweep reloads the CSV tables from disk and then restores the real column names from a reference table. The name index was not updated with them, so it still held the placeholder names the loader assigns. The first lookup by name then threw (`Column Ee_2 not found in table`) and killed the run.

The same staleness made `has()` report real columns as absent, which was the worse symptom: the guarded columns in the wave-port voltage and port-Z impedance printers were quietly skipped for the whole restarted run, with no error at all.

- The index is now rebuilt where the names are reassigned, which also drops the loader's placeholder entries and keeps `insert`'s duplicate detection honest.
- Lookups additionally fall back to scanning the columns when the index misses, so a name reassigned anywhere else cannot reintroduce the same failure.
- Added unit tests; they fail against the current code.

Existing unit tests stop at the `Initialize*` calls and never print after a reload, and no regression input sets `"Restart"`, which is why this went unnoticed.


palace/utils/tablecsv.hpp
```diff
@@ -120,9 +123,20 @@ class Table
...
   {
-    return name_to_index.count(std::string(name)) > 0;
+    if (name_to_index.count(std::string(name)) > 0)
+    {
+      return true;
+    }
+    // Column names can be reassigned after a file is loaded, without insert, which leaves
+    // the map stale, so fall back to the columns themselves.
+    return std::any_of(cols.cbegin(), cols.cend(),
+                       [name](const Column &col) { return col.name == name; });
```
hughcars: With the rebuild at the only rename site, the fallbacks here and in `operator[]` aren't needed, and they stop the index being the source of truth: `insert()` still only consults `name_to_index`, so under a stale index `has(n)` is true while `insert(n, ...)` adds a duplicate, and the `operator[]` repair adds the new name without dropping the loader's `col_i` key. They also leave the actual fix untested: I deleted the `RebuildNameIndex()` call in `postoperatorcsv.cpp` and `[tablecsv]`/`[postoperatorcsv]` still pass. I'd drop both and let a stale index throw (the `<algorithm>` include goes with this):

```suggestion
    return name_to_index.count(std::string(name)) > 0;
```

palace/utils/tablecsv.cpp
```diff
@@ -113,11 +113,31 @@ bool Table::insert(Column &&column)
...
+    // Column names can be reassigned after a file is loaded, without insert, which leaves
+    // the map stale, so fall back to the columns themselves and repair the entry.
+    for (std::size_t i = 0; i < cols.size(); i++)
+    {
+      if (cols[i].name == name)
+      {
+        name_to_index[std::string(name)] = i;
+        return cols[i];
+      }
+    }
```
hughcars: Same here, so a stale index throws:

```suggestion
```

test/unit/test-tablecsv.cpp
```diff
@@ -295,3 +295,59 @@ TEST_CASE("TableCSV_LoadFromFile", "[tablecsv][Serial]")
     CHECK(table_w.table[1].data == std::vector<double>{1, 1, 1, 1, 1, 1});
   }
 }
+
+// The restart path reloads a table from disk and then restores the real column names from a
+// reference table. Looking columns up by those names used to fail, because the name index
+// still held the placeholder names the loader assigns.
+TEST_CASE("TableCSV_NamesRestoredAfterReload", "[tablecsv][Serial]")
```
hughcars: Without the fallbacks this test needs a `RebuildNameIndex()` call, at which point it duplicates the next one; I'd fold them together. What would catch a regression in the real path is a name-based check (`has("inc1_1")`, `table["idx"]`) in the existing `restart1_restart_in_middle` fixture in `test-postoperatorcsv.cpp`, which already drives `MoveTableValidateReload` but only checks names by position.

hughcars (changes requested):
Restarts have broken on name lookups since `d821d68ca9` moved `operator[]` onto the index in v0.15.0, and the rebuild in `MoveTableValidateReload` fixes it. Drop the fallbacks (inline) and add a Bug Fixes entry to the CHANGELOG, something like: Fixed driven-sweep restarts (`"Restart"` > 1) aborting with `Column ... not found in table`, because reloaded CSV tables kept stale column-name indices.

PR #946, opened 2026-09-18
Use C++17 trait variable templates and static_cast

nikosavola:
Two mechanical respellings. No expression changes.

- `std::is_same/is_base_of/is_integral<...>::value` -> the `_v` forms, and `std::conditional<...>::type` -> `std::conditional_t` with the redundant `typename` dropped (71 sites)
- C-style casts -> `static_cast` (50 sites)

Palace already targets C++17 and already uses both spellings in places, so this is consistency rather than a new requirement: `spaceoperator.cpp` had `is_same_v` and `is_same::value` side by side, and the tree had 399 `static_cast` against 50 C-style casts. Casts that transcribe a foreign C or Fortran API's argument type (`a_int`, `HYPRE_Int`, `HYPRE_PtrToParSolverFcn`, `task_info_t`) and the C libCEED `qfunctions/` headers are left alone.

First of a stack; the branches above this one build on it.


hughcars (approved):
Checked the `(CeedSize)a * b * c` sites keep the cast on the first factor, and no pointer casts are touched.

hughcars:
Rebase on main to pick up the libCEED pin from #940, the Spack failures are that.

nikosavola:
Rebased on main, thanks!

PR #947, opened 2026-09-18
C++17 cleanups: make_unique, structured bindings, inline constexpr, filesystem

nikosavola:
Four small C++17 cleanups. None change behaviour.

- `unique_ptr::reset(new T(...))` -> `std::make_unique<T>(...)`, 5 sites in 2 files. The project's `.clang-tidy` already enables `modernize-make-unique`.
- two `for (auto it = m.begin(); it != m.end(); ++it)` loops -> structured bindings, where both halves are used
- `static constexpr` -> `inline constexpr` for the four namespace-scope constants in `utils/constants.hpp`, making them one entity instead of one copy per TU
- `utils/filesystem.hpp` drops its pre-C++17 fallback: the build floor is C++17, so the `<experimental/filesystem>` branch and its `#error` are unreachable. The detection was half-broken anyway, since `__cpp_lib_filesystem` is only defined after including `<filesystem>`.


hughcars (approved):

hughcars (changes requested):
Retracting my approve, this doesn't build: `SurfaceFunctional`'s constructors are private with `BoundaryPhysicalTraceCache` and `PointFieldEvaluator` as friends, so `reset(new ...)` compiles inside those classes but `std::make_unique` does the `new` from a non-friend. See `build-test-macos-spack`. Keep `reset(new ...)` for those five sites; the rest is fine.

nikosavola:
@hughcars good catch.

Dropped that commit entirely rather than just reverting it: all five sites were `SurfaceFunctional` constructions, so there was nothing left of it. Branch force-pushed to 28d1bc81, PR retitled/rewritten to drop the `make_unique` bullet.

hughcars (approved):
Dropping the `make_unique` commit clears my earlier request for changes; the structured bindings, `inline constexpr` constants and `filesystem.hpp` change preserve behaviour.

Optional, here or as a follow-up: `check_std_fs_support` in `palace/cmake/CheckCompilerFeatureSupport.cmake` still falls back to `<experimental/filesystem>`, so a toolchain with only that header passes configure and then fails on `filesystem.hpp` instead of at the configure guard. Keep the `-lstdc++fs` retry and replace the `#if` block in the test program with:

```cpp
#include <filesystem>
namespace fs = std::filesystem;
```

PR #948, opened 2026-09-18
Use std::pair class template argument deduction

nikosavola:
Modernization for std::pair


hughcars (approved):

hughcars:
Rebase on main to pick up the libCEED pin from #940, the Spack failures are that.

nikosavola:
Rebased on main, thanks!

